Issue #19, opened 2020-07-05
Revamp


HenkMutsaerts:
Improve inline comments and headers

PR #21, opened 2020-07-06
#19 Revamp quantification scripts

HenkMutsaerts:


jan-petr:
Looks good. But I had to add some argument checking, fixed a minor bug with brackets and removed double %% on subsequent lines.

PR #29, opened 2020-07-08
#28 Fix the issue with loading Lesion masks into CAT.

### LINKED ISSUE #28: xASL lesions are not correctly loaded to CAT
The lesions from Lesion.nii should be loaded to CAT12 and used for cost function masking. This was working for CAT12.5 but not for the new CAT12.7 because it was added only in cat_run_job1070.m but not to cat_run_job.m that is now used in CAT12.7 instead of the older version.

jan-petr:
Close #28


jan-petr:
> Looks good, I assume you tested this with both a case with and without a lesion? Testing on _x.Quality=0_ should be sufficient.

Damn. Yes, tested, but another patient now fails for another reason. Not sure if this is connected or not, so please wait with the pull...

PR #34, opened 2020-07-10
#4 Added different Siemens import scenarios

### LINKED ISSUE #4: Siemens import issue in CCLL order
Dcm2nii creates several separates control/label files instead of a 4D volume for certain Siemens datasets - especially in the GENFI cohort. These files need to be merged in a single 4D volume in the import. There's is a non-trivial syntax of the files names that can be used to reconstruct the correct file order.

jan-petr:
Close #4


jan-petr:
It has no conflicts with other edits in import, so it can be auto-merged.

I have tested it with my synthetic data.

And Maurice also tested it with his Siemens data.

It shouldn't affect any other datasets (non-Siemens, or data without multiple files created by dcm2nii). So all should be fine.

jan-petr:
Comments inline:

> Should this be a separate script? At least I would assign a different category than "adm".
> Perhaps we should simply have an "imp" or "bids" category for this.

xASL_bids_MergeNifti

> You want to check PhaseEncodingDirection as well as PhaseEncodingAxis?
> In case of PePolar, you want both to be identical across series. though Philips doesn't provide these, you have to add them manually..

OK. Added PEDirection.

> I think we can change "nii_files" into "NIfTiPaths" to comply with our coding practice

Agree. nii_files was the original parameter name. Changed to NiftiPaths - as Nifti is used everywhere.

> Check the indentation, you add 1 tab too much for each if/for/case etc

Matlab standard settings - 1 tab=4 spaces, changed this to 2 if you like it more.

> What is "bCheck"?I would specify it, e.g. "bCheckPePolar"
Check the consistency of all files and parameters, not just PEpolar -> bCheckConsistency
 
> What does "NT" stand for? checkNT, tmpNT, less readable then the full name (or semi-full:)
Time dimension -> dimTime, checkTimeDim

> I would change the name xASL_adm_MergeNiiFiles_Merge to make it less similar with the actual main function name
That was on purpose - so that it would be clear that this is a subfunction that is not usable outside.

> fList = xASL_adm_GetFileList(Fpath,'^__parms.m$','List',[],false); Regular expression here should be ._ right, not *
> Also here, would use FileList instead of fList ;)
FileList changed. Don't understand - the regexp works?!

> xASL_adm_MergeNiiFiles_Delete: you delete the .nii & .json, do you also need to delete the _parms.mat here?
No, the parms.mat is there only once, so it is enough to move it. Or not?

> ```
> 			% Get the size of the 4th dimension
> 			tmpNT = xASL_io_ReadNifti(nii_files{iFile});
> 			if length(tmpNT.dat.dim) > 3
> 				tmpNT = tmpNT.dat.dim(4);
> 			else
> 				tmpNT = 1;
> 			end
> ```
> 
> -> this looks lengthy, why not simply tmpNT = size(xASL_io_Nifti2Im ,4) -> this will always default to 1 without error

Lengthy yes, but more efficient. Your code reads the whole volume which is much more time consuming.

> % Only works with Siemens files
> -> unclear, you mean here that the lines above will fail for non-Siemens?
> why not doing this within a switch or if-statement? That seems more readable and less prone to bugs?

Can't do switch or if. I want to run the Siemens first, and if this fails try a generic merging - made this clear in the comments.

> So "bCheck" actually means "bSiemensSpecificSorting", shouldn't this be even a separate subfunction, to make clear that this curation part will only be done for Siemens?

Sure - I did three separate functions for M0, Siemens ASL, normal ASL - makes the code more clear

> Sorry for nagging about these documentation/good coding practice issues, but our available time will get fewer and fewer so the least we want to do is create scripts that we don't understand in a few years time ;)

Sure - these were good tips. Thanks!

HenkMutsaerts:
## I did some edits:
* putting everything in the correct time ("take" instead of "takes"
* Fix switch indentation (otherwise double indentation compared to if-else)
* Making clear what are subfunctions by extra spacing & heading
* Listing subfunctions in main header for clarification
* Some code shortening
* Subfunction 3 (merge any ASL files) is this any non-Siemens or any? is "any" here the same as "generic" for subfunction 1 (M0)?
* I thought we agreed to always have a space behind a comma
* Beginning of header should repeat the function name I believe, per Matlab convention
* Added that pathOut should contain NiftiPaths (input) if nothing happened (as described in header)

## Answers to comments (in reverse order ;)
>Lengthy yes, but more efficient. Your code reads the whole volume which is much more time consuming.
Sure but you added 5 lines and the time consuming is not * 5 timelength ;) I would always go for readability, especially in this stage. If needed, we can create a separate function xASL_io_ReadNiftiSize

>FileList changed. Don't understand - the regexp works?!
Could be, but it is incorrect:
^ = the expression to the right needs to be at the beginning of the line
. = means anything
* = means the expression to the left can be repeated 0-inf times
So ^.*_parms will find anything before _parms
Another thing that went wrong was the *.mat*, in a regular expression the period (.) needs to be escaped to be interpreted as character instead of expression
So:
>FileList = xASL_adm_GetFileList(Fpath, '^*_parms.mat$', 'List', [], false);
should be
>FileList = xASL_adm_GetFileList(Fpath, '^.*_parms\.mat$', 'List', [], false);
You did this correctly in *'\d*'* elsewhere, which means 0 or more digits
>xASL_adm_MergeNiiFiles_Delete: you delete the .nii & .json, do you also need to delete the _parms.mat here?
Each specific NIfTI has its own .json & _parms.mat sidecar, if these are all deleted separatedly, it should be fine
>dimTime
I would call it dim4, but dimTime works as well. It needs to be readable without having written the code ;)
>Tab indentation
Indeed for Matlab default 1 tab is 4 spaces, but careful in GitHub online GUI/website, this can differ. I think it was set by default at my side for 8 spaces, which is why I thought you used 8 spaces :)

jan-petr:
Agree with everything. Yes, I understand, but sometimes too lazy :) Thanks for correcting.

Yes - thanks for the pathOutput.

You're right about repeating function name in the header - didn't know that.

The generic ASL subfunction - yes like for M0 - it takes any vendor including Siemens that was not catched by the specific Siemens merge.

OK for the dimensions - I fixed this though to take the 4th dimension only.

PR #40, opened 2020-07-15
Update xASL_adm_GzipNifti.m

HenkMutsaerts:
Minor bugfix

### DIFF
--- a/External/SPMmodified/xASL/xASL_adm_GzipNifti.m
+++ b/External/SPMmodified/xASL/xASL_adm_GzipNifti.m
@@ -11,7 +11,7 @@ function pathOut = xASL_adm_GzipNifti(pathIn, bOverwrite)
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
 % DESCRIPTION: Take the input file, zips it, overwriting any existing zipped file and return the path of the zipped file.
 %
-% EXAMPLE: pathOut = xASL_adm_UnzipNifti('test.nii',1)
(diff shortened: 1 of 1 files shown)


jan-petr:
You introduced a bug. I have fixed it in my recent commit. See if this is still fine and then we can approve this.

jan-petr:
See my commit here. Unfortunately, I couldn't relate it to this branch since there was no issue associated!
https://github.com/ExploreASL/ExploreASL/commit/acd465a2ad3f600b4f1c33b112e8ba8a2d041eef

jan-petr:
No, you have introduced a bug with your bugfix, so I had to revert it :D

if isGz == 1
then the file is file.nii.gz and name0 = file.nii
so I put breakpoint in the code and your fix created file.nii.nii

jan-petr:
xASL_fileparts and fileparts behave differently!!!

[path0,name0,~] = xASL_fileparts('/test/test/test.nii.gz')

path0 =

    '/test/test'


name0 =

    'test'

>> [path0,name0,~] = fileparts('/test/test/test.nii.gz')

path0 =

    '/test/test'


name0 =

    'test.nii'

HenkMutsaerts:
Ah OK, you're right. Then we can revert my change, and add the comment that this had the .nii (just that I won't complain about it later ;) Then we can squash & merge (or rebase? ;)

You want to do this or shall I do this?

jan-petr:
I would not revert - it fixed other good things. I would keep as it is now - after my last fix, all is good.

Regarding merging - I would do squash in this case - it was a really minor edit, so we don't need to have the long list of comments in between.

Issue #46, opened 2020-07-20
 error in Compiling ExploreASL


sandeepganji:
Compiling ExploreASL
{Error using <a href="matlab:matlab.internal.language.introspective.errorDocCallback('ExploreASL_make_standalone', '~\Development\ExploreASL_make_standalone\ExploreASL_make_standalone.m', 223)" style="font-weight:bold">ExploreASL_make_standalone</a> (<a href="matlab: opentoline('~\Development\ExploreASL_make_standalone\ExploreASL_make_standalone.m',223,0)">line 223</a>)
Error: The specified file <Modules\SubModule_Structural> was not found. If you wrapped the file name with double quotes, try replacing
the double quotes with single quotes and rerun the command.

sandeepganji:
The issue is coming from the ExploreASL_make_standalone.m file. The paths for 'SubModule_Structural', 'SubModule_ASL', 'SubModule_Population', 'isnear','AtlasesNonCommercial' are not appended by the ExploreASLPath.

Adding that fixed the compilation error.

HenkMutsaerts:
Great, in which version did you notice this? In the master version or (also) in the develop branch?
Also: please assign somebody to this issue, label it and put a milestone in (I now did this for you :)

sandeepganji:
Found that in Release version 1.1.2.

MichaelStritt:
I see the problem. Will fix it in a second. Wanted to merge my other branch anyway.

HenkMutsaerts:
@MichaelStritt Can you put this in a separate branch, so we can merge this irrespective of your branch, let's keep things separated for overview ;)

PR #49, opened 2020-07-21
Closes #32 docker integration and #46 compilation error

### LINKED ISSUE #32: Docker integration
Allowing more flexible input and output, for integration with a docker. 
*Note that in principle*, a Docker should always wrap around software, the software should not be adapted to a specific environment. The software ExploreASL should be as general as possible. Therefore, we should only do general improvements here, that will work in more situations.

MichaelStritt:
Main changes are within ExploreASL_make_standalone and some of the import scripts. This is a preliminary solution for a first demo docker version.


MichaelStritt:
### ExploreASL_MakeStandalone

- Fixed the \n in the error message
- The -a option can be used for files and folders https://www.mathworks.com/help/compiler/mcc.html In this example I wanted to reduce the size of the separate compiled version which only uses a really small subset of ExploreASL functionality. It actually wouldn't even need SPM if the initialize scripts wouldn't check if spm is there for example.

MichaelStritt:
### xASL_dcm_Import

- We can integrate this into the ExploreASL_Import. This would make a lot of things easier for me. I just thought that ExploreASL will only be used in this docker/mTRIAL use case from a very limited number of people.
- Some things are just temporarily static. We can change this in a new update
- Right now mTRIAL automatically detects ASL and T1 files. It is a bit problematic to automatically detect the M0 files (robustly) though. This is why this functionality will come with a later update.

MichaelStritt:
### ExploreASL_ImportConfig

- I thought there was a problem if I don't use tokenVisitAliases at all.
- Other issues were fixed, see above commit.

MichaelStritt:
Hey guys, sorry for the mess with all these comments and revision commits. Hope the code has improved though. Will hopefully test the integration with the new json import file within the next days.

jan-petr:
ExploreASL_ImportConfig.m line 853 - I added more sessions than 2 - to avoid issues in the future.

Line 109-114. You don't need to do this by hand, but rather use function xASL_bids_parms2BIDS that takes structure in xASL format and in BIDS format and merges them and you can set which of them has the priority and if it outputs as xASL or BIDS. All these conversions and differences are taken care of here and it's better to use this one so that any future changes are done only there and not in all functions.

Line 131-135 - better to use spm_jsonread so that we use a single function to read JSONs consistently in the whole code. xASL_bids_parms2BIDS will take care of the manufacturer-vendor conversion.

Line 150-154 - simply use spm_jsonwrite

MichaelStritt:
I don't know. The result seems to be ok.
Annotation 2020-07-31 105728

jan-petr:
:+1:

jan-petr:
I edited two small things. For converting x.Manufacturer to x.Vendor, I use a function dedicated to that, rather than doing it manually (so that all these conversion are kept within that function).

Second - it would have crashed when MRAcquisitionType would not be present - circumvented that.

PR #51, opened 2020-07-21
#48 xASL_io_Nifti2Im.m detect wrong scaling

### LINKED ISSUE #48: Detect NIfTIs with erroneously high scale slopes and/or image intensities
Came across a T1w image with too high intensities, that were clipping in the MriCron viewer. I would suggest to rescale in the function _xASL_im_ClipExtremes_ to a max of 4096 to avoid this.

HenkMutsaerts:
Fixes #48

### DIFF
--- a/External/SPMmodified/xASL/xASL_io_Nifti2Im.m
+++ b/External/SPMmodified/xASL/xASL_io_Nifti2Im.m
@@ -135,20 +135,20 @@ end
 % warning & try to correct this
 
 if exist('imOut', 'var') && exist('nii', 'var')
-    if nii.dat.scl_slope>16
-        % this if-clause speeds up in all other cases that don't have this potential Philips import issue
-        MaxIm = max(imOut(isfinite(imOut)));
-        if MaxIm>1e9
+	if nii.dat.scl_slope>16
+		% this if-clause speeds up in all other cases that don't have this potential Philips import issue
+		MaxIm = max(imOut(isfinite(imOut)));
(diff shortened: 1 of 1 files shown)


HenkMutsaerts:
## Change the following (discussed with Jan)
* Remove the error
* Values need to be higher than 10^8
* Structural only (T1/FLAIR) automatic fix
* Warning for any regexp
* Ignore the rescale slope

jan-petr:
Did the requested changes.

HenkMutsaerts:
Thanks, though:
* it again has the annoying tabs...
* why not resetting the max value to e.g. 4096? these are structural images
* And we need to remove the _if rescale>_

jan-petr:
> * it again has the annoying tabs...

Again - the tabs are not my problem :) That's the viewer. I add indentation as Tabs, not as 4 spaces. In some programs, this is displayed as 8 spaces. When displaying in a GitHub web interface, put '?ts=4' at the end of the URL - it will show normally. I can of course enter hard 4 spaces instead of 1 tab, but that's less elegant and I though we have agreed on tab.

> * why not resetting the max value to e.g. 4096? these are structural images

OK. Done.

> * And we need to remove the _if rescale>_

OK. Done.

HenkMutsaerts:
Great, yes we agreed on tabs and I fixed the annoying default displaying of GitHub.com to 4 instead of 8 spaces per tabs, for our repository (though still only in the development for now).

PR #64, opened 2020-07-28
Feature#3 ASL registration using affine/DCT

### LINKED ISSUE #3: ASL registration using Affine and DCT registration
The ASL can be sometimes deformed with respect to T1w (or with respect to MNI in case a direct ASL-MNI registration without T1w is done). In that case, you might want to do a affine or affine+DCT registration on top of the Rigid body registration.

jan-petr:
Close #3 

- Implements the DCT+affine registration option for T1w-ASL registration.
- Cleans up the previous affine registration.
- Fixes several bugs in the decision process of rigid/affine/affine+DCT registration of T1w-ASL.


jan-petr:
> * _Functions/xASL_im_CreatePseudoCBFm_
>   bExtraScaling is in the explanation but not in the function call?

Sorry. This was an input option, but I have removed it already previously. Now normal scaling is done for non-PVC, and extraScaling is always done for PVC and only for PVC. I have left the bExtraScaling by error in the function description. Typo fixed.

> * You changed the 1. 2. etc in the header comments into 1) 2), but for Michael’s documentation crawler this needs to be periods e.g. 1. 2.

OK, fixed. Thanks for noting this.

> * Please reset the double tabs to single tabs (or 4 spaces instead of 8), e.g.
> 
> if nargin < 3 || isempty(bPVC)
> bPVC = false;
> end
> 
> This is the case throughout. Also makes for many differences, more difficult for me to review, as all these extra tabs are seen as differences. Could be that you are editing partly in the browser, that resets it to 8 instead 4 spaces?

Sorry, but all is fine on my side. There's only one single TAB. It is the GitHub that displays it wrong. Open the file in GitHub, click edit, set to 4 and you'll see it. 
Or, add '?ts=4' in the URL for viewing...
https://github.com/ExploreASL/ExploreASL/blob/Feature%233_ASLprereg/Functions/xASL_im_CreatePseudoCBF.m?ts=4


> * Line 97: if ~bPvec, you don’t remove/recreate the x.P.Path_rc1T1, do you know for sure
>   that this is the correct one, in ASL space? Or should this be the Path_PVGM?

This is how it was before. rc1T1 was created and deleted. The only thing I have changed is that for PVC, it deletes it on line 137 and not 97.

>   I would recreate this temporary file. This may take a bit more computation time, but at least you know for sure that you have the exact tissue maps that you want (less possibilities for bugs).

But it is recreated on lines 82-87.

> * Why do you convert NaNs in zeros? The NaNs have a specific meaning, i.e. outside of the FoV?
>   Why do we want to remove this meaning? Because PVC doesn’t deal with NaNs? Then you need to
>   let PVC automatically do PVC only in kernels that don’t have NaNs. We only set NaNs outside of the FoV,
>   or e.g. by vascular masking. But also then, we don’t want to convert this to 0.

I don't remember - it was a serious problem for something. Probably the affine registration. You sure you want the NaNs back?

> * RatioMax = (maxPseu-minPseu)/(maxPWI-minPWI);
>   If we do this, we first need to clip below 0, as we don’t want to include
>   the negative noise tail in the range here right?

It worked fine like this. We can of course remove the min completely, but then remove both minPseu and minPWI. Up to you.

> * xASL_spm_affine: warning('Cannot modify the headers in otherList if bDCT is true');
>   What does this mean? A warning should be clear right ;)

Was clear to me :D

warning('otherList not empty and bDCT==1. DCT produces a _sn.mat file with transformation parameters, it cannot apply the transformation to the files in the otherList');

> * x.bAffineRegistration -> turned off by low quality
>   I would not change the processing with low quality, if possible, only do fewer iterations
>   for faster testing. Otherwise the low quality testing won’t be sufficient for thorough
>   testing.

Sure. Removed this. Don't know how to decrease easily the number of iterations though. 

> * Why do we have both “bDCT” and “bDCTRegistration”?

I have used bDCT registration in xASL_wrp_RegisterASL as there was already the bAffine. In xASL_spm_affine, there's not the x-struct, so I had to pass the parameter. I thought that here, bDCT is clear enough. But I can of course rename it to bDCTregistration as well if you think it would make things more clear.

> * ~isfield(x,'bUseMNIasDummyStructural') -> missing the isempty

Thanks. Fixed.

> * Additions to the BaseOtherList: ensure here that this is correct, registrations will be applied to these images as well throughout.

Yes - that's what I wanted.

> * Why do you add this comment:
>   % By default use the GM, WM and T1 from the template
>   Only for the rT1 and not the rc1T1/rc2T2 (placing of the comment?)

Good point. I have reorganized the comments a bit.

> * We have 1) rigid 2) affine 3) DCT, why are affine & DCT combined in the affine registration?
>   Makes it difficult to read. We won’t do DCT in many cases, where we will only do affine,
>   so I would split the two. Will iterate the affine a bit, which is not terrible,
>   but makes code much more readable: 1, 2, 3 all unrelated. Not 2 & 3 intermingled, becomes then more difficult to debug. You do something similar here as you did with the ROI stats ;)

Fixed as we discussed.

Please, let me know about the remaining issues above.

HenkMutsaerts:
* The minPseu/minPWI: we need to keep it for cases where the min()>0. Only for cases where min()<0, this adds noise
to the registration. So I would simply clip below zero before this, and leave this range otherwise. This should already have been done in the creation of the registration-dummy ASL image, but since we explicitly define a range here, it would be good practice to explicitly clip below zero.

* Low quality: indeed ‘old normalise’ doesn’t allow setting the affine registration iterations (at least not by default).
However, this goes pretty fast already as the cost function has low-computation cost. The DCT takes a bit more time
so you could consider setting the non-linear iterations to a lower number if x.Quality==0, but if you are worried that this would not work, then we can keep this as is.

* Yes the NaNs need to be kept, as they have a meaning, e.g. to retrace the original FoV in the new resampled space.
The SPM registration shouldn’t matter for the NaNs, right? If it is an issue for the registration, we need to create a NaN-mask that we can re-apply at the end. For PVC, you can ignore NaNs right? Setting NaNs to zero before PVC will really disturb PVC right, if you only fo this for the ASL image and not for the tissue priors.

* I would still vouch for increasing the PVC kernel size or making the globally, for sequences other than 3D spiral.
At the very least, insert a warning when DCT/affine is requested for non-tested datasets (e.g. not 3D spiral).

* Did you add the parameters to DataParTemplate.m?
Also, in this template file, it should be clear that you need bAffine to do bDCT.
Or should we throw a warning if bDCT==1 && bAffine==0, and reset bAffine==1?

* Shall we add a DCT example to the 10 testdatasets? Would be good to always add new features to the testing
(not necessarily by adding a new datasets, can be implemented in an existing dataset), in this case we can 
set bDCT==1 in the DataPar.json of a 3D spiral dataset.

* Reverting the change if TC is not improved: this would differ for bCT right?
As in bDCT we don’t have to reset any files (no files are changed), but we do have to delete the new sn_mat.

* % Remove the temporary files - Clipped_ORI and rPWI  -> this only had _ORI, not rPWI? Does this still needs to be deleted? Actually, I would not use rPWI (too general), but use the same filename as originally (mean_PWI_something) with _DCT suffixed

jan-petr:
First batch of edits:

Added feature - low quality DCT with 4 basis functions only.

Doesn't backup files for DCT, but only deletes _sn.mat.

bDCTRegistration option explain in DataParTemplate. A warning is thrown when bAffine is 0 and bDCT 1, but we cannot turn on the bAffine, because it has options 1 and 2, so don't want to impose under which mode is the affine and thus DCT run. But you can if you want... At least it is now all explained. Anyway - DCT is not for kids ;)

jan-petr:
Second batch of edits.
x.P.Path_mean_rPWI changed to x.P.Path_mean_PWI_Clipped_DCT, and deleted at the end.

Regarding the min/max. This is not so clear issue. You previously had this in the code:
DiffMin = minPWI - minPseu;
PWIim = PWIim-DiffMin;
RatioMax = maxPseu/maxPWI;
So you were previously manipulation the minimums, in that case, I found it wrong to scale by max only and therefore, I have scaled by the full range (max-min). I can revert to scaling by max only, but then we have to removing the shifting of the mins as well - up to you how you want to solve this. But currently it works rather nicely.

Removed the NaNs and put them only to the end.

PR #68, opened 2020-07-29
#67 Fixed the issue with overwritting arrays in xASL_adm_LoadParms

### LINKED ISSUE #67: LoadParms crashes on merging arrays
- xASL_adm_LoadParms.m crash when reporting unequal merging of arrays.

jan-petr:
Close #67 

- Fixed the issue with overwriting arrays in xASL_adm_LoadParms

### DIFF
--- a/Functions/xASL_adm_LoadParms.m
+++ b/Functions/xASL_adm_LoadParms.m
@@ -204,9 +204,11 @@ end
 
 for iField=1:length(Qfields)
     if isfield(x,Qfields{iField})
-        if isfield(x.Q,(Qfields{iField})) && ~min((x.Q.(Qfields{iField})==x.(Qfields{iField})))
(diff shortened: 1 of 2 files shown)


jan-petr:
> * Should we deal with these small issues in the "minor issues" category, or better to keep reviewing them like this?

Yes. Very minor can be fixed directly, this is a borderline case. But I don't see a large overhead in reviewing those - there's no need to put a lot of comments. Just have a look for 30 sec and say OK ;)

> * bConcatenate: while I understand this is useful, the implementation now is very specific to JSON reading right? Shouldn't bConcatenate be an input string, for the separator (e.g. can be ',', but also '\t' which would include a tab, or ' ' for a space. If empty, it is defaulted to ''.

I though about it previously. But this would need two extra parameters, bConcatenate and strDelimiter. Because you might want to concatenate with '' as well or not concatenate at all. So I would leave it as it is now and only change when needed. I agree that this might be useful, but we should spend time on every minor detail. Agree?

jan-petr:
The delimiter options is now tweaked to the last detail.

PR #72, opened 2020-08-03
#71 Correct WMH_SEGM for lesion ROIs/masks

### LINKED ISSUE #71: Mask WMH_SEGM for custom lesion masks
Lesions such as tumors have hyperintensity on FLAIR, and would be automatically segmented as WMH_SEGM.nii. Instead, they should be masked out in case any custom - e.g. manually segmented - lesion masks/ROIs are provided, either in FLAIR or T1w space. Also, a warning should be issued, especially if the custom lesion mask is not binary.

Lesion masks: ^Lesion_(FLAIR|T1)_\d*\.nii$

HenkMutsaerts:
Fixes #71

### DIFF
--- a/External/SPMmodified/xASL/xASL_im_SaveOriginal4CAT.m
+++ b/External/SPMmodified/xASL/xASL_im_SaveOriginal4CAT.m
@@ -30,7 +30,7 @@ else
 	% The parameters of the CAT12 space are taken from the original space and modified
 	volCAT = spm_vol(T1Path); % Loads the parameters of the original space
 	volOriginalvoxelSize    = sqrt(sum(volCAT.mat(1:3,1:3).^2)); % calculates the voxelsize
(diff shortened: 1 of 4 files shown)


HenkMutsaerts:
* I wanted the new part to be functional on itself, didn't test it separately. So if you want the loading & saving done once instead of twice, you have to test this ;)
* One of the functionalities of xASL_delete is that it only deletes a file if it exists

jan-petr:
> * I wanted the new part to be functional on itself, didn't test it separately. So if you want the loading & saving done once instead of twice, you have to test this ;)

It is saving it not twice, but several times :D It works - it rewrites the variable and newly save it on each occasion - absolutely the same functionality to save it only on the last occurrence - but I understand that through testing it was good to have it in each iteration ;)

> * One of the functionalities of xASL_delete is that it only deletes a file if it exists

I know - for transparency, I put it also in the function. But you can remove that one.

PR #83, opened 2020-08-11
#82 Fixed BIDS validator issue for 3D files from 4D sequences.

### LINKED ISSUE #82: BIDS validator rejects 4D files with NT==1
Files that are acquired using a protocol that is otherwise 4D (e.g. BOLD, but not T1w), but has the fourth dimension == 1 and is thus only a 3D file is rejected by BIDS validator.

In NIFTI header two fields, dim and pixdim, are given in a format [ND NX NY NZ NT ...] to specify the number of voxel in the dimension and the size of the voxel in that dimension. Additionally, dim[0] (ND) specifies the number of dimensions.

The trouble with 4D files with NT==1 is that pixdim[4] == TR, but dim[4] == 1. At the same time, when dim[0] == ND == 3, the BIDS validator thinks he has 3D file only and complains because he sees that pixdim[4] is non-zero.

This needs to be fixed in NIFTI writing routines to set dim[0] to 4 in these cases.

jan-petr:
Close #82

### DIFF
--- a/Development/BIDS/load2bids.m
+++ b/Development/BIDS/load2bids.m
@@ -745,6 +745,9 @@ for ii = 1:length(fList)
 				if scaleFactor
 					imNii = imNii .* scaleFactor;
 					xASL_io_SaveNifti(fullfile(inSesPath,[aslLabel '.nii']),[aslOutLabel '_asl.nii.gz'],imNii,[],1,[]);
+				elseif size(imNii,4) == 1
+					% The fourth dimension is 1, so we have to write the file again, to make sure the 
+					xASL_io_SaveNifti(fullfile(inSesPath,[aslLabel '.nii']),[aslOutLabel '_asl.nii.gz'],imNii,[],1,[]);
 				else
 					% Copy the ASL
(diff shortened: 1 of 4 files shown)


jan-petr:
So of course BIDS complains about this only for M0 and ASL files, the T1w has to be the other way around - facepalm.

Wait with the merge until I fix this.

jan-petr:
Fixed. This is now saving only ASL and M0 and not T1w as 4D file.

jan-petr:
@HenkMutsaerts I have noticed, that I wasn't correctly converting BIDS to xASL structures in the BIDS examples - I have failed to recalculate the milliseconds to seconds - or I have done it for a few fields manually, but not for all.

This is now fixed, but I have modified also the xASL_bids_parms2BIDS as the conversion was not applied to some fields. This probably wasn't a problem before because those fields (PLD and LabDur) were always given in the xASL struct and not in BIDS. But this needs to be properly checked. Also there probably was a historical confusion between parms.m, parms.json (in xASL) and parms.json (in BIDS) - so hopefully all is fixed now, but it needs a thorough testing!

jan-petr:
All should be fine and ready to merge.

PR #95, opened 2020-08-18
#90 Fixes #90

### LINKED ISSUE #90: xASL_io_Nifti2Im Error
Hey guys, just stumbled upon another error. 

Background info: current develop branch, Windows 10, compiled ExploreASL

Has to do with io of Niftis, so I also assigned Jan, hope that's ok.

Annotation 2020-08-14 204659

HenkMutsaerts:
This is part of the xASL_io_Nifti2Im that loads an image and automatically checks if it needs to be scaled. But if the image is empty for whatever pipeline reason, it cannot find a maximum and crashed :) So this can be safely skipped.

I suggest putting this as hotfix in release 1.2.1.

### DIFF
--- a/External/SPMmodified/xASL/xASL_io_Nifti2Im.m
+++ b/External/SPMmodified/xASL/xASL_io_Nifti2Im.m
@@ -136,17 +136,19 @@ end
 
 if exist('imOut', 'var')
 	MaxIm = max(imOut(isfinite(imOut)));
-	if MaxIm>1e9 && exist('Fpath', 'var') && exist('Ffile', 'var')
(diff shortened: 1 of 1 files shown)


MichaelStritt:
Hey, just commented in the original issue https://github.com/ExploreASL/ExploreASL/issues/90#issuecomment-675548811
Definitely better now, but there are still some minor issues. I'd approve this as a more or less temporary fix though.

jan-petr:
> Hey, just commented in the original issue [#90 (comment)](https://github.com/ExploreASL/ExploreASL/issues/90#issuecomment-675548811)

I agree Michael and answered in the original issue as well - it seems that this fix fixes the symptoms and not the disease...

PR #98, opened 2020-08-19
#90 Fixes #90

### LINKED ISSUE #90: xASL_io_Nifti2Im Error

HenkMutsaerts:
Fixes #90 (or only partly)

This is part of the xASL_io_Nifti2Im that loads an image and automatically checks if it needs to be scaled. But if the image is empty for whatever pipeline reason, it cannot find a maximum and crashed :) So this can be safely skipped.


jan-petr (approved):
Fine to merge this to the hotfix - given that the other part of the issue will be fixed later.

Issue #99, opened 2020-08-19
Loading metadata crashes in some conditions


HenkMutsaerts:
Zeroes should also be able to be loaded, and strings, and always nicely referred to as indices internally in ExploreASL

HenkMutsaerts:
## xASL_str2num
* Header added
* Allows converting strings inside a cell array to number array or keeping them in a cell array
* Allows choosing between resetting nonnumerical strings to NaN or keeping them (when keeping cells only)
* Other minor improvements

HenkMutsaerts:
## xASL_init_LoadMetadata
* Allowing strings as well as numbers, make this flexible (& hopefully still readable :)
* avoid creating warnings for xASL specific MAT-files
* Cosmetic changes

PR #100, opened 2020-08-19
Bug #99 loading metadata

HenkMutsaerts:
## xASL_str2num
* Header added
* Allows converting strings inside a cell array to number array or keeping them in a cell array
* Allows choosing between resetting nonnumerical strings to NaN or keeping them (when keeping cells only)
* Other minor improvements

## xASL_init_LoadMetadata
* Allowing strings as well as numbers, make this flexible (& hopefully still readable :)
* avoid creating warnings for xASL specific MAT-files
* Cosmetic changes


jan-petr (approved):
OK.

PR #103, opened 2020-08-20
#102 Fixes #102

### LINKED ISSUE #102: xASL_wrp_CreatePopulationTemplates bug
## By Maurice Pasternak

Please be aware that a different bug appears downstream at line 323 within the xASL_wrp_CreatePopulationTemplates.m file that was altered. 'Undefined function or variable 'Sets2Check'

I see that the variable is attempted to be defined earlier in that script file but in some user cases, like mine, it will never be defined. Specifically, on line 118 in xASL_wrp_CreatePopulationTemplates.m, if bCompute4Sets is 0, then Sets2Check is will not have the opportunity to be defined. Should probably have an "else" statement around that area to define Sets2Check as an empty container if the variable does not exist.

HenkMutsaerts:
@MauricePasternak you want to check this branch out and test if xASL_wrp_CreatePopulationTemplates now works for you?
This should fix #102


jan-petr (approved):
OK

PR #104, opened 2020-08-20
Fixed the results table creation for TestCases

HenkMutsaerts:
Actually this was a bug, not a feature ;)
But anyway this is a developer feature.
I have now recreated this change from the hotfix branch, so we can insert this in release 1.2.1.
Makes for easier testing ;)
I cannot assign myself as reviewer though :(


jan-petr (approved):
Looks good. The guy that wrote it knows his job.

Issue #55, opened 2020-07-23
ExploreASL GUI in Python


MauricePasternak:

HenkMutsaerts:
## Short notes of our the ExploreASL_GUI kickoff meeting
>Present: Vera Keil, Maurice Pasternak, Michael Stritt, Jan Petr, Henk Mutsaerts

* Development stages:
    * A) Co-development (current stage)
    * B) developer testing (developers test non-compiled version)
    * C) user testing (e.g. Vera Keil tests compiled version)

* ToDo Henk
    * Allow choosing to run individual submodules (e.g. by adding 11, 12, 13, 14, etc for structural submodules to the current iModule option)
    * test "weights" for each submodule -> compute the percentage length for x.Quality=0 & x.Quality=1, by adding tic toc throughout & writing this to a script, after running the individual submodules
    * Create metadata parameter file (JSON), where each option should have childs:
        * data format (numerical/dropdown box (ordinal)/string/checkbox (Boolean)

* To be added:
    * Submodules
    * Lock file management
    * Progress visualization
    * Compilation
    * Multi-OS support

MauricePasternak:
I believe it is time to launch this. I've tested it on my datasets and the TestDataSet that comes with ExploreASL (very glad it has FLAIR for last-minute troubleshooting).

In a recent and important update, it is independent from the MATLAB Engine API, making compilation a reality. At startup, the program calls out for matlabroot. If that variable is not present, the program will not start. You can verify this by trying the program on a machine lacking an activated MATLAB.

Of the to-do list, it should have: 

- Submodules
- Lock file management (it will be located under "Modify for Re-run" in the Executor)
- Progress visualization and error management (though the latter should be expanded for the sake of developers...because it's free of the Engine API, it cannot visualize pipeline print output until the process finishes through stdout. I will create a separate issue for how this can be elegantly addressed from your end.)
- Multi-OS support (so long as a call to the system produces the key words: Darwin, Linux, Windows).

Regarding compilation; I am able to freeze-compile it on Windows and package it into an installer. The installer ends up at 85MB and when unpacked, the "frozen" source code + binaries is 344 MB. I have not bothered to optimize imports, so we should be able to bring those numbers down if needed. The freeze-compiler (PyInstaller) must be run separately on each operating system to create system-specific binaries. The python code, however, is obviously agnostic to OS and will work on any OS so long as it has Python and the necessary libraries present (ex. PySide2, seaborn, numpy, etc.). 

[...]

PR #107, opened 2020-08-31
Feature #55 x asl gui python

MauricePasternak:
All relevant additions are located in: ExploreASL\External\ExploreASL GUI


MichaelStritt:
Hey, don't know how to actually review all of this, but maybe it would make sense for one of us (Henk-Jan, Jan, Sandeep or me) to actually **test** the functionality, too. As said before, the code looks really clean and nicely written. It's great that you also thought of writing a markdown readme. This could be really useful for future users. Since all of this basically does not change the main ExploreASL code and is only located in **External** I'd approve the merge request though. Great work!

MichaelStritt (changes requested):
Can we get rid of the .idea folder?

HenkMutsaerts (changes requested):
Great work!
I would address the following points:

* Files with different/separate license: why don't you also create a folder called _External_ inside your _ExploreASL_GUI_ folder, and put the code from different licenses there? This would be clear and avoid any license clashes/conflicts. Saw you already did this for dcm2nii!
* While spaces in file/folder names shouldn't be an issue anymore, they can still be an annoying source of bugs that take a long time to find. I would recommend renaming _ExploreASL GUI_ to _ExploreASL_GUI_ (did this for you)
* In the README.md: I wouldn’t specify the Q&A for your availability (the last one you specified).
You just don’t want to bring them on ideas, or suggest anything that may change.
But that’s up to you, this is just my recommendation :)
* I have changed some names in _External/ExploreASL GUI/JSON_LOGIC/ExecutorTranslators.json_

* The actual merging:
Just tested it, we cannot simply rebase as there happened too much in the develop branch in the meantime. So best is to repair this branch by cherry-picking, such that it doesn't contain any conflicts with the current develop branch. I can do this. Preferably I put everything in a single squashed commit, if you're OK with this?

MauricePasternak:
@MichaelStritt 

Regarding .idea and __pycache__ , those are incidental byproducts of the IDE (Pycharm) I'm using. Not essential to the core program; it just helps the IDE understand project structure (i.e when I'm importing classes/functions from one .py file to another). I'll purge it on the next commit.

Regarding you being able to review this...I can look into preparing a setup.py file as if I were to publish the project. That way, if you have Python, you would be able to do a `python setup.py install` on your command line. Alternatively, I do have a compiled version via installer that works for Windows. The downside would be basically no ability to give good troubleshooting feedback.

@HenkMutsaerts 

All these sound good. Just to clarify, at no point anywhere have I altered the source code of any library I used. This is particularly relevant, as PySide2 uses LGPLv3. And for everyone's records and awareness, here is a list of the external libraries used and their licenses:
- PySide2: LGPLv3
- Pydicom: MIT
- watchdog: Apache 2.0
- nibabel: MIT
- nilearn: BSD-3
- seaborn: BSD-3
- numpy: BSD-3
- pandas: BSD-3
- more_itertools: MIT
- matplotlib: Python Software Foundation License (PSF); which is basically what Python itself uses

MichaelStritt:
Hey Maurice,
just to clarify, I don't think the .idea and the pycache within the ExploreASL_GUI are a problem. It's just about the .idea within the main ExploreASL folder, since all of the GUI related stuff should probably stay in the external folder.
Regarding the testing: I use python myself and the testing of the GUI in general is possible. It's probably more a question for the dev group on thursday. I'm just not sure how we will keep on testing different functionalities of ExploreASL. Some parts of the whole ExploreASL project are just not tested within Henks pipeline tests, but since ExploreASL_GUI is external for now, I'm not sure if we have to test it at all. Right now we are experiencing that we have to test more and more for each release/hotfix, so we have to automate a lot of these things as good as possible.

HenkMutsaerts:
* Great, I agree with Michael to keep everything in External/ExploreASL_GUI.
* Ideally, there is a separate folder inside ExploreASL_GUI containing the compiled version.
* For the installation for our testing, it would indeed be good to have a setup.py for the installation of all your external packages. If they are easily downloaded (which they are) and you didn't edit them, I indeed wouldn't include them.
* When you create the setup.py: is there a version dependency? E.g. users may already have some of the libraries with a different version installed? I would make all of this clear in e.g. a setup.md.

HenkMutsaerts (approved):

HenkMutsaerts (changes requested):
* remove External/ExploreASL_GUI/.gitignore
* cleanup .gitignore
* cleanup reviewers
* ask Henk to merge ;)

MichaelStritt:
Hey Maurice, I finally found some time to test the GUI properly, but I think I'm doing something wrong.
Hope it's okay if I ask some possibly stupid questions? I guess I just have overlooked something.

So in your install instructions it says that I have to unzip the repository and run the setup.exe, but I don't really know which folder to unzip here. But I guess that's not a problem, since I have python on my system anyway. So I created a virtual environment and used your ```requirements.txt``` as well as ```python setup.py install``` to set it up. Running python ```xASL_GUI_run.py``` sadly lead to the following error message:

```
Traceback (most recent call last):
  File "xASL_GUI_run.py", line 1, in <module>
    from ExploreASL_GUI.xASL_GUI_MainWin import xASL_MainWin
  File "...\ExploreASL\External\ExploreASL_GUI\ExploreASL_GUI\xASL_GUI_MainWin.py", line 5, in <module>
    from ExploreASL_GUI.xASL_GUI_Executor import xASL_Executor
  File "...\ExploreASL\External\ExploreASL_GUI\ExploreASL_GUI\xASL_GUI_Executor.py", line 8, in <module>
    from more_itertools import peekable
ModuleNotFoundError: No module named 'more_itertools'
```

I am using python 3.7.6 on windows server. I guess I'm just not using it right, so I thought I'd ask right away ;)

MauricePasternak:
@MichaelStritt 

Today I reminded myself that more_itertools is apparently not part of the standard Python library (it honestly should be...). 
In which case I'm attaching the updated requirements.txt (I'll push this once I get back on my workstation desktop). I also apologize from my end if I've forgotten any parts of the python setup, as I've never done this before. I believe I also forgot to include the blank __init__.py file inside ExploreASL_GUI/ExploreASL_GUI to let setuptools know that it is a package.

There is an additional important consideration that Henk & I discovered yesterday. You should have MATLAB version 2019a or later if you are running Windows. This is because the main Executor function that launches ExploreASL through a shell does so with a flag `-batch` . The purpose of this flag is to stop any operating system from launching unresponsive MATLAB cmd line windows and to keep all processing output confined to a shell's stdout and stderr. I was very disappointed to see that MATLAB did not support running their software non-interactively via shell until 2019a.

https://www.mathworks.com/help/matlab/ref/matlabwindows.html?s_tid=srchtitle

I cannot wait until ExploreASL is compiled into it's own exe so that the GUI can interact with it via shell without this version restriction.

It is possible to switch the `-batch` flag with `-r`, but this will result in unresponsive MATLAB command line windows popping up (they will still be processing successfully but it just looks incredibly unprofessional). Again, I'll blame MATLAB for this as well because they don't support the C++ multithreading that PySide2 is performing with quick window switching (that is why the program doesn't freeze during processing). 

As for the setup.exe installer for Windows x64 ... I was about to push it with that installer located in the folder ExploreASL_GUI/compiled but Henk rightfully advised against that since pushing ~85MB each time will exhaust Github's capacity for a repository very quickly.

Since I can't attach it via GitHub here is the restricted Google Drive link to it: https://drive.google.com/file/d/1w8nfNi5uwlPoalgYEi0gtsmrr-TWRSbX/view?usp=sharing

[requirements.txt](https://github.com/ExploreASL/ExploreASL/files/5178511/requirements.txt)

MichaelStritt:
Hey Maurice, no worries regarding the whole installation process. This is your first big pull request in this repository and all in all this whole process is there so that we can find these little bugs. As a part of my internship I wrote a little script, which was optimized by Henk. It's in the development folder and called ExploreASL_make_standalone. I'm using the standalone version for the docker and it's working quite well. So if you want to use an exe version, there are compiled ones in the Revamp folder on the ExploreASL drive: https://drive.google.com/drive/folders/1z6fbW6GnlmPhicXvXcRtGZQxptujN7o1?usp=sharing
They are not that up to date, but I can add an updated one with the hotfix next week.

MichaelStritt:
> There is an additional important consideration that Henk & I discovered yesterday. You should have MATLAB version 2019a or later if you are running Windows. 

I don't think we need to provide the same backwards compatibility for the GUI & for ExploreASL in general, so I wouldn't worry too much about that.

> EDIT: Actually, on that note I have a question for you & @HenkMutsaerts on whether the compiled ExploreASL will even accept command line arguments, since that is the current means by which my GUI launches the MATLAB-confined ExploreASL_Master.m scripts.

I don't know what you mean. If you use the current compiled windows version, so the ExploreASL exe, then you can call it similar to the normal ExploreASL_Master function. You simply run the exe and pass the DataParFile as a separate string. Let me know if anything is unclear there.

HenkMutsaerts:
## Matlab backwards compatibility of ExploreASL
Michael is right, if the -batch option is the only issue, then let's keep the requirement for using Matlab 2019a or later. In fact, your GUI should work with the compiled version, as using Matlab can still be daunting to some clinicians ;)

## Using ExploreASL with arguments
The compiled version of ExploreASL works identical as the non-compiled version, i.e. it can take the same input arguments. In the future, we should add the option for individual submodules to the ExploreASL input. Or remove the inputs of ExploreASL altogether, and move them all to the DataPar.json to keep it clean?

MichaelStritt:
> In the future, we should add the option for individual submodules to the ExploreASL input. Or remove the inputs of ExploreASL altogether, and move them all to the DataPar.json to keep it clean?

Good ideas. Both are valid options and we could discuss this on thursday in the dev group :)
We could rethink some of the methods for rerunning a study in general, too. If you would define that you are doing a rerun and you only want to run the population module for example, then ExploreASL could remove the "lock/status" files itself. This would only require two input fields in the DataPar file and a small script which removes some of the files.

Coming back to this pull request. How do we move forward? I think because of the conflicts, it would be good if @HenkMutsaerts  could do the merge. Testing the GUI can be done before a future push/merge, too. I'd say we help Maurice with this initial merge now and take notes of what needs to be improved for the next version.

HenkMutsaerts (approved):
Fine for now, stays in separate folder and we will perform alpha and beta tests before giving this to users. Thanks & kudos to @MauricePasternak for this great piece of work!

Issue #121, opened 2020-09-03
Save and Load x.mat for QC output after each ASL submodule


jan-petr:
* Load and save the x.mat to keep track of Output before and after each wrapper.

* Make sure that no spurious x is added in either of the submodules that would not be worthy of saving (in most submodules, the x modifications were discarded and not returned, so there can be some weird thins - although in some modules, this is returned).

* Move some of the QC things to the correct submodule.

HenkMutsaerts:
You fixed this right, should we move this issue to milestone 1.2.1 and close it?

jan-petr:
No. I have fixed it for one ASL submodule only, but not for all the other submodules.

As written in the description - the other submodules only modify x inside, but throw away all the changes. So I have to write loading and saving of x before/after each submodule. But at the same time, I have to verify that nothing important is changed inside the x. Right now, the x-struct changes are always discarded, so we can't just output and save it without checking.

So:
1) This needs some time, but is not at all urgent.
2) After done, needs proper testing to avoid introducing bugs - i.e. not good to do it in a hotfix.

jan-petr:
### Reworked ToDo after discussion in #112 and #122:

* Load and save the QC*.json to keep track of Output before and after each submodule wrapper in the ASL module. Replaces the loading and saving of x.mat.
* x.mat is currently saved for xASL_wrp_registerASL, but this can be removed and replaced by QC*.json, as the x.mat was saved solely for the purpose of QC and not for other things in the x.mat.
* At the start of ASL module - next to the output cleaning in x.mat, make sure that the QC*json is cleaned and created as well.
* Move some of the QC things to the correct submodule.
* The question still is, if the same output is saved to x.mat in parallel, or kept only in QC*.json?

HenkMutsaerts:
* OK, then we only keep QC*.json
* we open and close QC*.json each time, and don't keep track what was opened and closed. Each submodule should be "on its own" in this regard.
* In the derivative stuff that now loads the x.mat (e.g. PDF), we should convert to QC*.json as well.
* Can become confusing that part of the output is moved to json, but the images are still inside the x.mat 
(there are both x.Output.(ScanType).* and x.Output_im.(ScanType).*

HenkMutsaerts:
## Copy of my comment before closing #112
>I agree to write the QC*.json instead of the x.mat. That will make stuff much easier, as the changes in the x.mat are almost always thrown away after each submodule.

[...]

PR #122, opened 2020-09-04
Feature #112 output tc for asl2t1w

### LINKED ISSUE #112: Saving Tanimoto Coefficient of the final ASL-T1w registration
Will be saved as one of the QC parameters.
Henk: Name the structure and field for saving this.
Jan: Implement.

jan-petr:
* Writing the Tanimoto coefficient for ASL2T1 registration to the x-struct and to the QC output.
* QC Output is cleaned at the start of the ASL module and not only in the QC submodule.
* Saving TC as registration statistics in the Population module

Close #112

### DIFF
--- a/Functions/xASL_adm_LoadX.m
+++ b/Functions/xASL_adm_LoadX.m
@@ -1,10 +1,10 @@
 function [x, IsLoaded] = xASL_adm_LoadX(x, Path_xASL, bOverwrite)
 %xASL_adm_LoadX Load x.mat file that keeps track of QC output
-% FORMAT: [x[, IsLoaded]] = xASL_adm_LoadX(x, Path_xASL[, bOverwrite])
+% FORMAT: [x[, IsLoaded]] = xASL_adm_LoadX(x[, Path_xASL, bOverwrite])
 %
 % INPUT:
 %   x           - structure containing fields with all information required to run this submodule (REQUIRED)
-%   Path_xASL   - path to the x.mat that contains the QC output (REQUIRED)
+%   Path_xASL   - path to the x.mat that contains the QC output (OPTIONAL, DEFAULT = x.SUBJECTDIR/x.mat)
 %   bOverwrite  - true to overwrite the current x structure with the
 %                 x.Output & x.Output_im from the x.mat (OPTIONAL, DEFAULT=false)
 %
@@ -24,13 +24,21 @@ function [x, IsLoaded] = xASL_adm_LoadX(x, Path_xASL, bOverwrite)
 % EXAMPLE: [x, IsLoaded] = xASL_adm_LoadX(x, fullfile(x.D.ROOT,'x.mat'), true);
 %
 % __________________________________
-% Copyright (C) 2015-2019 ExploreASL
+% Copyright (C) 2015-2020 ExploreASL
 
 
 
 %% -------------------------------------
 %  Admin
 IsLoaded = false; % default
+
+if nargin<2 || isempty(Path_xASL)
+	if ~isfield(x,'SUBJECTDIR')
+		error('Path_xASL not provided and x.SUBJECTDIR not defined.');
+	end
+	Path_xASL = fullfile(x.SUBJECTDIR,'x.mat');
+end
+
 if nargin<3 || isempty(bOverwrite)
     bOverwrite = false;
 end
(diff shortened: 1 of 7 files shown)


HenkMutsaerts (changes requested):
## xASL_wrp_VisualQC_ASL.m
I would keep the cleaning here. Each submodule should load and save x.mat on its own.

## xASL_module_ASL
I would remove the cleaning of x.Output & x.Output_im. This will not work nicely if you do a rerun and skip certain submodules.

## xASL_stat_GetRegistrationStatistics
* Replace “CSV” to “TSV” throughout
* add comments (e.g. that you initialize with NaNs)
* Line 63: Here you load the QC_collection_<SUBJECT_ID>.json, and include all parameters, not only the TC fields? Perhaps it is easier to load the x.mat’s and insert any fields that include “TC”? E.g.:
if regexp(CurrentField, ‘^TC.*Perc$’

jan-petr:
> ## xASL_wrp_VisualQC_ASL.m
> I would keep the cleaning here. Each submodule should load and save x.mat on its own.

After the Whatsapp discussion - the agreement is to keep it as it is now?
 
> ## xASL_module_ASL
> I would remove the cleaning of x.Output & x.Output_im. This will not work nicely if you do a rerun and skip certain submodules.

We need to clean it somewhere - keeping it on that line which had other cleaning stuff was the most reasonable to me. Do we keep it or change it?

> ## xASL_stat_GetRegistrationStatistics
> * Replace “CSV” to “TSV” throughout

Done.

> * add comments (e.g. that you initialize with NaNs)

Done.

> * Line 63: Here you load the QC_collection_<SUBJECT_ID>.json, and include all parameters, not only the TC fields? Perhaps it is easier to load the x.mat’s and insert any fields that include “TC”? E.g.:
>   if regexp(CurrentField, ‘^TC.*Perc$’

I can use x.mat instead of the QC.json. But I find that using QC.json is better and safer as that should include the final list of QC values and nothing else = a small file. While x.mat is a 1MB zipped file with a thousand of things. So the JSON is much better for extracting the same values.

I don't take all parameters but only those listed at line 38 (practice taken from the other xASL_stat_Get*.m from the Population module.

The regexp stuff won't work, because the list of fields to include might differ between subjects - so it would mess up the columns and we would have to create a dynamic list of columns and update at each step. This would make things much less flexible. So including a predefined fields is much easier for now.

HenkMutsaerts:
## xASL_wrp_VisualQC_ASL & xASL_module_ASL
Yes, let's keep this for now. In the near future, we should revamp it in such a way that each submodule loads and saves.

## xASL_stat_GetRegistrationStatistics - use QC_*.json instead of x.mat
Good point. Currently, this is not yet completely proof, as the QC output of xASL_wrp_RegisterASL will be saved immediately to x.mat and only much later to QC*.json. So your solution is good, but would need a bypass altogether of the x.mat. So for QC: instead of loading & writing the x.mat, we should load & write the QC*.json, do you agree?

HenkMutsaerts (approved):
See comments above, if you agree, we can add these ideas as new issues, and merge this PR.

jan-petr:
Yes, I agree, this has already an issue #121

I agree to write the QC*.json instead of the x.mat. That will make stuff much easier, as the changes in the x.mat are almost always thrown away after each submodule. 

There's only one question then - should this QC*.json loading and saving happen inside the submodule. Or should the submodule output a QC structure, which is then added to the QC*.json. I support the second option and I have wrote it like that to the #121 - can you please verify that there while we have it fresh in our heads, so that we reach an agreement that I can later implement?

HenkMutsaerts:
>I agree to write the QC*.json instead of the x.mat. That will make stuff much easier, as the changes in the x.mat are almost always thrown away after each submodule.
Correction: the changes in x.mat are not thrown away, but the changes in x, right? At least that's the idea. Anyway, I agree changing QC*.json is more specific.
>There's only one question then - should this QC*.json loading and saving happen inside the submodule. Or should the submodule output a QC structure, which is then added to the QC*.json. I support the second option and I have wrote it like that to the #121 - can you please verify that there while we have it fresh in our heads, so that we reach an agreement that I can later implement?
I agree with you here. This should be standardized over all submodules.
Will merge this now!

Issue #20, opened 2020-07-05
Implement BASIL


HenkMutsaerts:
# The following steps are needed as pre-work for BASIL
1. BASIL prefers masks to limit the voxels on which to run the modeling, but ExploreASL currently masks after the quantification. The masking wrapper should therefore be moved to before the quantification wrapper
    1. CBF will be replaced by PWI in the xASL_wrp_CreateAnalysisMask, after which this part can be moved in front of the Quantification
2. For better modeling BASIL requires the full 4D timeseries. Currently the resample wrapper does the averaging, unless there is a PWI4D option that states to keep the 4D timeseries. Instead, the averaging should move from the resample wrapper to the quantification wrapper, to ensure that the quantification wrapper always has access to the full 4D timeseries. The control-label subtraction can stay in the resample wrapper. The PWI4D flag can stay, but would be applied in the quantification wrapper only, where it would disable deleting of the temporary time series (this is for advanced users that want to keep the 4D time series for modelling purposes).
    1. Keep subtraction in xASL_wrp_Resample
    2. Move averaging to xASL_wrp_Quantify
    3. Remove PWI4D option from xASL_wrp_Resample
    4. Keep PWI4D option in xASL_wrp_Quantify

HenkMutsaerts:
See also discussion with Martin Craig [HERE](https://docs.google.com/document/d/13QfpVda2JB0kxRAUdZRQbrMiuG7n9LB7q6lncVIPeF0/edit?usp=sharing). 
Thanks @ Martin Craig *martin.craig@eng.ox.ac.uk*

HenkMutsaerts:
## 1 - Switch order quantification & mask creation
* xASL_module_ASL: switch order xASL_wrp_Quantify (now last) & xASL_wrp_CreateAnalysisMask (now first)
* Move creation of standard space masked image from xASL_wrp_CreateAnalysisMask to xASL_wrp_Quantify

## 1 - Replace CBF by PWI in xASL_wrp_CreateAnalysisMask (as CBF is not yet quantified now)
* xASL_wrp_CreateAnalysisMask -> replace 'CBF' by 'PWI'
* xASL_im_MaskNegativeVascularSignal -> replace 'CBF' by 'PWI'
* xASL_im_MaskPeakVascularSignal -> replace 'CBF' input by 'PWI' input
* Cosmetic: use 1. instead of 1)

HenkMutsaerts:
[...]

PR #124, opened 2020-09-06
Feature #20 basil implementation

HenkMutsaerts:
* Still need to test this
* Still need to implement changes from hotfix1.2.1
* Otherwise this should be good to implement BASIL

The following steps were applied:
## The following steps are needed as pre-work for BASIL
1. BASIL prefers masks to limit the voxels on which to run the modeling, but ExploreASL currently masks after the quantification. The masking wrapper should therefore be moved to before the quantification wrapper
    1. CBF will be replaced by PWI in the xASL_wrp_CreateAnalysisMask, after which this part can be moved in front of the Quantification
2. For better modeling BASIL requires the full 4D timeseries. Currently the resample wrapper does the averaging, unless there is a PWI4D option that states to keep the 4D timeseries. Instead, the averaging should move from the resample wrapper to the quantification wrapper, to ensure that the quantification wrapper always has access to the full 4D timeseries. The control-label subtraction can stay in the resample wrapper. The PWI4D flag can stay, but would be applied in the quantification wrapper only, where it would disable deleting of the temporary time series (this is for advanced users that want to keep the 4D time series for modelling purposes).
    1. Keep subtraction in xASL_wrp_Resample
    2. Move averaging to xASL_wrp_Quantify
    3. Remove PWI4D option from xASL_wrp_Resample
    4. Keep PWI4D option in xASL_wrp_Quantify

### DIFF
--- a/Functions/xASL_im_MaskNegativeVascularSignal.m
+++ b/Functions/xASL_im_MaskNegativeVascularSignal.m
@@ -1,20 +1,20 @@
-function [NegativeMask, TreatedCBF] = xASL_im_MaskNegativeVascularSignal(x, IsSpace)
+function [NegativeMask, TreatedPWI] = xASL_im_MaskNegativeVascularSignal(x, IsSpace)
 %xASL_im_MaskNegativeVascularSignal Segment ASL image clusters with
 %significantly negative signal
 %
-% FORMAT: [NegativeMask, TreatedCBF] = xASL_quant_DetectNegativeVascularSignal(x)
+% FORMAT: [NegativeMask, TreatedPWI] = xASL_quant_DetectNegativeVascularSignal(x)
 %
 % INPUT:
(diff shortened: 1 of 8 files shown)


MichaelStritt (approved):
Hey, just looked through all the changes and couldn't find major Matlab or pipeline problems. But I'm not an expert in if these changes could crash anything, so maybe Jan should review as well. Have you tested it already?

HenkMutsaerts:
I will 
1) test now
2) merge
3) notify Martin Craig so he can continue implementing BASIL!

HenkMutsaerts:
Found several errors when testing this, will fix them later.

jan-petr:
I will fix several more bugs in the develop branch and then we will need to merge all the hotfix and develop changes to this branch to allow testing.

I will create a new feature branch, cherry pick all the commits to it and delete the old branch and this PR.

HenkMutsaerts:
I cherry-picked the commits for #20 to the new #20 branch, but we cannot reopen this PR, so have to create a new one, I will first test the this branch.

PR #135, opened 2020-09-17
#119 Fixing the unlocking of Structural module

### LINKED ISSUE #119: Structural Lock files are not created for the last subject
At least not in the Craniosynostosis study where dummy T1files are created.

jan-petr:
Unlocking it also for the last patient when no T1 is present for the patient.

Close #119


HenkMutsaerts (approved):
Good for now, but this module has at different locations an explicit test for the presence of certain files, without which it will skip the module (see bContinue etc, line 124). We can manage this better in the future, don't you agree?

PR #142, opened 2020-09-21
#119 Fixing the unlocking of Structural module

jan-petr:
Unlocking it also for the last patient when no T1 is present for the patient


jan-petr:
The previous branch was based on the incorrect develop. So the easiest was to delete the branch and recreate both branch and the edit. Henk can you please review it again.

HenkMutsaerts (approved):

PR #144, opened 2020-09-22
#120 Fixes the native space population parallel processing

### LINKED ISSUE #120: Bug in xASL_im_CreateAnalysisMask in native space mode in parallel execution.
Can crash if run in native space mode and in parallel.

The line 170 that creates smoothed version of the templates operates on the original templates directory - creating a conflict when multiple instances are accessing it at once.

This, and potentially other similar parts of the code have to make sure that the file is copied to patient directory and processing is done locally:

Error on line 170:
pathTmpPreSmooth = xASL_im_PreSmooth(x.P.Path_PWI,listMasks{kk},[],x.S.optimFWHM_mm,[],x.P.Path_mean_PWI_Clipped_sn_mat, 1);

jan-petr:
Closes #120 

Does not write temporary files to the shared template folder but only to the individual subjects folder

### DIFF
--- a/Modules/SubModule_Population/xASL_im_CreateAnalysisMask.m
+++ b/Modules/SubModule_Population/xASL_im_CreateAnalysisMask.m
@@ -167,7 +167,9 @@ if x.bNativeSpaceAnalysis
 						switch (listType(kk))
 							case 1
 								% Pre-smooth the mask before downsampling to native space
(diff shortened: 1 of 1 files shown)


HenkMutsaerts (approved):
Looks like too much code that we have to fix when we will fix the native processing in the population module ;) joking

Issue #130, opened 2020-09-13
Improved testing


HenkMutsaerts:
Need to improve the ExploreASL testing, at least allowing serial & parallel testing for all OSes.
@Michael: do we want to add more features, as per what we discussed at our last dev-meeting?

MichaelStritt:
> @michael: do we want to add more features, as per what we discussed at our last dev-meeting?

I think the most important part would be a script which calls all individual unit tests as well as a template for an individual test. We should probably agree on a folder structure for this, too.

MichaelStritt:
Hey @HenkMutsaerts what do you think about a structure like the one in f774ff7 ?

PR #146, opened 2020-09-23
Feature #130 improved testing

MichaelStritt:
Introduce unit testing framework for other developers, so that we can all work on this together.


HenkMutsaerts (approved):
Nice work!

I know that I had added git pulling by default in the test script, but I wouldn’t do this in a unit testing function (or provide the option). You may want to test a certain branch in certain condition.

The JSON loading could be done with our wrapper around the spm_jsonread.

Try to use readable words, what is “val” ;)
Same for “fg”

PDF printing: 600 dpi is a bit high perhaps, if you have many results. 300 should be more than enough

Please avoid using “string” but replace by “char” or something else ;)
There’s no need for this, and it breaks backward compatibility.

Why do you have the “outdated” files? Do they need a cleanup?

Eager to try out the unit testing workflow!

Issue #151, opened 2020-09-24
Minor bugfixes when trying out TopUp


HenkMutsaerts:
xASL_exist: manage Type

xASL_io_SaveNifti: Remove temp files if they exist

xASL_adm_CleanUpBeforeRerun: Skip empty lock folders, avoids deleting folder content!

xASL_bids_Add2ParticipantsTSV: Correct row

HenkMutsaerts:
Apologies, forgot to add one commit

PR #152, opened 2020-09-24
Bug #151 minor bugfixes

HenkMutsaerts:
Found these minor fixes when working with TopUp


jan-petr (approved):
Looks good. Can be merged.

HenkMutsaerts:
Sorry, added the latest commit for you to review

jan-petr (changes requested):
xASL_module_ASL.m 
line 282 - the edits on all three lines have to be reverted.

This loading of x, outputting x by the register function and saving x was there so that we can save the TC for the visual QC - that was a part of a previously approved branch.

In the future, we will not save it through X, but directly to the JSON-QC file, but we are not yet there, so this edit has to be reverted.

If you agree, I can do that, but wanted to reach a consensus on that first :)

jan-petr (approved):

Issue #139, opened 2020-09-18
x fields are case sensitive


jan-petr:
While even if you give x.Vendor = 'Siemens', then it can happen that vendor name is also extracted automatically from the DICOMs as 'SIEMENS'. In the ExploreASL code, this is a problem since the vendor field is case-sensitive - for example on line 157 or xASL_wrp_RegisterASL.m

1. Fix import to convert SIEMENS to Siemens in vendor, and also for all vendors.
2. Fix Load_json parms to make the vendor field case insensitive.
3. Fix also potentially all the code to make it vendor insensitive.

Let Francesco Neri <nericesco@gmail.com> know when we have a fixed version.

HenkMutsaerts:
1. I wouldn't do this conversion, SIEMENS should be fine for now (don't complicate BIDS more than needed :)
2. actually, let's make all fields case insensitive, by regexp(pattern in lowercase, lower())
3. let's not make it vendor insensitive ;)

jan-petr:
Conclusions of a H+J discussion:
1. Let's not alter anything in the import.
2. Let's not alter anything in the load-parms, because that could be saved and so altering the JSON.
3. Let's only make sure that the string comparison is case-insensitive during the comparison.
4. Let's fix this not only for vendor, but also for other fields like sequencetype etc.

jan-petr:
Let's make automaticly extracted things from DICOM case insensitive (x.Vendor, x.Sequence, x.LabelingType, x.readoutDim)

But also our parameters x.M0, x.WMHSegmAlg, x.Seg.Meth will be done case insensitive. Right?

I'll search for 'strcmp(x' and 'regexp(x' and the same with a space before...

HenkMutsaerts:
yes that should be fine!

jan-petr:
xASL_wrp_spm_GLM - line 92, should there be '3D_spiral' or '3D spiral'?

Otherwise, I have edited all the switches, strcmps and regexps for all the quantification and processing choices.

Though I have not edited this for the sets2check.

HenkMutsaerts:
# 3D_spiral is OK
See https://github.com/ExploreASL/ExploreASL/blob/master/DataParTemplate.m:
>% x.Sequence - string containing the sequence used (REQUIRED for ASL)
%            - options:
%              - '3D_spiral'
%              - '3D_GRASE'
%              - '2D_EPI'

# strcmpi & regexpi
[...]

PR #155, opened 2020-09-25
#139 String parameters made case insensitive

jan-petr:


HenkMutsaerts (approved):
# 3D_spiral is OK
See https://github.com/ExploreASL/ExploreASL/blob/master/DataParTemplate.m:
>% x.Sequence - string containing the sequence used (REQUIRED for ASL)
%            - options:
%              - '3D_spiral'
%              - '3D_GRASE'
%              - '2D_EPI'

# strcmpi & regexpi
I didn't know these, I used lower() inside strcmp or regexp on multiple locations, but these are better indeed.
Also backwards compatible ;)

# Additional enters/whitespace
You added a lot of extra empty lines as well with this last commit, was this on purpose? Each new empty line can cause a potential conflict, just to avoid too much unnecessary work ;)

Issue #106, opened 2020-08-24
Improvement creation average maps for territorial ASL


HenkMutsaerts:
We need to improve a few details that will help to create average maps

PR #157, opened 2020-09-26
Feature #106 create average maps

HenkMutsaerts:
Just for you to review, not to merge yet, as I may add more functionality.
Or I could merge it already, and keep this feature open?
This is for the territory atlas project with Reinoud Bokkers.


jan-petr (approved):
I made a minimal edit in the header. Otherwise looks good. Up to you when to merge, but I would do it rather fast and start testing and leave features for later.

Issue #123, opened 2020-09-05
030_FLAIR_BiasfieldCorrection.status not created during re-run


MauricePasternak:
Minor compatibility issue with the GUI, which expects 030_FLAIR_BiasfieldCorrection.status to be created during a re-run if the user deletes all Structural Module status files. ExploreASL doesn't create this status file if the dataset is re-run:  

    WMH segmentation already existing, or FLAIR.nii missing, skipping FLAIR biasfield correction

Around the time that message is made, could 030_FLAIR_BiasfieldCorrection.status please be created [EDIT: only if WHM_SEGM.nii is detected] so that the GUI doesn't interpret the process as an error?

In other words, to re-create this:
1) Process the subject in TestDataSet as usual.
2) Delete all locked folders and .status files
3) Re-run
4) Observe that 030_FLAIR_BiasfieldCorrection.status is not made during the re-run.

@MichaelStritt something to keep in mind as you are testing the GUI, especially on the subject with a FLAIR image in TestDataSet

MichaelStritt:
> In other words, to re-create this:
>     1. Process the subject in TestDataSet as usual.
>     2. Delete all locked folders and .status files
>     3. Re-run
>     4. Observe that 030_FLAIR_BiasfieldCorrection.status is not made during the re-run.
> 
> @MichaelStritt something to keep in mind as you are testing the GUI, especially on the subject with a FLAIR image in TestDataSet

Tested it with Henks recent fix a minute ago. Seems to work fine now.

PR #158, opened 2020-09-26
#123 CreateMissingFLAIRStatus

HenkMutsaerts:
Ensure that FLAIR status files are also created without FLAIRs or WMHs, to avoid confusing external progress trackers


MichaelStritt:
Additionally tested it without FLAIR right now. Seems to work as intended.

MichaelStritt (approved):

Issue #143, opened 2020-09-21
GE images don't output VBA mask in Population module


HenkMutsaerts:
Perhaps the absence of a susceptibility mask will skip the creation of a VBA mask.

MauricePasternak:
I think I found the issue and it might be present in more than just GE due to the nature of the error...

xASL_module_Population breakpoint line 81.  Files such as 'MaskSusceptibility_bs-mean.nii' or 'FoV_bs-mean_Unmasked.nii' do not exist.

Instead, they are named, for example, FoV_n23_bs-mean_Unmasked.nii

HenkMutsaerts:
* xASL_vis_Imwrite: Manage empty masks:
  * Cosmetics; *Warning instead of crashing when input images are empty or contain NaNs only

* xASL_im_CreateAnalysisMask: Manage susceptibility mask for 3D spiral: 
  * If susceptibility mask is missing for 3D spiral, this is now fine. Create a dummy mask then, and within the complete function the assumption of the presence of a susceptibility mask is now changed accordingly

*xASL_adm_GetFileList: Warning when too few or many files found:
  * When too many or too few files or folders were found, this threw an error, which is now changed to a warning instead

* Load FoV masks using regular expression
* xASL_im_CreateAnalysisMasks: Load masks by regular expression:
  * These masks were previously saved without specifying the number of images averaged in the filename. Now this is mentioned in the filename, the filename of the population-based masks changes with the number of images that were averaged. These new regular expressions should deal with this

HenkMutsaerts:
## **MAURICE PASTERNAK COMMENTED:**

I will re-test with a larger sample size, but from the mini-batch of 4 subjects (2 with ASL and 2 without), the pipeline fails due to the following condition in xASL_wrp_CreatePopulationTemplates starting @ line 108:

```
if x.nSubjectsSessions<6
    % With too small datasets, created templated won't be reliable
    fprintf('%s\n',['Too few images (' num2str(x.nSubjectsSessions) ') for creating templates/parametric images, skipping...']);
    return;
end

Size1 = sum(x.WBmask(:));

x.D.TemplatesStudyDir = fullfile(x.D.PopDir, 'Templates');
xASL_adm_CreateDir(x.D.TemplatesStudyDir);
```

So the snowball effect is this:

1) analysis/Population/Templates is never created (due to early cancellation at low subject number)

[...]

PR #161, opened 2020-09-27
Bug #143 masks missing

HenkMutsaerts:
See the issue, perhaps not the nicest programming but should be fine for now


jan-petr (approved):

MauricePasternak (approved):
Works. I took the liberty to make the tiniest changes to lines 56-59 in xASL_im_CreateAnalysisMask.m , as the filenames were not matched.

Note that the filepaths by the time the code reaches those lines is, for example:

PV_pGM_n23_bs-mean.nii.gz

So, I added in a pattern to account for "n_\d+" and the "\.gz" extension which may or may not be present. With this addition the module proceeds to completion and VBA_mask_final.nii is made

jan-petr:
@MauricePasternak thanks for the edit! BTW: '.gz' should be handled by the GetFileList function automatically. Otherwise - good catch.

Issue #162, opened 2020-09-27
TestDataSet has bNativeSpaceProcessing==true


HenkMutsaerts:
@jan-petr why does the test dataset native space processing on true?
I know that this is an option that we should offer by default, but this option actually crashes the population module for me, it is less flexible than bNativeSpaceProcessing==false. So I would suggest to put this in the TestDataSet after we have corrected this option, that it has the "modularity" that it needs to have :)

jan-petr:
I agree. But can you put an issue for the crash - maybe it will be easy to fix this right away.

HenkMutsaerts:
It works fine, though it seems to change the TotalGM & DeepWM maps, so this won't work when ExploreASL is read-only (as it should). Let's try to fix this in 1.4.0, it really has priority because we promised this in the paper ;)

Now removed the bNativeSpaceProcessing in the TestDataSet

jan-petr:
@HenkMutsaerts I agree. I have added this comment to the issue about nativeSpaceProcessing revamp. I will thus review, approve, merge and close.

jan-petr:
@HenkMutsaerts Can you also check the conflicts (use the webeditor above). It seems to me that the develop branch is more correct in all cases - but you're be in a better position to confirm this because you just did all these edits (just to avoid mistakes because Maurice was also editing those).

PR #167, opened 2020-09-29
Bug #162 b native space processing in test data set

jan-petr:
Henk, I agree with the commit and approve to merge the branch. But when I create the PR, you have to approve it and I can't...


jan-petr:
@HenkMutsaerts Can you also check the conflicts (use the webeditor above). It seems to me that the develop branch is more correct in all cases - but you're be in a better position to confirm this because you just did all these edits (just to avoid mistakes because Maurice was also editing those).

PR #183, opened 2020-10-14
#182 WM overlay for QC with contours and template WM.

### LINKED ISSUE #182: Visualization of CBF+WM - contour and template WM
For the visualization of the CBF alignment with the WM mask in the standard space, add two extra options:

x.bVisualQCCBFvsWMTemplate - if on, then produces an additional file that use the template WM instead of the individual WM - to be used when the WM segmentation is questionable.
x.bVisualQCCBFvsWMContour - if on, then produces the the WM as a contour, not a full ROI - 1 pixel thickness (contour calculated per slice basis).

These options work on top of the standard output and produce files with different names - so that it could be recognized. Once this functionality is done - we can discuss all the names and the exact logistics.

jan-petr:
Closes #182 
* Overlays in ASLCheck can now use also the template version of GM and WM and not only individual segmentations.
* Overlays in ASLCheck can now be also contours and not full ROIs.

### DIFF
--- a/DataParTemplate.m
+++ b/DataParTemplate.m
@@ -301,10 +301,10 @@ function x = DataParTemplate(x)
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
 % VISUALIZATION PARAMETERS
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
-% x.bVisualQCCBFvsGMWMTemplate - if == 1, uses the template GM and WM to make visualization overlays with CBF and other derived maps
+% x.vis.bVisualQCCBFvsGMWMTemplate - if == 1, uses the template GM and WM to make visualization overlays with CBF and other derived maps
 %                              if == 0, uses the individual GM and WM maps of the subject to make visualizations
 %                              (OPTIONAL, DEFAULT = 0)
-% x.bVisualQCCBFvsGMWMContour -  if == 1, then produces the GM and WM as a contour in the overlay with CBF and derived maps, not a full ROI - 1 pixel thickness (contour calculated per slice basis).
+% x.vis.bVisualQCCBFvsGMWMContour -  if == 1, then produces the GM and WM as a contour in the overlay with CBF and derived maps, not a full ROI - 1 pixel thickness (contour calculated per slice basis).
 %                              if == 0, then a standard full GM or WM ROI is used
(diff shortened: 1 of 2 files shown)


HenkMutsaerts (changes requested):
* The DataParTemplate.m needs to have a clearer description
* The contour visualization function should not be inside the wrapper but in a function.
Adding it as option to xASL_vis_CreateVisualFig would be best

DataParTemplate.m
```diff
@@ -236,6 +236,15 @@
 % 							   - 1 = enabled
 % 							   - 0 = disabled
 %
+% -----------------------------------------------------------------------------------------------------------------------------------------------------
+% VISUALIZATION PARAMETERS
+% -----------------------------------------------------------------------------------------------------------------------------------------------------
+% x.bVisualQCCBFvsGMWMTemplate - if == 1, uses the template GM and WM to make visualization overlays with CBF and other derived maps
+%                              if == 0, uses the individual GM and WM maps of the subject to make visualizations
+%                              (OPTIONAL, DEFAULT = 0)
+% x.bVisualQCCBFvsGMWMContour -  if == 1, then produces the WM and GM as a contour in the overlay with CBF and derived maps, not a full ROI - 1 pixel thickness (contour calculated per slice basis).
```
HenkMutsaerts, line 245: GM and WM is switched compared to above (WM and GM). Perhaps better to put these booleans in `x.vis.` or `x.stats.`

Modules/SubModule_ASL/xASL_wrp_VisualQC_ASL.m
```diff
@@ -286,6 +334,14 @@ function xASL_wrp_VisualQC_ASL(x)
     end
 end
 
+% Delete the temporary contours if they exist
+if exist(fullfile(x.SESSIONDIR,'pGMC.nii'),'file')
```
HenkMutsaerts, line 338: `xASL_delete` already checks for existance, so you can remove the `if exist` lines

HenkMutsaerts (changes requested):

HenkMutsaerts:
@jan-petr I fixed my last requests, so you're good to go

HenkMutsaerts (approved):

PR #188, opened 2020-10-19
Feature #20 basil implementation

HenkMutsaerts:


HenkMutsaerts:
I have now finished the pipeline preparations for BASIL implementation and tested these changes with `xASL_qc_TestExploreASL` on macOS Matlab 2016b. With thanks to martin.craig@eng.ox.ac.uk, check our discussion [HERE](https://docs.google.com/document/d/13QfpVda2JB0kxRAUdZRQbrMiuG7n9LB7q6lncVIPeF0/edit?usp=sharing). 

### Changes
>1. `070_Quantification` & `080_CreateAnalysisMask` are now switched to: `070_CreateAnalysisMask` & `080_Quantification`.

* BASIL needs a mask before quantification, to avoid running the full model on each voxel. 
* This order change should not affect the reproducibility.
* This order reversal is now appropriately tested and all dependencies fixed.
* In the respective masking functions, the CBF variables are replaced by PWI variables (as CBF is created only in the subsequent module)
* `xASL_wrp_Quantify`: now has an additional final step to create standard space `qCBF_masked`
* Definitions of `qCBF_masked`, MaskVascular and MaskSusceptibility are now moved to `xASL_init_FileSystem`

>2. xASL_wrp_Resample now always saves both PWI4D and PWI.

* The previous situation was to only save PWI4D if this was requested by `x.SavePWI4D`. 
* This option is now removed and PWI4D is always saved, both in native and standard space. These files are removed at the end of `xASL_wrp_VisualQC_ASL`, as this function can use PWI4D for QC parameters.
* `x.SaveCBF4D` is created, to allow keeping quantified time series of paired subtractions.

>3. While at it, I revamped `xASL_wrp_Resample`
* Layout is now improved for readability
* Resampling of motion correction now has 4th-degree B-spline resampling. This is the same as the warping to standard space and is the default of SPM. We previously used a 2nd-degree B-spline here, which is a good compromise for images that are not used for quantification (e.g. images for registration).
* Bilateral filter is now disabled as this needs testing


>4. While at it, I revamped some other minor things:

* `xASL_adm_LoadParms`: `x.M0=no_background_suppression` now issues a warning (should be `x.M0=UseControlAsM0`)
* `xASL_im_Presmooth`: now loads image matrices as single precision instead of double, perhaps this speeds up the `convn` function
* `xASL_im_Upsample`: now issues an error when input image has too many dimensions
* `xASL_qc_CreatePDF`: minor bugfix to allow printing booleans

MichaelStritt (approved):

Functions/xASL_im_MaskPeakVascularSignal.m
```diff
@@ -70,8 +75,8 @@
 		warning('No usable M0 image, using CBF only');
 		M0_im = 1;
 	else
-		M0_im = xASL_io_Nifti2Im(Path_M0);
-		if ~(min(size(TreatedCBF)==size(M0_im))) % if CBF & M0 images arent the same size (which is why rM0 should be the input path in case of different sizes)
+		M0_im = xASL_io_Nifti2Im(Path_M0); % allow both a path or image matrix
+		if ~(min(size(PWI)==size(M0_im))) % if CBF & M0 images arent the same size (which is why rM0 should be the input path in case of different sizes)
```
KoenBaas, line 79: Small detail: replace CBF with PWI in comments?

KoenBaas (approved):

jan-petr:
martin.craig1@gmail.com has to approve

Functions/xASL_im_MaskPeakVascularSignal.m
jan-petr: Done.

jan-petr (approved):
Testing shows no errors and consistent results.

PR #221, opened 2020-11-11
Feature #193 compare bid sfolders

### LINKED ISSUE #193: Comparing BIDS folders
Function that compares two BIDS folders with several subfolders and studies and prints the differences. Specifically:
* It goes through all folders and subfolders in the first path and checks if there's an existing counterpart.
* It checks as well all possible extra files and directories in the second path if they are in the first path and reports them.
* My idea was that if it finds that two directories exist in both paths, then it iteratively calls the same function to check if their content is the same as well - that should make the programming easier.
* It checks if the list of files exist in both directories and reports the missing/extra files.
* If both files exist and are JSONs or NIIs or TSV, then loads them and compare the content.
* It can probably just output differences on the screen, but you might think of a more smart aggregating of the outputs in the function output so that they can be gathered through the iterative call and a sum returned by the parent function.

Take the BIDS as the use-case - i.e. focus on things that are in BIDS (NII, JSON, TSV), but probably no need to check the other file types. But you don't need to have any really BIDS specific functionality - it doesn't need to check the quality of the data, just the presence and similarity - i.e. doesn't need to interpret BIDS>

jan-petr:
Closes #193 

* Compares two directories and their subdirectories.
* Reports missing files and directories in each of the paths.
* Compares content of JSON, NII, TXT, TSV and CSV files and reports differences.

### DIFF
--- a/Development/BIDS/xASL_bids_compareStructures.m
+++ b/Development/BIDS/xASL_bids_compareStructures.m
@@ -1,12 +1,13 @@
-function [identical,results] = xASL_bids_compareStructures(pathDatasetA,pathDatasetB,printReport)
+function [identical,results] = xASL_bids_compareStructures(pathDatasetA,pathDatasetB,printReport,thresh)
 %xASL_bids_compareStructures Function that compares two BIDS folders with several subfolders and studies and prints the differences.
 %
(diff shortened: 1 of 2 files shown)


MichaelStritt (approved):
Looks good 👍

Issue #223, opened 2020-11-11
Phoenix Protocol Reader


MichaelStritt:
- Implementation of a function to read and a function to analyze the Siemens phoenix protocol (private tag 0x29,0x1020)

HenkMutsaerts:
@MichaelStritt can you please provide a description to this issue ;)

Also, we should probably have a similar issue for reading the Philips XX files (when Sandeep is ready with this)

jan-petr:
Yes, issue 194 lists all these advanced BIDS import options (including Philips XX), we have isolated the Phoenix thing because we can easily do it already...

MichaelStritt:
Hey @jan-petr, I added the option for DCMTK here 1d8050a but can you show me the code snipped how to do this?

MichaelStritt:
- Issue: There are two parameters which include the word **lrepetitions** in 3D GRASE PCASL datasets
    - Possible solution: Search parameter list only for exactly matching parameters instead of using include
- We could still improve the logic for the interpretation of the **parameters** regarding the **xasl** struct

jan-petr:
Copying my review also here (as we have agree previously to have all copied also in the issue branch):

Looks very good. Some minor changes to do:

- Explain in the header what is the alternative to bUseDCMTK - i.e. Python
- Why is the phoenixProtocol output a cell and not a string?
- I already perform the reading of the Phoenix in my import. It would be thus good to split the reading itself and the parsing of the protocol to two independent functions. I mean that you can read+parse in a single function, but I need the parsing also separately.
- Please be more verbose in the description of the function so that the inputs/outputs/function is more clear.

Smart to have the ASL and 'raw parameter' output - I guess there's still a plenty of work ahead of us.

jan-petr:
Copying the answer from Michael also here (sorry, my fault not starting the discussion here):

Hey @jan-petr, I tried to fix as much as possible in e27e33e.

> Explain in the header what is the alternative to bUseDCMTK - i.e. Python

> Please be more verbose in the description of the function so that the inputs/outputs/function is more clear.

The header description should be a bit better now.

> Why is the phoenixProtocol output a cell and not a string?

In Matlab I personally find it easier to work with cell arrays. I added a separate raw output now 👍 

> I already perform the reading of the Phoenix in my import. It would be thus good to split the reading itself and the parsing of the protocol to two independent functions. I mean that you can read+parse in a single function, but I need the parsing also separately.

[...]

PR #230, opened 2020-11-22
Feature #223 phoenix protocol reader

MichaelStritt:
1. We could improve the logic
    - Interpret **parameters** for **xasl** structure variables
2. We could add specific PCASL parameters
    - An exemplary PCASL phoenix protocol is in the google drive document

Otherwise I think it seems fine to be merged 👍


MichaelStritt:
Hey @jan-petr, I've read your e-mail and added the additional parameters with 4ebfd3a. 👍 

Regarding the interpretation:
- **xAslulMode:** maybe we can get this one. The GSP table says that if this is 4, we should have Q2TIPS. In some other cases we have 2, but I'm not sure what technique was used there.
- alTI0 and alTI1: The GSP table uses these to assess if the dataset is a normal ASL or a PseudoM0. Maybe we can use this to find out if we have an M0 scan? Maybe the **M0Mode** will help here, too?

I'm moving the code to the Functions folder now.

jan-petr (changes requested):
Looks very good. Some minor changes to do:

- Explain in the header what is the alternative to bUseDCMTK - i.e. Python
- Why is the phoenixProtocol output a cell and not a string?
- I already perform the reading of the Phoenix in my import. It would be thus good to split the reading itself and the parsing of the protocol to two independent functions. I mean that you can read+parse in a single function, but I need the parsing also separately.
- Please be more verbose in the description of the function so that the inputs/outputs/function is more clear.

Smart to have the ASL and 'raw parameter' output - I guess there's still a plenty of work ahead of us.

MichaelStritt:
Hey @jan-petr, I tried to fix as much as possible in e27e33e.

> Explain in the header what is the alternative to bUseDCMTK - i.e. Python

> Please be more verbose in the description of the function so that the inputs/outputs/function is more clear.

The header description should be a bit better now.

> Why is the phoenixProtocol output a cell and not a string?

In Matlab I personally find it easier to work with cell arrays. I added a separate raw output now 👍 

> I already perform the reading of the Phoenix in my import. It would be thus good to split the reading itself and the parsing of the protocol to two independent functions. I mean that you can read+parse in a single function, but I need the parsing also separately.

Are you thinking of a separate script or just a separate function? I implemented a separate function, but I'd be happy to change it if you were thinking of something else.

jan-petr (approved):
OK

Issue #35, opened 2020-07-11
Calculation background suppression efficiency for pseudo-M0


HenkMutsaerts:
When no M0 is obtained, and the control image needs to be used as pseudo-M0, the suppression efficiency needs to be calculated. For 2D EPI this is slice-wise. When the option "UseControlAsM0" is selected, and background suppression is on, then this parameter - that should be a vector for 2D multislice - should be checked and applied when the pseudo-M0 is created.

HenkMutsaerts:
## General
* In practice, only some earlier Siemens 3D GRASE and Philips 2D EPI sequences have background suppression but miss a separate M0 image;
* For both: we set "M0" : "UseControlAsM0", which will create a "pseudo-M0" from the (mean) control image (of course throwing an error if this doesn't exist, we need the raw control-label intermediate images to do this correction);
For 3D: compute a single correction value and apply this to the "pseudo-M0"
For 2D: compute a slice-wise correction value (i.e. adding the SliceReadoutTime iteratively) and apply these to the "pseudo-M0" slices

## For Philips 2D EPI
Usually 2 background suppression pulses are played, that are specified in the examcard/protocol text file. From these it should be easy to compute its effect, but we need to compute this for each slice.

## For Siemens 3D GRASE (quoting David Thomas):
>The way Siemens calculates the background suppression pulse timings is using equation [1] in Matthias' 2005 paper (DOI 10.1002/mrm.20580). For this, you choose a pair of T1 values that you want to suppress: T1opt and 2*T1opt. For 3T, I think they use a T1opt value of 0.7s, so then T1opt represents WM and 2*T1opt is GM. With those two T1 values, you can then use eq [1] to calculate the timing of 2 BS pulses for any given TI. In fact, Siemens choose to null the T1opt and 2*T1opt signals 100ms before readout (i.e. at TI-0.1s), to avoid any issues with signal cancellation due to negative (pre-null point) signals.
>
[...]

PR #233, opened 2020-11-25
Feature #35 b sup calculation

HenkMutsaerts:


HenkMutsaerts:
For Yeva (Jan's MSc student) to test the background suppression correction without smoothing:

If you avoid adding `x.M0_conventionalProcessing` to your DataPar.m, `xASL_wrp_ProcessM0` will copy the standard space M0 after quantification but before masking/smoothing.

MichaelStritt (changes requested):
Finally found some time to check this 👍 

The changes from **BackGrSupprPulses** to **BackgroundSuppressionNumberPulses** are really straight forward and seem to be fine to me :)

The changes in the **io_mTRIAL** header are not necessary. I removed this part from the header and the whole script is going to change as soon as we move to BIDS anyway. This just leads to a conflict right now.

Regarding the header in **xASL_quant_M0**:

I know you didn't change that much there, it's just that maybe we should write down which fields are added in the x struct. 
So some information in the _OUTPUT:_ part, besides that the x struct comes out, would be nice. Would make it easier for reviewers to test it, too.

Additionally, every xASL header need an _EXAMPLE:_ part. If we don't have an example we should just use `EXAMPLE: n/a.` This helps the **docu crawler.**

I manually tested **xASL_quant_M0** and especially **xASL_quant_RevertBsupFxControl**. I couldn't find/produce a nice test data set for it, but as far as I can see it right now, we don't get any errors :)

```
[M0IMtest] = xASL_quant_M0(M0IM, x);
x.M0_usesASLtiming didnt exist, set to 1, because of absence background suppression, same timing in mean control (used as M0) as in ASL
Warning: ...
x.Q.PresaturationTime is missing or empty, using default value
This assumes that a pre-saturation pulse has been played at the start of the sequence
For the average slice: Control image divided by 0.40875 to correct for background suppression
Using BackgroundSuppressionPulseTime=...
with presaturation time=..., tissue T1=...
And SliceReadoutTime= ...
This converts the control image to allow its use as a pseudo-M0 image
Background suppression pulses have saturated the pseudo-M0 signal, no need for correcting incomplete T1 relaxation
```

jan-petr:
> Finally found some time to check this 

Thanks. Finally found some time to reply :)

> The changes in the **io_mTRIAL** header are not necessary. I removed this part from the header and the whole script is going to change as soon as we move to BIDS anyway. This just leads to a conflict right now.

OK. Reverted those so that I don't have to do conflict solving.
 
> Regarding the header in **xASL_quant_M0**:
> I know you didn't change that much there, it's just that maybe we should write down which fields are added in the x struct.
> So some information in the _OUTPUT:_ part, besides that the x struct comes out, would be nice. Would make it easier for reviewers to test it, too.

Very good point! Filled in all the info.
 
> Additionally, every xASL header need an _EXAMPLE:_ part. If we don't have an example we should just use `EXAMPLE: n/a.` This helps the **docu crawler.**

Thanks. Didn't know that. But in this case - which function is missing it?

> I manually tested **xASL_quant_M0** and especially **xASL_quant_RevertBsupFxControl**. I couldn't find/produce a nice test data set for it, but as far as I can see it right now, we don't get any errors :)

Thanks.

Final edit to do for Yeva is to make sure that BSup-M0 correction works with Background timings defined after the start of labeling before the readout (following the new BIDS definition) - I will make this change in the next commit.

MichaelStritt:
> Thanks. Didn't know that. But in this case - which function is missing it?

I can quick fix this in develop if there are some problems right before the release. So this should be okay 👍 

> Very good point! Filled in all the info.

Thanks :) Let's merge!

MichaelStritt (approved):

PR #242, opened 2020-11-29
Closes #241 Add warning when chosen loading data without x structure output

### LINKED ISSUE #241: Add warning when loading data without x output structure
When you run ExploreASL and choose to load data only, if you don't add an output such as `x = ExploreASL;` then it could be confusing that the data was loaded but not stored in a Matlab structure in memory. Now this issues a warning and suggestion to replace `ExploreASL;` with `x = ExploreASL;`

HenkMutsaerts:
Closes #241

### DIFF
--- a/ExploreASL.m
+++ b/ExploreASL.m
@@ -32,5 +32,9 @@ end
 
 x = ExploreASL_Master(var1, var2, var3, var4, var5, var6);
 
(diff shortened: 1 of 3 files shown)


MichaelStritt (approved):
Tested and looks good 👍 

image

PR #245, opened 2020-11-30
Closes #56 Atlas Option

### LINKED ISSUE #56: Add atlas startup options for mTrial
* Precuneous in Hammers atlas -> correct spelling
* Add options to population module
* Create option to specify different atlases (and put MNI structural back to default)
* For mTrial demo, add vascular atlas & HO_cort_CONN

MichaelStritt:
What we still have to fix/work on:
1. Issue with ROI statistics (volumetric images/ASL-specific masks)
2. If 1. is fixed, we can remove the susceptibility masking part of ExploreASL_Initialize.m 
3. Document the changes (atlas option) in the DataPar template
4. Atlases for ASPIRE? Which are usable now and which have the necessary regions? (anterior & posterior cingulate cortex, precuneus)


MichaelStritt:
> Nice work! Let me please review this again before we merge it

I thought we agreed on working on it together before merging it, so I wouldn't have done this anyway ;)

>x.S.bMasking: wasn't this already predefined somewhere else?

We already talked about this before. Two weeks ago you wrote: _Now I see the reason for the error at your side: you're trying to run volumetric analyses with the HOcort and somehow it doesn't know that this is not an ASL scan, so it still tries to find ASL-specific masks for the anatomical/structural scans._
I've never used different atlases in ExploreASL before. I know they have different usage rights and depict different brain regions, but which are volumetric and which aren't? And why do we get this weird susceptibility warning? Can you explain to me what we have to do here or just help me fix it? The main reason I started with this was because you said we would need open source atlases for ASPIRE. I thought you said we have to use HOcort and HOcon because they have the dementia related brain regions and are open source?

> x.bGetAtlasROIsInNativeSpace -> there seem to be more parameters like this one, in the Population module?

We actually also talked about this before. You said: _Also shows that an option needs to be self-explanatory/really clear. Perhaps Jans "bNativeSpaceAnalysis" was not really clear as well, because "analysis" can mean anything. Your "AtlasesNativeSpace" is also not completely clear :) We should probably have only "bGetAtlasROIsInNativeSpace" or something, do you agree?_

> the string issue with the tmpatlases, I would remove this, just issue an error if strings are added; people should use cells & chars instead.

Definitely, but strsplit and strrep work perfectly fine with chars from the JSON file. So this part...

```
tempAtlases = strsplit(x.Atlases,',');
tempAtlases = strrep(strrep(tempAtlases,'{',''),'}','');
x.Atlases = strrep(tempAtlases,'''','');
```

... work perfectly fine with a JSON file like this...

```
{
"x": [{
	"name": "ExampleDataSet",
	"subject_regexp": "^Sub-\\d{3}$",
	"Quality": 0,
	"bGetAtlasROIsInNativeSpace": 0,
	"Atlases": "{'TotalGM','DeepWM','Hammers','HO_cortex','HO_subcortical'}"
}]
}
```

... and I have to get the cell array from the chars of the JSON file in some way. How would you do it?

> Nice to have an x.Atlases parameter, perhaps better to add this to x.S where all the other "statistical" settings "live"?

Good idea, I can change this right away 👍

MichaelStritt:
Things we came up with:
- Move (fallback) atlas definition to the `DefineDataIndependentSettings` function
- Move path definitions to `initDefinePaths` in `ExploreASL_Initialize`
- Reading of the JSON file: Can we improve reading cell arrays there? How are we currently doing this?
- Reproduce error regarding masking and show it to Henk

MichaelStritt:
Hey @HenkMutsaerts,
I did some major changes. Please check out the **Files changed** tab in **GitHub**.

- I moved the atlas path definitions in `ExploreASL_Initialize` to `xASL_init_DefinePaths`. This didn't really work since the `initFileSystem` overwrites x.P definitions. So I moved it there.
- Checking the matlab version in `xASL_import_json` is unnecessary. Now we just convert the path to char every time. This should be more efficient.
- I added an automatic detection for cell arrays in `xASL_import_json`. If there are chars like `"Atlases": "{'TotalGM','DeepWM','Hammers','HO_cortex','HO_subcortical'}"` in the x structure or one of it's substructures (x.S or x.Q), the char is automatically converted to a cell array. For me this worked quite nicely, but we should test this thoroughly. A recursive solution would have been even better, but I wasn't clever enough to come up with one right now.

The JSON file I used to test this looked like this:
```
{
	"x": [{
			"name": "ExampleDataSet",
			"subject_regexp": "^Sub$",
			"Quality": 0,
			"S":
			{
				"Atlases": "{'TotalGM','DeepWM','Hammers','HO_cortex','HO_subcortical'}"
			}
		}]
}
```

I just ran the code as it is right now and everything "worked", can you check if I'm doing the new part in the Population Module correct, though? I just removed the "volumetric" part of it (`x.S.IsVolume`). Is that correct or do I need to use the "volumetric" option for some of the Atlases?

MichaelStritt:
Hey @HenkMutsaerts 

Just ran a test with the TestDataSet and these settings:

```
{
"x": [{
	"name": "ExampleDataSet",
	"subject_regexp": "^Sub-\\d{3}$",
	"Quality": 0,
	"S":
	{
	"Atlases": "{'TotalGM','DeepWM','Hammers','HO_cortex','HO_subcortical','DKT31_CMA_labels_MNI152'}"
	}
}]
}
```

Everything besides the mindboggle atlases seemed to work. ExploreASL asked me to specify the tsv file for the atlas regions. I think I have to define it first, but can you tell me how to **integrate the tsv file** in the workflow? Getting a GUI request would be pretty bad in the automated workflow. Additionally, we probably either have to use **.nii instead of .nii.gz** files for the atlas and/or **resample to 1.5mm**. Otherwise ExploreASL always creates the corresponding unpacked and resampled atlas Niftis.

gui

HenkMutsaerts:
@MichaelStritt shall we organize a Teams meeting to discuss this together, you're now making new folder structures which use are not immediately clear to me.

* We have all completely free atlas in the SPMmodified/MapsAdded subfolder, to ensure that these work when running our SPM modifications without ExploreASL. I agree though that it would be nice to move these to a general Atlas folder
* I do agree that it would be nice to have two Atlas folders in /External, one that is restricted to non-commercial use and one that is not.
* The license or README for each atlas, I would add as sidecar, not in a separate folder, for users to have everything in the same place. 
* Some of the dummy license files are somewhere else, e.g. `/ExploreASL/External/AtlasesNonCommercial/README_Atlases.txt`
* Good work on finding the MindBoggle atlas, let's check if these atlas regions are as nice as the other default atlases, best to check this visually with a jpg containing the ROIs or something similar, and perhaps asking a neuroradiologist, e.g. Vera Keil if these are ROIs that we want in specific diseases (e.g. AD)
* The suffix `-MNI: I would be careful with new/custom suffixes as they can also suggest the opposite, e.g. the suffix MNI suggests that the other atlases are perhaps not in MNI (whereas all should be in the same space!)
* Again, great work!

HenkMutsaerts (changes requested):
@MichaelStritt shall we organize a Teams meeting to discuss this together, you're now making new folder structures which use are not immediately clear to me.

* We have all completely free atlas in the SPMmodified/MapsAdded subfolder, to ensure that these work when running our SPM modifications without ExploreASL. I agree though that it would be nice to move these to a general Atlas folder
* I do agree that it would be nice to have two Atlas folders in /External, one that is restricted to non-commercial use and one that is not.
* The license or README for each atlas, I would add as sidecar, not in a separate folder, for users to have everything in the same place. 
* Some of the dummy license files are somewhere else, e.g. `/ExploreASL/External/AtlasesNonCommercial/README_Atlases.txt`
* Good work on finding the MindBoggle atlas, let's check if these atlas regions are as nice as the other default atlases, best to check this visually with a jpg containing the ROIs or something similar, and perhaps asking a neuroradiologist, e.g. Vera Keil if these are ROIs that we want in specific diseases (e.g. AD)
* The suffix `-MNI: I would be careful with new/custom suffixes as they can also suggest the opposite, e.g. the suffix MNI suggests that the other atlases are perhaps not in MNI (whereas all should be in the same space!)
* Again, great work!

MichaelStritt:
@HenkMutsaerts: Would be cool if you could check out the changes. I hope everything is a bit more clear with the comments below.

> * We have all completely free atlas in the SPMmodified/MapsAdded subfolder, to ensure that these work when running our SPM modifications without ExploreASL. I agree though that it would be nice to move these to a general Atlas folder
> 
> * I do agree that it would be nice to have two Atlas folders in /External, one that is restricted to non-commercial use and one that is not.

I first thought the same, but it is kind of risky to say "**here is this folder of freely available atlases, use them however you want to**". Right now, I think the best way would be to have a single Atlases folder. There we have the label files (tsv), the actual atlas (nii/gz) and the license (txt). If you're completely averse to this opinion, we can change it back though. 

> * The license or README for each atlas, I would add as sidecar, not in a separate folder, for users to have everything in the same place.

That should be done. Check out the current folder. We have one tsv, one nii and one txt now. All have the same name. For the Harvard-Oxford the text seemed a bit misleading, but I copy pasted the whole text to all of them and the Thalamus one.

> * Good work on finding the MindBoggle atlas, let's check if these atlas regions are as nice as the other default atlases, best to check this visually with a jpg containing the ROIs or something similar, and perhaps asking a neuroradiologist, e.g. Vera Keil if these are ROIs that we want in specific diseases (e.g. AD)

I checked the regions you mentioned once and they should be in it. I think they use a different way to store them in the NIFTI, which is why we get this huge amount of empty regions. I'd rather optimize this at a later point in time though. This PR is getting a bit too big and too old already. There are even some rebased commits from before 1.4.0, which could lead to some conflicts... Hopefully we can just squash it nicely.

> * The suffix `-MNI: I would be careful with new/custom suffixes as they can also suggest the opposite, e.g. the suffix MNI suggests that the other atlases are perhaps not in MNI (whereas all should be in the same space!)

Removed it.

HenkMutsaerts:
> I first thought the same, but it is kind of risky to say "here is this folder of freely available atlases, use them however you want to". Right now, I think the best way would be to have a single Atlases folder. There we have the label files (tsv), the actual atlas (nii/gz) and the license (txt). If you're completely averse to this opinion, we can change it back though.

I think that is fine, as long as we don't change the `/SPMmodified/MapsAdded` (which contains the "truly free" atlases?

Now checking your file additions

HenkMutsaerts (changes requested):
* `Development/Mindboggle/get_mindboggle_labels_tsv.m`: would try to keep the same naming practice throughout (which would be something like `xASL_vis_GetMindboggleLabelsTSV.m`)

* Why do we have the `FutureUse` folder? It could go into /develop/, but otherwise best to decide now if we will use it or not (it only takes up Github space). It seems as this is the origin for the Mindboggle atlas to use, so we don’t need to keep this. (Perhaps best to incorporate a reference to these source data somewhere, so we can retrieve them if necessary)

* `External/Atlases/Mindboggle.nii` should be converted to `.nii.gz`, and cloned also into `.nii.mat`, which is a mat-file (which is automatically compressed for uint8 data formats by Matlab), and ExploreASL automatically reads this sidecar instead of unzipping the .gz on disk. We only keep the .nii.gz copy as this is easier to load in lightweight NIfTI viewers such as FSLview or MRICron.

* Also, the labels of `External/Atlases/Mindboggle.nii`should be increasing and adjacent integers, that correspond to the columns in the TSV sidecar. Now there are huge gaps of unused labels (e.g. 50-700, 800-900, 1100-1900). It seems that one hemisphere is multiplied by 2, so you can merge them to have bilateral regions (`xASL_wrp_GetROI*` will split them). Otherwise, this NIfTI is in the correct space, and contains mutually exclusive regions, this is very nice!

* The .txt License files, e.g. `/External/Atlases/HOsub_CONN.txt`, perhaps
rename them to e.g. `HOsub_CONN_LICENSE.txt` (like with other license files?)
, to make it extra clear that this file should be read ;)

* Note that you also committed a chmod (from 755 to 644)… for 
`External/Atlases/HammersCAT12.txt`. Not a big deal, but probably nicer
if all files have similar default permissions.

* External/Atlases/Mindboggle.tsv: sorry, perhaps I didn’t explain this well
enough; but this format is invalid. It should be a single row with all ROI 
names, corresponding to the integer labels in the nii.gz. The cloning for B, L, and R, is done by the `xASL_wrp_GetROI*` functions, the ROIs in the .nii.gz and .tsv sidecar should all be bilateral only. Now you also did this B, L, R cloning in the TSV :) And some ROIs don’t have names?

* Why do we have a separate README folder, instead of keeping those readme.txt as sidecars as well?

* The doc crawler, `README.md` etc, can we not set up that this is processed/edited
automatically when merging? (By Github actions)

* `xASL_import_json`: would ensure that if the if-statements are not fulfilled,
that you have an alternative (e.g. `warning('this field should be a character array')`; for: line 100 if `ischar(x.(sFields{n})) % Make sure the field is a character array` (and similar throughout)

* `xASL_init_FileSystem:`
`x.P.Atlas.TotalGM                           = fullfile(x.D.MapsSPMmodifiedDir, 'TotalGM.nii');` -> I would standardise this, such that the filenames are 
always the same as the atlas NIfTI filenames. This makes it easier for everyone
(including the users).


* `xASL_vis_CreateVisualFig:`
`MaskIn       - cell structure containing 2 binary masks matrices (OPTIONAL, DEFAULT=no masking)`
You added “2” here, but they only have to be 2, if ImIn is 2 as well (e.g. a T1w
with an image overlay)

* But I don't know what you changed here, or what came from other commits, perhaps better to rebase or cherry-pick, that I only have to review changes you did :)

Issue #248, opened 2020-12-15
Temporary fix native space processing 


HenkMutsaerts:
There were some bugs for native space processing, and/or it didn't work for all atlases in the population module.

Issue #179, opened 2020-10-04
Finish creation average maps for CICERO


HenkMutsaerts:
There are a few bugs and feature(s) still to fix, nearly there. In the previous release, this was addressed by issue #106.

HenkMutsaerts:
For CICERO, we want to create average maps where we choose which hemisphere to use, and flip them accordingly, such that you get an average image at one hemisphere. E.g., when creating an average atlas for a fetal CoW variant, which subject A has on the left side and subject B on the right side, then we want the average of both on the same side (left by default, which means flipping those on the right).

1) all labels are currently transferred to standard space as they are
2) in standard space, need to allow: Scantype = ICA. 
   * If _L -> use ICA_L NIfTI
   * If _R -> use ICA_R NIfTI & flip to the left side
   * When both are desired, they should simply be added as two separate rows
3) Add color-scale for the figures created from the NIfTis, with legend.
4) Equalize color-scales between figures, either by:
   * Combining all NIfTIs in single Figure
   * Disabling auto-clipping and clip all manually to the max nParticipants
Later:
5) Add Dice/Tanimoto Coefficient for differences/similarities between participants/normal-variant groups
6) Denote for each voxel if it differs significantly between normal-variants?

* Update the [tutorial](https://docs.google.com/document/d/1-b272yYb6zyS4QoB1BciBs4-kuBS1PkmTi7LJ1MC1u0/edit) accordingly

HenkMutsaerts:
## ToDo List by Yuanyuan Shen
1. Merge POS-L and POS-R into POS(or named BA), but flip according to group afterwards;
2. Add possibility scale in visualize maps, keep the scale (color) consistent between maps;
3. Add statistical analysis: compare maps between groups by Tanimoto coefficient (image similarity ).

HenkMutsaerts:
I rewrote the input management for bSkipWhenMissingScans, this is how I would do it (with the new x-field having priority but still allowing to input it as separate argument), but this becomes quite non-transparant and confusing; I believe we should just remove the input argument from the header, and adapt all scripts accordingly. Or even easier: we don't need this as parameter that is specified at the DataPar right; we actually only wanted to change that this parameter is off by default right?

Then I would suggest retrieving to the original state (keeping it in the header as it was, removing it from the DataParTemplate, and only setting the default to false... What do you think?

HenkMutsaerts:
Completely rebased & merged

PR #253, opened 2020-12-19
#252 Minor bug fixes

### LINKED ISSUE #252: Population modules analysis masks - minor errors
Several minor bugs in the population module in creation of the templates.

- Main problem is that xASL_adm_GetFileList returns a cell array and not char array. And only some functions can handle it.
- Some other minor problems with Atlas name definitions.

jan-petr:
Critical error in the Population module. Closes #252

### DIFF
--- a/DataParTemplate.m
+++ b/DataParTemplate.m
@@ -202,7 +202,7 @@ function x = DataParTemplate(x)
 %                  - 0 = run CAT12
 % x.bHammersCAT12 - boolean specifying if CAT12 should provide Hammers volumetric ROI results (OPTIONAL, DEFAULT = 0)
 % x.bFixResolution - resample to a resolution that CAT12 accepts (OPTIONAL, DEFAULT=false)
-% x.bSkipWhenMissingScans - skip creating population templates when more than 10% scans are missing (OPTIONAL, DEFAULT=1)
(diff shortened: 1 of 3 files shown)


HenkMutsaerts (approved):
Look s good, my mistake in #248.
I've added another commit to simply things & add the PathSusceptibility (this is how it was before I removed these {1} lines). The reason I did this in #248, is that I replaced .* by the actual n= nSubjectSessions/nSubjects, so it would always find a single file (or no files).

If you agree with my commit then you can merge.

HenkMutsaerts:
Looks good, but then the option should have a more specific name at the beginning of the pipeline:

x.bSkipTemplateCreationWhenMissingScans

Inside xASL_wrp_CreatePopulationTemplates we can then use the x as well, not having to create the “isfield” statement in xASL_module_Population (I would like to keep the modules simple). But this would actually complicate feature#179, where I’m already changing this function; and Reinoud’s group is using this function for creating vascular territory atlases.

Actually, I suggest disabling this feature by default. I believe we should then print in the NIfTI filename the n of the included subjects/sessions (rather than the total), this would be better; but then we should first delete the existing template NIfTI. And we should change this again in the CreateMasks script :)
Do you agree?

HenkMutsaerts:
@jan-petr I've fixed all in #179. If you approve my latest commits in #179, you can discard this #252 issue & PR #253.

PR #256, opened 2021-01-05
Feature #179 cicero

HenkMutsaerts:
@jan-petr You can check the stuff from #253 and #252 that I added here in the latest commits.
Not sure why this branch cannot "automatically merge", it seemed to have a conflict in CHANGES.md, but this seemed to be the same.


jan-petr:
> I rewrote the input management for bSkipWhenMissingScans, this is how I would do it (with the new x-field having priority but still allowing to input it as separate argument), but this becomes quite non-transparant and confusing; I believe we should just remove the input argument from the header, and adapt all scripts accordingly. Or even easier: we don't need this as parameter that is specified at the DataPar right; we actually only wanted to change that this parameter is off by default right?
> Then I would suggest retrieving to the original state (keeping it in the header as it was, removing it from the DataParTemplate, and only setting the default to false... What do you think?

Agree and done.

PR #258, opened 2021-01-07
Feature #163 dicom2 bids

### LINKED ISSUE #163: Create DICOM2ASL-BIDS app
* Create script based on ExploreASL_Import.m that outputs to ASL-BIDS
* Create a conversion script ExploreASL_Import2BIDS
* Insert some functionality from GSP flavors

jan-petr:
Close #163


HenkMutsaerts (changes requested):
### Dicom2BIDS ASL App

* As app, and the different license, the functions of this DCM2BIDS app should be in a separate External/AppName folder
* They should then work without ExploreASL but with the xASL-version of SPM (SPM_modified)
* We should try if this works compiled; preferably with a separate testing person

### Minor changes
* Why do some of the functions use system('cp') or system('mv') rather than xASL_SysCopy or xASL_SysMove?
The way you coded this now, it doesn't work with Windows (the Sys... uses the same for Linux/Mac but is also Windows 
compatible).
* It does contain some German still (privatItem rather than privateItem ;)
* Use of spaces is not consistent yet (it has both “( ( privatItem ) )” and “((privatItem))”
* S = num2str(json,12) -> num2str only recently got the second input argument to specify the number of floating point digits, best to use xASL_num2str instead for backward compatibility
* Functionname ExploreASL_bids_JsonCheck should be xASL_bids_JsonCheck
* Also check number of spaces, indentation should have 4 spaces/1 tab, not 8 spaces (2 tabs) (in ExploreASL_bids_JsonCheck
* Matlab doesn’t need unnecessary brackets (e.g. “switch (jsonIn.Manufacturer)” should be “switch jsonIn.Manufacturer”) (same in other functions)
* Function xASL_adm_Hex2Num has a “return” in the middle of the function (line 61), isn’t this bad practice?
* in xASL_bids_CompareStructures you give the error: error('The path of structure A is neither a char array not a string.’). But char==string with jsons right? And we always use chars in ExploreASL, so there shouldn’t be strings?
* xASL_bids_Par2JSON: confuses “parms” “parameters” “json”, please check for consistent use. The json file should be called “json sidecar” and the parms.mat file is the legacy ExploreASL parameter file.
* there should be a space after a comma in Matlab scripts
* xASL_bids_PhoenixProtocolAnalyzer: why did you use the term “rawPar”? In BIDS terminology, the rawdata folder is the BIDS folder ;) The sourcedata folder contains what we previously called “raw” in ExploreASL (i.e. the DICOMs)

MichaelStritt (commented):
Just read Henks review. I have some spare time this morning and will try to implement some of the change requests.

MichaelStritt:
@jan-petr & @HenkMutsaerts 

> - It does contain some German still (privatItem rather than privateItem ;)
> - Use of spaces is not consistent yet (it has both “( ( privatItem ) )” and “((privatItem))”

This should be done, but the corresponding **.cpp** files probably need to be recompiled.

> - S = num2str(json,12) -> num2str only recently got the second input argument to specify the number of floating point digits, best to use xASL_num2str instead for backward compatibility

I changed the corresponding lines to **xASL_num2str**, but the functionality of **xASL_num2str** does not seem to be exactly the same. With `a = xASL_num2str(1.23456,3);` we get `a = '1.23'` but we probably rather want `a = '1.234'`. Should we change this in **xASL_num2str**?

> - Functionname ExploreASL_bids_JsonCheck should be xASL_bids_JsonCheck

I renamed these functions. I'm not 100% sure that I found all the code lines where these functions are used. This should probably be tested.

> - Also check number of spaces, indentation should have 4 spaces/1 tab, not 8 spaces (2 tabs) (in ExploreASL_bids_JsonCheck)

I used smart indent on the function and can't see any problems right now. Should be okay, I think.

> - Matlab doesn’t need unnecessary brackets (e.g. “switch (jsonIn.Manufacturer)” should be “switch jsonIn.Manufacturer”) (same in other functions)

Fixed this in **VendorFieldCheck** and **CompareStructures**. I guess this should be fine as well.

> - Function xASL_adm_Hex2Num has a “return” in the middle of the function (line 61), isn’t this bad practice?

Not necessarily, but I think it's not "best" practice in this case ;) I changed the first two usages of the **return** command to an **error**, because that's probably what we want here? The third usage (input is already a char) seems to be fine, because then you should **return** the input value?

> - In xASL_bids_CompareStructures you give the error: error('The path of structure A is neither a char array not a string.’). But char==string with jsons right? And we always use chars in ExploreASL, so there shouldn’t be strings?

That was my fault. First I wrote the code based on a really up-to-date version of Matlab and wrote it with char and string compatibility. Then you reviewed it once and we removed all string usage. This error message just wasn't updated. Fixed it.

> - xASL_bids_Par2JSON: confuses “parms” “parameters” “json”, please check for consistent use. The json file should be called “json sidecar” and the parms.mat file is the legacy ExploreASL parameter file.

Maybe @jan-petr should fix this. I know it's not much code, but I don't want to add more confusion to the confusion 👍 

> There should be a space after a comma in Matlab scripts

Fixed this whenever I stumbled upon it. The remaining ones shouldn't stop us from merging ;)

> - xASL_bids_PhoenixProtocolAnalyzer: why did you use the term “rawPar”? In BIDS terminology, the rawdata folder is the BIDS folder ;) The sourcedata folder contains what we previously called “raw” in ExploreASL (i.e. the DICOMs)

My fault, didn't think about BIDS here. I changed **rawPar** to **sourcePar**.

--------------------

We should definitely test all the changes, to make sure I didn't break some parts of the code.

HenkMutsaerts:
> > I renamed these functions. I'm not 100% sure that I found all the code lines where these functions are used. This should probably be tested.

You can easily do a search and replace.

> > Not necessarily, but I think it's not "best" practice in this case ;) I changed the first two usages of the return command to an error, because that's probably what we want here? The third usage (input is already a char) seems to be fine, because then you should return the input value?

The functionality is good, but the usage of "return" in the middle of a function decreases readability. Rather replace it by if-statements.

Otherwise, very good! You don't need to apologize, you're not a Canadian ;) Jan can fix the rest. And he/we should think about backward compatibility for the new BIDS names

jan-petr:
> I renamed these functions. I'm not 100% sure that I found all the code lines where these functions are used. This should probably be tested.

OK - checked this. I have checked, fixed, and tested also the other things - with the updated Flavors - and all works well.

I still need to fix this: xASL_bids_Par2JSON: confuses “parms” “parameters” “json”, please check for consistent use. The json file should be called “json sidecar” and the parms.mat file is the legacy ExploreASL parameter file.

Making a separate DICOM2BIDS app - I have created a new issue for this...

HenkMutsaerts:
@jan-petr did you fix the following:

> xASL_bids_Par2JSON: confuses “parms” “parameters” “json”, please check for consistent use. The json file should be called “json sidecar” and the parms.mat file is the legacy ExploreASL parameter file.

If yes, we can merge this PR, or should we test it first?

MichaelStritt:
> If yes, we can merge this PR, or should we test it first?

I think Jans changes in xASL_bids_Par2JSON should be fine. I ran some tests of the dcm2bids workflow yesterday, to be able to test if they work with your bids2legacy code. Looks really good so far. Let's merge everything that has been done so far and improve it later.

MichaelStritt (approved):

PR #264, opened 2021-01-11
#39 Added the option to output PV-corrected maps in ASL native space.

### LINKED ISSUE #39: Create PV-corrected GM & WM CBF maps in native space
For some users it would be useful to select their own ROI and position it over a PV corrected CBF map, rather than having to use this on the ROI level. @jan-petr: can you add the creation of PV-corrected CBF maps in native space as default option to the quantification submodule?

Thanks!

jan-petr:
Close #39


HenkMutsaerts (changes requested):
Please put the wrapper in the same "state 8" as the quantification, in the ASL module.

MauricePasternak (changes requested):
Overall, seems to work. I tried both the Gaussian and Square kernels at their recommended voxel sizes and the CBFwm/gm niftis were newly made. 

However, in this pull request there are a few oppurtunities I see to address a few things that were detected by the GUI:

1) x.M0PositionInASL4D isn't clear in what its default value should be if the parameter is present (and the GUI always spits it out). I assumed a default value of 0 is sufficient to indicate that this should be skipped. Could a 2-3 line condition be introduced to omit M0 extraction out of the ASL volume if that parameter's value is 0? I prefer a 0 over an empty vector since the other possible values are integers as well. In xASL_module_ASL.m, something like:

```
    % make sure to split the M0 from the ASL time-series if needed
    if x.M0PositionInASL4D ~= 0
        xASL_io_SplitASL_M0(x.P.Path_ASL4D, x.M0PositionInASL4D);
    end
```


2) I noticed "080_CreateAnalysisMask.status" and "070_Quantification.status" switched places (i.e. "070_Quantification.status" is now "080_Quantification.status") relative to the master branch. Since the GUI is based off the master branch, this throws errors. Should I accomodate to this newer filename scheme and just endure the errors until master follows the same scheme?

3) 090_VisualQC_ASL.status is deleted and re-created during every run, resulting in a false-positive signal detected by my file watchers. Either I can give that file a value of 0 towards the progressbar or that behavior is avoided.

jan-petr:
> Overall, seems to work. I tried both the Gaussian and Square kernels at their recommended voxel sizes and the CBFwm/gm niftis were newly made.

Good to hear that.
 
> However, in this pull request there are a few oppurtunities I see to address a few things that were detected by the GUI:

Thanks. Let's discuss here, but some of our conclusions should end up by creating a new issue.
 
> 1. x.M0PositionInASL4D isn't clear in what its default value should be if the parameter is present (and the GUI always spits it out). I assumed a default value of 0 is sufficient to indicate that this should be skipped. Could a 2-3 line condition be introduced to omit M0 extraction out of the ASL volume if that parameter's value is 0? I prefer a 0 over an empty vector since the other possible values are integers as well. In xASL_module_ASL.m, something like:

Yes. Good point. I think that a default value of 0 is acceptable. @HenkMutsaerts - do you agree? I will create a new issue for this - #272.

> 1. I noticed "080_CreateAnalysisMask.status" and "070_Quantification.status" switched places (i.e. "070_Quantification.status" is now "080_Quantification.status") relative to the master branch. Since the GUI is based off the master branch, this throws errors. Should I accomodate to this newer filename scheme and just endure the errors until master follows the same scheme?

Yes - they have switched places because this was needed to be able to accommodate BASIL in the future. This is something that was already merged to develop and we want to release version 1.4.0 on Friday - so you can carry on.

Contrary to this - the new status 085 for PVC will be deleted and merged together with 080.

> 2. 090_VisualQC_ASL.status is deleted and re-created during every run, resulting in a false-positive signal detected by my file watchers. Either I can give that file a value of 0 towards the progressbar or that behavior is avoided.

@HenkMutsaerts Any suggestion here? It is always deleted if something changes - so the behavior can't changed. On the other hand - we plan to move all functions for calculating QC to the respective modules in the future - so this module will only assemble everything and will be lightweight. So IMHO - can be given a weight 0.

HenkMutsaerts:
Looks good, I would check for naming consistency though, mostly we use PVC, here you introduce PVCorrection.
We discussed that we wouldn't introduce a new status, but would include `xASL_wrp_PVCorrection` inside the current `080_quantification`.

HenkMutsaerts:
> 1. x.M0PositionInASL4D isn't clear in what its default value should be if the parameter is present (and the GUI always spits it out). I assumed a default value of 0 is sufficient to indicate that this should be skipped. Could a 2-3 line condition be introduced to omit M0 extraction out of the ASL volume if that parameter's value is 0? I prefer a 0 over an empty vector since the other possible values are integers as well. In xASL_module_ASL.m, something like:

>Yes. Good point. I think that a default value of 0 is acceptable. @HenkMutsaerts - do you agree? I will create a new issue for this - #272.

I would argue against this in general. 0 is a value that is used quite a lot, so 'undefined' or 'empty' should rather be []. I don't know how this is in JSON, if we can't do "[]" because it has to be numeric, then I would go for NaN, if it has to be integer, then I would go for 9999, which is more often used as "missing value".



Contrary to this - the new status 085 for PVC will be deleted and merged together with 080.

> 2. 090_VisualQC_ASL.status is deleted and re-created during every run, resulting in a false-positive signal detected by my file watchers. Either I can give that file a value of 0 towards the progressbar or that behavior is avoided.

Status files/submodules that are dependent of previous status files/submodules, are automatically deleted when a status is finished, to ensure that they are updated for a rerun. You can find this behavior by looking for `x.mutex.DelState` inside `xASL_module_Structural` and `xASL_module_ASL`.

BTW: let's try to organize these discussions a bit, such that they fit the subject of the issue. This issue is about native space PVC, not about general GUI-related questions ;) So this discussion should actually go to a GUI issue.

HenkMutsaerts (changes requested):
The default of PVC is 0 because it takes quite some time right? Then I would add this to the DataParTemplate explanation, otherwise, it's weird to have it there as an option in the first place.

The name of the regression-PVC function has "kernel", isn't that weird, as all PVC's essentially have a different kernel? E.g., the Bspline one essentially has a Bspline kernel. So then it makes more sense to give it a name that really separates it from the Bspline one?

I would provide an explanation with what "flat" means (e.g. 2D rectangular kernel)
Actually in xASL_wrp_PVC you call this a "square" kernel, which one is more correct? :)

%% 9    Visual QC
iState = 9;	iState = 9;
if ~x.mutex.HasState(StateName{iState}) && x.mutex.HasState(StateName{iState-2})	if ~x.mutex.HasState(StateName{iState}) && x.mutex.HasState(StateName{iState-3})

This last change `iState-2` to `iState-3` should be reverted right, as you reverted an extra submodule?

HenkMutsaerts (approved):
Looks good, see my comments in the issue. Note that I added 1 commit for progress tracking.

Issue #274, opened 2021-01-14
ExploreASL_GUI-Update_Jan_2021


MauricePasternak:
Time for a much-needed update.

Some of the major changes, apart from aesthetics, include:

- The ability to stop processing mid-way
- Processing in the ASL_Module now gives feedback to which runs completed a step
- Updated Modification Jobs Prior/Post run
    -  Alter participants.tsv --> Alters the participants.tsv file in a given study with metadata present in container. A useful addition for cases where copying only certain rows from a very large metadata spreadsheet would be very tedious.
    - Change Json Sidecars --> Allows the user to batch-edit the json sidecars of a given study, making modifications as necessary. Batch-editing can be done manually or instructions (in the form of a spreadsheet) loaded in from a file
    - Merge Study Directories --> Allows the user to create a common directory from several studies, merging all the files, folders, etc. Can be done with symlinks instead to save diskspace.
- The ability to reorganized directory levels of a study to make it compatible with the Import Module (i.e. can pull apart a directory subject_visit into SUBJECT/subject/VISIT/visit)

HenkMutsaerts:
Nice, though please merge squashed, add the issue number, and sign your merge ;)
Just follow our GitHub flow, and ask us if needed; just to keep overview.
You can reuse an issue though, you merge, close the issue for e.g. 1.5.0, and re-open it later for 1.6.0, and so on.

HenkMutsaerts:
We decided to remove /External/ExploreASL_GUI and keep this in a /ExploreASL/ExploreASL_GUI repository forked from /MauricePasternak/ExploreASL_GUI.

PR #275, opened 2021-01-14
Major GUI Update for Jan 2021

MauricePasternak:
The virtual environment was obviously not included, as it is too bulkly. I trust one can be generated from the requirements.txt.

Main Questions to ask:

1) The following is an example of the default parameters now generated by the ParmsMaker:

[DataPar.zip](https://github.com/ExploreASL/ExploreASL/files/5817352/DataPar.zip)

Anything safe to remove? Bad defaults? NB: the filepaths under "MyPath" and "ROOT" will be actual, existent filepaths when this file is normally made.

2) FYI JSON_LOGIC now is now the source of most GUI tooltips, error messages. But the main thing to check is if the Translators are correct or if measures should be taken to accomodate new modules. DARTEL? LongReg? If so, what would be the status files made and what would be the DataPar key to assess when trying to anticipate those files being made?

3) (Optional) If you have time, I guess give a virtualenv (must be Python 3.8 or above) a go?


HenkMutsaerts (changes requested):
You add ExploreASL_GUI/.gitignore to .gitignore while at the same time you ask Git to version control it (or perhaps a different folder)? I would at your end also keep External/ExploreASL_GUI (same as on the server) and only specify in .gitignore files that should be ignored, not the same files in a different folder.

HenkMutsaerts:
Thanks! The venv should either be included, or included in a separate docker, as we had too many version issues; rather have a bulky docker than us humans need to waste time on PyPa versions :)

I would start without DARTEL/LongReg, these are optional

### Safe to remove from DataPar.json, because they are deprecated are:
>>   "SESSIONS": [
      "ASL_1"
   ],
   "session": {
      "options": [
         ""
      ]
   },



### Safe to remove because they are defaulted, are:
>>    "exclusion": [],
   "M0_conventionalProcessing": 0,
   "M0_GMScaleFactor": 1,
   "M0PositionInASL4D": 0,
   "DELETETEMP": 1,
   "SkipIfNoFlair": 0,
   "SkipIfNoM0": 0,
   "SkipIfNoASL": 1,
   "SegmentSPM12": 0,
   "bRunModule_LongReg": 0,
   "bRunModule_DARTEL": 0,
   "bHammersCAT12": 0,
   "bFixResolution": 0,
   "SPIKE_REMOVAL": 1,
   "SpikeRemovalThreshold": 0.01,
   "motion_correction": 1,
   "bRegistrationContrast": 2,
   "bAffineRegistration": 2,
   "bDCTRegistration": 0,
   "bRegisterM02ASL": 1,
   "bUseMNIasDummyStructural": 0,
   "bPVCorrectionNativeSpace": 0,
   "bPVCorrectionGaussianMM": 0,
   "PVCorrectionNativeSpaceKernel": [
   "FSLdirectory": "",
   "MakeNIfTI4DICOM": 0,
   "ApplyQuantification":
 "Lambda": 0.9,
      "T2art": 50,
      "TissueT1": 1240,
      "nCompartments": 1
 "bMasking": 

And if no background suppression was used, you can remove those fields, for 3D you can remove the SliceReadoutTime.

Now working on reading this all from BIDS, so this won't be necessarily hopefully soon!

HenkMutsaerts:
* Allowing Matlab 2016+ on MacOS
* Allowing different Linux versions (GLIBC error) or having a docker instead?

HenkMutsaerts:
@MauricePasternak Great work! Minor question: when you do a commit, it should begin with mentioning the issue, which facilitates keeping track of the repository history. Plus, the commit message should be specific, otherwise e.g.  `MacOS compatibility update` could refer to anything. Note that the link with the issue only remains if you mention the issue in your commit message, and the link with the branch is gone after merging. So your commit should be e.g. `#274 ExploreASL_GUI MacOS compatibility update`. We can edit this for your last commits here.

HenkMutsaerts (changes requested):
@MauricePasternak Great work! Minor question: when you do a commit, it should begin with mentioning the issue, which facilitates keeping track of the repository history. Plus, the commit message should be specific, otherwise e.g. MacOS compatibility update could refer to anything. Note that the link with the issue only remains if you mention the issue in your commit message, and the link with the branch is gone after merging. So your commit should be e.g. #274 ExploreASL_GUI MacOS compatibility update. We can edit this for your last commits here.

MichaelStritt:
@HenkMutsaerts: Are we going to close this PR since we removed the **GUI** from `ExploreASL/External`?

MauricePasternak:
Yes, I believe it should be closed, as the file contents will go to the forked repo instead.

PR #284, opened 2021-01-20
#283 xASL_stat_GetROIstatistics Improve feedback on missing images

### LINKED ISSUE #283: xASL_stat_GetROIstatistics: provide more feedback on missing images
Edited the code such that it doesn't only provide feedback when the input CBF image .* mask in computation functions issued by xASL_stat_GetROIstatistics is empty, but also before this function is called, e.g. when the CBF image in memory is empty, or when a mask is empty, and for which subject. 

This should help understanding why these images are skipped (e.g. in my case here we wanted to re-quantify and recreate qCBF images while accounting for Hct for Blood T1 calculation, and Hct information wasn't present in all cases.

HenkMutsaerts:
Closes #283


jan-petr (approved):
Looks good.

PR #287, opened 2021-01-20
#282 xASL_module_Population: iWorker>1 error fixed

### LINKED ISSUE #282: xASL_module_Population error for iWorker>1
When not worker 1, give warning only and not crash with error.
When running ExploreASL parallel (multiple Matlab instances, through the header iWorker & nWorkers), the Population module should only run serial. It does this now, but it also gives an error instead of nicely cleanly closing.

HenkMutsaerts:
Closes #282 

@MDijsselhof  Another nice one for you ;) You can also merge this one, see if that works for you! Otherwise, we can do this together the first couple of times.


MDijsselhof (dismissed):
Looks great.

PR #289, opened 2021-01-22
Fix #273 default option gzip nifti

### LINKED ISSUE #273: Default option xASL_adm_GzipNifti
Hey, only a minor issue. In the header it says, that the default for bOverwrite is **false**, but it seems like it's **true**.
pathOut = xASL_adm_GzipNifti(pathIn, bOverwrite)

```
bOverwrite - indicates if we should overwrite the .NII file (OPTIONAL, default = FALSE)
```
...
```
% Check for the optional parameter overwrite
if nargin < 2 || isempty(bOverwrite)
    bOverwrite = 1;
end
```

MichaelStritt:


jan-petr (approved):
Looks good. Changed the header instead of the functionality. Fixed and can be merged.

HenkMutsaerts:
Yes, I agree, if you change the default, you really need to check how this affects the pipeline. The defaults are usually set to benefit the pipeline.

PR #295, opened 2021-02-01
Closes #294 initial test framework

### LINKED ISSUE #294: Initial Test Framework
- Move test script framework from Development to a dedicated folder
- Add functionality to call test repository
- Define simple tests with regard to test data

Notice: This is only supposed to be a minor issue for the initial set up

MichaelStritt:
Hey, I just moved the existing scripts from the Development folder to a dedicated Testing folder. I added a first unit test for the tsvReader and I updated the template. This could almost be treated as **External**, since it shouldn't affect the general ExploreASL workflow at all.

Edit: Added an tsvWriter test.


jan-petr (commented):
Looks super good! 
Can we make a first working example? I.e. minimal edits in the header of those 3 example functions so that they don't look like templates but like finished testing functions. And can we make a single execution script that actually runs the test on these three and report the results - by having this starting script, we could get rid of the function that brings out a dialog to select a folder - because we probably want to start things on a single click.

But all this can be made into a new issue.

MichaelStritt:
@jan-petr thanks! I quickly updated the headers. There's a master script, that's `xASL_ut_UnitTesting`. If you run this one and if the **ExploreASL/Testing** repository is in the same folder as you **ExploreASL/ExploreASL** repository, it should run out of the box 👍

Example:

```
>> UnitTests = xASL_ut_UnitTesting;
Testing repository found...
================================= TEST RESULTS =================================
TEST:		xASL_tsvRead
RESULTS:	Passed
TEST:		xASL_tsvWrite
RESULTS:	Passed
TEST:		xASL_io_Nifti2Im
RESULTS:	Passed
================================================================================
```

The resulting structure looks like this right now, but we could also use a table or a tsv:

image

jan-petr:
Thanks. Looks amazing!

Minor fixes. Sorry.
- xASL_ut_UnitTest_function_tsvRead (line 36) - hardcoded path, won't work on Linux
- Can you add the ExploreASL/ExploreASL/Testing directory to the basic path during ExploreASL_Initialize - so that the testing function can be called right away. 
- For discussion with Henk - does it make sense to have the input files in one repository and the code in another? We can probably have both in the ExploreASL/ExploreASL.

MichaelStritt:
No need to say your sorry ;) 

> xASL_ut_UnitTest_function_tsvRead (line 36) - hardcoded path, won't work on Linux

Will fix that right away 👍 

> Can you add the ExploreASL/ExploreASL/Testing directory to the basic path during ExploreASL_Initialize - so that the testing function can be called right away.

Sure 👍 

> For discussion with Henk - does it make sense to have the input files in one repository and the code in another? We can probably have both in the ExploreASL/ExploreASL.

Hm... I personally wouldn't do that, because in the end there will be a number of test files or half-processed datasets, maybe some DRO data etc., in there, but I'm fine with both.

MichaelStritt:
Really like these tables right now, so I added that as well:

```
> [UnitTests,UnitTestsTable] = xASL_ut_UnitTesting;
Testing repository found...
================================= TEST RESULTS =================================
TEST:		xASL_tsvRead
RESULTS:	Passed
TEST:		xASL_tsvWrite
RESULTS:	Passed
TEST:		xASL_io_Nifti2Im
RESULTS:	Passed
================================================================================
>> UnitTestsTable

           name              unit          tests        passed
    __________________    __________    ____________    ______

    'xASL_tsvRead'        'Function'    [1×2 struct]    true  
    'xASL_tsvWrite'       'Function'    [1×2 struct]    true  
    'xASL_io_Nifti2Im'    'Function'    [1×1 struct]    true 
```

jan-petr (approved):
Works perfectly now. Eager to start using it!

MichaelStritt:
@jan-petr Oh sorry, I forgot to check paths the individual tests. Thanks for the fix in 7745c3f  👍

PR #303, opened 2021-02-04
#302 Remove SPM/LST/CAT12 server calls

### LINKED ISSUE #302: Remove CAT12 server calls
While debugging some work from Maurice Pasternak, I discovered some instances where CAT12 tries to call a server to note versions that are using CAT12. In this issue, I will comment these sections.

HenkMutsaerts:
Closes #302
(Tested this with /External/TestDataSet)


jan-petr (approved):
Very good.

PR #307, opened 2021-02-04
Bug #305 x asl adm unix path

### LINKED ISSUE #305: xASL_adm_UnixPath: bug with Windows+WSL
If `ispc` is true and WSL is present, `xASL_adm_Unix` is not functioning well.
I will make two changes:
* skipping this function if Windows & WSL are present, by default, unless this is specifically requested (i.e. when running `xASL_fsl_*` functions);
* the list of illegal characters should not include the colon ':' when `ispc` is true

HenkMutsaerts:
Closes #305
Guys, I selected you all, as this a simple but quite relevant step.
Can also easy cause quite some time to find this potential bug (like Maurice and I did today :)

@MDijsselhof: can I ask you to test this in a subject with FSL TopUp?
@MauricePasternak: can I ask you to test this on your Windows+WSL?


MichaelStritt (commented):
I don't see any Matlab errors in the code and everything seems okay regarding our xASL guidelines. Would be cool if Mathijs & Maurice could test it thouh 👍

jan-petr (changes requested):
Good fix. But I have reservations to the implementation. The functionality is good, but it is a bit confusing.

1. The options makes it look as if bWSL should be always 1 for PC+WSL, but in practice, you only use it for FSL, and not for all cases when PC+WSL. The option thus shouldn't be named bWSL, but rather bForceUnix.
2. The input parameter admin is rather confusing - working well, but confusing. You assign a default value, but in case a value was given (true or false), it tries to confirm it. That's very confusing. Solution would be to first rename bWSL to bForceUnix (per point 1). And then either explicitly say that you are confirming it, or that if bForceUnix is true, you seek the WSL installation and then set a secondary bWSL that is used afterwards.

Otherwise - the changes and functionality looks good.

External/SPMmodified/xASL/xASL_adm_UnixPath.m
```diff
@@ -24,11 +28,20 @@
 % __________________________________
 % Copyright 2021 ExploreASL
         
+    if nargin<2 || isempty(bWSL)
+        bWSL = false;
+    elseif ispc
```
jan-petr, line 33: This statement is only reached if bWSL is defined. Either TRUE or FALSE - this doesn't really make sense.
Either this needs to be changed, or the input parameter has to change from bWSL to bForceUnix, so that the setting of bWSL below makes sense.

MauricePasternak:
I have not looked into the code, but I can confirm that the missing nT1.nii error followed by CAT12 failing (and backup methods entering a perpetual loop of some kind) is gone and the module goes on to produce the expected segmentations, report, etc.

MauricePasternak (approved):

External/SPMmodified/xASL/xASL_adm_UnixPath.m
HenkMutsaerts: I will rename the booleans such that this makes sense. But this does already make sense. You'll see ;)

jan-petr: Looks good now.

jan-petr (approved):

jan-petr (approved):
Looks good now. I made a minimal change with no practical consequences. Please test and merge.

HenkMutsaerts:
Looks good to me!

MDijsselhof (approved):

PR #310, opened 2021-02-07
Bug #309 toohigh values

### LINKED ISSUE #309: Non-linear registration fails with too high T1w values
The `T1.nii` of BIDS test data `Philips_PCASL_2DEPI` has too high values, which chokes the CAT12 non-linear registration.
This fix will normalize the maximum value of T1w and FLAIR NIfTIs to 4096 (12^2, 12 bit max) before any processing.

HenkMutsaerts:
Closes #309


jan-petr (approved):
OK - if that works for all old and new. I made a minor fix in the header. And also - X^(2^0.5) returns a complex number for a negative X, so I made abs(IM) before IM^(2^0.5).

PR #317, opened 2021-02-08
Closes #272

### LINKED ISSUE #272: M0PositionInASL4D can be zero
Maurice: x.M0PositionInASL4D isn't clear in what its default value should be if the parameter is present (and the GUI always spits it out). I assumed a default value of 0 is sufficient to indicate that this should be skipped. Could a 2-3 line condition be introduced to omit M0 extraction out of the ASL volume if that parameter's value is 0? I prefer a 0 over an empty vector since the other possible values are integers as well. In xASL_module_ASL.m, something like:

    % make sure to split the M0 from the ASL time-series if needed
    if x.M0PositionInASL4D ~= 0
        xASL_io_SplitASL_M0(x.P.Path_ASL4D, x.M0PositionInASL4D);
    end

MichaelStritt:
Hey, I added a second line with a default for `x.M0` if `x.M0PositionInASL4D` is not empty. Is it correct like this? I couldn't find another default value in the DataParTemplate.

### DIFF
--- a/DataParTemplate.m
+++ b/DataParTemplate.m
@@ -24,7 +24,7 @@ function x = DataParTemplate(x)
 %
 % EXAMPLE: ExploreASL_Master('//StudyFolder/analysis/DataPar_HiQ.json');
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
-% Copyright 2015-2019 ExploreASL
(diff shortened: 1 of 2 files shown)


MichaelStritt:
@HenkMutsaerts: I implemented the minor changes you described in #272  👍

PR #331, opened 2021-02-14
Bug #325 x asl adm clean up before rerun

### LINKED ISSUE #325: xASL_adm_CleanUpBeforeRerun feeds cell instead of string to xASL_delete
Occurs in Section 3: "3) Remove lock files/folders for reprocessing"

At this block:

        LockedDir = xASL_adm_GetFileList(CurrentDir{iCurrent}, 'locked', 'FPListRec', [0 Inf], true);
        if isempty(LockedDir) % keep locked dir for mutex
            xASL_delete(CurrentDir{iCurrent});
        end

Since, xASL_delete expects a string, you can see the issue that happens, given:

```
CurrentDir =
  1×1 cell array
    {1×1 cell}
    
CurrentDir{iCurrent}
ans =
  1×1 cell array
    {'/home/mpasternak/Documents/Some_Study/Philips/2D_EPI/analysis/lock/xASL_module_ASL/Sub-001/xASL_module_ASL_ASL_1'}
```
I believe the intention was to delete `LockDir` instead?

HenkMutsaerts:
Closes #325

### DIFF
--- a/External/SPMmodified/xASL/xASL_delete.m
+++ b/External/SPMmodified/xASL/xASL_delete.m
@@ -25,6 +25,15 @@ elseif ischar(bFolderContent) || ~(bFolderContent==0 || bFolderContent==1)
     error(['Invalid parameter bFolderContent:' bFolderContent]);
 end
 
+% Dealing with cell input (e.g. accidental output from xASL_adm_GetFileList)
+if iscell(InputPath) && numel(InputPath)>1
+    error('InputPath should be a path to the file/folder to be deleted, but was a cell');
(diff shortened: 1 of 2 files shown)


jan-petr (approved):
Henk, you have found the right cause for the error, but your fix wasn't working under certain circumstances - I have fixed all these now.

HenkMutsaerts:
Looks good, the only issue could be that `xASL_adm_GetFileList` gives a row of cells, whereas you now concatenate a column of the lists (`;` instead of `,`), so this may go wrong. Unless `xASL_adm_GetFileList` always gives a single cell output.

Should also still need testing.

MauricePasternak (approved):
Clean behavior now. Looks good to me!

PR #332, opened 2021-02-15
Closes #326

### LINKED ISSUE #326: Parse warnings/errors
### Rationale
When testing, especially when testing in parallel, it takes a lot of time to go through all log-files one-by-one. Rather, they should be read in, and their warnings and errors should be parsed.

In these log files, that were started inside ExploreASL using the Matlab `diary on` & `diary off` functionality:

### Warning
A warning contains `Warning:`, starts with `[` and ends with `]`. Then there are several references to the (sub-)functions that were responsible for this warning, starting with `In <a href=` and ending with `line \d*</a>)]`
Example:

```
[Warning: Registration off, spatial CoV detection unreliable] 
[> In <a href="matlab:matlab.internal.language.introspective.errorDocCallback('xASL_wrp_RegisterASL>xASL_im_GetSpatialCovNativePWI', '/Users/henk/ExploreASL/ExploreASL/Modules/SubModule_ASL/xASL_wrp_RegisterASL.m', 645)" style="font-weight:bold">xASL_wrp_RegisterASL>xASL_im_GetSpatialCovNativePWI</a> (<a href="matlab: opentoline('/Users/henk/ExploreASL/ExploreASL/Modules/SubModule_ASL/xASL_wrp_RegisterASL.m',645,0)">line 645</a>)
  In <a href="matlab:matlab.internal.language.introspective.errorDocCallback('xASL_wrp_RegisterASL', '/Users/henk/ExploreASL/ExploreASL/Modules/SubModule_ASL/xASL_wrp_RegisterASL.m', 392)" style="font-weight:bold">xASL_wrp_RegisterASL</a> (<a href="matlab: opentoline('/Users/henk/ExploreASL/ExploreASL/Modules/SubModule_ASL/xASL_wrp_RegisterASL.m',392,0)">line 392</a>)
  In <a href="matlab:matlab.internal.language.introspective.errorDocCallback('xASL_module_ASL', '/Users/henk/ExploreASL/ExploreASL/Modules/xASL_module_ASL.m', 292)" style="font-weight:bold">xASL_module_ASL</a> (<a href="matlab: opentoline('/Users/henk/ExploreASL/ExploreASL/Modules/xASL_module_ASL.m',292,0)">line 292</a>)
  In <a href="matlab:matlab.internal.language.introspective.errorDocCallback('xASL_Iteration>runIteration', '/Users/henk/ExploreASL/ExploreASL/Functions/xASL_Iteration.m', 361)" style="font-weight:bold">xASL_Iteration>runIteration</a> (<a href="matlab: opentoline('/Users/henk/ExploreASL/ExploreASL/Functions/xASL_Iteration.m',361,0)">line 361</a>)
  In <a href="matlab:matlab.internal.language.introspective.errorDocCallback('xASL_Iteration', '/Users/henk/ExploreASL/ExploreASL/Functions/xASL_Iteration.m', 100)" style="font-weight:bold">xASL_Iteration</a> (<a href="matlab: opentoline('/Users/henk/ExploreASL/ExploreASL/Functions/xASL_Iteration.m',100,0)">line 100</a>)
  In <a href="matlab:matlab.internal.language.introspective.errorDocCallback('ExploreASL_Master', '/Users/henk/ExploreASL/ExploreASL/ExploreASL_Master.m', 141)" style="font-weight:bold">ExploreASL_Master</a> (<a href="matlab: opentoline('/Users/henk/ExploreASL/ExploreASL/ExploreASL_Master.m',141,0)">line 141</a>)] 
```

### Error
An error starts with `ERROR: Job iteration terminated!` after which a similar syntax follows as with the warning, and it ends with `CONT: but continue with next iteration!
Example:

```
ERROR: Job iteration terminated!

ans =


Error using <a href="matlab:matlab.internal.language.introspective.errorDocCallback('xASL_module_Structural', '/Users/henk/ExploreASL/ExploreASL/Modules/xASL_module_Structural.m', 198)" style="font-weight:bold">xASL_module_Structural</a> (<a href="matlab: opentoline('/Users/henk/ExploreASL/ExploreASL/Modules/xASL_module_Structural.m',198,0)">line 198</a>)
piet

Error in <a href="matlab:matlab.internal.language.introspective.errorDocCallback('xASL_Iteration>runIteration', '/Users/henk/ExploreASL/ExploreASL/Functions/xASL_Iteration.m', 361)" style="font-weight:bold">xASL_Iteration>runIteration</a> (<a href="matlab: opentoline('/Users/henk/ExploreASL/ExploreASL/Functions/xASL_Iteration.m',361,0)">line 361</a>)
				[result, x] = jobfn(x); % [result, x] = jobfn(x, job_ex);

Error in <a href="matlab:matlab.internal.language.introspective.errorDocCallback('xASL_Iteration', '/Users/henk/ExploreASL/ExploreASL/Functions/xASL_Iteration.m', 100)" style="font-weight:bold">xASL_Iteration</a> (<a href="matlab: opentoline('/Users/henk/ExploreASL/ExploreASL/Functions/xASL_Iteration.m',100,0)">line 100</a>)
	[bAborted, xOut] = runIteration(dbSettings);

Error in <a href="matlab:matlab.internal.language.introspective.errorDocCallback('ExploreASL_Master', '/Users/henk/ExploreASL/ExploreASL/ExploreASL_Master.m', 106)" style="font-weight:bold">ExploreASL_Master</a> (<a href="matlab: opentoline('/Users/henk/ExploreASL/ExploreASL/ExploreASL_Master.m',106,0)">line 106</a>)
        [~, x] = xASL_Iteration(x,'xASL_module_Structural');

Error in <a href="matlab:matlab.internal.language.introspective.errorDocCallback('ExploreASL', '/Users/henk/ExploreASL/ExploreASL/ExploreASL.m', 33)" style="font-weight:bold">ExploreASL</a> (<a href="matlab: opentoline('/Users/henk/ExploreASL/ExploreASL/ExploreASL.m',33,0)">line 33</a>)
x = ExploreASL_Master(var1, var2, var3, var4, var5, var6);


CONT: but continue with next iteration!
```

Important to note that we should delete the log-files before rerunning a test.

MichaelStritt:
Hey @HenkMutsaerts & @jan-petr 

Can you give this a try?

I added a comment to issue #326 as well.

Thanks!


HenkMutsaerts (changes requested):
Very nice!

First review (will test tomorrow):

%        rootDir            - Case root directory (OPTIONAL)
-> OPTIONAL should always have a DEFAULT (in this case, it defaults to a user input).
-> Note that uigetdir requires JVM and will crash when using no GUI; see how I fixed this in the `ExploreASL_Initialize`, that you allow a Linux dialogue instead of GUI browser, if needed.

* I would output to JSON or TSV using our usual workflow, the Table is something Matlab specific, not sure if this easily saves to a JSON or TSV. We want it soon to be communicating back to GitHub so human readable/non-Matlab is probably preferred :)

* The `nargin <` should also have ` || isempty() as input check; such that you can input a third input argument without having to specify the second argument but omitting this by `[]`

* Better to use `xASL_adm_GetFileList` with `dir` option instead of `dir`, our xASL function has better cross-platform performance etc.

* `curFileArr` -> good practice these days to use full words, to increase code readability

* `contains(lower(` could be replaced by `regexpi`

* we try to avoid `break`,`continue` etc for readability for debugging, here it should be fine

* We may want to find a different prefix, `qc` is more used for visual/image quality control, whereas this is more `code/pipeline` quality control. Should we have a new prefix called `test`?

* Nice that you created `xASL_qc_RemoveLogFilesForRerun` ! Wasn't my intention, also this already happens inside `xASL_adm_CleanUpBeforeRerun`. It was more a caution warning :)

MichaelStritt:
> % rootDir - Case root directory (OPTIONAL)
> -> OPTIONAL should always have a DEFAULT (in this case, it defaults to a user input).
> -> Note that uigetdir requires JVM and will crash when using no GUI; see how I fixed this in the `ExploreASL_Initialize`, that you allow a Linux dialogue instead of GUI browser, if needed.

CLI user input is implemented in 1298684 👍 

> I would output to JSON or TSV using our usual workflow, the Table is something Matlab specific, not sure if this easily saves to a JSON or TSV. We want it soon to be communicating back to GitHub so human readable/non-Matlab is probably preferred :)

TSV export is implemented in 1298684 👍 

> The `nargin <` should also have `|| isempty() as input check; such that you can input a third input argument without having to specify the second argument but omitting this by`[]`

Done 👍 

> Better to use `xASL_adm_GetFileList` with `dir` option instead of `dir`, our xASL function has better cross-platform performance etc.

I know we had this discussion before. I use `xASL_adm_GetFileList` whenever possible, but `dir` works recursively and `xASL_adm_GetFileList` not, as far as I know. This is the reason I had to use `dir` here.

> `curFileArr` -> good practice these days to use full words, to increase code readability

Done 👍 

> `contains(lower(` could be replaced by `regexpi`

I don't see any benefits.

> We may want to find a different prefix, `qc` is more used for visual/image quality control, whereas this is more `code/pipeline` quality control. Should we have a new prefix called `test`?

I know what you mean. This is more manual debugging and stuff, but it's really not testing. If you insist on it, we can change the name though. 

> Nice that you created `xASL_qc_RemoveLogFilesForRerun` ! Wasn't my intention, also this already happens inside `xASL_adm_CleanUpBeforeRerun`. It was more a caution warning :)

Ah okay, then I understood you wrong in your original message. If you want we can delete `xASL_qc_RemoveLogFilesForRerun`. Otherwise I'd merge this.

HenkMutsaerts:
* `xASL_adm_GetFileList` does work recursively, check its file header, there is an option 'FPListRec' 

* The header options should express the input format (e.g. Booleans in your case)

* "logContent.tsv" could be more specific, including the function names, that it actually parses warnings/errors, if you like. We will have many log files in the end ;)

* `contains(lower())` is two functions and less backward compatible than `regexpi`, which has been along for ages. `regexpi` is also more flexible, if we need to add more cases for string containment; hence it's just better practice :)

* We can try a different name, I'm fine with any other name? I thought the idea is actually for testing (as well). Hence why I would make a TSV file output as default. In the end, this will all be done automatically, and the output of this function sent back to Github Actions, wouldn't that be cool? ;)

* Yes, we can probably keep this function `xASL_qc_RemoveLogFilesForRerun`, it's quite nifty to have anyway perhaps for users. However, I would still change the prefix `qc` also here.

HenkMutsaerts (changes requested):
* `xASL_adm_GetFileList` does work recursively, check its file header, there is an option 'FPListRec' 

* The header options should express the input format (e.g. Booleans in your case)

* "logContent.tsv" could be more specific, including the function names, that it actually parses warnings/errors, if you like. We will have many log files in the end ;)

* `contains(lower())` is two functions and less backward compatible than `regexpi`, which has been along for ages. `regexpi` is also more flexible, if we need to add more cases for string containment; hence it's just better practice :)

* We can try a different name, I'm fine with any other name? I thought the idea is actually for testing (as well). Hence why I would make a TSV file output as default. In the end, this will all be done automatically, and the output of this function sent back to Github Actions, wouldn't that be cool? ;)

* Yes, we can probably keep this function `xASL_qc_RemoveLogFilesForRerun`, it's quite nifty to have anyway perhaps for users. However, I would still change the prefix `qc` also here.

MichaelStritt:
> The header options should express the input format (e.g. Booleans in your case)

Done 👍 

> "logContent.tsv" could be more specific, including the function names, that it actually parses warnings/errors, if you like. We will have many log files in the end ;)

The table and the exported TSV now include corresponding filename and line number. I think this should work now, as long as Matlab doesn't change the warning or error formatting. 👍 

> contains(lower()) is two functions and less backward compatible than regexpi, which has been along for ages. regexpi is also more flexible, if we need to add more cases for string containment; hence it's just better practice :)

Yeah... sorry, but I really hate writing regular expressions. I know `contains` is 2016b and newer. I implemented a method with `strfind` & `lower`, this should work for 2006a and newer.

> xASL_adm_GetFileList does work recursively, check its file header, there is an option 'FPListRec'

It uses `xASL_adm_GetFileList` now 👍 

> We can try a different name, I'm fine with any other name? I thought the idea is actually for testing (as well). Hence why I would make a TSV file output as default. 

Changed the prefix to `test` and the TSV output default to true 👍 

> Yes, we can probably keep this function xASL_qc_RemoveLogFilesForRerun, it's quite nifty to have anyway perhaps for users. However, I would still change the prefix qc also here.

Changed to `adm` 👍

MichaelStritt:
See comments in #326

HenkMutsaerts (changes requested):
Very nice!

Can you fix the following minor issues:

### `xASL_adm_RemoveLogFilesForRerun`
* replace `dir` by `xASL_adm_GetFileList`
* replace `delete` by `xASL_delete`
* input check needs `isempty()` as well

### `xASL_test_GetLogContent`
* I don't understand the Matlab checks, what you use doesn't look like something that would break before 2017...
* Why do you have both `sub` and `population`? Why is this version dependent? (either this should be clear from the comments or you need to tell me ;)
* `writetable` was introduced in 2016, and not sure if it would work when running without JVM. Also principally I am much more in favor of `.tsv`, nothing can go wrong there, whereas `.xlsx` is designed for higher level data, can take longer, assumes the subject can read `xlsx` (e.g., we will use this script for getting feedback back to Github, which won't be able to read `xlsx` right? Unless there are specific advantages for `.xlsx`, as Excel can read `.tsv` easily with the same layout... (unless of course you have some specific layout?)
* The code looks very bulky, having both before and after 2017 code with minor changes (easy to write one code that works in both cases right?)
* A real programmer like you should use regexp, it is designed for a reason ;)

MichaelStritt:
> Can you fix the following minor issues:
> #### `xASL_adm_RemoveLogFilesForRerun`
> 
>     * replace `dir` by `xASL_adm_GetFileList`
> 
>     * replace `delete` by `xASL_delete`
> 
>     * input check needs `isempty()` as well

Done in 69abbf0 

> #### `xASL_test_GetLogContent`
> 
> I don't understand the Matlab checks, what you use doesn't look like something that would break before 2017...

`contains` was introduced in 2016b

> Why do you have both `sub` and `population`? Why is this version dependent? (either this should be clear from the comments or you need to tell me ;)

Because multiple subjects can be part of the same population? You only have a list of files from a BIDS directory and it worked nicely on the results of the full bids pipeline test.
In the original implementation I used `contains`. We can stick to the old version or the new version, I don't care.

> `writetable` was introduced in 2016, and not sure if it would work when running without JVM. Also principally I am much more in favor of `.tsv`, nothing can go wrong there, whereas `.xlsx` is designed for higher level data, can take longer, assumes the subject can read `xlsx` (e.g., we will use this script for getting feedback back to Github, which won't be able to read `xlsx` right? Unless there are specific advantages for `.xlsx`, as Excel can read `.tsv` easily with the same layout... (unless of course you have some specific layout?)

On default you get a TSV. It is just an option to export an XLSX. You don't have to. I don't understand the problem here.

> The code looks very bulky, having both before and after 2017 code with minor changes (easy to write one code that works in both cases right?)

Yeah, depends on the `contains` thing. You can always use the old way to do it or we stick to 2017 and newer.

> A real programmer like you should use regexp, it is designed for a reason ;)

I'm an engineer and we are known for getting things to work. If there is no benefit of using regexp I wont use it. If I have to validate an e-mail address which is parsed from a JavaScript input form to a PHP script running on some server, then of course I use regular expressions. That's kind of telling someone else to use a different knife to cut an apple, even if the first knife worked perfectly fine.

MichaelStritt:
@HenkMutsaerts: I did a revamp of some parts of the code. The messages should have a nice style now. Hope it works on every OS. Check out the example files below. 

[logContent.zip](https://github.com/ExploreASL/ExploreASL/files/6020905/logContent.zip)

Functions/xASL_adm_RemoveLogFilesForRerun.m
```diff
@@ -0,0 +1,49 @@
+function xASL_adm_RemoveLogFilesForRerun(rootDir)
+%xASL_adm_RemoveLogFilesForRerun Removes all log files from xASL directory
```
jan-petr, line 2: Better call it something else - xASL directory can be confused with the ExploreASL code. Directory with ASL data in Legacy format, ASL data in BIDS? - Not sure which of those two applies.

Functions/xASL_adm_RemoveLogFilesForRerun.m
```diff
@@ -0,0 +1,49 @@
...
+% FORMAT: xASL_adm_RemoveLogFilesForRerun(rootDir);
+%
+% INPUT:
+%        rootDir            - Case root directory
+%
+% OUTPUT:
+%        n/a
+%
+% -----------------------------------------------------------------------------------------------------------------------------------------------------
+% DESCRIPTION:      Removes all log files from xASL directory.
```
jan-petr, line 13: Same as above.

Functions/xASL_adm_RemoveLogFilesForRerun.m
```diff
@@ -0,0 +1,49 @@
+function xASL_adm_RemoveLogFilesForRerun(rootDir)
+%xASL_adm_RemoveLogFilesForRerun Removes all log files from xASL directory
+%
+% FORMAT: xASL_adm_RemoveLogFilesForRerun(rootDir);
+%
+% INPUT:
+%        rootDir            - Case root directory
```
jan-petr, line 7: Root directory of the data (REQUIRED)

Functions/xASL_adm_RemoveLogFilesForRerun.m
jan-petr, line 2: Better call it something else - xASL directory can be confused with the ExploreASL code. Directory with ASL data in Legacy format, ASL data in BIDS? - Not sure which of those two applies.

Functions/xASL_adm_RemoveLogFilesForRerun.m
jan-petr, line 13: Same as above.

Functions/xASL_adm_RemoveLogFilesForRerun.m
```diff
@@ -0,0 +1,49 @@
...
+% EXAMPLE:          rootDir = '.\Test_Runs\TestDataSet';
+%                   xASL_adm_RemoveLogFilesForRerun(rootDir);
+%
+% REFERENCES:       ...
+% __________________________________
+% Copyright @ 2015-2021 ExploreASL
+
+    %% Input Check
+    if nargin < 1 || isempty(rootDir)
+        error('Missing xASL directory...');
```
jan-petr, line 26: Same as above.

Functions/xASL_test_GetLogContent.m
```diff
@@ -0,0 +1,308 @@
+function [logContent] = xASL_test_GetLogContent(rootDir, printContent, storeRelativePath, exportTable)
+%xASL_test_GetLogContent Get warnings and errors from log files
+%
+% FORMAT: [logContent] = xASL_test_GetLogContent(rootDir, [printContent], [storeRelativePath], [exportTable])
```
jan-petr, line 4: In other files, we wrote - not sure how relevant
[printContent, storeRelativePath, exportTable]
instead of 
[printContent], [storeRelativePath], [exportTable]

Functions/xASL_test_GetLogContent.m
```diff
@@ -0,0 +1,308 @@
+function [logContent] = xASL_test_GetLogContent(rootDir, printContent, storeRelativePath, exportTable)
```
jan-petr, line 1: For a broader discussion. 
We have _ut_ for unit tests and _test_ for other tests? Should we have these two or unify under a single prefix. Or it doesn't make sense as this is not really part of the unit testing?

Functions/xASL_test_GetLogContent.m
```diff
@@ -0,0 +1,308 @@
...
+%
+% OUTPUT:
+%        logContent         - table containing warnings and errors
+%
+% -----------------------------------------------------------------------------------------------------------------------------------------------------
+% DESCRIPTION:      Get warnings and errors from log files.
+%
+% -----------------------------------------------------------------------------------------------------------------------------------------------------
+%
+% EXAMPLE:          % Run the GetLogContent script, do not print the messages, store the full path and export an XLSX
```
jan-petr, line 21: Very good to describe what actually this example does! We haven't done that in most other functions :)

jan-petr (changes requested):
Works very good - already saves a lot of work during testing. 

The format is not optimal - as we discussed - but we can look at this later (can you create a new issue for that please). 

I made more comments in the code - lets see if it is easy to work with that kind of commenting. 

Three more issues (don't have to be fixed now - can be added to the new issue mentioned above)
1) ExploreASL_Import creates these kind of files in the analysis folder 'import_log_Philips_PCASL_3DGRASE_R5.4_TopUp_20210217_191213.txt' - it would be good to read these as well. You can probably also change the name upon saving in the import module if this will make your work easier.

2. Inside that file, there are both standard Matlab errors and warnings, but also normal text I produce for the conversion. Example:
"Warning: Empty protocol name(s) (0018,1030)"
It would be good to collect those as well. We can give a rule that if this is found, starting with "Warning:" it reads and outputs a single line. It is only warnings, no errors.

3. When writing a TSV, it should delete XLSX and vice versa - so that we don't have both with possibly different content (if one called later).

MichaelStritt:
> For a broader discussion.
> We have ut for unit tests and test for other tests? Should we have these two or unify under a single prefix. Or it doesn't make sense as this is not really part of the unit testing?

That's a question for Henk. First I just used QC, because I thought we mostly use this for our own quality control. Henk suggested to use test. I'm fine with both or something completely different, but I wouldn't use ut.

MichaelStritt:
> ExploreASL_Import creates these kind of files in the analysis folder 'import_log_Philips_PCASL_3DGRASE_R5.4_TopUp_20210217_191213.txt' - it would be good to read these as well. You can probably also change the name upon saving in the import module if this will make your work easier.

@jan-petr: can you send me one of those files, so that I can try it? I should have some as well, but they have no errors and it would be cool if I don't have to reproduce errors right now

jan-petr:
> > ExploreASL_Import creates these kind of files in the analysis folder 'import_log_Philips_PCASL_3DGRASE_R5.4_TopUp_20210217_191213.txt' - it would be good to read these as well. You can probably also change the name upon saving in the import module if this will make your work easier.
> 
> @jan-petr: can you send me one of those files, so that I can try it? I should have some as well, but they have no errors and it would be cool if I don't have to reproduce errors right now

[import_log_Philips_PCASL_3DGRASE_R5.4_TopUp_20210217_191213.txt](https://github.com/ExploreASL/ExploreASL/files/6021989/import_log_Philips_PCASL_3DGRASE_R5.4_TopUp_20210217_191213.txt)

jan-petr:
Another error:
ExploreASL_Master, line 152
====================================================================================================
Error using fprintf
Function is not defined for 'cell' inputs.

Error in xASL_tsvWrite (line 76)
            fprintf(FileID,DelimiterIs, xASL_num2str(InputCell{iX,iY}));

Error in xASL_test_GetLogContent (line 149)
        xASL_tsvWrite(logContentCell, fullfile(rootDir,'logContent.tsv'), true);
 
 With the attached log file
[xASL_module_Population.log](https://github.com/ExploreASL/ExploreASL/files/6022025/xASL_module_Population.log)

MichaelStritt:
@jan-petr: Most of it should be fixed with 201e397. There was a bug with the export of the "main message" if there are multiple errors/warnings in one log file.

MichaelStritt:
@jan-petr: good lord this took longer than expected. The problem with some of these error and warning messages is that they have no clear "end" identifier. With the ones in our main pipeline we always know that the last script mentioned in the message is the ExploreASL Master script, so we can find the end there pretty easily.
Here we have these custom warnings which just begin with "warning: " and the end of the message can look different each time. And then there are these warnings which are normal Matlab warnings, but hey don't start with "warning: " or are executed from another workflow (import e.g.). All of this leads to code that's a bit messy. For me it worked though.

There's an example attached below:

[logContent.zip](https://github.com/ExploreASL/ExploreASL/files/6023572/logContent.zip)

jan-petr (approved):
Great. It works perfectly with all my files. No crashes. It can handle the weird warnings from dcm2nii and so on. So ready to merge.

There are some minor things where the formatting might improve. I have listed them in another issue - but we will have to see if these are fixable without affecting the functionality.

Functions/xASL_adm_RemoveLogFilesForRerun.m
```diff
@@ -0,0 +1,49 @@
...
+    %% Get all log files within root directory
+    fileList = xASL_adm_GetFileList(rootDir, '^.+\.log$', 'FPListRec');
+    
+    %% Iterate over log files
+    for iFile=1:numel(fileList)
+        % Get current file path
+        curFile = fileList{iFile};
+        % Delete file
+        fprintf('Delete: %s\n',curFile);
+        xASL_delete(curFile)
```
HenkMutsaerts, line 39: * `xASL_adm_RemoveLogFilesForRerun`: line 31 `xASL_delete(curFile)` misses a `;`

Functions/xASL_adm_RemoveLogFilesForRerun.m
```diff
@@ -0,0 +1,49 @@
+function xASL_adm_RemoveLogFilesForRerun(rootDir)
+%xASL_adm_RemoveLogFilesForRerun Removes all log files from a directory containing .log files.
+%
+% FORMAT: xASL_adm_RemoveLogFilesForRerun(rootDir);
+%
+% INPUT:
+%        rootDir            - Case root directory (REQUIRED)
```
HenkMutsaerts, line 7: * What do you mean by `Case` (input `rootDir`?)

Functions/xASL_test_GetLogContent.m
HenkMutsaerts: Good idea to discuss this elsewhere;)

HenkMutsaerts (changes requested):
Looks very good! Just a few stylistic changes:

* The function `xASL_test_GetLogContent` is very long and unreadable. Can you either split it into separate shorter functions, or sections, that you refer to in the description?
Look at other scripts for examples, such as the structure of `xASL_init_FileSystem`.  Someone unfamiliar with this function should be able to quickly read what it does.
* Let's see when using this function if the exported format is nice to use, I still vouch for a human-readable format such as JSON or TSV instead of Matlab Tables ;)

jan-petr:
> Looks very good! Just a few stylistic changes:
> 
> * The function `xASL_test_GetLogContent` is very long and unreadable. Can you either split it into separate shorter functions, or sections, that you refer to in the description?
>   Look at other scripts for examples, such as the structure of `xASL_init_FileSystem`.  Someone unfamiliar with this function should be able to quickly read what it does.

Yes - '%%' and Section headers like in xASL_init_FileSystem will make it easier to grasp. 

> * Let's see when using this function if the exported format is nice to use, I still vouch for a human-readable format such as JSON or TSV instead of Matlab Tables ;)

For me, the printout in standard output looks pretty good for a quick check. And the TSV/XLSX **are** produced and are fine. I would keep the function output in Matlab-Tables - that can be handy for collecting output for several directories and then printing it jointly somewhere.

MichaelStritt:
@HenkMutsaerts & @jan-petr: I'm out of office today and won't be able to work on it tomorrow. If it's okay like it is right now, I'd merge it for 1.5.0, so that we can test on Thursday/Friday.

MichaelStritt:
@jan-petr: I'd do a signed squash merge. We don't need these individual commits anyway and then Beatriz commit should be fixed.

MichaelStritt:
@HenkMutsaerts can you approve it like this?

jan-petr:
@MichaelStritt Very good point. A completely new function - a squash-merge is a good choice. I've reactivated the review from Henk and put it to his folder.

jan-petr (approved):
OK.

PR #333, opened 2021-02-15
Closes #306 PhaseEncodingAxis

### LINKED ISSUE #306: JSON encoding error: PhaseEncodingAxis "i" -> 0+1i
The PhaseEncodingAxis "i" is recognized incorrectly as numerical (imaginary number, 0+1i) instead of a char array "i", either by `dcm2niiX` or by our own `ExploreASL_Import: parms2json` functionality? This char is meant to denote that axis "i" is used for phase encoding (rather than "j" or "k").

* I will fix this in the ASL and M0 sidecars in `/External/TestDataSet/`
* @MDijsselhof you observed this as well right?
* @jan-petr can you fix this?

MichaelStritt:
See notes in #306

### DIFF
--- a/External/TestDataSet/analysis/Sub-001/ASL_1/ASL4D.json
+++ b/External/TestDataSet/analysis/Sub-001/ASL_1/ASL4D.json
@@ -26,7 +26,7 @@
 "AcquisitionMatrixPE":62,
 "ReconMatrixPE":64,
 "PixelBandwidth":3054,
-"PhaseEncodingAxis":0+1i,
+"PhaseEncodingAxis":"i",
 "ImageOrientationPatientDICOM":[0.988042,-0.143706,0.0558729,0.154183,0.918592,-0.363889],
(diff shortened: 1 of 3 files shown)


jan-petr (changes requested):
What if the direction '-i' is read as 0-i - we might check and fix that option as well - for the future.

MichaelStritt:
> What if the direction '-i' is read as 0-i - we might check and fix that option as well - for the future.

Added it in 6c12710 👍

jan-petr:
> > What if the direction '-i' is read as 0-i - we might check and fix that option as well - for the future.
> 
> Added it in [6c12710](https://github.com/ExploreASL/ExploreASL/commit/6c1271043616b7627dcf367d8e0853d3e24ff491) 
> 
> Edit: Hope this works for you, I personally never experienced this bug.

OK. Thanks - I haven't experienced it either. But if the +i is a problem, then -i has to be as well, in theory :)

jan-petr (approved):
OK.

PR #340, opened 2021-02-17
#339 Fixed MEX spm_jsonread problem with UTF-8 

### LINKED ISSUE #339: spm_jsonread cannot read special characters saved by spm_jsonwrite
Philips_PCASL_2DEPI_Bsup_AD4
the ASL.json
The error is in saving - if saved in a different program, it works, if saved with spm_jsonwrite, then spm_jsonread cannot read it after itself. It is this character that breaks everything ö

jan-petr:
Close #339  Please compile for windows and test on the simple JSON example with UTF-8 characters - it should crash before fix, but work after the fix.


MichaelStritt (approved):
See comments in #339

PR #358, opened 2021-02-23
Feature #354 xASL_io_SplitASL_M0: Remove asl dummy

### LINKED ISSUE #354: Option to skip dummy ASL scans
Create an option for skipping ASL scans.
These can be used to skip the dummy and noise scans Wibeke has.

### Release summary
Added an option x.DummyPositionInASL4D that removes marked dummy scans when splitting ASL to ASL+M0+dummy

jan-petr:
Closes #354

### DIFF
--- a/DataParTemplate.m
+++ b/DataParTemplate.m
@@ -88,6 +88,13 @@ function x = DataParTemplate(x)
 %                     - example for Philips 3D GRASE = '[1 2]' % (first control-label pair)
 %                     - example for Siemens 3D GRASE = 1 % first image
 %                     - example for GE 3D spiral = 2 % where first image is PWI & last = M0
+% x.DummyScanPositionInASL4D - indicates the position of Dummy scans in TimeSeries if they are integrated by the vendor in the 
+%                              DICOM export. This allows to remove the dummy scans or noise scans that are part of the Timeseries.
+%                              A new ASL4D.nii is saved with dummy scans removed and the original is backed-up. Works in a similar way 
+%                              as M0PositionInASL4D, both can be entered at the same time and both indicate the original position 
+%                              in the Timeseries independend of each other. (OPTIONAL, DEFAULT = [] (no M0 in timeseries))
+%                            - example for Siemens 2D EPI = '[79 80]' % Skip the control-label pair used for noise measurements
+%                            - example for certain Siemens 3D GRASE = 2 % Skip the first dummy control image
 
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
 % SEQUENCE PARAMETERS
@@ -316,6 +323,7 @@ x.name = ExampleDataSet;
 x.subject_regexp = '^Sub-\d{3}$';
 x.M0 = 'separate_scan';
 x.M0PositionInASL4D = '[1 2]';
+x.DummyScanPositionInASL4D = [];
 x.readout_dim = '2D';
 x.Quality = 0;
 x.DELETETEMP = 1;
--- a/DataParTemplate.m
+++ b/DataParTemplate.m
@@ -82,19 +82,21 @@ function x = DataParTemplate(x)
 %                      because ExploreASL pragmatically smooths the M0 a lot, assuming that
 %                      head motion and registration between M0 & ASL4D will differ between
 %                      patients and controls. 
-% x.M0PositionInASL4D - indicates the position of M0 in TimeSeries, if it is integrated by the vendor in the 
+% x.M0PositionInASL4D - A vector of integers that indicates the position of M0 in TimeSeries, if it is integrated by the vendor in the 
 %                       DICOM export. Will move this from ASL4D.nii to M0.nii(OPTIONAL, DEFAULT = [] (no M0 in timeseries))
 %                     - Note that the x.M0PositionInASL4D parameter is independent from the x.M0 parameter choice.
 %                     - example for Philips 3D GRASE = '[1 2]' % (first control-label pair)
 %                     - example for Siemens 3D GRASE = 1 % first image
 %                     - example for GE 3D spiral = 2 % where first image is PWI & last = M0
-% x.DummyScanPositionInASL4D - indicates the position of Dummy scans in TimeSeries if they are integrated by the vendor in the 
+%                     - Empty vector should be given (= [] or = null (in JSON)) if no action is to be taken and nothing is removed
+% x.DummyScanPositionInASL4D - A vector of integers that indicates the position of Dummy scans in TimeSeries if they are integrated by the vendor in the 
 %                              DICOM export. This allows to remove the dummy scans or noise scans that are part of the Timeseries.
 %                              A new ASL4D.nii is saved with dummy scans removed and the original is backed-up. Works in a similar way 
 %                              as M0PositionInASL4D, both can be entered at the same time and both indicate the original position 
 %                              in the Timeseries independend of each other. (OPTIONAL, DEFAULT = [] (no M0 in timeseries))
-%                            - example for Siemens 2D EPI = '[79 80]' % Skip the control-label pair used for noise measurements
+%                            - example for Siemens 2D EPI = [79 80] % Skip the control-label pair used for noise measurements
 %                            - example for certain Siemens 3D GRASE = 2 % Skip the first dummy control image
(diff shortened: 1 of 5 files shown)


DataParTemplate.m
```diff
@@ -87,6 +87,13 @@
 %                     - example for Philips 3D GRASE = '[1 2]' % (first control-label pair)
 %                     - example for Siemens 3D GRASE = 1 % first image
 %                     - example for GE 3D spiral = 2 % where first image is PWI & last = M0
+% x.DummyScanPositionInASL4D - indicates the position of Dummy scans in TimeSeries if they are integrated by the vendor in the
```
HenkMutsaerts, line 90: here the format of the parameter is missing (numerical, scalar or vector)

Functions/xASL_io_SplitASL_M0.m
```diff
@@ -1,32 +1,61 @@
-function xASL_io_SplitASL_M0(InPath,iM0)
-%xASL_io_SplitASL_M0 Splits ASL & M0 images, when they are within the same NIfTI
+function xASL_io_SplitASL_M0(inPath, iM0, iDummy)
```
HenkMutsaerts, line 1: Would rename this function to `xASL_io_SplitASL`

Functions/xASL_io_SplitASL_M0.m
```diff
@@ -1,32 +1,61 @@
...
+%              The dummy scans can be excluded from the ASL sequence during the splitting. Both iM0 and iDummy
+%              are the absolute positions of both in the original time series
 %
 %              Vendor product sequence examples:
 %              GE 3D spiral sometimes puts the M0 at the last volume of the series -> iM0 = [2];
 %              Philips 3D GRASE puts the M0 as control-label volume pair -> iM0 = [1 2];
 %              Siemens 3D GRASE puts the M0 as the first volume -> iM0 = 1;
+%              Some Siemens 3D GRASE puts a second Dummy control image -> iDummy = 2;
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
 % EXAMPLE:
```
HenkMutsaerts, line 25: This should be:
`%EXAMPLE for moving the first two volumes to M0.nii`

Functions/xASL_io_SplitASL_M0.m
```diff
@@ -1,32 +1,61 @@
...
 % xASL_io_SplitASL_M0('/data/RAD/share/EPAD500/010EPAD00001/ASL_1/ASL4D.nii', [1 2]);
 % EXAMPLE for concatenating files only:
 % xASL_io_SplitASL_M0('/data/RAD/share/EPAD500/010EPAD00001/ASL_1/ASL4D.nii', []);
+% EXAMPLE for both excluding the Dummy scans and splitting M0:
```
HenkMutsaerts, line 29: This is incorrect right, in this example, the M0 is not moved but only the dummy volume is removed?
Also, I would make it more specific by writing "removed" rather than "excluded".

Functions/xASL_io_SplitASL_M0.m
```diff
@@ -1,32 +1,61 @@
...
 
+    %% -----------------------------------------------------------------------------------------------------------------------------------------------------
+    %% Input parameter admin
+	% Can be char or number on input, output should be a horizontal vector
 
-    [Fpath, Ffile] = xASL_fileparts(InPath);
+	if nargin < 2 || isempty(iM0)
+		iM0 = [];
+	else
+		if ischar(iM0)
```
HenkMutsaerts, line 41: you can probably omit the `if ischar` as `xASL_str2num` has this check built-in

Functions/xASL_io_SplitASL_M0.m
```diff
@@ -1,32 +1,61 @@
...
+			iM0 = xASL_str2num(iM0);
+		end
+		iM0 = iM0(:)';
```
HenkMutsaerts, line 44: This could allow a matrix instead of vector right?

Functions/xASL_io_SplitASL_M0.m
HenkMutsaerts: Same below

Modules/xASL_module_ASL.m
```diff
@@ -92,7 +92,9 @@
 end
 
 % Make sure to split the M0 from the ASL time-series if needed
-xASL_io_SplitASL_M0(x.P.Path_ASL4D, x.M0PositionInASL4D, x.DummyScanPositionInASL4D);
+if ~isempty(x.M0PositionInASL4D) || ~isempty(x.DummyScanPositionInASL4D)
```
HenkMutsaerts, line 95: Is this part close to settings the default for these parameters (should also be the beginning of `xASL_module_ASL`)? Just to avoid bulky code

HenkMutsaerts (approved):
Few minor things, looks good otherwise

DataParTemplate.m
jan-petr: Updated of M0Position. And added an example for null entry (as discussed with Maurice).

Functions/xASL_io_SplitASL_M0.m
jan-petr: Agree. Done.

Functions/xASL_io_SplitASL_M0.m
jan-petr: Done.

Functions/xASL_io_SplitASL_M0.m
jan-petr: Good point. Updated examples.

Functions/xASL_io_SplitASL_M0.m
jan-petr: Agree. Thanks.

Functions/xASL_io_SplitASL_M0.m
jan-petr: yep a matrix will be converted to a vector. This should be fine, right?

Modules/xASL_module_ASL.m
jan-petr: Defaults are defined above. This was only checking if it makes sense to run the function or not. But this is now moved inside this wrapper. If no indices are provided, it will not continue running the SplitASL function.

jan-petr:
All comments fixed. Check the two unresolved comments.

Functions/xASL_io_SplitASL_M0.m
HenkMutsaerts: No, my point was that a matrix should give a warning or error. So it is fine to force the vector direction (horizontal vs vertical), but this way you explicitly allow a wrong input (i.e. a matrix).

E.g., if iMo = [1 2 3;1 2 3] it will become [1 2 3 1 2 3] without a warning or error

DataParTemplate.m
HenkMutsaerts: What does "Updated of M0Position" mean? ;)

The null entry is fine if it is JSON and ExploreASL `[]` standard, I would only ensure that advanced parameters are not in the example JSON file.

jan-petr: Haha - that's a fusion of "Update of M0Position" and "Updated M0Position" and I wanted to say that I have updated the description of the M0PositionInASL4D. It was late :) Comment sloppy, update itself not sloppy :)

Agree - DummyPositionInASL4D wasn't added to the example JSON.

Functions/xASL_io_SplitASL_M0.m
jan-petr: Yes, good point. If size(iM0,1)>1 && size(iM0,2)>1 then a warning... 
I'll update that.

jan-petr:
All should be fixed. Need your formal approval before merging.

HenkMutsaerts (approved):
Great, yes this was rather something I noted. We should do a formal input parameter check as Pieter suggested before. Let's put this on the list :)

jan-petr:
> Great, yes this was rather something I noted. We should do a formal input parameter check as Pieter suggested before. Let's put this on the list :)

Yes, we should. But we do not suck so much in parameter checking anymore. So other things - unit testing, Flavors testing, new features have much higher priority :)

Issue #359, opened 2021-02-23
BIDS for post-contrast T1w


jan-petr:
Post-contrast T1w images are imported in xASL legacy format as T1c. However, in BIDS they are normally labeled as t1w and have an additional ce-<label> where label is the name of the contrast agent. We need to make an import of DICOM2BIDS and BIDS2Legacy to work with those.

PR #360, opened 2021-02-24
Feature #351 processing t2and t1c

### LINKED ISSUE #351: Import and basic processing of T2 and T1Gad
Make sure that the following sequences are imported and go through a basic processing. It only needs to be aligned with T1 (that is aligned with MNI) - nothing more. 
More details to what to fix and where:

- xASL_wrp_LinearReg_T1w2MNI: line 50 (OtherList)
- T2 to T1 registration
- T1Gad - correct BIDS name
- T1Gad import
- T1Gad registration to T1
- xASL_wrp_LinearReg_FLAIR2T1 -> rename into xASL_wrp_LinearReg_Others2T1 (here we add postGd-T1w & T2, but only with the FLAIR coreg the WMH_SEGM)

### Release summary
- T2 and T1c files are now also aligned to the T1w and outputted to standard space.

jan-petr:
Add T2 and T1c (post-gadolinium T1) to ExploreASL.

It will now handle the import of them from DICOM2Legacy and do basic processing on them - alignment with T1 etc. But it won't use them for ROIs or do any evaluation on them

Close #351


Functions/xASL_adm_Dicom2Parms.m
```diff
@@ -93,7 +93,8 @@
 	% this should not give an error, but multiple ASL scale slopes should give an error
 	% Check if FLAIR occurs
 	isASL = true; % by default, assume these are ASL DICOMs
-	if ~isempty(findstr(FileList{1},'FLAIR')) || ~isempty(findstr(FileList{1},'T1'))
```
HenkMutsaerts, line 96: This edit could be replaced by ~isempty(regexpi(FileList{1}, '(FLAIR|T1|T2|T1c)'))

Functions/xASL_adm_Load4DMemMapping.m
```diff
@@ -19,10 +19,10 @@
 
 
 %% Admin, specify for each modality how to treat it
-ModalitiesList  = {'ASL' 'ASL_notScaled' 'ASL_untreated' 'ASL_PVEc'  'ASL_HctCohort'    'ASL_HctCorrInd' 'PWI' 'M0' 'mean_control' 'SD' 'SNR' 'TT' 'PV_pGM' 'PV_pWM'  'FLAIR'   'T1'  'WMH_SEGM'  'c1T1' 'R1'};
```
HenkMutsaerts, line 22: This wrapper/function can be phased out, we don't use it anymore

Functions/xASL_bids_Dicom2JSON.m
```diff
@@ -115,7 +115,8 @@
 	% this should not give an error, but multiple ASL scale slopes should give an error
 	% Check if FLAIR occurs
 	isASL = true; % by default, assume these are ASL DICOMs
-	if ~isempty(findstr(FileList{1},'FLAIR')) || ~isempty(findstr(FileList{1},'T1'))
```
HenkMutsaerts, line 118: Same as above, replace by ~isempty(regexpi())

Functions/xASL_init_FileSystem.m
```diff
@@ -50,6 +50,9 @@
 xASL_adm_CreateDir(x.D.FLAIR_CheckDir);
 % xASL_adm_CreateDir(x.D.FLAIR_REGDIR);
 
+xASL_adm_CreateDir(x.D.T1c_CheckDir);
+xASL_adm_CreateDir(x.D.T2_CheckDir);
+
 xASL_adm_CreateDir(x.D.ASLCheckDir);
```
HenkMutsaerts, line 56: Need to be careful not to always create these QC folders, 95% of the cases won't have a T2 or T1c.

Modules/xASL_module_Structural.m
```diff
@@ -185,14 +185,30 @@
...
+    if xASL_exist(x.P.Path_FLAIR, 'file') % Fix multiple FLAIR iterations
         IM = xASL_io_Nifti2Im(x.P.Path_FLAIR);
         if size(IM,4)>1 || size(IM,5)>1 || size(IM,6)>1 || size(IM,7)>1
-            warning(['Too many dims, using first: ' x.P.Path_T1]);
+            warning(['Too many dims, using first: ' x.P.Path_FLAIR]);
             xASL_Move(x.P.Path_FLAIR, x.P.Path_FLAIR_ORI, true);
             xASL_io_SaveNifti(x.P.Path_FLAIR, x.P.Path_FLAIR, IM(:,:,:,1,1,1,1), [], false);
         end
-    end
+	end
```
HenkMutsaerts, line 195: Here we iterate the same code multiple times, better to iterate over {x.P.Path_FLAIR x.P.Path_T1 x.P.Path_T2 ...}

Same below; to avoid the code growing out of hand

Modules/xASL_module_Structural.m
```diff
@@ -218,22 +234,29 @@
...
-    if xASL_exist(x.P.Path_FLAIR, 'file')
-
-        xASL_wrp_LinearReg_FLAIR2T1w(x, x.bAutoACPC);
+    if xASL_exist(x.P.Path_FLAIR, 'file') || xASL_exist(x.P.Path_T1c, 'file') || xASL_exist(x.P.Path_T2, 'file')
+		
+		if xASL_exist(x.P.Path_FLAIR, 'file')
+			xASL_wrp_LinearReg_FLAIR2T1w(x, x.bAutoACPC);
+		end
 
+		if xASL_exist(x.P.Path_T1c, 'file') || xASL_exist(x.P.Path_T2, 'file')
```
HenkMutsaerts, line 246: Nice, I would also insert the other admin here (i.e. the folder creation for QC etc.), given how unlikely it is that this part will be run

HenkMutsaerts (changes requested):
* did you check that the names are equal to those used in BIDS?
* we need to make sure that these new scantypes are only added when needed (to avoid having verbose T2-related logging every time ExploreASL runs). Likewise, for the QC folders?
* Also see my in-line comments, I would rewrite to ensure that all the new T2/T1c code is together, since it will be used very rarely.

Functions/xASL_adm_Dicom2Parms.m
jan-petr: The whole thing was useless as isASL parameter is never used :) Deleted completely.

Functions/xASL_adm_Load4DMemMapping.m
jan-petr: This function:
Development/StatsAnalysis/SPM_VBA_Stats/xASL_wrp_spm_GLM.m
was still using it. So I have deleted the ASL_adm_Load4DMemMapping.m file and moved the function to that wrapper. Can be then deleted with it. But can't delete function that is referenced from, even an obscure one, function.

Functions/xASL_bids_Dicom2JSON.m
jan-petr: Same as above - was completely deleted.

Functions/xASL_init_FileSystem.m
jan-petr: Shall we only create when needed? When this would go then?

Modules/xASL_module_Structural.m
jan-petr: Good point. Fixed

Modules/xASL_module_Structural.m
jan-petr: Done.

Functions/xASL_init_FileSystem.m
jan-petr: I see below. Done.

jan-petr:
> * did you check that the names are equal to those used in BIDS?

Yes, T2 is similar to T1. So fine in BIDS and legacy. T1c is difficult, in BIDS, it is sub..._ce-<labeltype>_t1.nii . This will require more complex changes to BIDS import and BIDS2Legacy. I have created a new issue for that, but for the DICOM2Legacy and Legacy processing, I have inserted it already in the simplified T1c form.

> * we need to make sure that these new scantypes are only added when needed (to avoid having verbose T2-related logging every time ExploreASL runs). Likewise, for the QC folders?

Sure - will be added in the import only when specified.

> * Also see my in-line comments, I would rewrite to ensure that all the new T2/T1c code is together, since it will be used very rarely.

Inline comments answered and implemented accordingly - see the last commit.

jan-petr:
Sanneke - missing T1c and T2 files in the Check-directories and in the population folder. I will implement that additionally.

jan-petr:
Fixed also the T1cCheck and T2Check but not sure if correctly.
@HenkMutsaerts Can you check xASL_wrp_VisualCheckCollective_Structural inside xASL_wrp_VisualQC_Structural
It doesn't crash, does OKish job, but not completely what I want.

jan-petr:
> See minor comments in-line.
> Also, is `t1c` the term used in BIDS?

All minor comments fixed - but please make them part of the review (a review comment rather than a simple comment) - because these can be resolved. Those you made now can't.

t1c issue - no, it is more complicated. See issue #359 for that.

Functions/xASL_adm_Dicom2Parms.m
HenkMutsaerts: This was probably used in the past

Functions/xASL_adm_Load4DMemMapping.m
HenkMutsaerts: Yes nice. This spm wrapper is quite useful for doing voxel-wise GLM analyses

Functions/xASL_bids_Dicom2JSON.m
HenkMutsaerts: > > * did you check that the names are equal to those used in BIDS?
> 
> Yes, T2 is similar to T1. So fine in BIDS and legacy. T1c is difficult, in BIDS, it is sub..._ce-_t1.nii . This will require more complex changes to BIDS import and BIDS2Legacy. I have created a new issue for that, but for the DICOM2Legacy and Legacy processing, I have inserted it already in the simplified T1c form.
> 
> > * we need to make sure that these new scantypes are only added when needed (to avoid having verbose T2-related logging every time ExploreASL runs). Likewise, for the QC folders?
> 
> Sure - will be added in the import only when specified.
> 
> > * Also see my in-line comments, I would rewrite to ensure that all the new T2/T1c code is together, since it will be used very rarely.
> 
> Inline comments answered and implemented accordingly - see the last commit.

`T2` should be `T2w`. I know we don't have this for `T1`, but would be good to phase to BIDS as good as possible.

Modules/SubModule_Structural/xASL_wrp_VisualQC_Structural.m
```diff
@@ -166,6 +166,9 @@ function xASL_wrp_VisualQC_Structural(x)
 end
 
 if  xASL_exist(x.P.Path_T2,'file')
+	% It is created only if the T2 file exists
+	xASL_adm_CreateDir(x.D.T2_CheckDir);
```
HenkMutsaerts, line 170: yes and let's make an issue to move this to the CreateFigure function instead. That makes debugging easier.

Modules/xASL_module_Structural.m
```diff
@@ -225,14 +225,6 @@
 		end
 
 		if xASL_exist(x.P.Path_T1c, 'file') || xASL_exist(x.P.Path_T2, 'file')
```
HenkMutsaerts, line 227: Why not moving this check to the beginning of `xASL_wrp_LinearReg_Others2T1w` and removing it here?

HenkMutsaerts (changes requested):

Functions/xASL_bids_Dicom2JSON.m
jan-petr: I made a new issue for that. Good to be close to BIDS, but bad to have T1.nii and T2w.nii. We can synchronize both in the future. Though when thinking about it. We might not need to as the legacy format will be phased out once we move to BIDS, so I would keep this as it is T1.nii and T2.nii.

Modules/SubModule_Structural/xASL_wrp_VisualQC_Structural.m
jan-petr: Created a new issue for that.

Modules/xASL_module_Structural.m
jan-petr: Good idea. Done

HenkMutsaerts (approved):
This looks good. The deletion of `xASL_adm_Load4DMemMapping` shouldn't be in this issue, though this is fine for now :)

line 231 could state "skipping registering them" instead of "skipping", just to add some understanding for the user.

jan-petr:
> This looks good. The deletion of `xASL_adm_Load4DMemMapping` shouldn't be in this issue, though this is fine for now :)
> 
> line 231 could state "skipping registering them" instead of "skipping", just to add some understanding for the user.

OK - will change this message and then merge. Sanneke will be happy - we need to merge with fixes from other branches :)

PR #362, opened 2021-02-24
Closes #355

### LINKED ISSUE #355: Confusing explanation `DataParTemplate.m`
`DataParTemplate.m` has some confusing statements as guideline for the `DataPar.json`. This needs to be checked, e.g., the following is probably wrong (line 22):

>> % Scalars can be inputted as scalars, but vectors need to be inputted as string, e.g. '[1 2]'

MichaelStritt:
Only minor changes. Everything is described in #355


jan-petr (approved):
OK.

PR #365, opened 2021-02-24
#356 Single function for parsing SliceReadoutTime

### LINKED ISSUE #356: Single function for parsing SliceReadoutTime - called at multiple locations
For the SliceReadoutTime, I still think there is a lot of redundancy, and it becomes quite difficult to read what is being done with this parameter throughout the pipeline. I would put this parsing in a separate function, though we might call it from different locations.

Essentially, the parameter can be either a vector or a single number. These are then applied differently in ASL and M0 quantification. The easiest will be to turn it to a vector at the start of ASL module and then always apply as a vector.

### Comment
This only internally changes how is the SliceTime handeled and where it is recalculated as a vector from a scalar. Only minor mention in the release text. However, this overlaps with another issue that was changing slice-readout times - so we can simply mention one common comment - see below.

### Release summary
Internally restructure SliceTime allowing ExploreASL now to work with multi-band 2D EPI as well or any other SliceTime order.

jan-petr:
Closes #356 

Created a single central function from parsing the SliceReadoutTime parameter. It checks for 2D and 3D readout and if the SliceReadoutTime is a vector or scalar and then produces a vector with the length corresponding to the number of slices in the native ASL. 

This issue does not change the functionality, but rather cleans the code and puts all the checks to a single function, while previously they were everywhere.

### Release changes
Move current SliceReadoutTime parsing from throughout the pipeline to a single function

### DIFF
--- a/Functions/xASL_quant_FEAST.m
+++ b/Functions/xASL_quant_FEAST.m
@@ -53,20 +53,19 @@ xASL_adm_CreateDir(x.D.TTCheckDir);
 for iSession=1:2
     % Load data
     CBF{iSession} = xASL_io_Nifti2Im(PathCBF{iSession});
-    Gradient{iSession} = xASL_io_Nifti2Im(fullfile(x.D.PopDir, ['SliceGradient_extrapolated_' x.P.SubjectID '_' x.SESSIONS{iSession} '.nii']));
+    SliceNumber = xASL_io_Nifti2Im(fullfile(x.D.PopDir, ['SliceGradient_extrapolated_' x.P.SubjectID '_' x.SESSIONS{iSession} '.nii']));
 	% Obtain the correct SliceTime
 	imASL = xASL_io_ReadNifti(x.P.Path_ASL4D);
 	nSlices = size(imASL.dat,3);
 	SliceTime = xASL_quant_SliceTimeVector(x,nSlices);
 	
     % Correct different PLD scales
-	imGradientRounded = round(Gradient{iSession});
-	imGradientRounded(imGradientRounded<1) = 1;
-	imGradientRounded(imGradientRounded>length(SliceTime)) = length(SliceTime);
-	PLD{iSession} = x.Q.Initial_PLD + SliceTime(imGradientRounded);
+	SliceNumber = round(SliceNumber);
+	SliceNumber(SliceNumber<1) = 1;
+	SliceNumber(SliceNumber>length(SliceTime)) = length(SliceTime);
+	PLD{iSession} = x.Q.Initial_PLD + SliceTime(SliceNumber);
 	
-    Gradient{iSession} = exp(PLD{iSession}./x.Q.BloodT1) / (2.*x.Q.LabelingEfficiency.*x.Q.BloodT1 .* (1- exp(-x.Q.LabelingDuration./x.Q.BloodT1)) );
-    CBF{iSession} = CBF{iSession}./Gradient{iSession};
+    CBF{iSession} = CBF{iSession}./(exp(PLD{iSession}./x.Q.BloodT1) / (2.*x.Q.LabelingEfficiency.*x.Q.BloodT1 .* (1- exp(-x.Q.LabelingDuration./x.Q.BloodT1)) ));
 end
 % Average different PLD scales
 PLD_combined= (PLD{1}+PLD{2})./2;
--- a/Functions/xASL_quant_FEAST.m
+++ b/Functions/xASL_quant_FEAST.m
@@ -25,7 +25,7 @@ function xASL_quant_FEAST(x)
 % EXAMPLE: xASL_quant_FEAST(x);
 % REFERENCE: JJ Wang, 2003 MRM; Y Chen, 2012 MAGMA
 % __________________________________
-% Copyright 2015-2020 ExploreASL
+% Copyright 2015-2021 ExploreASL
 
 
 %% -------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


Functions/xASL_quant_SliceTimeVector.m
```diff
@@ -0,0 +1,75 @@
...
+if ~isfield(x,'readout_dim')
+	error('x.readout_dim field is missing');
+end
+
+if strcmpi(x.readout_dim, '3D')
+	% For 3D sequences, zero is returned
+	SliceTime = 0;
+	return;
+end
+
```
HenkMutsaerts, line 43: Why do we check a lot of x fields here that are not used below (unless except for the shortest_TR function)?
Doesn't it make more sense to keep this function "x-free"? Or should it also output x?

Functions/xASL_quant_FEAST.m
```diff
@@ -54,15 +54,17 @@ function xASL_quant_FEAST(x)
...
-		imGradientRounded(imGradientRounded>length(x.Q.SliceReadoutTime)) = length(x.Q.SliceReadoutTime);
-		PLD{iSession} = x.Q.Initial_PLD + x.Q.SliceReadoutTime(imGradientRounded);
-	end
+	% Obtain the correct SliceTime
+	imASL = xASL_io_ReadNifti(x.P.Path_ASL4D);
+	nSlices = size(imASL.dat,3);
+	SliceTime = xASL_quant_SliceTimeVector(x,nSlices);
+	
+    % Correct different PLD scales
+	imGradientRounded = round(Gradient{iSession});
```
HenkMutsaerts, line 63: Perhaps we should rename `Gradient` to `SliceNumber` or something more self-explanatory

Functions/xASL_quant_M0.m
```diff
@@ -133,39 +133,33 @@
...
+
+		SliceIM = zeros(size(M0IM));
+		for iZ=1:size(M0IM,3)
+			SliceIM(:,:,iZ) = iZ;
+		end
+
+		if  x.M0_usesASLtiming
+			% in this case, the M0 readout has the exact same timing as the ASL readout
+			% this is the case e.g. for Philips 3D GRASE
+			NetTR = x.Q.LabelingDuration+x.Q.Initial_PLD+SliceTime(SliceIM);
```
HenkMutsaerts, line 151: Is the "new" SliceTime always a vector that starts with 0?
This should be checked for all options in all *_quant_Q functions

HenkMutsaerts (approved):
Some minor comments, main thing is that the SliceTime is tested and in all cases provides the same answer (which is now always a vector with the first element being 0 right?)

Functions/xASL_quant_SliceTimeVector.m
jan-petr: 1) We cannot make it x-free. We would have to provide the x.readout_dim and other parameters as parameters, and we would have to check those before calling this function. This was the exact behavior you wanted to avoid, so we hid all these inside.
2) Not true that we check for unnecessary x-fields. readout_dim is needed to return 0 for 3D. SliceReadoutTime is always needed. Nothing else is checked.
3) I wouldn't output x. Because we input SliceReadoutTime (legacy field) and output SliceTime (BIDS field). These will change once we turn to full BIDS later - then we can redo all these everywhere in the code. But to make things easier and more transparent now, I have the legacy input and BIDS output (that way the xASL functions used after this function are already in turned to BIDS compliant) - but I didn't want to mix the inside x-struct now.

Hope this answers all questions ;)

jan-petr:
> Some minor comments, main thing is that the SliceTime is tested and in all cases provides the same answer (which is now always a vector with the first element being 0 right?)

Exactly! And it checks that for 3D this goes to 0 immediately. So we have cleaned and checked the SliceTime in the same way before using it in all functions.

Functions/xASL_quant_FEAST.m
jan-petr: Yes! It was a bit ambiguous if this was in standard or native space etc. So SliceNumber would be totally helpful. Shall I do it now?

Functions/xASL_quant_M0.m
jan-petr: Yes. It is. That's what the new function does. Checks it for all cases. No need to check it afterwards, because the new function does exactly that - so you can then peacefully use it and know that it was checked already.

Functions/xASL_quant_FEAST.m
jan-petr: It's loaded from NII called 'SliceGradient_extrapolated'. Should I still rename?!

HenkMutsaerts: Yes, SliceNumber would work. Native and standard spaces are recognized by their folder (subject or population).
There is indeed a separate extrapolated SliceGradient, which avoids the NaNs outside the original FoV when resampling. This one is useful for the quantification, while the one without the extrapolation is useful for QC (displaying the native spaces slices in standard space).

jan-petr: OK. I will change the name in this file and then merge.

jan-petr: I have renamed it both for quant_FEAST and quant_SinglePLD -> gradient to slice-number.

In SinglePLD, I have avoided creating a new matrix for the rounded version of SliceNumber -> it can replace the old one.
in FEAST, I have done the same and also simplified certain things not to keep unnecessary copies of certain variables.

Issue #371, opened 2021-02-25
BIDS Flavors testing - fix processing of legacy data


jan-petr:
The correct paths need to be inserted for processing the Flavors in legacy format in function:
xASL_test_BIDSFlavorsFull.m

PR #372, opened 2021-02-25
Bug #371 flavors legacy processing

jan-petr:
Fixing several bugs:
- Range checking of SliceReadoutTime as a vector wasn't working properly.
- JSON_import was not accepting logical type as valid.
- The BIDS flavors testing script needed to fix correctly the path in the legacy processing step.
Close #372

### DIFF
--- a/Development/xASL_test_BIDSFlavorsFull.m
+++ b/Development/xASL_test_BIDSFlavorsFull.m
@@ -66,19 +66,29 @@ if bTest(1)
 	fprintf('Copying the source data to a temporary folder: %s\n',conversionPath);
 	if ~exist(conversionPath, 'dir')
 		mkdir(conversionPath);
-	end
-	
-	fList = xASL_adm_GetFileList(flavorsPath, [], 'List', [], 1);
+    end
+    
+    if xASL_exist(flavorsPath)
+        fList = xASL_adm_GetFileList(flavorsPath, [], 'List', [], 1);
+    else
+        error('No source directory...');
+    end
 	for iList = 1:length(fList)
 		if ~exist(fullfile(conversionPath,fList{iList}), 'dir')
 			mkdir(fullfile(conversionPath,fList{iList}));
 		end
 		if ~exist(fullfile(conversionPath,fList{iList}, 'sourcedata'), 'dir')
 			mkdir(fullfile(conversionPath,fList{iList}, 'sourcedata'));
-		end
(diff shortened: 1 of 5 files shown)


MichaelStritt (approved):
Everything I've seen looks good. Will run a small test in a few minutes.

MichaelStritt:
@jan: did some minor changes for the windows users

jan-petr:
> Everything I've seen looks good. Will run a small test in a few minutes.

I have tested it as well - it runs fine with the whole Flavors. In the meaning that it starts all the studies - not that it finishes them :) But that's something to test properly in develop with all the other merges.

Development/xASL_test_BIDSFlavorsFull.m
```diff
@@ -66,19 +66,29 @@ function xASL_test_BIDSFlavorsFull(pathExploreASL,pathTest,bTest)
 	fprintf('Copying the source data to a temporary folder: %s\n',conversionPath);
 	if ~exist(conversionPath, 'dir')
 		mkdir(conversionPath);
-	end
-	
-	fList = xASL_adm_GetFileList(flavorsPath, [], 'List', [], 1);
+    end
+    
+    if xASL_exist(flavorsPath)
```
jan-petr, line 71: You might add 'dir' to the exist function.

jan-petr:
@MichaelStritt Can you please fix the minor thing I commented on - I want to avoid conflicts in parallel editing.

Development/xASL_test_BIDSFlavorsFull.m
MichaelStritt: Good idea :)

MichaelStritt:
@jan-petr: done 👍

Development/xASL_test_BIDSFlavorsFull.m
```diff
@@ -126,6 +140,9 @@ function xASL_test_BIDSFlavorsFull(pathExploreASL,pathTest,bTest)
 				end
 			else
 				warning('Here we expect a unix-ish system');
+                diary('off');
+                fclose('all'); % ensure that no file is locked
+                xASL_delete(fullfile(ListFolders{iList}, 'derivatives'));
```
MichaelStritt, line 145: @jan-petr: I think I forgot to add **true** for the second argument here, so that it deletes the folder content as well. Can we fix this?

PR #375, opened 2021-02-26
Bug #373 parameter parsing in dicom2 bids

### LINKED ISSUE #373: Parameter parsing in DICOM2BIDS
Several minor parameters parsing errors have to be fixed:
1) TotalAcquiredPairs from config has priority over DICOM.
2) TotalAcquiredPairs from DICOM is taken as a maximum.
3) Characters with ASCII code 0 are replaced by a space. However, the strrep function returns a warning if a numeric value zero is provided as a string - it works, but it complains. The solution is to do a prior explicit type conversion.
4) dcm2nii are catched and reported as an extra Matlab warning. This is not necessary, as warnings are printed on screen and log anyway, and warnings do not crash anything, so it is an extra duplicitous output.

### Release commits
All minor fixes no need to mention. Removed some warnings in DICOM2BIDS.

jan-petr:
Closes #373


MichaelStritt (approved):
Checked the code, great fix 👍

Issue #384, opened 2021-03-02
xASL_io_ExportVTK


MichaelStritt:
There's a pretty nice 3D software called [Paraview](https://www.paraview.org/) that I used a lot at University. It requires files in `VTK` format. I saw that there's a [Matlab script](https://www.mathworks.com/matlabcentral/fileexchange/47814-vtkwrite-exports-various-2d-3d-data-to-paraview-in-vtk-file-format) with [MIT License](https://github.com/joe-of-all-trades/vtkwrite/blob/master/license.txt) which can convert image data to `structure_points` in `VTK` format. I implemented a small wrapper which would work nicely in ExploreASL. What do you think about adding it?

### How to test
```
xASL_io_ExportVTK(x.MyPath,niftiPath);
```

The script will print the license information and export a `export.vtk` file to the corresponding directory. There's also a unit test called `xASL_ut_UnitTest_function_ExportVTK` which shows all the different options.

### Exemplary M0 image in ParaView

![image](https://user-images.githubusercontent.com/46593074/109628284-dbc34e80-7b42-11eb-81ca-c1d7d08204a7.png)

MichaelStritt:
```
              name                  unit          tests        passed
    _________________________    __________    ____________    ______

    'xASL_tsvRead'               'Function'    [1×2 struct]    true  
    'xASL_tsvWrite'              'Function'    [1×2 struct]    true  
    'xASL_io_Nifti2Im'           'Function'    [1×1 struct]    true  
    'xASL_test_GetLogContent'    'Function'    [1×2 struct]    true  
    'xASL_io_ExportVTK'          'Function'    [1×5 struct]    true  
```

MichaelStritt:
### Change Requests:
- Replace xASL_im_ResampleLinearFair with xASL_spm_Resample

MichaelStritt:
### Simple Test

```matlab
>> test_mask = '.\ExploreASL\External\SPMmodified\toolbox\cat12\templates_volumes\brainmask.nii';
>> test_image = '.\ExploreASL\External\SPMmodified\toolbox\cat12\templates_volumes\brainmask.nii';
>> xASL_io_ExportVTK('C:\...\ExploreASL',test_image,test_mask,'C:\...\testVTK\export.vtk');
We assume that image and mask have the same orientation...
Export vtk as structured points...
```

PR #386, opened 2021-03-02
Closes #361 bids2legacy

### LINKED ISSUE #361: DRO test with bids2legacy
### Requirements

- Add a test script called `xASL_ut_UnitTest_function_BIDS2Legacy.m`.
- This unit test should load the bids structure in `Testing/UnitTesting/dro_files/test_patient/`.
- Check if all necessary files are part of the resulting legacy structure (**T1w, ASL & M0 NIFTI & JSON** as well as data par file).
- (The requirement above could require a dummy data par file for the DRO).
- This test does **not** have to include the processing of the resulting legacy structure in ExploreASL.
- This unit test should be added to `xASL_ut_UnitTesting.m`.

###  Tests

- UnitTests(1) =`xASL_ut_UnitTest_function_BIDS2Legacy(TestRepository);` should return a unit test structure according to the template.
- If `xASL_ut_UnitTest_function_BIDS2Legacy.m` works correctly, the test should be passed.

MichaelStritt:
Minor test for BIDS2LEGACY using DRO test data. Check out #361 for a more detailed explanation.


MichaelStritt:
There could be a minor merge conflict with #384 since both update the main testing script.

Functions/xASL_bids_PrepareDRO.m
```diff
@@ -0,0 +1,96 @@
+function xASL_bids_PrepareDRO(droTestPatient,droSubject)
+%xASL_bids_PrepareDRO Prepare DRO test patient for BIDS2Legacy conversion.
+%
+% FORMAT: xASL_bids_PrepareDRO(droTestPatient,droSubject)
```
jan-petr, line 4: OPTIONAL parameters should be in [] here

Functions/xASL_bids_PrepareDRO.m
```diff
@@ -0,0 +1,96 @@
+function xASL_bids_PrepareDRO(droTestPatient,droSubject)
```
jan-petr, line 1: Can you change the name to xASL_test_PrepareDROTest or something like this - it is not preparing DRO, but preparing a DRO test, right? So that's more about testing than about BIDS.

Functions/xASL_bids_PrepareDRO.m
```diff
@@ -0,0 +1,96 @@
...
+%xASL_bids_PrepareDRO Prepare DRO test patient for BIDS2Legacy conversion.
+%
+% FORMAT: xASL_bids_PrepareDRO(droTestPatient,droSubject)
+% 
+% INPUT:
+%   droTestPatient      - Path to the DRO (CHAR ARRAY, REQUIRED)
```
jan-petr, line 7: You probably needs to give more details of what exactly are the inputs - names and structure of the files and what exactly the function does with it.

Functions/xASL_bids_PrepareDRO.m
```diff
@@ -0,0 +1,96 @@
...
+function [json] = createDatasetDescriptionTemplate(name)
+
+    % Create dummy dataset_description.json
+    json = struct;
+    if nargin > 0
+        json.Name = name;
+    else
+        json.Name = "RandomText";
+    end
+    json.BIDSVersion = "1.5.0";
```
jan-petr, line 83: Call bidsPar = xASL_bids_Config() and use bidsPar.BIDSVersion

jan-petr (changes requested):
Great to start filling these up. Some minor changes in names :)

Functions/xASL_bids_PrepareDRO.m
MichaelStritt: Oh, sorry, I forgot that. Fixed in commit 12959c4.

Functions/xASL_bids_PrepareDRO.m
MichaelStritt: I added this in 3e2b8e1. Is there a script that creates a dataset description template already, like I do in the subfunction `createDatasetDescriptionTemplate`? If not, maybe such a helper function could be useful for other applications as well.

Functions/xASL_bids_PrepareDRO.m
MichaelStritt: I added more descriptive text in 3e2b8e1.

Functions/xASL_bids_PrepareDRO.m
MichaelStritt: So my idea was to use the output of the asldro python script exactly like it is and then change some file names (add `sub-...` etc. in `lines 45-55`) and add all necessary sidecars & missing fields (dataset description in `lines 61-74`) so that the `BIDS2Legacy` script can convert it to something that can be read by **ExploreASL**.
Having a script for this could be quite useful, since we could test other parts of **ExploreASL** with the dro data as well.
I'm fine with renaming and/or improving the script in general and it will be used almost exclusively for testing, but not just for this individual test, right?

jan-petr: I see, I get the general idea - and I completely agree. Do I get it right that what you do here is fixing all the small BIDS issues for DRO? i.e. things that Aaron should fix in the future because DRO output should be (in theory and in future) in BIDS?

We should then adapt the name of the function to something like xASL_bids_DRO2BIDS. And this function should become obsolete once Aaron fixes everything, shouldn't it?

Functions/xASL_bids_PrepareDRO.m
jan-petr: Thanks. Regarding Default description - yes, there should be a function. And this function better has some normal text and not RandomText :) My fault mainly as I used these for the ImportBIDS to have some values for testing. But according to BIDS, only name and version is required.

So lets leave this as it is now and I have created a new issue to fix it for all cases. #443

jan-petr (changes requested):
See my answers to the comments - almost there.

Functions/xASL_bids_PrepareDRO.m
MichaelStritt: Exactly that was the idea. I just wasn't sure if that's Aaron making mistakes or just me not using it correctly (but I set everything to default). And it is correct that the files have to start with `sub-Sub1...` right?
I'll change the name to `xASL_bids_DRO2BIDS` for now, sound way much better 👍

MichaelStritt: Fixed in 9f0a3ab.

MichaelStritt:
@jan-petr: Thanks for the feedback. In af8c4d4 I converted the inline function for the `dataset_description` template to a separate script, with an input parser, called `xASL_bids_CreateDatasetDescriptionTemplate`. This can be used to create templates pretty easily and you can set all parameters, if necessary, also manually. `xASL_bids_PrepareDRO` was renamed to `xASL_bids_DRO2BIDS`.
Are there other open change requests?

Functions/xASL_bids_CreateDatasetDescriptionTemplate.m
```diff
@@ -0,0 +1,105 @@
+function [json] = xASL_bids_CreateDatasetDescriptionTemplate(varargin)
```
jan-petr, line 1: Contemplating what is the best input type... 
Mostly, we will not be filling the varargin ourself, but the varargin values will be given to us in a JSON already? 

Basically, we will receive an incomplete dataset with a JSON containing dataset_description fields among other fields. And our task will be to filter out the useful information.

So my guess is that we will call this function and give it a JSON-struct on the entry and it will filter out the good parameters, check them and provide the missing ones. Restructuring the JSON-input into separate input variables would be an extra work then.

Do you agree, or you had a different use case in your mind?

Functions/xASL_bids_CreateDatasetDescriptionTemplate.m
```diff
@@ -0,0 +1,105 @@
...
+%
+% OUTPUT:
+%   json                - JSON structure which can be saved using spm_jsonwrite (dataset_description.json template)
+%                         
+% -----------------------------------------------------------------------------------------------------------------------------------------------------
+% DESCRIPTION:  This script creates a JSON structure which can be saved
+%               using spm_jsonwrite to get a dataset_description.json template.
+%
+% -----------------------------------------------------------------------------------------------------------------------------------------------------
+% EXAMPLE:      [json] = xASL_bids_CreateDatasetDescriptionTemplate('Test Patient','1.5.0','1.2.3','Test','Test License','Author Names','Thanks');
```
jan-petr, line 24: My proposition would be to have a single struct on the entry with some correct and some incorrect fields.

Functions/xASL_bids_CreateDatasetDescriptionTemplate.m
```diff
@@ -0,0 +1,105 @@
...
+
+    % Create dummy dataset_description.json
+    json = struct;
+    
+    % Required fields
+    for iCell = 1:numel(bidsPar.datasetDescription.Required)
+         json.(bidsPar.datasetDescription.Required{1,iCell}) = 'Undefined';
+    end
+    % Recommended fields
+    for iCell = 1:numel(bidsPar.datasetDescription.Recommended)
```
jan-petr, line 41: Recommended and optional fields should be rather not assigned at all - i.e. skip them in the struct instead of assigning 'undefined' or ''.

jan-petr: But overall, it is great that you use the 'required' and 'recommended' lists from the BIDS config - it will make our live easier during an update :)

Functions/xASL_bids_CreateDatasetDescriptionTemplate.m
```diff
@@ -0,0 +1,105 @@
...
+    end
+    % Optional fields
+    for iCell = 1:numel(bidsPar.datasetDescription.Required)
+         json.(bidsPar.datasetDescription.Optional{1,iCell}) = 'Undefined';
+    end
+    
+    % Parse the input parameters
+    p = inputParsing(varargin{:});
+    
+    % Optionally change the parameter values
```
jan-petr, line 52: This is a bit strange. Above, you assign fields according to the config files, but here, you use only specific selected ones. You want to allow the function to assign any possible dataset_description parameter. That's why it would be better to have a struct as input where you would be able to go through all the lists and 
1. Pick all REQUIRED fields from input and give warning when missing.
2. Pick all RECOMMENDED and OPTIONAL fields from input and not assign when missing.
3. Ignore all other fields in the input struct.

Functions/xASL_bids_CreateDatasetDescriptionTemplate.m
```diff
@@ -0,0 +1,105 @@
...
+    validName = @(variable) ischar(variable);
+    validBIDSVersion = @(variable) ischar(variable);
+    validHEDVersion = @(variable) ischar(variable);
+    validDatasetType = @(variable) ischar(variable);
+    validLicense = @(variable) ischar(variable);
+    validAuthors = @(variable) ischar(variable);
+    validAcknowledgements = @(variable) ischar(variable);
+    
+    % Define defaults
+    defaultName = 'Undefined';
```
jan-petr, line 82: Does not make sense to define here the default values etc. Because it doesn't follow the bids_Config, and also it is done already above in the parent function. 
But my recommendation would be to get rid of this parameter parsing anyway.

jan-petr (commented):
> @jan-petr: Thanks for the feedback. In [af8c4d4](https://github.com/ExploreASL/ExploreASL/commit/af8c4d4e529ea87582f903e01b5ee1044799f919) I converted the inline function for the `dataset_description` template to a separate script, with an input parser, called `xASL_bids_CreateDatasetDescriptionTemplate`. This can be used to create templates pretty easily and you can set all parameters, if necessary, also manually. `xASL_bids_PrepareDRO` was renamed to `xASL_bids_DRO2BIDS`.
> Are there other open change requests?

jan-petr (changes requested):
Thanks. All my comments addressed, but I am afraid that we will have to redo the datasetdescription function. It is perfect that it loads the requirements from the config script, but it clashes a bit with the previous implementation - see my inline comments.

Functions/xASL_bids_CreateDatasetDescriptionTemplate.m
MichaelStritt: Ahhh, now I understand what you want to use it for 👍 My use case here was a bit more simple so I used some copy paste from the Initialize script I'm working on. But I think I understand what you have in mind and I can implement it.

Functions/xASL_bids_CreateDatasetDescriptionTemplate.m
MichaelStritt: Check out bcfe3e6.

Functions/xASL_bids_CreateDatasetDescriptionTemplate.m
MichaelStritt: Check out bcfe3e6.

Functions/xASL_bids_CreateDatasetDescriptionTemplate.m
MichaelStritt: Check out bcfe3e6.

Functions/xASL_bids_CreateDatasetDescriptionTemplate.m
MichaelStritt: Removed in bcfe3e6.

Functions/xASL_bids_CreateDatasetDescriptionTemplate.m
MichaelStritt: Fixed in bcfe3e6.

MichaelStritt:
> 
> 
> Thanks. All my comments addressed, but I am afraid that we will have to redo the datasetdescription function. It is perfect that it loads the requirements from the config script, but it clashes a bit with the previous implementation - see my inline comments.

No worries @jan-petr :) I think I just had a different use case in mind. Would you mind checking out bcfe3e6? 

A simple test for the functionality would be something like this:
```
draft.Name = 'DRO_Digital_Reference_Object';
draft.Test = 'Test';
draft.License = 'Test_License';
[json] = xASL_bids_CreateDatasetDescriptionTemplate(draft);
```

jan-petr:
> No worries @jan-petr :) I think I just had a different use case in mind. Would you mind checking out [bcfe3e6](https://github.com/ExploreASL/ExploreASL/commit/bcfe3e6d4e1abe97b1fe0f287bd09cedc9bfaf37)?
> 

Test works! Also excluding unwanted fields. There were two small bugs. And also the concept with extra 'transfer' structure was two complex. Simplified :)

You can check if this works. But otherwise ready to merge.

jan-petr (approved):
OK.

PR #392, opened 2021-03-03
#382 Reorder CCLL to CLCL for PARREC import

### LINKED ISSUE #382: PAR/REC control/label order
Incorrect control/label merging. PAR/REC files are read by dcm2nii as two separate NIIs with controls and labels. They should be merged while interlacing the dynamics and not all controls and then all labels.

This needs to be fixed and the tested on the Philips PAR/REC dataset from Flavors.

### Release summary
Reorders NIFTIs during the import of PAR/REC to have the correct control/label ordering of dynamics.

jan-petr:
Close #382

### DIFF
--- a/ExploreASL_Import.m
+++ b/ExploreASL_Import.m
@@ -556,7 +556,7 @@ for iSubject=1:nSubjects
                                 [nii_files, scan_name, first_match, MsgDcm2nii] = xASL_io_dcm2nii(scanpath, destdir, scan_name, 'DicomFilter', imPar.dcmExtFilter, 'Verbose', imPar.bVerbose, 'Overwrite', imPar.bOverwrite, 'Version', imPar.dcm2nii_version, 'x', x);
 
                                 % If dcm2nii produced a warning or error, catch this & store it
-                                if ~isempty(MsgDcm2nii) && ~isempty(regexp(lower(MsgDcm2nii),'.*(warning|error).*')) % if it contains a warning/error
+                                if ~isempty(MsgDcm2nii) && ~isempty(regexp(lower(MsgDcm2nii),'.*(error).*')) % if it contains a warning/error
                                     dcm2niiCatchedErrors = CatchErrors('xASL_io_dcm2nii', MsgDcm2nii, dbstack, ['dcm2nii_' imPar.dcm2nii_version], pwd, scan_name, scanpath, destdir, dcm2niiCatchedErrors, imPar);
                                 end
 
@@ -572,12 +572,23 @@ for iSubject=1:nSubjects
                             first_match = xASL_adm_GetFileList(scanpath, ['.*' imPar.dcmExtFilter],'FPList',[0 Inf]);
                             if  ~isempty(first_match); first_match = first_match{1}; end
                         end
-                    end
-
-                    %% Merge NIfTIs if there are multiples
-					if length(nii_files)>1
-						1;
 					end
+
+					%% In case of a single NII ASL file loaded from PAR/REC, we need to shuffle the dynamics from CCCC...LLLL order to CLCLCLCL... order
+					[~,~,scanExtension] = xASL_fileparts(scanpath);
+					if ~isempty(regexpi(scanExtension,'par|rec')) && length(nii_files)==1 && ~isempty(regexpi(scan_name,'ASL'))
+						% For a PAR/REC files that produces a single ASL4D NIFTI
+						imASL = xASL_io_Nifti2Im(nii_files{1});
+						% If multiple dynamics
+						if size(imASL,4) > 1
+							% Then reshuffle them
+							imASLreordered = zeros(size(imASL));
+							imASLreordered(:,:,:,1:2:end) = imASL(:,:,:,1:ceil(size(imASL,4)/2));
+							imASLreordered(:,:,:,2:2:end) = imASL(:,:,:,ceil(size(imASL,4)/2)+1:end);
+							xASL_io_SaveNifti(nii_files{1},nii_files{1},imASLreordered);
+						end
+					end
+                    %% Merge NIfTIs if there are multiples
 					% For ASL or M0, merge multiple files
 					if length(nii_files)>1 
 						if ~isempty(strfind(scan_name,'ASL4D'))
--- a/ExploreASL_Import.m
+++ b/ExploreASL_Import.m
@@ -457,7 +457,7 @@ for iSubject=1:nSubjects
 
                 % convert scan ID to a suitable name and set scan-specific parameters
                 if size(imPar.tokenScanAliases,2)==2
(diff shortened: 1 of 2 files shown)


ExploreASL_Import.m
```diff
@@ -556,7 +556,7 @@ function ExploreASL_Import(imPar, bCopySingleDicoms, bUseDCMTK, bCheckPermission
                                 [nii_files, scan_name, first_match, MsgDcm2nii] = xASL_io_dcm2nii(scanpath, destdir, scan_name, 'DicomFilter', imPar.dcmExtFilter, 'Verbose', imPar.bVerbose, 'Overwrite', imPar.bOverwrite, 'Version', imPar.dcm2nii_version, 'x', x);
 
                                 % If dcm2nii produced a warning or error, catch this & store it
-                                if ~isempty(MsgDcm2nii) && ~isempty(regexp(lower(MsgDcm2nii),'.*(warning|error).*')) % if it contains a warning/error
+                                if ~isempty(MsgDcm2nii) && ~isempty(regexp(lower(MsgDcm2nii),'.*(error).*')) % if it contains a warning/error
```
HenkMutsaerts, line 559: better to use `regexpi()` instead of `regexp(lower())`

ExploreASL_Import.m
```diff
@@ -572,12 +572,23 @@ function ExploreASL_Import(imPar, bCopySingleDicoms, bUseDCMTK, bCheckPermission
                             first_match = xASL_adm_GetFileList(scanpath, ['.*' imPar.dcmExtFilter],'FPList',[0 Inf]);
                             if  ~isempty(first_match); first_match = first_match{1}; end
                         end
-                    end
+					end
 
-                    %% Merge NIfTIs if there are multiples
-					if length(nii_files)>1
-						1;
+					%% In case of a single NII ASL file loaded from PAR/REC, we need to shuffle the dynamics from CCCC...LLLL order to CLCLCLCL... order
```
HenkMutsaerts, line 577: This looks good, beware of the space ` ` after each comma: `, ` instead of `,`

`scanExtension` would capture `.PAR` or `.REC` here, so you want to compare with `'^\.(par|rec)$'` instead of `'par|rec'` . Also, you were missing some brackets here

ExploreASL_ImportBIDS.m
```diff
@@ -1380,10 +1380,21 @@ function ExploreASL_ImportBIDS_DCM2NII(imPar, bCopySingleDicoms, bUseDCMTK, bChe
 							end
 						end
 						
-						% Merge NIfTIs if there are multiples
-						if length(nii_files)>1
-							1;
+						%% In case of a single NII ASL file loaded from PAR/REC, we need to shuffle the dynamics from CCCC...LLLL order to CLCLCLCL... order
+						[~,~,scanExtension] = xASL_fileparts(scanpath);
```
HenkMutsaerts, line 1384: Same here as above

HenkMutsaerts (changes requested):
See inline.
Plus:
1. Why is `ExploreASL_ImportBIDS` a copy of `ExploreASL_Import`, can we not keep this transparent? I thought that the main difference between the two comes AFTER the original `ExploreASL_import` (as you go from legacy to BIDS right?). This would help people like me, understanding the new code, and less cumbersome during the transition phase.

2. We should beware of this code becoming bulky. It would be better to have a subfunction like `ParsePARREC` for all specific flavors, and keep the main code easy to read!

I understand that the stuff I mention above creates extra work for you, but think of the future work it will save you, as you would have to explain every line of code to the person that uses your code after you!

ExploreASL_Import.m
jan-petr: Good catch. I have replaced that in the BIDS version only.

ExploreASL_Import.m
jan-petr: No need to - this works. It returns the index with occurences - no need for brackets and stuff...
Check regexpi('.rec','par|rec')

ExploreASL_ImportBIDS.m
jan-petr: Also not needed - simplicity...

jan-petr:
> See inline.

See my answers to unresolved comments to see if you agree.

> Plus:
> 
> 1. Why is `ExploreASL_ImportBIDS` a copy of `ExploreASL_Import`, can we not keep this transparent? I thought that the main difference between the two comes AFTER the original `ExploreASL_import` (as you go from legacy to BIDS right?). This would help people like me, understanding the new code, and less cumbersome during the transition phase.

Theoretically, you are absolutely right and this is what is happening. Practically, I choose this way - will explain. I needed to write this BIDS script that will act in a similar way as ExploreASL_Import, but with important tweaks here and there. I didn't want to break anything in the process so I have created a copy and edited that. Not the best approach, but the fastest and most error prone. Now, the BIDSImport contains parts identical to Import - these can be taken from both and put to subfunctions. I could have done that previously, but wanted to avoid introducing errors. However, it would not have been possible to call ExploreASL_Import from ExploreASL_ImportBIDS as the inputs and outputs were slightly different. 

So that is the explanation - theoretically it is as you say. Practically, I chose this approach. But it is a VERY good idea to unify these. Best is to take similar portions of code and turn them into separate functions. I made an issue for that and will do that in future if you agree - no need to rush this for 1.5.0. See #394 

> 2. We should beware of this code becoming bulky. It would be better to have a subfunction like `ParsePARREC` for all specific flavors, and keep the main code easy to read!

Totally agree. I already turned some parts to subfunctions - can work on that more in the new issue - see again #394 

> I understand that the stuff I mention above creates extra work for you, but think of the future work it will save you, as you would have to explain every line of code to the person that uses your code after you!

Of course - I totally agree. I avoid creating bulky code - but in this case, it was already bulky before I touched it :) This import - it is really a disaster, but the first goal was to fix it, now the second goal is to make it more modular - totally agree. See #394 

The great thing is that we can play with #394 as much as we want and have the Flavors that can test that not a single thing changed...

Any further stuff or approve?

ExploreASL_Import.m
HenkMutsaerts: Yes, it works, but this would also work with an extension of `.par.zip` or an extension of `par` (without extension period) which would in both cases crash the script. Here you want to be explicit, we don't want the code to work with occurrences that are different than `.PAR` or `.REC` right?

ExploreASL_Import.m
HenkMutsaerts: Why not in both (if you want to sync them later anyway)?

ExploreASL_Import.m
HenkMutsaerts: Also, you omitted the comma comment (first line of my reaction here :)

HenkMutsaerts (approved):

ExploreASL_Import.m
jan-petr: Regexp fixed with your comment.

Sorry - added comma - we want the comma, not to remove it :)

PR #414, opened 2021-03-12
Closes #368 spm_jsonwrite: Remove digits for integers

### LINKED ISSUE #368: Too many digits for some of the numbers in JSON
- Probably an issue in `spm_jsonwrite`

### Release note
Fix problem with writing trailing zeros in real numbers in spm_jsonwrite

MichaelStritt:


MichaelStritt:
@HenkMutsaerts: what do you think about this fix? It works with 2016b and newer and only needs minor code changes.

jan-petr (approved):
OK.

MichaelStritt:
> OK.

@jan-petr: do you know if there's a nice alternative for the matlab strip command? Mathijs just said that he still uses matlab<2016b, so this wouldn't work for him. We could only do this for newer versions, but then I'd have to add a version check to line 71 in xASL_str2num.

PR #431, opened 2021-03-17
#430 xASL_adm_ReplaceSymbols: Warning not error with unknown symbols

### LINKED ISSUE #430: xASL_adm_ReplaceSymbols crashes with PhoenixProtocol
If Phoenix protocol is loaded in x-struct, which will not happen for DICOM->BIDS->Legacy as this is removed from JSON, but might happen with the legacy import, then PhoenixProtocol gets into x-struct and this then tries to find and replace symbols inside the Phoenix protocol. Because PhoenixProtocol with XML have many strings that resemble these symbols. This will fail to find the symbol to replace and report an error and crash.

So we need to report a warning instead and end the replacing procedure without doing any replacement and returning the input string.

### Release notes
Fix xASL_adm_ReplaceSymbols crash when trying to replace sub-strings in PhoenixProtocol field.

jan-petr:
Close #430

### DIFF
--- a/Functions/xASL_adm_ReplaceSymbols.m
+++ b/Functions/xASL_adm_ReplaceSymbols.m
@@ -22,15 +22,15 @@ function strOut = xASL_adm_ReplaceSymbols(strIn, symbolTable, bracketLeft, brack
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
 %
 % __________________________________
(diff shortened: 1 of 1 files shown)


HenkMutsaerts (approved):
Looks good, I removed the double mentioning of the function name.

PR #434, opened 2021-03-18
Close #352 Improve Log Parsing

### LINKED ISSUE #352: Improving parsing of errors and warnings from log files.
Certain cases of errors and warnings are still not parsed optimally. See the examples below.
I understand that the warnings are too non-standardized, so we will need to check if this can be improved or not. If not, then the current state is also good.

### Release notes: Improve parsing of errors and warnings from log files in xASL_test_GetLogContent

(I think the title is already a pretty good release note :) )

MichaelStritt:
Closes #352


jan-petr (approved):
OK.

PR #444, opened 2021-03-23
#433 Enable ASL split on re-run

### LINKED ISSUE #433: xASL_io_SplitASL does not account for re-runs
First time through, the function will behave as intended and split according to `x.M0PositionInASL4D`. 

Upon attempting re-runs, however, if both ASL4D.nii.gz and ASL4D_Source.nii.gz exist, then no splitting will occur. This behavior prevents users from re-running the ASL module if they accidentally specified an incorrect value for `x.M0PositionInASL4D` the first time around.

Attaching [DataPar.zip](https://github.com/ExploreASL/ExploreASL/files/6159977/DataPar.zip) just in case the input was incorrect (but then it wouldn't have worked the first time...)

jan-petr:
Closes #433

### DIFF
--- a/Functions/xASL_io_SplitASL.m
+++ b/Functions/xASL_io_SplitASL.m
@@ -84,17 +84,17 @@ function xASL_io_SplitASL(inPath, iM0, iDummy)
     
     ASLlist = xASL_adm_GetFileList(Fpath, ['^' Ffile '(|_.)'  '(|_\d)' '\.nii$'], 'FPList', [0 Inf]);
   
-    if isempty(ASLlist) && xASL_exist(BackupName)
-        fprintf('ASL4D.nii didnt exist but can restore from ASL4D_Source.nii\n');
+    if xASL_exist(BackupName)
+        fprintf('Restoring ASL4D.nii from backup in ASL4D_Source.nii\n');
         ASLlist{1} = ASLname;
         xASL_Move(BackupName, ASLlist{1}, true);
         % Restore json as well
         if exist(ASLJSONPath,'file') && exist(BackupJSONPath,'file')
             xASL_Move(BackupJSONPath, ASLJSONPath, true);
         end
-        % Delete M0, allowing to restore this from the backup
-        if xASL_exist(Path_M0,'file')
-            warning('Deleting M0 to restore it from ASL4D');
+        % Delete M0, allowing to restore this from the backup in case M0 is in the ASL timeseries
+        if xASL_exist(Path_M0,'file') && ~isempty(iM0)
+            fprintf('Deleting M0 to restore it from ASL4D\n');
             xASL_delete(Path_M0);
             xASL_delete(M0JSONPath);
         end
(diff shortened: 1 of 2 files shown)


MauricePasternak (approved):
Worked for me when I alternated M0 positions back and forth on a GE test set.

Modules/xASL_module_ASL.m
```diff
@@ -119,12 +103,40 @@
...
+StateName{ 9} = '090_VisualQC_ASL';
+StateName{10} = '100_WADQC';
 
 %% Change working directory to make sure that unspecified output will go there...
 oldFolder = cd(x.SESSIONDIR);
 
+%% Split ASL and M0 within the ASL time series
+% Run this when the data hasn't been touched yet
+% The first three states are here, because the first two are run only conditionally
+if ~x.mutex.HasState(StateName{1}) && ~x.mutex.HasState(StateName{2}) && ~x.mutex.HasState(StateName{3})
```
HenkMutsaerts, line 124: Come to think of it, we probably need to enforce saving also the first two states even if they were not run, like with the structural module, to keep things the same throughout the pipeline, don't you agree?
If yes, let's make a separate issue for this (and notify Maurice)

HenkMutsaerts (changes requested):

Functions/xASL_io_SplitASL.m
```diff
@@ -84,17 +84,17 @@ function xASL_io_SplitASL(inPath, iM0, iDummy)
     
     ASLlist = xASL_adm_GetFileList(Fpath, ['^' Ffile '(|_.)'  '(|_\d)' '\.nii$'], 'FPList', [0 Inf]);
   
-    if isempty(ASLlist) && xASL_exist(BackupName)
-        fprintf('ASL4D.nii didnt exist but can restore from ASL4D_Source.nii\n');
+    if xASL_exist(BackupName)
```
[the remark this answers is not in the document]
jan-petr: Fixed.

Functions/xASL_io_SplitASL.m
```diff
@@ -84,17 +84,17 @@ function xASL_io_SplitASL(inPath, iM0, iDummy)
...
+        fprintf('Restoring ASL4D.nii from backup in ASL4D_Source.nii\n');
```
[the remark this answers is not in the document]
jan-petr: Fixed.

Modules/xASL_module_ASL.m
jan-petr: I think that it is fine to not save states that were not run because of lack of data. But it wouldn't be completely incorrect to save them either. We can ask Maurice if either of those is problematic for him.

I have a slight preference to keep things as they are. The problem can be that if we save states for non-existing data, then if you later on add TOP-UP to an already processed data, it will ignore it if the state will be pre-saved.

HenkMutsaerts: True, I agree!

HenkMutsaerts (approved):
Looks good! Somehow I couldn't find my requests and how you addressed them, which could save time in the re-review stage ;) Should I be able to see this easily?

jan-petr:
> Looks good! Somehow I couldn't find my requests and how you addressed them, which could save time in the re-review stage ;) Should I be able to see this easily?

Yes. Two ways. Either go to 'files changed' and you will see the requests there. Or check above in the comments. Note that I have answered and closed comment as they were all minor, so I didn't need re-review of the answers. You thus have to click the 'show resolved'. 

However, you can't really see the comments in the new commit. Only in the heap of all changes... So you can't see the newest commit with the answers...

PR #446, opened 2021-03-24
#415 xASL_Copy: Allow recursive copying of folders

### LINKED ISSUE #415: xASL_Copy: allow recursive copying
When a folder is copied into another folder, but the other folder already exists, this needs to have `bOverwrite` set to true. Now we do this automatically when the destination folder is empty, and provide a warning when it is not empty (as this could merge contents).

jan-petr:
Close #415


MichaelStritt (approved):
Just tested the changes on a folder with subfolders and a lot of files. Worked nicely and was quite fast. I'll merge it now.

Issue #403, opened 2021-03-08
xASL_quant_M0: x.M0_usesASLtiming incorrect explanation 


MDijsselhof:
xASL_quant_M0: x.M0_usesASLtiming explains the following:

'% in this case, the M0 readout has the exact same timing as the ASL readout
% this is the case e.g. for Philips 3D GRASE'

The explanation only has to be adjusted to a 2D sequence as this step in the code is only for 2D sequences. 

### Release note
Very minor edit in comments, can be aggregated with other commits under a text: Improved inline comments in the code.

PR #448, opened 2021-03-24
#403 changes x.M0_usesASLtiming explanation fix

MDijsselhof:
xASL_quant_M0: x.M0_usesASLtiming explains the following:

'% in this case, the M0 readout has the exact same timing as the ASL readout
% this is the case e.g. for Philips 3D GRASE'

The explanation only has to be adjusted to a 2D sequence as this step in the code is only for 2D sequences. 

### Release note
Very minor edit in comments, can be aggregated with other commits under a text: Improved inline comments in the code.


jan-petr (approved):
OK to merge.

PR #450, opened 2021-03-24
Bids #418 slice timing consistency

### LINKED ISSUE #418: Warning if SliceReadoutTime isn't consistent between slices
In most cases where SliceTiming is a vector, SliceReadoutTime will be a scalar (unless we move fully to BIDS). When we move fully to BIDS, the warning should still be there, but then if length(unique(diff(SliceTiming(:)))>1.
In other words, if the timing difference between slices varies between slices, this should issue a warning, as this is only the case in 5% of cases.

### Release note
Verifies the SliceTiming parameter if timing difference is consistent between slices.

jan-petr:
Close #418

### DIFF
--- a/Functions/xASL_quant_SliceTiming.m
+++ b/Functions/xASL_quant_SliceTiming.m
@@ -1,5 +1,6 @@
 function SliceTiming = xASL_quant_SliceTiming(x, inputIm)
-%xASL_quant_SliceTiming Calculates a vector with the timings of slice readouts relative to the start of readout
+%xASL_quant_SliceTiming Takes the SliceReadoutTime (xASL-legacy or BIDS) and calculates a vector with the timings of slice readouts relative 
+% to the start of readout (BIDS definition)
 %
 % FORMAT: SliceTiming = xASL_quant_SliceTiming(x, inputIm)
 %
@@ -10,10 +11,11 @@ function SliceTiming = xASL_quant_SliceTiming(x, inputIm)
 %   SliceTiming  - a vector of slice times defining the start of readout per slices with respect to the start of the readout of the first slices
 %                  with the length corresponding to the third dimension in inputIm
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
-% DESCRIPTION: This function takes the x.SliceReadoutTime, which can be a vector (of start of readout per slices relative to the 
-%              readout of the first slice) or scalar (with difference in readout times between two consecutives slices) and creates a vector
-%              (of the relatives timings for each slices) out of it with the correct length corresponding to the number of slices in the inputIm. 
-%              It also checks the x.readout_dim, and for 3D readouts it returns 0.
+% DESCRIPTION: This function takes the x.Q.SliceReadoutTime, which can be:
+%              i) a vector of slice readout-timings relative to the start of readout of the first slice or (i.e. the BIDS definition of SliceTiming)
+%              ii) a scalar with difference in readout times between the consecutives slices (i.e. the xASL legacy definition of SliceTiming)
+%              The function creates a vector (of the relatives timings for each slices) out of it with the correct length corresponding to the number of slices in the inputIm 
+%              corresponding to the BIDS definition. It also checks the x.readout_dim, and for 3D readouts it returns 0.
 %              It loads the image from inputIm and calculates the SliceTiming according to the number of slices in the third dimension
 %              If a path is given, it also checks if it can find a JSON sidecar, then it loads the JSON sidecar, and looks for SliceTiming inside it. If
 %              SliceTiming/SliceReadoutTime is found in the JSON sidecar, it prioritize it over the value in the x-struct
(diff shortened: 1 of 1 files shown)


Functions/xASL_quant_SliceTiming.m
```diff
@@ -111,6 +112,43 @@
...
+		% See if the slice-timing difference between slices is equal for all slices
+		[SliceTimingDiffUnique,SliceTimingDiffIndex] = uniquetol(SliceTimingDiff,0.01);
+		
+		% Make sure that the positive values are listed first
+		[SliceTimingDiffUnique,SliceTimingDiffIndexSort] = sort(SliceTimingDiffUnique,'descend');
+		SliceTimingDiffIndex = SliceTimingDiffIndex(SliceTimingDiffIndexSort);
+		
+		if length(SliceTimingDiffUnique) > 2
+			fprintf('Warning: Inconsistent SliceTiming between slices\n');
+		elseif length(SliceTimingDiffUnique) == 2
```
HenkMutsaerts, line 129: Are you sure multiband can only have 2 values? I thought it could also have higher acceleration factors, leading to multiple SliceReadoutTimes

Functions/xASL_quant_SliceTiming.m
```diff
@@ -111,6 +112,43 @@
 	SliceTiming = (0:1:(nSlices-1)) * x.Q.SliceReadoutTime;
 elseif length(x.Q.SliceReadoutTime) == nSlices
 	SliceTiming = x.Q.SliceReadoutTime;
+	
+	% Check for SliceTiming consistency between slices if a vector was provided
+	if nSlices > 1
```
HenkMutsaerts, line 118: The nomenclature should be clear throughout this function:

`SliceTiming` == absolute values of slice time after 0 (BIDS)
`SliceReadoutTime` == difference of SliceTiming (xASL legacy)

So `length(unique(SliceTiming(:)))>1` should always be true, whereas below this seems not to be the case?

I see now that you use `SliceTimingDiff` instead of `SliceReadoutTime`, why not using the old term? There is no reason not to continue using the old `SliceReadoutTime` right?

Most important is that this is clearly described in the header!

Functions/xASL_quant_SliceTiming.m
```diff
@@ -111,6 +112,43 @@
...
+		
+		% Make sure that the positive values are listed first
+		[SliceTimingDiffUnique,SliceTimingDiffIndexSort] = sort(SliceTimingDiffUnique,'descend');
+		SliceTimingDiffIndex = SliceTimingDiffIndex(SliceTimingDiffIndexSort);
+		
+		if length(SliceTimingDiffUnique) > 2
+			fprintf('Warning: Inconsistent SliceTiming between slices\n');
+		elseif length(SliceTimingDiffUnique) == 2
+			% This could be multi-band
+			bIsMB = false;
```
HenkMutsaerts, line 131: would write out `MultiBand` instead of `MB` to keep code readable

HenkMutsaerts (changes requested):

Functions/xASL_quant_SliceTiming.m
jan-petr: Haha - yes and no. Multiband usually has higher acceleration factors, but we don't look here at the SliceTiming, but at SliceTimingDiff!!
So an example of multiband with factor 3, SliceTime 30, and 12 slices  would be
0 30 60 90 0 30 60 90 0 30 60 90
So it would create 3 blocks of 4 slices. So 4 different times. BUT the differences of the neighboring slices are only 2: 30 and -90.

So the '2' here doesn't refer to the number of blocks or acceleration factor, but to the fact that the SliceTime increases equally within block and decreases equally between blocks = 2 different changes.

Functions/xASL_quant_SliceTiming.m
jan-petr: Good comment. Changed on all occurences.

Functions/xASL_quant_SliceTiming.m
jan-petr: I agree that this has to be clearly described in the headers and comments.

Though my current use is mostly correct. 
`SliceTimingDiff` is a temporary variable that describes the difference between neighboring slices, has the same length as the number of slices - 1. And can be negative for Multi-Band. It is used only internally to determine if the SliceTiming is consistent and/or multi-band. 

`SliceReadoutTime` can be either a single value (positive or zero) that describes the between slice difference or has the same length as the number of slices. This makes it different from SliceTimingDiff.

This function does the job of unifying SliceReadoutTime (legacy or BIDS) to SliceTiming (BIDS). Checks if the possibly manual input is fine.

This was all described in the header, but I have improved it - is it now more clear? 

length(unique(SliceTiming(:)))>1 - yes, this should be always true - where do you think it is not?

HenkMutsaerts:
Looks good, I thought only that we had `SliceReadoutTime` and `SliceTiming`, it seems as if this all is complicated a bit by multi-band:)

### I would only add an explanation of the different terms in the header:
% In ExploreASL, we currently use the following terminology:
% `SliceTiming`: BIDS parameter denoting the actual value when the slice is read out, with respect to the start of ... bla bla (s)
% `SliceReadoutTime`: legacy ExploreASL parameter providing a single value ... (will be phased out)
% `SliceTimingDiff`: internal parameter in this function for ...

HenkMutsaerts (changes requested):
Looks good, I thought only that we had `SliceReadoutTime` and `SliceTiming`, it seems as if this all is complicated a bit by multi-band:)

### I would only add an explanation of the different terms in the header:
% In ExploreASL, we currently use the following terminology:
% `SliceTiming`: BIDS parameter denoting the actual value when the slice is read out, with respect to the start of ... bla bla (s)
% `SliceReadoutTime`: legacy ExploreASL parameter providing a single value ... (will be phased out)
% `SliceTimingDiff`: internal parameter in this function for ...

(after this you can ask me to approve this without reading the code ;)

jan-petr:
OK. Header improved. Please approve.

HenkMutsaerts (approved):
Ah sorry I just didn't understand why `SliceReadoutTime` should `"contain a vector matching the BIDS definition"`, as this is not legacy at all ;) Probably has to do with multiband :) Thanks for clarifying!

jan-petr:
> Ah sorry I just didn't understand why `SliceReadoutTime` should `"contain a vector matching the BIDS definition"`, as this is not legacy at all ;) Probably has to do with multiband :) Thanks for clarifying!

Yes, to be able to deal with multi-band and other sequences that are non-standard (i.e. single difference number is not enough to describe it), we had to accommodate for this option also in the DataPar. So SliceReadoutTime can be indeed given either as legacy and as BIDS.

PR #453, opened 2021-03-25
#452 added atlas descriptions to DataParTemplate

### LINKED ISSUE #452: Provide all atlas options in DataParTemplate
List all atlas options in the DataParTemplate.m file and provide a bit of information about them

I have added the descriptions, but this needs to be checked and all my comments below about the current atlases answered - there are some doubts about the atlas integrity.

### Release note
Provided descriptions of available atlas options.

jan-petr:
Closes #452

### DIFF
--- a/DataParTemplate.m
+++ b/DataParTemplate.m
@@ -314,24 +314,27 @@ function x = DataParTemplate(x)
 %                       Exemplary notation within the JSON parameter file:
 %                       "x": [{
 %                       ...
-%                       "S": {"Atlases": ["TotalGM","DeepWM","Hammers","HOcort_CONN","HOsub_CONN","Mindboggle_OASIS_DKT31_CMA"]}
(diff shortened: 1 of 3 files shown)


DataParTemplate.m
```diff
@@ -319,6 +318,21 @@
 %                       ...
 %                       }]
 %                       (OPTIONAL, DEFAULT={'TotalGM','DeepWM'})
+%                       Available atlases (please check the atlas NIfTI and accompanying files for more information):
```
HenkMutsaerts, line 321: If we do this, we should add which atlases are open source (those in SPMmodified subdirectories and Mindboggle) and free for non-commercial use (e.g., HammersCAT12)

HenkMutsaerts (changes requested):

DataParTemplate.m
jan-petr: Good point. Modified that in the code

jan-petr:
I have fixed your comments and I am fine with all your answers. From my side, we can merge. Check my latest edit only 0f82dbb

HenkMutsaerts (approved):

PR #456, opened 2021-03-25
Optimization #62 json read write

### LINKED ISSUE #62: Standardize JSON reading and writing
Currently we have xASL_import_json.m, spm_jsonread, spm_jsonwrite for reading and writing of JSONs
We should:
* rename these into xASL_jsonRead & xASL_jsonWrite (like xASL_tsvRead & xASL_tsvWrite)
* move them to SPMmodified/xASL
* change throughout
* xASL_import_json is already a wrapper, that needs cleaning & renaming
* spm_jsonwrite needs a wrapper. We can move some stuff outside of our edited spm_jsonwrite perhaps

NEEDs further discussion. spm_jsonread and spm_jsonwrite work perfectly and they do not really need wrappers because no extra functionality is need - that is an unnecessary overhead. All our JSON files should adhere to the JSON format and they should not need wrappers - this only makes things more complicated and introduce potential errors (several of them fixed already that could have been avoided).

There are one exception to this - the dataPar and studyPar files. There, the content of JSONs need an extra interpretation (removing extra x-level), and fixing (numerical arrays entered as strings). However, the correct JSON fields generated by spm_jsonwrite do not save numerical arrays incorrectly as strings - so this behavior is really needed only for the incorrectly edited studyPar files.

Proposed solution:
1) Use spm_jsonwrite everywhere - that's JSON standard and no wrapper is needed.
2) Use spm_jsonread everywhere where possible - to read automatically and correctly saved JSON files. Go through code and change the calls of xASL_import_json to spm_jsonread where possible.
3) Only leave xASL_import_json for those files that contain strange manually edited constructions in JSON and thus need parsing - like studyPar and DataPar. But make sure that the JSON of those templates and all examples in Flavors, TestDataSet, 10-TestDataset are corrected. 
4) Renames xASL_import_json to xASL_io_ReadDataPar - make a joint function for this with m-files. And make sure to report a warning for all incorrect constructions, so that people edit JSONs correctly and we can phase out this function in the future.

### Release note
JSON i/o unified and using spm_jsonread and spm_jsonwrite for all operations.

jan-petr:
Close #62


MichaelStritt:
Calling it `xASL_io_ReadDataPar` instead of `xASL_import_json` really does make sense 👍

HenkMutsaerts:
> Calling it `xASL_io_ReadDataPar` instead of `xASL_import_json` really does make sense 👍

Agreed!

Functions/xASL_io_ReadDataPar.m
```diff
@@ -155,30 +155,38 @@
 % Iterate over fields
```
HenkMutsaerts, line 166: Same question here as in other function (sorry reviewed latest commit first, going back in time :)
These things may be useful for JSONs in general, not only for DataPar?

Functions/xASL_io_ReadDataPar.m
```diff
@@ -155,30 +155,38 @@
...
-			TempField = StructIn.(listFields{iField}); % Do not convert complex numbers
+	% Checks if conversion should be attempted. This check is done also inside the xASL_str2num, however,
+	% it is good to do it also beforehand, because you avoid unnecessary processing of numeric strings which
+	% can result in false-warnings about numeric being given inside a string. 
+	% i.e. this check avoids extra processing and warnings eventhough functionally not necessary
+	if ~isempty(StructIn.(listFields{iField})) && ~isnumeric(StructIn.(listFields{iField}))
+		% Convert string to num
+		TempField = xASL_str2num(StructIn.(listFields{iField}));
+		
+		% Check if we got an imaginary number
```
HenkMutsaerts, line 177: This check should probably go to `xASL_str2num`, in general we don't want text to convert to an imaginary number right?

Functions/xASL_io_ReadDataPar.m
```diff
@@ -155,30 +155,38 @@
...
-		TempField = TempField';
-	end
-	
-	% Check if there is a numerical vector inside the string
-	if min(isnumeric(TempField)) && min(~isnan(TempField))
-		StructOut.(listFields{iField}) = TempField;
-		if ~isequal(TempField,StructIn.(listFields{iField}))
-			fprintf('Warning: A numerical field provided as a string (%s). Please avoid this.\n', StructIn.(listFields{iField}));
+		
+		% Check if conversion was correct
```
HenkMutsaerts, line 184: Same here, shouldn't this go to `xASL_str2num`?

Functions/xASL_io_ReadDataPar.m
```diff
@@ -0,0 +1,250 @@
+function [x] = xASL_io_ReadDataPar(DataParFile)
+% xASL_io_ReadDataPar This function reads DATA_PAR .json or .m file and creates the x structure.
```
HenkMutsaerts, line 2: `DATA_PAR` is a meaningful term to us, but not to other first-time code readers. We don't need to keep calling it DataPar, but at least let's add some explanation:

"This deals with reading/parsing the data parameter file, which is a file containing settings specific to processing a certain dataset or study. Was historically abbreviated to DataPar".

Functions/xASL_io_ReadDataPar.m
```diff
@@ -0,0 +1,250 @@
...
+% 			"subject_regexp": 		"^Sub-\\d{3}$",
+% 			"M0": 					"separate_scan",
+% 			"BackgroundSuppressionNumberPulses": 2,
+% 			"readout_dim": 			"2D",
+% 			"QUALITY": 				false,
+% 			"Vendor": 				"Philips",
+% 		}]
+% }
+%
+% Don't forget to escape the backslashes!
```
HenkMutsaerts, line 42: Can we add an automatic escaping of backslashes (including a warning as well), at least for the `subject_regexp` field, which is generally the problem? Saves us some user questions ;)

Functions/xASL_io_ReadDataPar.m
```diff
@@ -0,0 +1,250 @@
...
+%	{"Atlases": ["TotalGM","DeepWM","Hammers","HOcort_CONN","HOsub_CONN","Mindboggle_OASIS_DKT31_CMA"]}
+%   ...
+%
+% -----------------------------------------------------------------------------------------------------------------------------------------------------
+% __________________________________
+% Copyright 2015-2021 ExploreASL
+
+%% Input Check
+if nargin < 1 || isempty(DataParFile) || ~exist(DataParFile, 'file')
+    error('DataParFile does not exist...');
```
HenkMutsaerts, line 57: would replace `DataParFile` with `Data parameter JSON file`

Functions/xASL_io_ReadDataPar.m
```diff
@@ -0,0 +1,250 @@
...
+%   "S":
+%	{"Atlases": ["TotalGM","DeepWM","Hammers","HOcort_CONN","HOsub_CONN","Mindboggle_OASIS_DKT31_CMA"]}
+%   ...
+%
+% -----------------------------------------------------------------------------------------------------------------------------------------------------
+% __________________________________
+% Copyright 2015-2021 ExploreASL
+
+%% Input Check
+if nargin < 1 || isempty(DataParFile) || ~exist(DataParFile, 'file')
```
HenkMutsaerts, line 56: variable `DataParFile` should be `DataParPath` or `pathDataPar` right?

Functions/xASL_io_ReadDataPar.m
```diff
@@ -0,0 +1,250 @@
...
+    error('DataParFile does not exist...');
+end
+
+% Input has to be a character array
+DataParFile = char(DataParFile);
+[Fpath, Ffile, Fext] = fileparts(DataParFile);
+
+%% Read an m-file
```
HenkMutsaerts, line 64: Perhaps good to separate this function into separate parts, and denote them with 1. 2. 3. etc like always, and adding this list to the header as well

Functions/xASL_io_ReadDataPar.m
```diff
@@ -0,0 +1,250 @@
...
+	end
+	% Write the JSON if it doesn't exist
+	if exist(PathJSON, 'file')
+		fprintf('Warning: m-files are deprecated, but cannot convert to json-file because a file with json extension exists in the same directory');
+	else
+		fprintf('Warning: m-files are deprecated, converted and saved to json-file. Please delete the original m-file');
+		spm_jsonwrite(PathJSON, x);
+	end
+
+elseif strcmp(Fext, '.json')
```
HenkMutsaerts, line 104: Should we do `strcmpi` to be forgiving if someone gets the capitals wrong (e.g. uses DataPar.Json and stalks us because the pipeline doesn't work ;)

Functions/xASL_adm_LoadParms.m
```diff
@@ -1,7 +1,7 @@
-function [Parms, x, Oldx] = xASL_adm_LoadParms(ParmsPath, x, bVerbose)
-%xASL_adm_LoadParmsMat Loads parameters from .mat or .json sidecar
+function [Parms, x] = xASL_adm_LoadParms(ParmsPath, x, bVerbose)
```
HenkMutsaerts, line 1: Are you sure we don't use the OldX anymore (I assume you are but just to be sure :)

HenkMutsaerts (changes requested):
Just some minor questions, looks very good! Great that we finally are able to clean up the pipeline a bit more :)

Functions/xASL_io_ReadDataPar.m
```diff
@@ -192,3 +203,48 @@
 end
 
 end
+
+%%xASL_io_ReadDataPar_ConvertCellArrays finds all the cell arrays entered as strings and fixes them
```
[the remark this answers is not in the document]
MichaelStritt: If we normally write json files using spm_jsonwrite, then this shouldn't happen anyway and it's really not according to official JSON format.
The DataPar is one of the only ones that people write by hand and even here I'd try to teach everyone in a tutorial or doc rather the correct way instead of allowing wrong formats.

Functions/xASL_bids_MergeNifti.m
```diff
@@ -187,7 +187,7 @@
 	end
 		
 	% If GE is not identified or ImageType field doesn't exist, then exits
-	if ~regexpi(varManufacturer,'GE') || ~isfield(jsonPar,'ImageType')
+	if isempty(regexpi(varManufacturer,'GE')) || ~isfield(jsonPar,'ImageType')
 		return;
```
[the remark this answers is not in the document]
MichaelStritt: I guess he needs to skip the rest of the processing pipeline and an error wouldn't be helpful. But maybe we should add a printf or a warning?

Functions/xASL_io_ReadDataPar.m
MichaelStritt: We use the same name in the master script though ;) so I think that's okay. Alternatively we can link to that

Functions/xASL_io_ReadDataPar.m
```diff
@@ -136,18 +136,29 @@
...
 	error('Unknown file extension');
 end
 
 end
 
+%% Goes through the JSON fields and make sure that incorrect entries are corrected
+function [StructOut] = xASL_io_ReadDataPar_FixFields(StructIn)
+
+% First fixes the numerical fields in the root structure and x.Q substructure
+StructOut = xASL_io_ReadDataPar_ConvertNumericalFields(StructIn);
```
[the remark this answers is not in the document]
jan-petr: Simple reason. The older function was written for these two only. The new function I wrote more robustly. The old function worked as these errors were only in root and Q, but they can in theory be put anywhere. So I wrote the new function more robustly.

Shall we change both to the simpler or more complex version. I would say we can leave it as it is as this function will be soon discontinued anyway and it is working with the current version and people should not be making these mistakes in the future anyway.

Functions/xASL_io_ReadDataPar.m
```diff
@@ -180,7 +191,7 @@
 		if min(isnumeric(TempField)) && min(~isnan(TempField))
 			StructOut.(listFields{iField}) = TempField;
 			if ~isequal(TempField,StructIn.(listFields{iField}))
-				fprintf('Warning: A numerical field provided as a string (%s). Please avoid this.\n', StructIn.(listFields{iField}));
+				fprintf('Warning: A numerical field provided as a string (%s). Converted, but should be avoided.\n', StructIn.(listFields{iField}));
```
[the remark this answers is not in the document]
jan-petr: Fixed

Functions/xASL_io_ReadDataPar.m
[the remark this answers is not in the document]
jan-petr: Exactly. Standard is a standard and no need to fix for deviations from that. We only do that for DataPar because we are responsible for providing the wrong instructions previously.

Functions/xASL_io_ReadDataPar.m
```diff
@@ -192,3 +203,48 @@
...
+% Iterate over fields
+for iField=1:length(listFields)
+	
+	% Finds the incorrect fields
+	if isstruct(StructIn.(listFields{iField}))
+		% Calls recursively the function for structs
+		StructOut.(listFields{iField}) = xASL_io_ReadDataPar_ConvertCellArrays(StructIn.(listFields{iField}));
+		
+	elseif ischar(StructIn.(listFields{iField})) && ~isempty(StructIn.(listFields{iField})) && ...
+			StructIn.(listFields{iField})(1) == '{' && StructIn.(listFields{iField})(end) == '}'
```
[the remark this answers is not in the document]
jan-petr: This works rather fine and is more lightweight. I would keep it.

Functions/xASL_io_ReadDataPar.m
```diff
@@ -192,3 +203,48 @@
...
+		% For chars entered in {} brackets, brakes the entries in '' to individuals arrays with a cell array
+		iFirst = 2;
+		iApostrophe = strfind(StructIn.(listFields{iField}),39);
```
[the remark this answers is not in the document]
jan-petr: Yes `'` would have been better, but `'\''` doesn't work and I couldn't find a better way to escape it. But I will add a commentary.

Functions/xASL_io_ReadDataPar.m
```diff
@@ -192,3 +203,48 @@
 end
 
 end
+
+%%xASL_io_ReadDataPar_ConvertCellArrays finds all the cell arrays entered as strings and fixes them
+function StructOut = xASL_io_ReadDataPar_ConvertCellArrays(StructIn)
+
```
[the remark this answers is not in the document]
jan-petr: Done.

Functions/xASL_io_ReadDataPar.m
```diff
@@ -192,3 +203,48 @@
...
+				% Else we just end a let the remains be written
+				iFirst = iApostrophe(iCurrent+1)+1;
+			end
+			iCurrent = iCurrent + 2;
+		end
+		% If some opened strings remain, write them to a last array field
+		if iFirst<=length(StructIn.(listFields{iField}))-1
+			StructOut.(listFields{iField}){1,length(StructOut.(listFields{iField}))+1} = StructIn.(listFields{iField})(iFirst:(length(StructIn.(listFields{iField}))-1));
+		end
+		fprintf('Warning: An incorrect format of a cell array provided (%s). Converted, but should be avoided.\n', StructIn.(listFields{iField}));
```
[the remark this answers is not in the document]
jan-petr: Done

Functions/xASL_bids_parseM0.m
```diff
@@ -1,27 +1,26 @@
 function xASL_bids_parseM0(pathASLNifti)
-%xASL_bids_parseM0 Check the ASL file in BIDS format for its M0 possibilities and 
-% convert the ASL NIFTI in ExploreASL legacy format
+%xASL_bids_parseM0 Check the ASL file in Legacy format, but with BIDS sidecars for its M0 possibilities and
```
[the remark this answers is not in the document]
jan-petr: ok. next time :)

Functions/xASL_io_ReadDataPar.m
jan-petr: As previously - we shouldn't fix JSON errors in general. People should stick to JSON format. We only fix DataPar where we gave the wrong template.

Functions/xASL_io_ReadDataPar.m
jan-petr: Improved the explanation.

Functions/xASL_io_ReadDataPar.m
jan-petr: I wouldn't add automatic escaping. Because you will never know if the backslash should be escaped, or if it was meant to escape some other character ;)

Functions/xASL_io_ReadDataPar.m
jan-petr: Replaced. But didn't put the 'JSON' in as it can be also an m-file.

Functions/xASL_io_ReadDataPar.m
jan-petr: No - I have tried that, but the structure of this function does not allow a nice separation to parts. The comments inside the code explain enough. I put in more comments, but numbering would not make it easier to read.

Functions/xASL_adm_LoadParms.m
jan-petr: Yes. Checked the code and searched through all uses of it everywhere.

Functions/xASL_io_ReadDataPar.m
jan-petr: Yes. Changed.

Functions/xASL_io_ReadDataPar.m
jan-petr: Yes. Done everywhere.

Functions/xASL_io_ReadDataPar.m
jan-petr: Right. But this is probably not really relevant for the other cases, so I slightly prefer to keep this here for this specific use.

Functions/xASL_io_ReadDataPar.m
jan-petr: As said above. I would keep the imaginary in this function. All the other checks are specific for this function to avoid calling the xASL_str2num to reduce overhead and warnings

Functions/xASL_bids_MergeNifti.m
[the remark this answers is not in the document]
jan-petr: Haha - you are right :) Though in this case, the return is used to break a for-cycle that goes through all files. So, we want to interrupt the execution of the for-cycle to stop loading more JSONs if the conditions are met to save processing time. At the same time, we cannot put this condition elsewhere, because we would have to pre-read all the JSONs.

Altogether - you are right, but to get rid of the `return` here, we would have to move this `if` elsewhere outside of the `for` cycle and this would result in more time-consuming code. So in this case, the `return` is justified and hard to replace.

jan-petr:
I have fixed some comments and provide answers for other comments. Please check the new code changes and either resolve/answer my other comments.

Functions/xASL_io_ReadDataPar.m
HenkMutsaerts: OK

Functions/xASL_io_ReadDataPar.m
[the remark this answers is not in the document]
HenkMutsaerts: OK, looks like `''''` does the job as well as `39`, but as long as it is understood in a comment that `39` means `'`

(I believe in Matlab a `'` is escaped by `''` rather than `\'`

Functions/xASL_io_ReadDataPar.m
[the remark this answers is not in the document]
HenkMutsaerts: OK for writing. When reading JSONs with wrong format, it would be nice if this is possible also for sidecars, not only DataPar, but of course giving a warning. We want users to use a formal JSON format, but remember that if they have a wrong sidecar .JSON, and our script crashes, it can take a while to find this out.

That's why I would make such warnings (and automatic fixes) general.

Unless this is a very specific fix to instructions we provided wrongly in the past.

Functions/xASL_io_ReadDataPar.m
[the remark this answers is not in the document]
HenkMutsaerts: I thought the "old function" was actually also new ;) At least I don't recognize both :) Anyway if this is discontinued soon then we shouldn't bother. I thought it wasn't much work to change. Happy either way, as long as it works and is easy readable :)

Functions/xASL_io_ReadDataPar.m
HenkMutsaerts: I still think the automatic escaping could be very nice, only for this field, and only when used in combination with the digits (then you know if it should or shouldn't be escaped).

HenkMutsaerts (approved):

Issue #451, opened 2021-03-24
Clean unused code in Development


jan-petr:
**This initial comment should contain all information - please reflect decisions reached in the discussion below in this list**
I have noticed that while Functions folder looks pretty good, the Development folder is a huge mess. So I propose that we at least try to clean old functions that are no longer used and will never be. We can either completely delete them, or move them to Custom scripts. This will make the code cleaner and easier to maintain.

We don't have to do everything, but at least the obvious files in the first round. I have checked all those below and here are my propositions:

REMOVE:
- `Development/BIDS/ZeroingOut.m` (:+1:)
- `Development/media/CreateMeanControl.m` (:+1:)
- `Development/media/CreateGif_Averaging.m` (:+1:)
- `Development/media/CreateGif_Averaging_4D_template.m` (:+1:)
- `CalculateDistanceFromCenter.m` (:+1:)

MOVE TO THE ROOT DEVELOPMENT FOLDER:
- `Development/ExploreASL_make_standalone/ConvertNiiGz2Mat.m` (:+1:)

LEGACY. These functions appear to be no longer used, but they are nicely written, not a mess. So lets make a folder CustomScripts/ExploreASLLegacy and move those there for now:
- `Development/xASL_bids_DataPar2JSON.m`  ( :+1: in #62)
- `Development/xASL_bids_InsertJSONFields.m` (:+1:)
- `Development/SaraElgayar_BIDS` (:+1:)

PROMOTE. Those functions do not longer need to be in the Development folder and can be moved to functions. But before that, the headers should be cleaned and the whole code roughly checked.
- `xASL_test_BIDSFlavorsFull` ( :+1: in #394)
- `xASL_im_HausdorffDist` (:+1:)
- `xASL_bids_TestBidsConversion` ( :+1: in #290)

### Release note
Clean unused code and cleaned the development directory.

MichaelStritt:
For `xASL_bids_TestBidsConversion` we should probably find a nice way to do this with regard to #290.

jan-petr:
> For `xASL_bids_TestBidsConversion` we should probably find a nice way to do this with regard to #290.

I agree to do that in #290 . If you do so, just make a note in the original commit at the top so that we have the list up-to-date :)

Can we get rid of the 'make_standalone' functions?

MichaelStritt:
> Can we get rid of the 'make_standalone' functions?

[...]

PR #458, opened 2021-03-26
Closes #290 ExploreASL_Master contains BIDS Import Option

### LINKED ISSUE #290: Option to run DICOM2BIDS & BIDS2Legacy in ExploreASL Master
An option to run DICOM2BIDS & BIDS2Legacy in ExploreASL Master would help to get rid of the additional compiled ExploreASL_Import and other users, which don't use the mTRIAL format, would be able to use the import functionality in the docker as well.

MichaelStritt:
Hey, so this is my initial implementation. I haven't tested it that extensively yet, but I wanted to show you my ideas on how to do some of these things.

@jan-petr: You suggested to move the import out of the initialization. The new code should do that. To load the data, we potentially have to re-initialize. I implemented that as well. Maybe you could check that out?

@HenkMutsaerts: I had this weird bug, where BIDS2LEGACY converted a bids compliant ASL file to an LEGACY format ASL4D. After that the parseM0 script is called which then says that ASL4D is not a valid BIDS file. This is why I added the try catch statement, which is certainly not the right fix, but I wanted to ask you how we should do this. (**section** `9. Parse M0`)

An additional question is about the DataPar file which is saved at the end of BIDS2LEGACY. Which fields should be stored there? I'm talking about **lines 85-101** in `ExploreASL_ImportWorkflow`.

### DIFF
--- a/Development/BIDS/ZeroingOut.m
+++ b/Development/BIDS/ZeroingOut.m
@@ -1,8 +1,8 @@
 %% Zeroing out imaging data for BIDS example datasets
 
-ExploreASL_Master('',0);
+ExploreASL_Master;
 
-Rdir = 'C:\Users\henkj\Google Drive\WorkGeneral\StudiesProjects\BIDS_ASL\TestDataSet_ImportBIDS\raw'
+Rdir = 'C:\Users\henkj\Google Drive\WorkGeneral\StudiesProjects\BIDS_ASL\TestDataSet_ImportBIDS\raw';
 Flist = xASL_adm_GetFileList(Rdir, '.*\.(nii|nii\.gz)$', 'FPListRec', [0 Inf], false);
 
 for iL=1:length(Flist)
--- a/Development/BIDS/ZeroingOut.m
+++ b/Development/BIDS/ZeroingOut.m
@@ -1,6 +1,6 @@
 %% Zeroing out imaging data for BIDS example datasets
 
-ExploreASL_Master;
+ExploreASL_Initialize;
 
 Rdir = 'C:\Users\henkj\Google Drive\WorkGeneral\StudiesProjects\BIDS_ASL\TestDataSet_ImportBIDS\raw';
 Flist = xASL_adm_GetFileList(Rdir, '.*\.(nii|nii\.gz)$', 'FPListRec', [0 Inf], false);
--- a/ExploreASL_ImportConfig.m
+++ b/ExploreASL_ImportConfig.m
@@ -9,34 +9,19 @@ function imPar = ExploreASL_ImportConfig(StudyRoot)
 % DESCRIPTION: Please read the help of ExploreASL_Import for more information
 % EXAMPLE:     n/a
 % __________________________________
-% Copyright 2015-2019 ExploreASL
-
-% Just configuring, no need to check if the directory exists yet
-%if nargin<1 || ~exist(StudyRoot,'dir')
-%    error('Please provide path to StudyRoot folder as input argument');
-%end
+% Copyright 2015-2021 ExploreASL
 
 if strcmp(StudyRoot(end),'\') || strcmp(StudyRoot(end),'/')
     StudyRoot = StudyRoot(1:end-1); % bugfix
 end
 
+% Get file path, name and extension
 [fpath, fname, fext] = fileparts(StudyRoot);
 
+% Set imPar fields accordingly
 imPar.studyID = [fname fext];
 imPar.AnalysisRoot = fpath;
 imPar.RawRoot = fpath;
-
-% Does not need to check for the RAW dir at this moment
-%RawDir = fullfile(StudyRoot,'raw');
-
-%if ~exist(RawDir,'dir')
-%    error(['Couldnt find raw DICOM data in ' RawDir]);
-%end
-
-% ExploreASL does not need to be initialized here
-% ExploreASL_Master;
-
-% fprintf('\n%s\n\n',['Found ' RawDir ' folder...']);
 
 %% -----------------------------------------------------------------------------
 %% Initialize the study-specific parameters
--- a/ExploreASL_ImportConfig.m
+++ b/ExploreASL_ImportConfig.m
@@ -34,7 +34,7 @@ imPar.RawRoot = fpath;
 %end
 
 % ExploreASL does not need to be initialized here
-% ExploreASL_Master('',0);
+% ExploreASL_Master;
 
 % fprintf('\n%s\n\n',['Found ' RawDir ' folder...']);
 
--- a/ExploreASL_ImportWorkflow.m
+++ b/ExploreASL_ImportWorkflow.m
@@ -46,7 +46,7 @@ function [x] = ExploreASL_ImportWorkflow(x)
                 warning('ImportArray was set to 1, but the sourceStructure file does not exist');
             end
         else
-            warning('ImportArray was set to 1, but there is not DataParPath, Import will not be executed');
+            warning('ImportArray was set to 1, but there is no DataParPath, Import will not be executed');
         end
     end
     
--- a/ExploreASL_ImportWorkflow.m
+++ b/ExploreASL_ImportWorkflow.m
@@ -1,0 +1,105 @@
+function [x] = ExploreASL_ImportWorkflow(x)
+%ExploreASL_ImportWorkflow Multi-step import workflow for DCM2NII, NII2BIDS & BIDS2LEGACY.
+%
+% FORMAT: [x] = ExploreASL_ImportWorkflow(x)
+%
+% INPUT:
+%   x             - Struct containing pipeline environment parameters, useful when only initializing ExploreASL/debugging
+%
+% OUTPUT:
+%   x             - Struct containing pipeline environment parameters, useful when only initializing ExploreASL/debugging
+%
+% -----------------------------------------------------------------------------------------------------------------------------------------------------
+% DESCRIPTION:    Multi-step import workflow for DCM2NII, NII2BIDS & BIDS2LEGACY.
+%
+% -----------------------------------------------------------------------------------------------------------------------------------------------------
+% EXAMPLE:        n/a
+% __________________________________
+% Copyright 2015-2021 ExploreASL
+
+    %% Import Workflow
+    
+    % We expect x.DataParPath to be the sourceStructure.json file, so the root directory should be the study directory
+    [x.StudyRoot,~,~] = xASL_fileparts(x.DataParPath);
+    
+    % Check if at least one of the three steps should be performed
+    if sum(x.ImportArray)>0
+        if ~isempty(x.DataParPath)
+            if exist(x.DataParPath,'file')
+                % DICOM TO NII
+                if x.ImportArray(1)==1
+                    ExploreASL_ImportBIDS(x.StudyRoot, x.DataParPath, [], [1 0 0], false, true, false, false, x);
+                end
+                % NII TO BIDS
+                if x.ImportArray(2)==1
+                    ExploreASL_ImportBIDS(x.StudyRoot, x.DataParPath, [], [0 1 0], false, true, false, false, x);
+                end
+                % ANONYMIZE
+                if x.ImportArray(3)==1
+                    ExploreASL_ImportBIDS(x.StudyRoot, x.DataParPath, [], [0 0 1], false, true, false, false, x);
+                end
+                % BIDS TO LEGACY
+                if x.ImportArray(4)==1
+                    x = xASL_Import_BIDS2LEGACY(x);
+                end
+            else
+                warning('ImportArray was set to 1, but the DataParPath does not point to a directory');
+            end
+        else
+            warning('ImportArray was set to 1, but there is not DataParPath, Import will not be executed');
+        end
+    end
+    
+    
+end
+
+%% xASL_Import_BIDS2LEGACY
+function [x] = xASL_Import_BIDS2LEGACY(x)
+
+    % Go through all studies
+    ListFolders = xASL_adm_GetFileList(x.StudyRoot, '^.+$', 'FPListRec', [0 Inf], 1);
+    for iList=1:numel(ListFolders)
+        % Convert only those containing raw data
+        [thisRootFolder,thisFolderName,~] = xASL_fileparts(ListFolders{iList});
+        if strcmp(thisFolderName,'rawdata') && exist(ListFolders{iList},'dir')
+            
+            % Currently, we clean the old data for unix only
+            if isunix
+                if exist(fullfile(ListFolders{iList}, 'derivatives'), 'dir')
+                    diary('off');
+                    fclose('all'); % ensure that no file is locked
+                    system(['rm -rf ' fullfile(ListFolders{iList}, 'derivatives')]);
+                end
+            else
+                fprintf('Delete existing derivatives folders...\n');
+                diary('off');
+                fclose('all'); % ensure that no file is locked
+                xASL_delete(fullfile(ListFolders{iList}, 'derivatives'),true);
+            end
+            
+            % Default dataPar.json for the testing that is fast to run
+            dataPar.x = x;
+
+            % Run the legacy conversion: Check if a dataPar is provided, otherwise use the defaults
+            fListDataPar = xASL_adm_GetFileList(ListFolders{iList},'(?i)(^dataPar.*\.json$)', 'FPList', [], 0);
+            if length(fListDataPar) < 1
+                % Fill the dataPars with default parameters
+                xASL_bids_BIDS2Legacy(thisRootFolder, 1, dataPar);
+            else
+                % Fill the dataPars with the provided parameters
+                dataPar = spm_jsonread(fListDataPar{1});
+                xASL_bids_BIDS2Legacy(thisRootFolder, 1, dataPar);
+            end
+        end
+    end
+    
+    % Overwrite DataParPath in x structure
+    fprintf('Overwriting x.DataParPath...\n');
+    ListDataParFiles = xASL_adm_GetFileList(x.StudyRoot, '^DataPar.+$', 'FPListRec', [0 Inf]);
+    x.DataParPath = ListDataParFiles{1,1};
+    
+end
+
+
+
+
--- a/ExploreASL_ImportWorkflow.m
+++ b/ExploreASL_ImportWorkflow.m
@@ -50,6 +50,9 @@ function [x] = ExploreASL_ImportWorkflow(x)
         end
     end
     
+    % Reset the import parameter (for the second initialization including the loading of the dataset)
+    x.ImportData = 0;
+    x.ImportArray = [0 0 0 0];
     
 end
 
@@ -79,26 +82,39 @@ function [x] = xASL_Import_BIDS2LEGACY(x)
 	            end
             end
             
-            % Default dataPar.json for the testing that is fast to run
-            dataPar.x = x;
+            % Default dataPar.json for the testing
+            warning('Are we supposed to write more fields to the DataPar file here?');
+            if isfield(x,'DataParPath')
+                dataPar.x.DataParPath = x.DataParPath;
+            else
+                dataPar.x.DataParPath = '';
+            end
+            if isfield(x,'subject_regexp')
+                dataPar.x.subject_regexp = x.subject_regexp;
+            else
+                dataPar.x.subject_regexp = '^sub-.*$';
+            end
+            if isfield(x,'DELETETEMP')
+                dataPar.x.DELETETEMP = x.DELETETEMP;
+            else
+                dataPar.x.DELETETEMP = 1;
+            end
 
             % Run the legacy conversion: Check if a dataPar is provided, otherwise use the defaults
             fListDataPar = xASL_adm_GetFileList(ListFolders{iList},'(?i)(^dataPar.*\.json$)', 'FPList', [], 0);
             if length(fListDataPar) < 1
                 % Fill the dataPars with default parameters
-                xASL_bids_BIDS2Legacy(thisRootFolder, 1, dataPar);
+                dataPar = xASL_bids_BIDS2Legacy(thisRootFolder, 1, dataPar);
             else
                 % Fill the dataPars with the provided parameters
                 dataPar = spm_jsonread(fListDataPar{1});
-                xASL_bids_BIDS2Legacy(thisRootFolder, 1, dataPar);
+                dataPar = xASL_bids_BIDS2Legacy(thisRootFolder, 1, dataPar);
             end
         end
     end
     
-    % Overwrite DataParPath in x structure
-    fprintf('Overwriting x.DataParPath...\n');
-    ListDataParFiles = xASL_adm_GetFileList(x.StudyRoot, '^DataPar.+$', 'FPListRec', [0 Inf]);
-    x.DataParPath = ListDataParFiles{1,1};
+    % Overwrite DataParPath
+    x.DataParPath = dataPar.x.DataParPath;
     
 end
 
--- a/ExploreASL_ImportWorkflow.m
+++ b/ExploreASL_ImportWorkflow.m
@@ -43,7 +43,7 @@ function [x] = ExploreASL_ImportWorkflow(x)
                     x = xASL_Import_BIDS2LEGACY(x);
                 end
             else
-                warning('ImportArray was set to 1, but the DataParPath does not point to a directory');
+                warning('ImportArray was set to 1, but the sourceStructure file does not exist');
             end
         else
             warning('ImportArray was set to 1, but there is not DataParPath, Import will not be executed');
@@ -63,7 +63,7 @@ function [x] = xASL_Import_BIDS2LEGACY(x)
         [thisRootFolder,thisFolderName,~] = xASL_fileparts(ListFolders{iList});
         if strcmp(thisFolderName,'rawdata') && exist(ListFolders{iList},'dir')
             
-            % Currently, we clean the old data for unix only
+            % Clean the old data
             if isunix
                 if exist(fullfile(ListFolders{iList}, 'derivatives'), 'dir')
                     diary('off');
@@ -71,10 +71,12 @@ function [x] = xASL_Import_BIDS2LEGACY(x)
                     system(['rm -rf ' fullfile(ListFolders{iList}, 'derivatives')]);
                 end
             else
-                fprintf('Delete existing derivatives folders...\n');
-                diary('off');
-                fclose('all'); % ensure that no file is locked
-                xASL_delete(fullfile(ListFolders{iList}, 'derivatives'),true);
+            	if exist(fullfile(ListFolders{iList}, 'derivatives'), 'dir')
+	                fprintf('Delete existing derivatives folders...\n');
+	                diary('off');
+	                fclose('all'); % ensure that no file is locked
+	                xASL_delete(fullfile(ListFolders{iList}, 'derivatives'),true);
+	            end
             end
             
             % Default dataPar.json for the testing that is fast to run
--- a/ExploreASL_ImportWorkflow.m
+++ b/ExploreASL_ImportWorkflow.m
@@ -67,19 +67,11 @@ function [x] = xASL_Import_BIDS2LEGACY(x)
         if strcmp(thisFolderName,'rawdata') && exist(ListFolders{iList},'dir')
             
             % Clean the old data
-            if isunix
-                if exist(fullfile(ListFolders{iList}, 'derivatives'), 'dir')
-                    diary('off');
-                    fclose('all'); % ensure that no file is locked
-                    system(['rm -rf ' fullfile(ListFolders{iList}, 'derivatives')]);
-                end
-            else
-            	if exist(fullfile(ListFolders{iList}, 'derivatives'), 'dir')
-	                fprintf('Delete existing derivatives folders...\n');
-	                diary('off');
-	                fclose('all'); % ensure that no file is locked
-	                xASL_delete(fullfile(ListFolders{iList}, 'derivatives'),true);
-	            end
+            if exist(fullfile(ListFolders{iList}, 'derivatives'), 'dir')
+                fprintf('Delete existing derivatives folders...\n');
+                diary('off');
+                fclose('all'); % ensure that no file is locked
+                xASL_delete(fullfile(ListFolders{iList}, 'derivatives'),true);
             end
             
             % Default dataPar.json for the testing
--- a/ExploreASL_Initialize.m
+++ b/ExploreASL_Initialize.m
@@ -2,7 +2,7 @@ function [x] = ExploreASL_Initialize(varargin)
 %ExploreASL_Initialize Initializes ExploreASL
 %
 % FORMAT: 
-%   [x] = ExploreASL_Initialize([DataParPath, ImportModules, ProcessModules, SkipPause, iWorker, nWorkers])
+%   [x] = ExploreASL_Initialize([DataParPath, ImportModules, ProcessModules, bPause, iWorker, nWorkers])
 %
 % INPUT:
 %   varargin    - This script accepts the same arguments as ExploreASL_Master. Check out the definitions there.
@@ -770,7 +770,7 @@ function p = inputParsing(varargin)
     validDataParPath = @(variable) ischar(variable) || isempty(variable);
     validImportModules = @(variable) ischar(variable) || isempty(variable) || isnumeric(variable);
     validProcessModules = @(variable) ischar(variable) || isempty(variable) || isnumeric(variable);
-    validSkipPause = @(variable) ischar(variable) || isempty(variable) || isnumeric(variable);
+    validbPause = @(variable) ischar(variable) || isempty(variable) || isnumeric(variable);
     validiWorker = @(variable) ischar(variable) || isempty(variable) || isnumeric(variable);
     validnWorkers = @(variable) ischar(variable) || isempty(variable) || isnumeric(variable);
     
@@ -778,7 +778,7 @@ function p = inputParsing(varargin)
     defaultDataParPath = [];
     defaultImportModules = [0 0 0 0];
     defaultProcessModules = [0 0 0];
-    defaultSkipPause = 0;
+    defaultbPause = 0;
     defaultiWorker = 1;
     defaultnWorkers = 1;
     
@@ -786,7 +786,7 @@ function p = inputParsing(varargin)
     addOptional(p, 'DataParPath', defaultDataParPath, validDataParPath);
     addOptional(p, 'ImportModules', defaultImportModules, validImportModules);
     addOptional(p, 'ProcessModules', defaultProcessModules, validProcessModules);
-    addOptional(p, 'SkipPause', defaultSkipPause, validSkipPause);
+    addOptional(p, 'bPause', defaultbPause, validbPause);
     addOptional(p, 'iWorker', defaultiWorker, validiWorker);
     addOptional(p, 'nWorkers', defaultnWorkers, validnWorkers);
     
@@ -803,7 +803,7 @@ function parameters = ExploreASL_Initialize_convertParsedInput(parameters)
     if isempty(parameters.DataParPath),     parameters.DataParPath = '';                                    end
     if ischar(parameters.ImportModules),    parameters.ImportModules = str2num(parameters.ImportModules);   end
     if ischar(parameters.ProcessModules),   parameters.ProcessModules = str2num(parameters.ProcessModules); end
-    if ischar(parameters.SkipPause),        parameters.SkipPause = str2num(parameters.SkipPause);           end
+    if ischar(parameters.bPause),           parameters.bPause = str2num(parameters.bPause);                 end
     if ischar(parameters.iWorker),          parameters.iWorker = str2num(parameters.iWorker);               end
     if ischar(parameters.nWorkers),         parameters.nWorkers = str2num(parameters.nWorkers);             end
     
@@ -820,9 +820,9 @@ function parameters = ExploreASL_Initialize_convertParsedInput(parameters)
                                      parameters.ProcessModules(1)];
     end
     
-    % Different default for deployed mode
+    % Make it impossible to set bPause to true in deployed mode
     if isdeployed
-        parameters.SkipPause = 1;
+        parameters.bPause = 0;
     end
 
 
@@ -836,7 +836,7 @@ function x = ExploreASL_Initialize_storeParsedInput(parameters)
     x.DataParPath = parameters.DataParPath;
     x.ImportModules = parameters.ImportModules;
     x.ProcessModules = parameters.ProcessModules;
-    x.SkipPause = parameters.SkipPause;
+    x.bPause = parameters.bPause;
     x.iWorker = parameters.iWorker;
     x.nWorkers = parameters.nWorkers;
     
@@ -856,7 +856,7 @@ function ExploreASL_Initialize_printSettings(x)
     fprintf('Run Structural Module\t%d\n', x.ProcessModules(1));
     fprintf('Run ASL Module\t\t\t%d\n', x.ProcessModules(2));
     fprintf('Run Population Module\t%d\n', x.ProcessModules(3));
-    fprintf('SkipPause\t\t\t\t%d\n', x.SkipPause);
+    fprintf('bPause\t\t\t\t\t%d\n', x.bPause);
     fprintf('iWorker\t\t\t\t\t%d\n', x.iWorker);
     fprintf('nWorkers\t\t\t\t%d\n', x.nWorkers);
     fprintf('==============================================================================================\n');
--- a/ExploreASL_Initialize.m
+++ b/ExploreASL_Initialize.m
@@ -831,7 +831,7 @@ function parameters = convertParsedInput(parameters)
     
     % Different default for deployed mode
     if isdeployed
-        parameters.SkipPause = 0;
+        parameters.SkipPause = 1;
     end
 
 
--- a/ExploreASL_Initialize.m
+++ b/ExploreASL_Initialize.m
@@ -155,7 +155,9 @@ function [x] = ExploreASL_Initialize(varargin)
 
     % Recheck the DataPar/sourceStructure file, which is possibly not a file or does not exist
     if ~exist(x.DataParPath,'file')
-        fprintf('DataPar file does not exist, ExploreASL will only be initialized...\n');
+        if x.ImportData || x.ProcessData
+            fprintf('DataPar file does not exist, ExploreASL will only be initialized...\n');
+        end
         x.ProcessData = 0;
         x.ProcessModules = [0 0 0];
     else % DataPar file/folder exists
--- a/ExploreASL_Initialize.m
+++ b/ExploreASL_Initialize.m
@@ -846,16 +846,30 @@ end
 %% Print chosen settings
 function ExploreASL_Initialize_printSettings(x)
 
+    % Fallbacks
+    dcm2nii = '';
+    nii2bids = '';
+    anonymize = '';
+    bids2legacy = '';
+    Structural = '';
+    ASL = '';
+    Population = '';
+    
+    % Texts
+    if x.ImportModules(1)==1,   dcm2nii = 'DCM2NII';            end
+    if x.ImportModules(2)==1,   nii2bids = 'NII2BIDS';          end
+    if x.ImportModules(3)==1,   anonymize = 'ANONYMIZE';        end
+    if x.ImportModules(4)==1,   bids2legacy = 'BIDS2LEGACY';    end
+    if x.ProcessModules(1)==1,  Structural = 'Structural';      end
+    if x.ProcessModules(2)==1,  ASL = 'ASL';                    end
+    if x.ProcessModules(3)==1,  Population = 'Population';      end
+    
+    % Printing
     fprintf('==================================== ExploreASL Settings =====================================\n');
     if length(x.DataParPath)>66,    fprintf('DataParPath\t\t\t\t...%s\n', x.DataParPath(end-66:end));
     else,                           fprintf('DataParPath\t\t\t\t%s\n', x.DataParPath); end
-    fprintf('Run DCM2NII\t\t\t\t%d\n', x.ImportModules(1));
-    fprintf('Run NII2BIDS\t\t\t%d\n', x.ImportModules(2));
-    fprintf('Run ANONYMIZE\t\t\t%d\n', x.ImportModules(3));
-    fprintf('Run BIDS2LEGACY\t\t\t%d\n', x.ImportModules(4));
-    fprintf('Run Structural Module\t%d\n', x.ProcessModules(1));
-    fprintf('Run ASL Module\t\t\t%d\n', x.ProcessModules(2));
-    fprintf('Run Population Module\t%d\n', x.ProcessModules(3));
+    fprintf('Import Modules\t\t\t%s %s %s %s\n', dcm2nii, nii2bids, anonymize, bids2legacy);
+    fprintf('Process Modules\t\t\t%s %s %s\n', Structural, ASL, Population);
     fprintf('bPause\t\t\t\t\t%d\n', x.bPause);
     fprintf('iWorker\t\t\t\t\t%d\n', x.iWorker);
     fprintf('nWorkers\t\t\t\t%d\n', x.nWorkers);
--- a/ExploreASL_Initialize.m
+++ b/ExploreASL_Initialize.m
@@ -50,19 +50,19 @@ function [x] = ExploreASL_Initialize(varargin)
     
     % Check if the ExploreASL pipeline should be run or not
     if sum(x.ImportModules)>0
-        x.ImportData = 1; % Importing data
+        x.bImportData = 1; % Importing data
     else
-        x.ImportData = 0; % Importing data
+        x.bImportData = 0; % Importing data
     end
     if sum(x.ProcessModules)>0
-        x.ProcessData = 1; % Loading & processing dataset
+        x.bProcessData = 1; % Loading & processing dataset
     else
-        x.ProcessData = 0; % Only initialize ExploreASL functionality
+        x.bProcessData = 0; % Only initialize ExploreASL functionality
     end
 
     % Check if the DataParPath is a file or a directory
     SelectParFile = false; % Fallback
-    if x.ProcessData
+    if x.bProcessData
         % Checkout the "Proceed with Initialization" section
         if (isempty(x.DataParPath) || ~exist(x.DataParPath,'file'))
             SelectParFile = true; % If the DataParPath is either empty OR the file does not exist, we have to select it later on (if processing is turned on)
@@ -121,7 +121,7 @@ function [x] = ExploreASL_Initialize(varargin)
     %% Check DataParFile
     
     % Check if pipeline should be run, but there is no DataParFile
-    if x.ProcessData && (~isfield(x,'DataParPath') || ~exist(x.DataParPath,'file'))
+    if x.bProcessData && (~isfield(x,'DataParPath') || ~exist(x.DataParPath,'file'))
         x.DataParPath = input('Please insert the path to your DataParFile: ');
     end
     
@@ -140,39 +140,39 @@ function [x] = ExploreASL_Initialize(varargin)
 
     % Recheck the DataPar/sourceStructure file, which is possibly not a file or does not exist
     if ~exist(x.DataParPath,'file')
-        if x.ImportData || x.ProcessData
+        if x.bImportData || x.bProcessData
             fprintf('DataPar file does not exist, ExploreASL will only be initialized...\n');
         end
-        x.ProcessData = 0;
+        x.bProcessData = 0;
         x.ProcessModules = [0 0 0];
     else % DataPar file/folder exists
         if strcmp(x.dataParType,'dataParFile') % It is a dataParFile, so do not run the BIDS import workflow
-            if x.ProcessData==0 || x.ProcessData==2
-                x.ProcessData = 2; % Initialize & load but do not process
+            if x.bProcessData==0 || x.bProcessData==2
+                x.bProcessData = 2; % Initialize & load but do not process
             end
         end
     end
 
     % Check output
-    if x.ProcessData>0 && nargout==0
+    if x.bProcessData>0 && nargout==0
         warning('Data loading requested but no output structure defined');
         fprintf('%s\n', 'Try adding "x = " to the command to load data into the x structure');
     end
     
     % Try to catch unexpected inputs
-    if strcmp(x.dataParType,'unknown') && x.ProcessData>0 && x.ImportData==0
+    if strcmp(x.dataParType,'unknown') && x.bProcessData>0 && x.bImportData==0
         fprintf('You are trying to process a dataset, without providing a DataPar file or running the import workflow...\n');
-        x.ProcessData = 0;
+        x.bProcessData = 0;
     end
     
     
     % Give some feedback
     reportProcess = '';
-    if x.ProcessData==0,        reportProcess = 'run the initialization';
-    elseif x.ProcessData==1,    reportProcess = 'run the processing pipeline';
-    elseif x.ProcessData==2,    reportProcess = 'load the dataset';
+    if x.bProcessData==0,        reportProcess = 'run the initialization';
+    elseif x.bProcessData==1,    reportProcess = 'run the processing pipeline';
+    elseif x.bProcessData==2,    reportProcess = 'load the dataset';
     end
-    if x.ImportData==1,         reportImport = 'will run the import workflow and ';
+    if x.bImportData==1,         reportImport = 'will run the import workflow and ';
     else,                       reportImport = '';
     end
     % Print feedback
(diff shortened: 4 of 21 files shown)


MichaelStritt:
**Note**

> @HenkMutsaerts: I had this weird bug, where BIDS2LEGACY converted a bids compliant ASL file to an LEGACY format ASL4D. After that the parseM0 script is called which then says that ASL4D is not a valid BIDS file. This is why I added the try catch statement, which is certainly not the right fix, but I wanted to ask you how we should do this. (**section** `9. Parse M0`)

Jan Petr fixed this in [PR 456](https://github.com/ExploreASL/ExploreASL/pull/456).

jan-petr:
In reaction to the comment above: No need to talk about this bug - I have created this bug by mistake when improving the headers and parameter admin. I have simply added an incorrect parameter check. I have removed that in a different issue (as said above). Due to the parallel fixes in multiple branches, we can now face some errors after merging everything. So I propose the following
1. Remove the try-catch statement - this is not needed anymore (though I understand you needed it in the moment to test everything).
2. We have several branches fixing several errors and it is now possible that separate branches might not work, so testing each branch won't really help that much.
3. So we should merge all the possibly conflicting branches that are there and run a very thorough testing of everything.

MichaelStritt:
> In reaction to the comment above: No need to talk about this bug - I have created this bug by mistake when improving the headers and parameter admin. I have simply added an incorrect parameter check. I have removed that in a different issue (as said above). Due to the parallel fixes in multiple branches, we can now face some errors after merging everything. So I propose the following
> 
>     1. Remove the try-catch statement - this is not needed anymore (though I understand you needed it in the moment to test everything).
> 
>     2. We have several branches fixing several errors and it is now possible that separate branches might not work, so testing each branch won't really help that much.
> 
>     3. So we should merge all the possibly conflicting branches that are there and run a very thorough testing of everything.

Totally agree 👍

MichaelStritt:
How do you want to do this? Do we merge #456 to develop and then merge develop back to #458?

ExploreASL_Master.m
```diff
@@ -1,37 +1,54 @@
-function [x] = ExploreASL_Master(DataParPath, ProcessData, SkipPause, iWorker, nWorkers, iModules)
-%ExploreASL_Master ExploreASL pipeline master wrapper calling the individual pipeline modules
+function [x] = ExploreASL_Master(varargin)
+%ExploreASL_Master ExploreASL pipeline master wrapper calling the individual import & pipeline modules
 %
-% FORMAT: [x] = ExploreASL([DataParPath, ProcessData, SkipPause, iWorker, nWorkers, iModules])
+% FORMAT: [x] = ExploreASL([DataParPath, ImportArray, ProcessArray, SkipPause, iWorker, nWorkers])
```
jan-petr, line 4: Good that you keep the parameter names at line four - makes it easier to see what's there.

ExploreASL_Master.m
```diff
@@ -1,37 +1,54 @@
-function [x] = ExploreASL_Master(DataParPath, ProcessData, SkipPause, iWorker, nWorkers, iModules)
-%ExploreASL_Master ExploreASL pipeline master wrapper calling the individual pipeline modules
+function [x] = ExploreASL_Master(varargin)
```
jan-petr, line 1: Seeing that the input parameters changed a bit - have you checked that this change is reflected in all functions that call the master script (ExploreASL_Master or ExploreASL)? There should be at least one, right - xASL_qc_Test...

ExploreASL_Initialize.m
```diff
@@ -1,16 +1,10 @@
-function [x] = ExploreASL_Initialize(DataParPath, ProcessData, iWorker, nWorkers)
+function [x] = ExploreASL_Initialize(varargin)
```
jan-petr, line 1: Same here - have you changed the calls of ExploreASL_Initialize everywhere if they match to this input parameter change?

ExploreASL_Initialize.m
```diff
@@ -1,16 +1,10 @@
...
 %
 % INPUT:
-%   DataParPath - path to data parameter file (OPTIONAL, required when ProcessData=true, will then be prompted if omitted)
-%   ProcessData - 0 = only initialize ExploreASL functionality (e.g. path management etc, REQUIRED, will be prompted if omitted)
-%               - 1 = initialize ExploreASL functionality, load data & start processing pipeline, 
-%               - 2 = initialize ExploreASL functionality, load data but no processing
-%               - (OPTIONAL, default=prompt the user)
-%   iWorker     - allows parallelization when called externally. iWorker defines which of the parallel ExploreASL calls we are (OPTIONAL, DEFAULT=1)
-%   nWorkers    - allows parallelization when called externally. nWorkers defines how many ExploreASL calls are made in parallel (OPTIONAL, DEFAULT=1)
+%   This script can accept the same arguments as ExploreASL_Master. Check out the definitions there.
```
jan-petr, line 7: Good point!

jan-petr: BTW: the guide for good practice in comments and commit messages says "can accept" -> "accepts"

Functions/xASL_bids_BIDS2Legacy.m
```diff
@@ -250,17 +250,27 @@ function xASL_bids_BIDS2Legacy(pathStudy, bOverwrite, dataPar)
 
 %% 9. Parse M0
 ListASL4D = xASL_adm_GetFileList(pathLegacy, '^ASL4D\.nii$', 'FPListRec');
-if ~isempty(ListASL4D)
-    for iList=1:numel(ListASL4D)
-        xASL_bids_parseM0(ListASL4D{iList});
+try
```
jan-petr: No longer needed. Good that you had it here for testing, but lets remove it now.

jan-petr (changes requested):
Looks all very good. Minor comments only. And crossing fingers that it all works :)

jan-petr:
> How do you want to do this? Do we merge #456 to develop and then merge develop back to #458?

1. Merge #456 to develop. Then rebase #458 (as normally done), but run test on the rebased before merging (quick test to make sure that nothing is completely broken).
2. Merge #458 to develop and test thoroughly.

ExploreASL_ImportWorkflow.m
```diff
@@ -0,0 +1,123 @@
...
+%
+% -----------------------------------------------------------------------------------------------------------------------------------------------------
+% EXAMPLE:        n/a
+% __________________________________
+% Copyright 2015-2021 ExploreASL
+
+    %% Import Workflow
+    
+    % We expect x.DataParPath to be the sourceStructure.json file, so the root directory should be the study directory
+    [x.StudyRoot,~,~] = xASL_fileparts(x.DataParPath);
```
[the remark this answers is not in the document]
MichaelStritt: I was almost waiting for feedback on this :)
I just thought it would be a good idea style-wise to create a separate variable for this, but maybe there is already a correct one in the x-struct. `ExploreASL_ImportBIDS` needs the folder containing the sourcedata, rawdata etc. which is the folder of the sourceStructure, which is why I defined this field. We can alternatively just remove it from the x-struct or put it in one of its fields, I guess, just let me know which one.

ExploreASL_ImportWorkflow.m
```diff
@@ -0,0 +1,123 @@
...
+
+    % Go through all studies
+    ListFolders = xASL_adm_GetFileList(x.StudyRoot, '^.+$', 'FPListRec', [0 Inf], 1);
+    for iList=1:numel(ListFolders)
+        % Convert only those containing raw data
+        [thisRootFolder,thisFolderName,~] = xASL_fileparts(ListFolders{iList});
+        if strcmp(thisFolderName,'rawdata') && exist(ListFolders{iList},'dir')
+            
+            % Clean the old data
+            if isunix
```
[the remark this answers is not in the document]
MichaelStritt: Should be fixed here 593ab7e.  I think this is a relic from the changes during the v1.5.0 release.

ExploreASL_Master.m
MichaelStritt: Similar question to the one about the new field for x.StudyRoot. I used x.SkipPause, x.iWorker and x.nWorkers to make the input parsing easier. If that's a problem or you want them in sub-fields of x we can easily do this. Separate variables are just a bit problematic and make the code really ugly there.

ExploreASL_Master.m
MichaelStritt: Not yet. I wanted to wait for feedback, so we don't have to change these scripts multiple times. I can do it know if you're happy with the way the Master scripts looks right now.

ExploreASL_Initialize.m
MichaelStritt: Same answer as above :)

Functions/xASL_bids_BIDS2Legacy.m
MichaelStritt: Removed it here b1758a0.

ExploreASL_Initialize.m
MichaelStritt: Fixed 2cf728a.

ExploreASL_Initialize.m
MichaelStritt: Hopefully fixed all in c70a671.

ExploreASL_Master.m
MichaelStritt: Hopefully fixed all in 8c1269d.

HenkMutsaerts:
>> @HenkMutsaerts: I had this weird bug, where BIDS2LEGACY converted a bids compliant ASL file to an LEGACY format ASL4D. After that the parseM0 script is called which then says that ASL4D is not a valid BIDS file. This is why I added the try catch statement, which is certainly not the right fix, but I wanted to ask you how we should do this. (section 9. Parse M0)

I need to fix this, so please don't put a try-catch statement here :)

ExploreASL_ImportWorkflow.m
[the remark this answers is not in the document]
HenkMutsaerts: It's good to think this a bit through. "DataPar" contained settings about data, this is not allowed in BIDS, and we don't need this anymore as all these settings are now in the json sidecars. So all we need still is a settings par file (which can still be called DataPar for backward compatibility). 

I'm also wondering if we shouldn't restructure the fieldnaming in the x struct while we're at it.
At least we should keep environmental (current choice, potentially platform-dependent) and data parameters (always the same, belong to data) separately. Also Pieter Vandemaele had good ideas about this, and drafted this in his docs on Google Drive

HenkMutsaerts (changes requested):
Just some questions for now, perhaps you should help me a bit with all these great changes you did ;)

ExploreASL_ImportWorkflow.m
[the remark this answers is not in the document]
MichaelStritt: Sure, just tell me what you want it to look like. For the parsing it's just quite useful to have all the arguments in the x struct, but I don't care that much about the field. In python people usually have `opts` (options) for the input arguments as a real only structure. The input parser does the same, you get p.results as read only. So I think we could use something like ...
```
x.opts.SkipPause
x.opts.iWorker
x.opts.nWorkers
```
... for the semi-important arguments and then we use ...
```
x.data.StudyRoot
x.data.DataParPath
```
... and maybe a separate processing ...
```
x.proc.ImportArray
x.proc.ProcessArray
```
I just want to make progress, so let me know what you'd like so I can go on implementing.

jan-petr (approved):
Fine to merge. We should go back to the issue of x-struct that wasn't resolved and discuss the new structure properly. But that shouldn't hold merging this. 

The merging and fixing conflicts might be really tricky - especially since some functions were renamed and edited at the same time. Let me know if you want help with that - we can probably reconstruct together all the double-edits.

ExploreASL_Initialize.m
```diff
@@ -697,111 +639,242 @@
...
+    for ii=1:length(subfolders_to_add)
+        addpath(fullfile(MyPath,subfolders_to_add{ii}));
+    end
+    
+end
 
 %% -----------------------------------------------------------------------
-%% 3) Print warnings
-fprintf('\n%s\n\n','---------------------------------------------');
+%% Define input parser
```
HenkMutsaerts, line 776: What is the advantage now of this, if the code only becomes longer with this way of input parsing? ;)
Also, it makes sense to read the input parsing at the first part of the code right. Though it's nice that it's separated here below in a subfunction, it becomes more difficult to debug.

ExploreASL_Initialize.m
```diff
@@ -697,111 +639,242 @@
...
+function p = inputParsing(varargin)
 
-field_symbol = {'subject_regexp'};
+    % Initialize input parser
+    p = inputParser;
+    
+    % Define valid input variables
+    validDataParPath = @(variable) ischar(variable) || isempty(variable);
+    validImportArray = @(variable) ischar(variable) || isempty(variable) || isnumeric(variable);
```
HenkMutsaerts, line 784: Why do all these numerical/logical/Boolean input values allow `ischar(var)`?

ExploreASL_Initialize.m
```diff
@@ -697,111 +639,242 @@
...
-    for iField=1:length(field_symbol)
-        if ~isfield(x.D,field_symbol{iField})
-            warning(['x.D.' field_symbol{iField} ' was not defined in DATA_PAR.m!'])
-        end
+%% -----------------------------------------------------------------------
+%% Convert parsed input
+function parameters = convertParsedInput(parameters)
+
+    % Check if inputs are empty or chars
+    if isempty(parameters.DataParPath),     parameters.DataParPath = '';                                    end
```
HenkMutsaerts, line 816: Is it confusing if the fieldname is `parameters` here and `x` somewhere else?

HenkMutsaerts, line 816: `,` should be `;` though `,` is perhaps also possible, `;` makes more sense right?

ExploreASL_Initialize.m
```diff
@@ -697,111 +639,242 @@
...
     end
+
+
 end
 
-if ~isempty(regexp(x.subject_regexp, '^(\^|)\.\*(\$|)$'))
-    warning('Subject regexp not specific! Check that no wrong folders are included as subjects');
+%% -----------------------------------------------------------------------
+%% Store parsed input
+function x = storeParsedInput(parameters)
```
HenkMutsaerts, line 846: Jan and I got the habit to prefix subfunctions by their main function name, so this would be `ExploreASL_Initialize_storeParsedInput` in this case. This makes it nicely readable since this immediately makes it clear that this is a subfunction. If you agree I would apply this for all subfunctions in this function as well (to keep style same throughout)

ExploreASL_Initialize.m
```diff
@@ -697,111 +639,242 @@
...
-if ~isfield(x,'M0')
-%     warning('M0 option missing!');
-else
-    fprintf('\n%s\n',['M0 option selected is "' num2str(x.M0) '"']);
 end
 
-fprintf('%s\n',['x.DELETETEMP = ' num2str(x.DELETETEMP) ' (delete temporary files)']);
-fprintf('%s\n',['x.Quality    = ' num2str(x.Quality) ' (0 = fast try-out; 1 = normal high quality)']);
+%% -----------------------------------------------------------------------
+%% Add ExploreASL Directory
```
HenkMutsaerts, line 747: Very nice that you restructured all this in subfunctions!

ExploreASL_Initialize.m
```diff
@@ -697,111 +639,242 @@
...
+function addExploreASLDirectory(MyPath)
 
+    % Define paths (should be equal when loading data or only initializing)
+    addpath(MyPath); % ExploreASL
+    subfolders_to_add = {'Functions', 'mex', 'Modules', 'Testing',...
```
HenkMutsaerts, line 752: perhaps good to order these alphabetically, easier to check

ExploreASL_Initialize.m
```diff
@@ -697,111 +639,242 @@
...
 
-fprintf('%s\n',['ASL sessions: ' num2str(x.nSessions)]);
+    fprintf('%s\n',['ASL sessions: ' num2str(x.nSessions)]);
 
-fprintf('\n%s\n','Ancillary data, sets:');
-if isfield(x.S,'SetsID')
-        fprintf('%s\n',[num2str(size(x.S.SetsID,2)) ' sets are defined for ' num2str(size(x.S.SetsID,1)) ' "SubjectsSessions":']);
+    fprintf('\n%s','Ancillary data, sets: ');
+    if isfield(x.S,'SetsID')
+            fprintf('%s\n',[num2str(size(x.S.SetsID,2)) ' sets are defined for ' num2str(size(x.S.SetsID,1)) ' "SubjectsSessions"']);
```
HenkMutsaerts, line 681: why did you delete the `:` here?

ExploreASL_Initialize.m
```diff
@@ -697,111 +639,242 @@
...
+        fprintf('%s\n',['Longitudinal timePoint ' num2str(iT) ' = ' num2str(x.nTimePointTotalSubjects(iT)) ' scans - ' num2str(x.nTimePointExcluded(iT)) ' exclusions = ' num2str(x.nTimePointSubjects(iT)) ' scans']);
+    end
 
-fprintf('%s\n',['ASL sessions: ' num2str(x.nSessions)]);
+    fprintf('%s\n',['ASL sessions: ' num2str(x.nSessions)]);
 
-fprintf('\n%s\n','Ancillary data, sets:');
-if isfield(x.S,'SetsID')
-        fprintf('%s\n',[num2str(size(x.S.SetsID,2)) ' sets are defined for ' num2str(size(x.S.SetsID,1)) ' "SubjectsSessions":']);
+    fprintf('\n%s','Ancillary data, sets: ');
```
HenkMutsaerts, line 679: Wasn't the `\n` there for a reason?

ExploreASL_Initialize.m
```diff
@@ -697,111 +639,242 @@
...
-%% 1) Set default settings if not defined
-if ~isfield(x,'Quality') || (x.Quality~=0 && x.Quality~=1)
-    x.Quality = 1;
-    fprintf('%s\n', 'Default Quality=1 used (optimal quality)');
-end
-if ~isfield(x,'DELETETEMP') || (x.DELETETEMP~=0 && x.DELETETEMP~=1)
-    x.DELETETEMP = 1;
-%     fprintf('%s\n','Default x.DELETETEMP=1 used (delete files temporarily used for processing)');
-end
+    %xASL_init_PrintCheckSettings Check whether pre-defined settings existed in DATA_PAR.m
```
HenkMutsaerts, line 644: Personal preference probably, but I don't see the benefit of starting a (sub-)function with a single indent, less will fit on the screen ;)

ExploreASL_Initialize.m
```diff
@@ -27,287 +21,249 @@
...
-        addpath(fullfile(x.MyPath,subfolders_to_add{ii}));
+    % Check if pipeline should be run, but there is no DataParFile
+    if x.ProcessData && (~isfield(x,'DataParPath') || ~exist(x.DataParPath,'file'))
+        x.DataParPath = input('Please insert the path to your DataParFile: ');
     end
-end
 
+    % Recheck the DataParFile, which is possibly not a file or does not exist
+    if ~exist(x.DataParPath,'file') && ~exist(x.DataParPath,'dir') % file = without BIDS Import, dir = BIDS import 
+        fprintf('DataPar file does not exist, ExploreASL will only be initialized...\n');
```
HenkMutsaerts, line 151: This will not be printed if you don't specify `DataParPath` at all right?
It should still be possible to initialize ExploreASL without getting this output, this output should only there with an invalid `DataParPath` indeed.

ExploreASL_Initialize.m
```diff
@@ -27,287 +21,249 @@
...
+    MasterScriptPath = fullfile(x.MyPath, 'ExploreASL_Master.m');
+
+    % Select the ExploreASL folder manually, if the script is not run in deployed mode
+    if ~isdeployed
+        if ~exist(MasterScriptPath,'file')
+            if bUseGUI
+                if ismac % somehow macOS doesnt show the title of the question dialogue
+                    fprintf('Select folder where ExploreASL is installed\n');
+                end
+                pathstr = uigetdir(CurrCD, 'Select folder where ExploreASL is installed');
```
HenkMutsaerts, line 97: We decided to get rid of the GUI stuff here right, then I would also always use the `input` here, irregardless of OS.

ExploreASL_ImportWorkflow.m
```diff
@@ -0,0 +1,115 @@
...
+    end
+    
+    % Reset the import parameter (for the second initialization including the loading of the dataset)
+    x.ImportData = 0;
+    x.ImportArray = [0 0 0 0];
+    
+end
+
+%% xASL_Import_BIDS2LEGACY
+function [x] = xASL_Import_BIDS2LEGACY(x)
```
HenkMutsaerts, line 60: Are you specifying some things that were already specified inside `xASL_bids_BIDS2Legacy`, like the default DataPar fields? I would specify this only once, to facilitate debugging.

HenkMutsaerts (changes requested):
Very very impressed, supernice work!

This of course triggers many new questions from me ;) Hope you can bear with them!


* The default behavior of `ExploreASL_Initialize` is just to initialize ExploreASL, so it should not print text `“Data Par file does not exist, ExploreASL will only be initialized…”`
Same for the statement `“ExploreASL will not run the import workflow …”` Rather it should say something like  `“Initializing ExploreASL…”` This is because `ExploreASL_Initialize` will be called from inside the pipeline or test scripts at multiple locations. This feedback is however nice for the main ExploreASL function, which will be called as full pipeline start.

* Throughout the pipeline where `ExploreASL_Master;` is now used for initialization only (mainly testing/QC functions), I would replace this by `ExploreASL_Initialize`. `ExploreASL_Initialize` should then by default initialize only, and not give any feedback about missing dataPar etc, whereas `ExploreASL_Master` can give this feedback (that's more the start of the fill pipeline).
* `ExploreASL` seems to become redundant, probably easier to remove this and replace by `ExploreASL_Master` (such that we only keep one `ExploreASL` function).
* I would rename `ImportArray` and `ProcessArray` to `ImportModules` and `ProcessModules`. That's more self-explanatory, from the rest of the explanation it's clear that these are arrays, but it should be clear that these arrays specify which modules to run.

When you’re done with all this, I should also check the new output you print on the screen/log. Again, very very nice work! Let me know if you need any help!

Development/BIDS/ZeroingOut.m
```diff
@@ -1,8 +1,8 @@
 %% Zeroing out imaging data for BIDS example datasets
 
-ExploreASL_Master('',0);
+ExploreASL_Master;
 
-Rdir = 'C:\Users\henkj\Google Drive\WorkGeneral\StudiesProjects\BIDS_ASL\TestDataSet_ImportBIDS\raw'
+Rdir = 'C:\Users\henkj\Google Drive\WorkGeneral\StudiesProjects\BIDS_ASL\TestDataSet_ImportBIDS\raw';
```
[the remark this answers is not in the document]
MichaelStritt: Check out this issue right here:  #451. Jan proposes to get rid of this script.

ExploreASL_ImportConfig.m
```diff
@@ -34,7 +34,7 @@
 %end
 
 % ExploreASL does not need to be initialized here
```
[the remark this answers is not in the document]
MichaelStritt: Ah sorry, I think we used to initialize at this part of the pipeline. Now we always initialize before and we shouldn't recursively call Initialize within Initialize. I can remove that comment ;)

ExploreASL_Initialize.m
```diff
@@ -27,287 +21,249 @@
...
 % __________________________________
-% Copyright 2015-2020 ExploreASL
+% Copyright 2015-2021 ExploreASL
 
 
+    %% Admin
 
-%% Admin
+    % Define input parser
+    p = inputParsing(varargin{:});
```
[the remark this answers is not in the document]
MichaelStritt: There's a comment in the header to check out ExploreASL_Master regarding the input arguments, since ExploreASL_Master & ExploreASL_Initialize share the same input arguments now. I can specify it in both scripts if you prefer that, but I actually prefer declaring them once, otherwise we always have to update both descriptions if there is a change.

ExploreASL_Master.m
```diff
@@ -1,37 +1,54 @@
...
+%                - NII2BIDS = Run the NIFTI to BIDS conversion (BOOLEAN, DEFAULT = 0)
+%                - ANONYMIZE = Run the defacing and full anonymization (BOOLEAN, DEFAULT = 0)
+%                - BIDS2LEGACY = Run the BIDS to LEGACY conversion (BOOLEAN, DEFAULT = 0)
+%
+%   ProcessArray - [STRUCTURAL, ASL, POPULATION] (OPTIONAL, BOOLEAN ARRAY) 
+%                - STRUCTURAL = Run the Structural Module (BOOLEAN, DEFAULT = 0)
+%                - ASL = Run the ASL Module (BOOLEAN, DEFAULT = 0)
+%                - POPULATION = Run the Population Module (BOOLEAN, DEFAULT = 0)
+%
+%   SkipPause    - TRUE = Pause workflow before ExploreASL pipeline (OPTIONAL, DEFAULT = FALSE)
```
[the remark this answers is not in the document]
MichaelStritt: So the default of SkipPause is False, at least for non-compiled mode. Ahhh sorry, It should be FALSE = ...

ExploreASL_Master.m
```diff
@@ -1,37 +1,54 @@
...
 % xASL_Module_Structure  - processes structural data, i.e. high-resolution
 %                          T1w and FLAIR scans, type help xASL_module_Structural for more information
 % xASL_Module_ASL        - processes ASL data, i.e. ASL scans and M0, type help xASL_module_ASL for more information 
 % xASL_Module_Population - processes data on population basis, mostly QC, but also template/parametric maps creation, etc. Type help xASL_module_Population for more information 
 % 
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
+%
+% EXAMPLES:
+%
+% [x] = ExploreASL_Master('MY_PATH',1,1,1,1,1);
```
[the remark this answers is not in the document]
MichaelStritt: 👍

ExploreASL_Master.m
```diff
@@ -1,37 +1,54 @@
...
-%               - 2 = initialize ExploreASL functionality, load data but no processing
-%               - (OPTIONAL, default=prompt the user)
-%   SkipPause   - true if calling pipeline without pausing for questioning for user prompting (OPTIONAL, DEFAULT=false)
-%   iWorker     - allows parallelization when called externally. iWorker defines which of the parallel ExploreASL calls we are (OPTIONAL, DEFAULT=1)
-%   nWorkers    - allows parallelization when called externally. nWorkers defines how many ExploreASL calls are made in parallel (OPTIONAL, DEFAULT=1)
-%   iModules    - scalar or vector allowing to select modules to run, e.g.
-%                  [1] = only structural, [2] = only ASL, [3] = only
-%                  Population. [1 2] = Structural & ASL module, etc.
-%                  OPTIONAL, DEFAULT = [1 2 3]
+%   DataParPath  - Path to data parameter file (OPTIONAL, DEFAULT = [])
```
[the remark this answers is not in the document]
MichaelStritt: 👍

ExploreASL_Master.m
```diff
@@ -1,37 +1,54 @@
...
 % 
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
+%
+% EXAMPLES:
+%
+% [x] = ExploreASL_Master('MY_PATH',1,1,1,1,1);
+% [x] = ExploreASL_Master('MY_PATH',[1 1 1 1],[1 1 1],1,1,1);
+% [x] = ExploreASL_Master('MY_PATH','[1 1 1 1]','[1 1 1]','1','1','1');
+%
+% OUTDATED EXAMPLES:
```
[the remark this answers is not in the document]
MichaelStritt: Yep, I just thought I'd keep them to recreate similar ones there, but you're right. Will remove them now.

ExploreASL_Master.m
```diff
@@ -49,61 +66,27 @@
...
-        if ischar(iModules)
-            iModules = str2num(iModules);
-        end
-    else
-        iModules = [1 2 3]; % by default we run all default modules (1) Structural, 2) ASL, 3) Population)
-    end
+    % Initialization
+    x = ExploreASL_Initialize(varargin{:});
     
+    % Import
```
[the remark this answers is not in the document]
MichaelStritt: Good idea.

ExploreASL_Master.m
```diff
@@ -49,61 +66,27 @@
...
+    end
     
     if x.ProcessData==0 || x.ProcessData==2
         if x.ProcessData==2 && nargout==0
             warning('Data loading requested but no output structure defined');
             fprintf('%s\n', 'Try adding "x = " to the command to load data into the x structure');
         end
         return; % skip processing
-    elseif ~isdeployed && ~SkipPause % if this exists, we skip the break here
+    elseif ~isdeployed && ~x.SkipPause % if this exists, we skip the break here
```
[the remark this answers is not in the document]
MichaelStritt: 👍

Functions/xASL_bids_BIDS2Legacy.m
```diff
@@ -72,15 +72,15 @@ function xASL_bids_BIDS2Legacy(pathStudy, bOverwrite, dataPar)
 end
 
 % Fills in important information in the dataPar if missing
-if ~isfield(dataPar,'x.subject_regexp')
+if ~isfield(dataPar.x,'subject_regexp')
```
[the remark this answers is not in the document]
MichaelStritt: I think I know what you mean. We could make super sure that the script doesn't break here using:
```matlab
if ~isfield(dataPar,'x')
    dataPar.x = struct;
end
if ~isfield(dataPar.x,'subject_regexp')
    ...
end
```
Other than that I don't know what you're talking about... Didn't you write the part with the dataPar.x ?

Functions/xASL_bids_BIDS2Legacy.m
```diff
@@ -260,7 +260,13 @@ function xASL_bids_BIDS2Legacy(pathStudy, bOverwrite, dataPar)
 end
 
 %% 10. Create DataPar.json
-spm_jsonwrite(fullfile(pathLegacy, 'DataPar.json'), dataPar);
+
+% Overwrite DataParPath in x structure
+fprintf('Overwriting x.DataParPath...\n');
+dataPar.x.DataParPath = fullfile(pathLegacy, 'DataPar.json');
```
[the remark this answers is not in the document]
MichaelStritt: In my head I'd always reinitialize at that point and we basically did exactly the same before, right? We always had `ExploreASL(DataParPath,...)` and if you'd call the pipeline like this, inside the script you'd have an absolute path as well.

Testing/UnitTests/xASL_ut_UnitTest_function_ExploreASL_Master.m
```diff
@@ -0,0 +1,269 @@
...
+% Define one or multiple test conditions here
+testCondition = true; % Fallback
+if ~exist(fullfile(testPatientDestination,'analysis'),'dir'),                   testCondition = false;          end
+if ~exist(fullfile(testPatientDestination,'analysis','Sub1'),'dir'),            testCondition = false;          end
+if ~exist(fullfile(testPatientDestination,'analysis','Sub1','ASL_1'),'dir'),    testCondition = false;          end
+if ~exist(fullfile(testPatientDestination,'analysis','Sub1','T1w_1'),'dir'),    testCondition = false;          end
+if ~exist(fullfile(testPatientDestination,'analysis','Sub1','ASL_1','ASL4D.json'),'file'),      testCondition = false;          end
+if ~exist(fullfile(testPatientDestination,'analysis','Sub1','ASL_1','ASL4D.nii'),'file'),       testCondition = false;          end
+if ~exist(fullfile(testPatientDestination,'analysis','Sub1','ASL_1','M0.json'),'file'),         testCondition = false;          end
+if ~exist(fullfile(testPatientDestination,'analysis','Sub1','ASL_1','M0.nii'),'file'),          testCondition = false;          end
```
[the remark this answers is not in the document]
MichaelStritt: I try to use as few ExploreASL functions within the UnitTesting as possible. Otherwise you test functions with functions you need to test. Basically nothing wrong with that, if you test xASL_exist first, but I try to get rid of these scenarios whenever possible.
In addition, I don't want it to find both .nii & .nii.gz here. We want the test to be strict. We always want the same results, not sometimes .nii and sometimes .nii.gz, right?

ExploreASL_ImportWorkflow.m
```diff
@@ -0,0 +1,105 @@
+function [x] = ExploreASL_ImportWorkflow(x)
```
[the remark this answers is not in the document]
MichaelStritt: No, sounds good to me. I'll rename it.

ExploreASL_Initialize.m
MichaelStritt: Length isn't everything ;)

I think the benefit is that it's modularized now. Of course we could write it in a more simplified way like before, but the use of varargin and the input parser are the most high level solutions described in here: https://docs.google.com/document/d/1M9GTXbRe1VngH9M3p6jyVABqtFgnSYDgpL13MnCelHQ/edit?usp=sharing 

And instead of multiple if-else-statements, you just change the default value or the limiting function handle instead. That seems a lot more sophisticated to me. Why should debugging be harder now?

ExploreASL_Initialize.m
MichaelStritt: The compiled version give us char arrays, don't you remember? :)

ExploreASL_Initialize.m
MichaelStritt: Hm... I see the x structure as our main structure for everything. So file names, data fields, etc. Here we just parse parameters of ExploreASL_Master/Initialize, so I thought parameters is a good variable.

MichaelStritt: I always use `,`. Saw both options here https://stackoverflow.com/questions/27561881/one-liner-for-if-then Seems like a style-wise choice.

ExploreASL_Initialize.m
MichaelStritt: Oh... here's a mistake in the comment anyway. I'll quickly check it later this afternoon.

ExploreASL_Initialize.m
MichaelStritt: Sure, makes my life easier :)
I thought you only wanted to get rid of the questdlg boxes and not of the uigetdir ones.

ExploreASL_ImportWorkflow.m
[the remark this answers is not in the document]
MichaelStritt: Okay, I added everything to the dev meeting notes. We can discuss this there and update it after the #290 merge in a separate issue.

ExploreASL_ImportConfig.m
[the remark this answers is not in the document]
MichaelStritt: Fixed in bb1a0d7.

ExploreASL_Master.m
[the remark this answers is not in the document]
MichaelStritt: Fixed in 329940e.

ExploreASL_Master.m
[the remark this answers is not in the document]
MichaelStritt: Fixed in 579fdb2.

ExploreASL_Master.m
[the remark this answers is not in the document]
MichaelStritt: Fixed in 579fdb2.

ExploreASL_Master.m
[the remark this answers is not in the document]
MichaelStritt: Fixed in 1fd818d.

ExploreASL_Master.m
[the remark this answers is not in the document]
MichaelStritt: Did it similar but a bit different in 579fdb2.

ExploreASL_Master.m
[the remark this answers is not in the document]
MichaelStritt: Changed in 1aa1ce0.

ExploreASL_ImportWorkflow.m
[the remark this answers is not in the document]
MichaelStritt: Done in 6fcf5db.

ExploreASL_Initialize.m
```diff
@@ -27,287 +21,249 @@
 % xASL_init_FileSystem             - dirty initialization of common filenames used throughout the pipeline
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
 % EXAMPLE for GUI: ExploreASL_Initialize
-% EXAMPLE for calling externally: ExploreASL_Initialize('//MyDisk/MyStudy/DataPar.m', true, 1, 1);
-% EXAMPLE for debugging/initialization only: [x] = ExploreASL_Initialize('',0);
+% EXAMPLE for calling externally: ExploreASL_Initialize('//MyDisk/MyStudy/DataPar.m');
```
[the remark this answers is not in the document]
MichaelStritt: Done in 6fcf5db.

ExploreASL_Initialize.m
```diff
@@ -27,287 +21,249 @@
...
+% EXAMPLE for debugging/initialization only: [x] = ExploreASL_Initialize('',0,0);
```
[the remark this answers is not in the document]
MichaelStritt: Fixed it in 6fcf5db. If the DataParPath is empty and processing is off, then it won't pause or ask for anything.

ExploreASL_Initialize.m
MichaelStritt: Done in 51f3f3b.

ExploreASL_Initialize.m
MichaelStritt: Thanks, maybe we can even move some to external scripts, to improve the ExploreASL_Initialize readability.

ExploreASL_Initialize.m
MichaelStritt: Fixed notation to camel case and improved the sorting in 7762591.

ExploreASL_Initialize.m
MichaelStritt: I changed the overall look a little bit. Check out the following example for loading a dataset. If you don't like it, we can change it back.

```
DataParPath		...pment\MATLAB\m.stritt\TestDataSet\analysis\DataParameters_LowQ.json
Run DCM2NII		0
Run NII2BIDS		0
Run ANONYMIZE		0
Run BIDS2LEGACY		0
Run Structural Module	0
Run ASL Module		0
Run Population Module	0
SkipPause		1
iWorker			1
nWorkers		1
ExploreASL v1.5.1 initialized ... 
Automatically defining sessions...
==================================== Additional Settings =====================================
1 scans - 0 exclusions, resulting in 1 scans of: 
Longitudinal timePoint 1 = 1 scans - 0 exclusions = 1 scans
ASL sessions: 1

Ancillary data, sets: 3 sets are defined for 1 "SubjectsSessions"
Set 1 = "LongitudinalTimePoint" options "TimePoint_1", codes for paired data
Set 2 = "SubjectNList" options "SubjectNList", codes for paired data
Set 3 = "Site" options "SingleSite", codes for two-sample data

x.D.ROOT		M:\SoftwareDevelopment\MATLAB\m.stritt\TestDataSet\analysis
x.DELETETEMP		1 (delete temporary files)
x.Quality		0 (0 = fast try-out; 1 = normal high quality)

```

ExploreASL_Initialize.m
MichaelStritt: Check out the comment above :)

ExploreASL_Initialize.m
MichaelStritt: Really? So in python you obviously have to do this, since this is how you write functions there. But I know this from C++ and JavaScript as well. There's even a wiki about this: https://en.wikipedia.org/wiki/Indentation_style I really like this style right here: Variant: 1TBS (OTBS), with this example:

```JavaScript
void checknegative(x) {
    if (x < 0) {
        puts("Negative");
    } else {
        nonnegative(x);
    }
}
```

ExploreASL_Initialize.m
MichaelStritt: Should be fixed in 5d88126. I also removed the quite verbose user feedback on default. It's really helpful during debugging, because there are so many different use cases.

MichaelStritt:
> The default behavior of `ExploreASL_Initialize` is just to initialize ExploreASL, so it should not print text `“Data Par file does not exist, ExploreASL will only be initialized…”`
> Same for the statement `“ExploreASL will not run the import workflow …”` Rather it should say something like  `“Initializing ExploreASL…”` This is because `ExploreASL_Initialize` will be called from inside the pipeline or test scripts at multiple locations. This feedback is however nice for the main ExploreASL function, which will be called as full pipeline start.

I know what you mean. The problem was that there are a lot of different use-cases. What happens if we call ExploreASL with a sourceStructure file, what if it's a DataPar file, what happens if the user only wants to load the data, what happens if the DataPar is invalid but the user set processing to true, ... etc. I've put the quite verbose statements into an if section, so that we can turn it on for debugging purposes. Check out 5d88126.

> Throughout the pipeline where `ExploreASL_Master;` is now used for initialization only (mainly testing/QC functions), I would replace this by `ExploreASL_Initialize`. `ExploreASL_Initialize` should then by default initialize only, and not give any feedback about missing dataPar etc, whereas `ExploreASL_Master` can give this feedback (that's more the start of the fill pipeline).

Definitely agree. Changed this in 33865d3.

> `ExploreASL` seems to become redundant, probably easier to remove this and replace by `ExploreASL_Master` (such that we only keep one `ExploreASL` function).

Sure, also within this PR?

> I would rename `ImportArray` and `ProcessArray` to `ImportModules` and `ProcessModules`. That's more self-explanatory, from the rest of the explanation it's clear that these are arrays, but it should be clear that these arrays specify which modules to run.

Good idea. I changed this in 14e9d8b. Hope I got all the different places in the code, but we'll have to extensively test anyway.

ExploreASL_Initialize.m
MichaelStritt: Damn... I know what you want to do, but it's kind of hard to code. So with the changes in c23eac5, there is no warning that the file does not exist, if you only initialize using `ExploreASL_Master();` or `ExploreASL_Initialize();`, since I check if either the import or the processing is turned on. 
If we turn on processing, then we get the CLI request and if this file doesn't exist either, the warning appears:
```
>> [x] = ExploreASL_Master('I dont exist', 0, 1, 1);
Please insert the path to your DataParFile: 'I dont exist either'
DataPar file does not exist, ExploreASL will only be initialized...
```
If the inserted file exists, the processing will start as usual:
```
>> [x] = ExploreASL_Master('I dont exist', 0, 1, 1);
Please insert the path to your DataParFile: '.\analysis\DataParameters_LowQ.json'
```
This doesn't work for the import workflow right now, meaning we don't ask a second time there:
```
>> [x] = ExploreASL_Master('I dont exist', 1, 0, 1);
DataPar file does not exist, ExploreASL will only be initialized...
```
And if processing and import are turned off, but the file doesn't exist, there is no warning (fallback solution for data loading?):
```
[x] = ExploreASL_Master('I dont exist', 0, 0, 1);
```
But there's the question if we need it there...

ExploreASL_ImportWorkflow.m
MichaelStritt: I'm not sure what you mean. I used the code from `xASL_test_BIDSFlavorsFull` and tried to make it work within this workflow. This needed a few adaptions, which is why we have this subfunction. If there's something specific within the function that you would do differently, please let me know.

ExploreASL_Initialize.m
MichaelStritt: Removed it: e68fa0d.

HenkMutsaerts:
https://github.com/ExploreASL/ExploreASL/commit/5d8812659ac96f95b8199baf8b746068a6319fd8
-> I responded here!

>> ExploreASL seems to become redundant, probably easier to remove this and replace by ExploreASL_Master (such that we only keep one ExploreASL function).

> Sure, also within this PR?

Haha, well you choose ;) To the user it is probably the same effort, in both functions they switch the usage. The use of `ExploreASL` as addition to `ExploreASL_Master` becomes a bit irrelevant because of this PR, that's why.

ExploreASL_Initialize.m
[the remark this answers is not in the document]
HenkMutsaerts: Yes that looked good

Functions/xASL_bids_BIDS2Legacy.m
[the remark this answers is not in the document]
HenkMutsaerts: Haha no then perhaps Jan wrote it?
I'm fine here, I just mean that when somewhere (either reading the dataPar.json or reading an input struct) only have `x.*` rather than `dataPar.x.*`, that it manages this correctly.

E.g., if the `dataPar.json` normally is loaded into `x`, and has `x.subject_regexp`, and now suddenly has `x.dataPar.x.subject_regexp` or `x.x.subject_regexp`

But perhaps this is unnecessary precaution ;)

Functions/xASL_bids_BIDS2Legacy.m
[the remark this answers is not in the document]
HenkMutsaerts: Yes, but we didn't save this path (well we probably did in x.mat but not intentionally :)
Most users like me also just entered

`ExploreASL` (or `ExploreASL_Master`)

which would pop up a GUI dialogue or CLI input dialogue, asking for this DataPar path (but that's basically the same)

ExploreASL_Initialize.m
HenkMutsaerts: That's true. I just thought of debugging somewhere in the script, and usually a script then more or less has a top-down order, where you see at the top which parameters are initialized where and why. So it is faster/easier to see how this works. Now, you have to search for the input parsing function, realize that this is a subfunction, search for the subfunction, and check what's going on there. 

It's also nice to have in 1 screen the function call, header, input parsing, and admin, rather than having the input parsing at the end of the script. Perhaps just intuitively weird. But practice probably will teach me otherwise ;)

ExploreASL_Initialize.m
HenkMutsaerts: True, but you turn `parameters` into `x` in a later stage. Unless fields need to co-exist.

ExploreASL_Initialize.m
HenkMutsaerts: Very nice! I would then only make it a bit more compact, by writing all ImportModules on 1 line, and all ProcessModules on 1 line.

Very cool!

ExploreASL_Initialize.m
HenkMutsaerts: Yes, I like this as well in small functions. In longer functions, you don't really use this first indent visually anymore, so it just takes up space. Anyway, this seems indeed a good style to use!

ExploreASL_Initialize.m
HenkMutsaerts: I like this example a lot, but do I now understand correctly that the first argument of `ExploreASL_Master` (i.e. the dataParFile) can be used as original dataPar file (for processing) as well as import-source file (for importing)?
That is perhaps asking for trouble... Unless we always use the "import-source file" for importing and then convert it later to an "original dataPar" file. 

(I know Jan uses a dataPar file for doing the import, and that this can be fine, but we are moving towards having the data parameters in JSON sidecars of the NIfTIs, and the study processing settings in the original dataPar (to make it confusing ;) so perhaps we should at least change the name of the original dataPar to processingPar or something :)

ExploreASL_ImportWorkflow.m
HenkMutsaerts: OK let me have a closer look (or please remind me ;)

ExploreASL_Master.m
```diff
@@ -17,7 +17,7 @@
 %                - ASL = Run the ASL Module (BOOLEAN, DEFAULT = 0)
 %                - POPULATION = Run the Population Module (BOOLEAN, DEFAULT = 0)
 %
-%   SkipPause    - TRUE = Pause workflow before ExploreASL pipeline (OPTIONAL, DEFAULT = FALSE)
+%   SkipPause    - FALSE = Pause workflow before ExploreASL pipeline (OPTIONAL, DEFAULT = FALSE)
```
HenkMutsaerts: Haha, this is my fault. I guess the name of a Boolean should always explain a positive thing.
So this Boolean should be called `bPause`. I called this Boolean `bSkipPause` as you by default want to pause (from the perspective that the user starts ExploreASL).

Let's rename this parameter to `bPause`, that is defaulted to FALSE (no pause). 
But in some situations we might want to default to bPause = TRUE, to give the user the chance to look at the screen settings before continuing. Or shall we just remove this parameter at all, and instruct the "advanced user" (the one that won't use Maurice's GUI) to start running ExploreASL using the "load data only" option (as a sort of dummy start of ExploreASL)?

HenkMutsaerts (changes requested):

MichaelStritt:
> * The first result of `ExploreASL_Initialize` (i.e. what it will always do) is the initialization.
> * I think it's always good to print what you're doing, my comment was just about printing what ExploreASL is NOT doing (e.g., "initializing ExploreASL but not importing BIDS") as this NOT statement can confuse if you're just initializing ExploreASL for some tests and suddenly you get a message about BIDS :)
> * So my solution would be: when running `ExploreASL_Initialize`, always print `Initializing ExploreASL`, and likewise, `run the processing pipeline` and `run the import workflow` I would always print if this is the case.
> * So it seems that I would set `debugFeedback = 1` and disable only the line of `x.ImportData==0` (as this states something we do NOT do).

Sounds good 👍  821f19f should do that now.

ExploreASL_Master.m
MichaelStritt: I'm fine with both. But using false on default already sounds like a really good idea ;) Should be done in 3c2529f.

Functions/xASL_bids_BIDS2Legacy.m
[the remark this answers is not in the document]
MichaelStritt: Ahhh! I know what you mean and I think I understand why you think that this should be possible. Earlier we read json files as text files and we built the struct ourselves. Now that we use `spm_jsonread` whenever possible, we get code like this:
```matlab
json = spm_jsonread('file');
```
And if we have a structure like this within the json file...
```json
{"x": [{
	"name": "ExampleDataSet",
	}]}
```
... we always get a variable json with a field called x with a sub-field called name etc.

If you dont want this, we would have to change our json files to something like this:
```json
{
"name": "ExampleDataSet",
"subject_regexp": "...",
...
}
```

... but I don't think that's worth the effort just for code that's a bit shorter ;)

Functions/xASL_bids_BIDS2Legacy.m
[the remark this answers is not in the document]
MichaelStritt: Hm... okay... Could this be problematic?
Let's try to get the use-cases:
1. A user runs the full pipeline: `[x] = ExploreASL_Master('sourceStructure.json',1,1);`. Here we need to find the DataParFile that was created during BIDS2LEGACY definitely.
2. A user runs the the conventional pipeline: `[x] = ExploreASL_Master('dataParFile.json',0,1);`. In the end the x-struct would contain a field with information about the absolute path of the `dataParFile.json`. If that's a problem we can remove it?
3. A user runs the conventional pipeline on a dataset: `[x] = ExploreASL_Master('dataParFile.json',0,1);`. Then he/she moves the dataset to another drive. He then does a rerun of the ASL module: `[x] = ExploreASL_Master('dataParFile.json',0,[0 1 0]);`. Everything's fine?
Sorry, I really don't get what would be the major problem here... :)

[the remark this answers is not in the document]
MichaelStritt: Hm... But there's probably a better solution anyway :)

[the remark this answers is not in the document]
MichaelStritt: Sorry, long story short, we can remove it. We just need the path for the following pipeline steps, but we export it using an output parameter anyway. I haven't tested it extensively yet, but I think this should work: cbf17dd.

ExploreASL_Initialize.m
MichaelStritt: Sure, if you prefer that. Maybe like this? d358bb6

```
>> [x] = ExploreASL_Master('.\analysis\DataParameters_LowQ.json', 1, 1);
ExploreASL will run the import workflow and will run the processing pipeline...


                    ## |                                                                      
                    ## |                                                                      
                    ##/  

DataParPath		...pment\MATLAB\m.stritt\TestDataSet\analysis\DataParameters_LowQ.json
Import Modules		DCM2NII NII2BIDS ANONYMIZE BIDS2LEGACY
Process Modules		Structural ASL Population
bPause			0
iWorker			1
nWorkers		1
ExploreASL v1.5.1 initialized ... 
```

```
>> [x] = ExploreASL_Master('.\analysis\DataParameters_LowQ.json', 0, 0);
ExploreASL will load the dataset...


                    ## |                                                                      
                    ## |                                                                      
                    ##/  

DataParPath		...pment\MATLAB\m.stritt\TestDataSet\analysis\DataParameters_LowQ.json
Import Modules			   
Process Modules			  
bPause			0
iWorker			1
nWorkers		1
ExploreASL v1.5.1 initialized ... 
Automatically defining sessions...
==================================== Additional Settings =====================================
1 scans - 0 exclusions, resulting in 1 scans of: 
Longitudinal timePoint 1 = 1 scans - 0 exclusions = 1 scans
ASL sessions: 1

Ancillary data, sets: 3 sets are defined for 1 "SubjectsSessions"
Set 1 = "LongitudinalTimePoint" options "TimePoint_1", codes for paired data
Set 2 = "SubjectNList" options "SubjectNList", codes for paired data
Set 3 = "Site" options "SingleSite", codes for two-sample data

x.D.ROOT		M:\SoftwareDevelopment\MATLAB\m.stritt\TestDataSet\analysis
x.DELETETEMP		1 (delete temporary files)
x.Quality		0 (0 = fast try-out; 1 = normal high quality)

```

ExploreASL_Initialize.m
MichaelStritt: That's actually pretty straight forward. We open the json. If the "root" variable is called "x", then it's a DataParFile.

MichaelStritt: Just a bit of additional explanation:
(1) If you run the ImportModules, ExploreASL expects the `sourceStructure.json` file.
(2) If all ImportModules are set to false, ExploreASL expects a `DataPar.json` file.
It's the best way, at least in my opinion. Originally, I gave the Import Modules only the directory and it searched for possible files, but that seems a bit more risky. Jan then told me to pass on the `sourceStructure.json` file and I think it works pretty well.

MichaelStritt:
> This looks good, then we could even make this a warning right?
> Since the user asks for importing or processing data and there is no data par file.
> Shouldn't it then be:
> `if any(x.ImportData>0) || any(x.ProcessData>0)`
> ?

Currently I'm using sum instead of any, to check if one of the values is set to true. Any basically does the same at this point as far as I can see.

MichaelStritt:
> Should `x.ImportData` be `x.bImportData` ?

Hm... I like it like it is right now. But if you want we can change it. I'd prioritize finding good fields to put these variables in instead.

MichaelStritt:
@HenkMutsaerts: I think 99% is fine now. If you're not totally unhappy, it would be cool if Jan & I could merge this tomorrow. We then test it and improve it (fields for the variable groups, additional warnings etc.). Testing it would be way much easier if I'd have some of the other recently closed PRs. I don't want to rebase even more, since we already expect quite a few conflicts. Please? :)

HenkMutsaerts:
> > This looks good, then we could even make this a warning right?
> > Since the user asks for importing or processing data and there is no data par file.
> > Shouldn't it then be:
> > `if any(x.ImportData>0) || any(x.ProcessData>0)`
> > ?
> 
> Currently I'm using `sum `instead of `any`, to check if one of the values is set to true. `any `basically does the same at this point as far as I can see.

Sure, but it looked as if you used something like `if x.ImportData || x.ProcessData`. 

> > Should `x.ImportData` be `x.bImportData` ?
> 
> Hm... I like it like it is right now. But if you want we can change it. I'd prioritize finding good fields to put these variables in instead.

I only meant prefixing variables that are Booleans by the letter "b" :)

HenkMutsaerts (approved):

MichaelStritt:
> 
> 
> > > This looks good, then we could even make this a warning right?
> > > Since the user asks for importing or processing data and there is no data par file.
> > > Shouldn't it then be:
> > > `if any(x.ImportData>0) || any(x.ProcessData>0)`
> > > ?
> > 
> > 
> > Currently I'm using `sum `instead of `any`, to check if one of the values is set to true. `any `basically does the same at this point as far as I can see.
> 
> Sure, but it looked as if you used something like `if x.ImportData || x.ProcessData`. I don't care for the difference between `sum` and `any`, perhaps `any` allows combining logical and numerical format? Otherwise both should be fine indeed.
> 
> > > Should `x.ImportData` be `x.bImportData` ?
> > 
> > 
> > Hm... I like it like it is right now. But if you want we can change it. I'd prioritize finding good fields to put these variables in instead.
> 
> I only meant prefixing variables that are Booleans by the letter "b" :)

Oh, I understand, makes sense. I changed that in bc39716 .

PR #459, opened 2021-03-29
Closes #449 Use spm_jsonwrite

### LINKED ISSUE #449: QC_collection NaN to null
If there are no correct values for fields within `QC_collection.json`, the value NaN is stored. This is **incorrect**. Correct JSON Notation would either be `"NaN"` or `null`. This breaks the JSON readers in some python and mTRIAL related scripts.

image

MichaelStritt:
Writing NaN to JSON is invalid. This easy fix converts all NaN to null. Tested it using the TestDataSet and it seems to work fine.

### DIFF
--- a/Functions/xASL_adm_SaveJSON.m
+++ b/Functions/xASL_adm_SaveJSON.m
@@ -73,14 +73,10 @@ function xASL_adm_SaveJSON(data, jsonFileName)
     end
 
     fclose all;
+    % Create directory if it does not exist already
     xASL_adm_CreateDir(fileparts(jsonFileName));
-    fid = fopen(jsonFileName,'w');
-    %fid=1;
-    for iD=1:length(data)
-        writeElement(fid, data(iD),'');
-    end
-    fprintf(fid,'\n');
-    fclose(fid);
+    % Write JSON fields to file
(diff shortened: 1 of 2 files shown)


jan-petr (changes requested):
The function xASL_adm_SaveJSON was removed and its calls were replaced by spm_jsonwrite everywhere as a part of #62.

Can you check if this issue with NaNs is solved by #62? If yes, then lets delete this branch and mark this issue as duplicit and close. #62 is ready and will be merged after a review, so no need for this provisory solution.

MichaelStritt:
> 
> 
> The function xASL_adm_SaveJSON was removed and its calls were replaced by spm_jsonwrite everywhere as a part of #62.
> 
> Can you check if this issue with NaNs is solved by #62? If yes, then lets delete this branch and mark this issue as duplicit and close. #62 is ready and will be merged after a review, so no need for this provisory solution.

Perfect, thanks. The issue of #459 was caused by xASL_adm_SaveJSON  from a call in xASL_qc_CollectParameters. You changed it to spm_jsonwrite in #62, so it should be fine. I'll mark it as duplicate and will remove branch & PR.

Issue #184, opened 2020-10-15
Low mean CBF values in TestDataSet


jan-petr:
In test subject Philips_2DEPI_Lesion_NoFLAIR_NoM0_HiQ has too low values
* Check the partial volume correction
* Check about the small FOV

### Release notes
Skip PVC in Population statistics, when this does not make sense for a given ROI.

### Special note
Remove 0-masking in Philips_2DEPI_Lesion_NoFLAIR_NoM0_HiQ in TestDataSets

HenkMutsaerts:
Update: the image voxel CBF values in GM are correct, though still the global GM CBF estimates are too low, so this really needs to be checked.

MDijsselhof:
I think the issue is that the ROI quantification is not also masked by the FOV, therefore only half of the brain has values while the whole ROI is used for calculating the mean.

MDijsselhof:
I suggest multiplying the x.WB mask with FoV at all times, what do you think?

jan-petr:
Yes. That is a good solution to me. Also - please try to re-run the same thing also with native-processing turned on and see what are the mean CBF values in the native processing. Because we might need to do the same thing for native processing to ensure that it follows a similar procedure. But it is possible that the FOV multiplication was already there.

A question for Henk - was the initial logic meant like this, or was the CBF image supposed to be set to NaN everywhere outside the FoV?

MDijsselhof:
I've hardcoded the vector to include all masks, the resulting values are:

![image](https://user-images.githubusercontent.com/75380250/113002697-56977d80-9172-11eb-934f-21156322654e.png)

Henk-Jan, Jan told me you could look into the reason why x.S.bmasking is set to 0 in this case.

MDijsselhof:
Quick fix: remove _bMasking=0_ from datapar

jan-petr:
I have recreated the branch so that we don't have several commits for a single change.

I have copied the commits - with some modifications - see the comments below.

jan-petr:
@MDijsselhof FYI:
I have excluded zeros from Mean computation, but also from SpatialCoV computation.

I have done that on mask, rather than on the imCBF image as past that point, imCBF is masked and no NaNs should be there anymore - this wouldn't be a problem for the mean calculation but could be for PVC.

jan-petr:
I have closed the PR and will create a new one, because I have recreated the branch.

jan-petr:
[...]

PR #468, opened 2021-04-06
#184 subject-specific masking vector changed

MDijsselhof:
In test subject Philips_2DEPI_Lesion_NoFLAIR_NoM0_HiQ has too low values

- Check the partial volume correction

- Check about the small FOV


jan-petr:
Some observations:
- It is only the mask x.S.bMasking = [0 0 1 0]; that fixes this. The others don't have large influence on the problem

jan-petr:
I have unlinked this PR and created a new PR, because I have recreated the branch.
This PR is thus closed without merging and marked as duplicit.

PR #469, opened 2021-04-06
Closes #341 Optimization xASL_TrackProgress

### LINKED ISSUE #341: xASL_TrackProgress is computationally intensive if used too much
Inserting `xASL_TrackProgress` in PVC (running it once for every kernel), took 26 seconds instead of 2.6 seconds that `xASL_im_PVCkernel` cost without this insertion. However, if running this progress tracker once for every "z-kernel" this was not a significant addition to the computation time.

Hence, we should go through our code and watch out for nested cases of `xASL_TrackProgress`, aimed to provide more accurate progress tracking, as this may slow down the code a lot.

### Release note: Reduce extensive usage of ExploreASL CLI progress tracker.

MichaelStritt:
Check out #341. I showed that most of the other scripts do not affect the overall computation time that much. If you disagree, we can remove xASL_TrackProgress in other scripts as well. I used the auto format option, since some lines weren't correctly aligned, which is why it shows a lot of changed lines, even though I only changes the content in 2 lines.

### DIFF
--- a/Functions/xASL_im_PVCkernel.m
+++ b/Functions/xASL_im_PVCkernel.m
@@ -1,4 +1,4 @@
-function [imPVC,imCBFrec,imResidual] = xASL_im_PVCkernel(imCBF, imPV,kernel,mode)
+function [imPVC,imCBFrec,imResidual] = xASL_im_PVCkernel(imCBF, imPV, kernel, mode)
 %xASL_im_PVCkernel Partial volume correction (PVC) of ASL data using GM-,WM-partial volume maps using linear regression
 %
 % FORMAT: [imPVC,imCBFrec,imResidual] = xASL_im_PVCkernel(imCBF, imPV [,kernel,mode])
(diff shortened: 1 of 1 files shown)


HenkMutsaerts (changes requested):
So sorry, but I guess we didn't communicate this well :( I told a while back that I made a mistake and that xASL_TrackProgres is only inefficient when it is inserted in the iX, iY, and iZ voxel loop (e.g., a typical image can be [121 145 121] voxels so this would really be inefficient and unnecessary).

As I also said in another comment, is that the whole goal of `xASL_TrackProgress`, is to enable tracking progress instead of having to look at the same text for 5 minutes, not knowing if it is actually doing something or crashed ;) So it is important to keep `xASL_TrackProgress` inserted in for-loops, but not too deep. So in the above example, it would be good to keep it inserted in the first for-loop layer rather than the last for-loop layer.

If you tested that it doesn't actually use up too much time, then we can simply keep it.
So does this mean that we don't change it, or was there a case where `xASL_TrackProgress` was really inefficient?

### Reformatting
Very nice idea! How can we do this auto-formatting, can we discuss this (and also discuss a default style)? I would not include formatting inside this PR because it is then two different changes in one commit, I'd rather keep them separated, I still strongly believe "one essential change per commit" is the way to go, also backed up by many texts online :) Do you mind doing the formatting in a separate commit?

MichaelStritt:
> So sorry, but I guess we didn't communicate this well :( I told a while back that I made a mistake and that xASL_TrackProgres is only inefficient when it is inserted in the iX, iY, and iZ voxel loop (e.g., a typical image can be [121 145 121] voxels so this would really be inefficient and unnecessary).
> 
> As I also said in another comment, is that the whole goal of `xASL_TrackProgress`, is to enable tracking progress instead of having to look at the same text for 5 minutes, not knowing if it is actually doing something or crashed ;) So it is important to keep `xASL_TrackProgress` inserted in for-loops, but not too deep. So in the above example, it would be good to keep it inserted in the first for-loop layer rather than the last for-loop layer.

No need to be sorry Henk. I totally get what you're saying. The thing is that here...

```
% For each pixel on the mask do the calculation
fprintf('\nCalculating PVC:    ');
for z=1:size(imCBF,3)
    xASL_TrackProgress(z, size(imCBF,3));
```

... you iterate over each individual voxel, right? So there's no outer- and inner loop structure. We can use an if-else-statement to control the number of calls, but that would make things more clunky and not necessarily faster. I'm fine with not changing it and I'm fine with changing it, here you kind of have to make a decision regarding user feedback vs. speed. With the profiling of the different functions I just wanted to show that the number of calls in xASL_im_PVCkernel is way much higher compared to the other ones.

> Very nice idea! How can we do this auto-formatting, can we discuss this (and also discuss a default style)? I would not include formatting inside this PR because it is then two different changes in one commit, I'd rather keep them separated, I still strongly believe "one essential change per commit" is the way to go, also backed up by many texts online :) Do you mind doing the formatting in a separate commit?

I wouldn't start doing separate PRs for these things. If you open them simultaneously you'll get weird merge conflicts. If you don't then everything takes way much longer. Cosmetics in the code shouldn't take that much time (at least in my opinion).

I mostly use this Matlab specific auto-indentation and I'll always use a single tab for everything within a function, so that it is one "level" above the function definition. Similar to python or other programming languages.

HenkMutsaerts:
>> ... you iterate over each individual voxel, right? So there's no outer- and inner loop structure. 

No if you read further, Jan (who wrote this code) iterates over voxels in Z, Y and X coordinates, in three inner loops, e.g.,:
`for iZ=1:size(image, 3)`
`.   for iY=1:size(image, 2)`
` . .       for iX=1:size(image, 1)`

a while back, I had `xASL_TrackProgress` inserted under the inner loop, which would really repeat this function for each voxel. Whereas now it is under the outer loop (Z-axis), so it only repeats the function for the number of slices (i.e. max ~100 times). I've added `xASL_TrackProgress` to `xASL_im_PVCkernel` as the latter function can take very long, then it is nice to have some progress feedback. So I would leave it like it was :)

>> I wouldn't start doing separate PRs for these things. If you open them simultaneously you'll get weird merge conflicts. If you don't then everything takes way much longer. Cosmetics in the code shouldn't take that much time (at least in my opinion).

Good point. Then at least I would put this in a separate commit. Some other code is completely unreadable because of poor cosmetics, unreadable code to me equals unusable code :) That's why I put so much emphasis on cosmetics. I also thought that it would be nicer to leave this auto-formatting out here, and do one huge branch merge/commit of auto-formatting of all functions, in a single version update or while releasing, to avoid conflicts (auto-formatting will give conflicts anyway...).


>> I mostly use this Matlab specific auto-indentation and I'll always use a single tab for everything within a function, so that it is one "level" above the function definition. Similar to python or other programming languages.

I see, that makes sense. From a practical perspective it just seems to limit your readability with 4 spaces ;) I also asked this somewhere else, which you can then ignore.

jan-petr:
Agree with Henk. We want to have it in for-loops, but at the highest level and not the deepest level. For PVC - this was already in deep level - it was called for each Z, but there was a X and Y loop inside taking much more time. So the PVC commit can be removed.

Regarding other uses - lets indeed focus on those that go to hundreds - there you can either move it up in the loop or alter it in such a way that it is called only on even 10th or 100th loop iteration and not on every one.
e.g.
for i=1:1000
if mod(x,100) == 0, xASL_TrackProgress(i,1000),end
end

MichaelStritt:
Ah okay, sorry, I didn't check the code to that level. Probably a bit lazy from my side. I reverted the previous commit and now only the fix for the format and indentation are left.

MichaelStritt:
As shown in the issue text, there are no other super significant usages. At least in my opinion.

HenkMutsaerts (approved):

PR #472, opened 2021-04-06
Bug #184 low mean cbf test data

### LINKED ISSUE #184: Low mean CBF values in TestDataSet

jan-petr:
Close #184

### DIFF
--- a/Functions/xASL_stat_ComputeMean.m
+++ b/Functions/xASL_stat_ComputeMean.m
@@ -1,5 +1,5 @@
 function [CBF_GM, CBF_WM] = xASL_stat_ComputeMean(imCBF, imMask, nMinSize, bPVC, bParametric, imGM, imWM)
-%xASL_stat_ComputeMean calculates mean CBF in the image with optional partial volume correction.
+%xASL_stat_ComputeMean calculates mean or median of CBF in the image across a mask with an optional partial volume correction.
 %
 % FORMAT:  [CBF_GM CBF_WM] = xASL_stat_ComputeMean(imCBF[, imMask, nMinSize, bPVC, bParametric, imGM, imWM])
 %
@@ -9,9 +9,9 @@ function [CBF_GM, CBF_WM] = xASL_stat_ComputeMean(imCBF, imMask, nMinSize, bPVC,
 %   nMinSize - minimal size of the ROI in voxels, if not big enough, then return NaN
 %            - ignore when 0 (OPTIONAL, default = 0)
 %   bPVC   - perform PV-correction (OPTIONAL, DEFAULT = 0)
-%            0 - don't do partial volume correction, just calculate a median on imMask
-%            1 - simple partial volume correction by dividing by the GM mask
-%            2 - partial volume correction using linear regression and imGM, imWM masks
+%            0 - don't do partial volume correction, just calculate a mean or median on imMask
+%            1 - simple partial volume correction by normalizaton by the GM volume - see Petr et al. 2018
+%            2 - partial volume correction using linear regression and imGM, imWM maps according to Asllani et al. 2008
 %   bParametric - performs parametric statistics (1 mean) or non-parametric when turned off (0 median) (OPTIONAL, DEFAULT 1) 
 %   imGM   - GM partial volume map with the same size as imCBF
 %            (OPTIONAL, REQUIRED for bPVC==2 and bPVC==1)
@@ -21,7 +21,7 @@ function [CBF_GM, CBF_WM] = xASL_stat_ComputeMean(imCBF, imMask, nMinSize, bPVC,
 %   CBF_GM - calculated CBF for options bPVC==0:2
 %   CBF_WM - calculated WM CBF for bPVC==2
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
-% DESCRIPTION: It calculates mean CBF over the mask imMask if the mask volume exceeds nMinSize. It calculates either
+% DESCRIPTION: It calculates mean or median of CBF over the mask imMask if the mask volume exceeds nMinSize. It calculates either
 %              a mean, a median, or a mean after PVC, depending on the settings of bPVC. For the PVC options, it needs also imGM and imWM and returns the
 %              separate PV-corrected values calculated over the entire ROI.
 %
--- a/Functions/xASL_stat_ComputeMean.m
+++ b/Functions/xASL_stat_ComputeMean.m
@@ -1,26 +1,28 @@
 function [CBF_GM, CBF_WM] = xASL_stat_ComputeMean(imCBF,imMask,nMinSize,bPVC,imGM,imWM)
-%Calculates mean CBF in the image with optional partial volume correction.
+%xASL_stat_ComputeMean calculates mean CBF in the image with optional partial volume correction.
 %
 % FORMAT:  [CBF_GM CBF_WM] = xASL_stat_ComputeMean(imCBF [,imMask,nMinSize,bPVC,imGM,imWM])
 %
 % INPUT:
 %   imCBF  - input CBF volume (REQUIRED)
-%   imMask - mask for the calculation (OPTIONAL, default = finite part of imCBF)
+%   imMask - mask for the calculation (OPTIONAL, DEFAULT = finite part of imCBF)
 %   nMinSize - minimal size of the ROI in voxels, if not big enough, then return NaN
 %            - ignore when 0 (OPTIONAL, default = 0)
-%   bPVC   - perform PV-correction (OPTIONAL, default = 0)
-%            0 - don't do partial volume correction, just calculate a median on imMask (+imGM and imWM)
+%   bPVC   - perform PV-correction (OPTIONAL, DEFAULT = 0)
+%            0 - don't do partial volume correction, just calculate a median on imMask
 %            1 - simple partial volume correction by dividing by the GM mask
 %            2 - partial volume correction using linear regression and imGM, imWM masks
 %   imGM   - GM partial volume map with the same size as imCBF
-%            (OPTIONAL, mandatory for bPVC==2 and bPVC==1)
+%            (OPTIONAL, REQUIRED for bPVC==2 and bPVC==1)
 %   imWM   - WM partial volume map with the same size as imCBF
-%            (OPTIONAL, mandatory for bPVC==2, if provided for bPVC=0 gives also CBF_WM)
+%            (OPTIONAL, REQUIRED for bPVC==2)
 % OUTPUT:
 %   CBF_GM - calculated CBF for options bPVC==0:2
-%   CBF_WM - calculated WM CBF for bPVC==2, and for ==0 if imWM provided
+%   CBF_WM - calculated WM CBF for bPVC==2
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
-% DESCRIPTION: It behaves in a similar way as VAR.
+% DESCRIPTION: It calculates mean CBF over the mask imMask if the mask volume exceeds nMinSize. It calculates either
+%              a mean, a median, or a mean after PVC, depending on the settings of bPVC. For the PVC options, it needs also imGM and imWM and returns the
+%              separate PV-corrected values calculated over the entire ROI.
 %
 % 1. Admin
 % 2. Mask calculations
@@ -34,8 +36,6 @@ function [CBF_GM, CBF_WM] = xASL_stat_ComputeMean(imCBF,imMask,nMinSize,bPVC,imG
 %          CBF_GM = xASL_stat_ComputeMean(imCBF,imMask,[])
 %          CBF_GM = xASL_stat_ComputeMean(imCBF,[],[],0)
 %          CBF_GM = xASL_stat_ComputeMean(imCBF,imMask,290,0)
-%          CBF_GM = xASL_stat_ComputeMean(imCBF,imMask,290,0,imGM)
-% [CBF_GM CBF_WM] = xASL_stat_ComputeMean(imCBF,imMask,290,0,imGM,imWM)
 %          CBF_GM = xASL_stat_ComputeMean(imCBF,imMask,[],1,imGM)
 %          CBF_GM = xASL_stat_ComputeMean(imCBF,imMask,[],2,imGM,imWM)
 % [CBF_GM CBF_WM] = xASL_stat_ComputeMean(imCBF,[],[],2,imGM,imWM)
@@ -50,13 +50,18 @@ function [CBF_GM, CBF_WM] = xASL_stat_ComputeMean(imCBF,imMask,nMinSize,bPVC,imG
 % Copyright (C) 2015-2021 ExploreASL
 
 %% 1. Admin
+if nargin<1
+	error('imCBF is a required input parameter');
+end
 
 if nargin<2 || isempty(imMask)
 	imMask = ones(size(imCBF));
 end
+
 if nargin<3 || isempty(nMinSize)
 	nMinSize = 0;
 end
+
 if nargin<4 || isempty(bPVC)
 	bPVC = 0;
 end
@@ -69,6 +74,10 @@ if nargin<6
 	imWM = [];
 end
 
+if nargout>1 && bPVC~=2
+	warning('Calculates CBF_WM only for PVC==2 option');
+end
+
 if (nargout>1) && isempty(imWM)
 	warning('Cannot calculate CBF_WM when imWM is not provided');
 end
@@ -78,6 +87,13 @@ if sum(isfinite(imCBF(:)))==0
     CBF_GM = NaN;
     CBF_WM = NaN;
     return;
+end
+
+if bPVC == 2
+	% If running PVC, then need imGM and imWM of the same size as imCBF
+	if ~isequal(size(imCBF),size(imGM)) || ~isequal(size(imCBF),size(imWM))
+		warning('When running PVC, need imGM and imWM of the same size as imCBF');
+	end
 end
 
 %% 2. Mask calculations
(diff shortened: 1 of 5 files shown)


Functions/xASL_stat_GetROIstatistics.m
```diff
@@ -553,7 +553,7 @@
         end
         
         % Initialize matrices with NaNs
-		FieldsAre = {'DAT_mean_PVC0' 'DAT_median_PVC0' 'DAT_mean_PVC2'}; % update this with fields filled below!
+		FieldsAre = {'DAT_mean_PVC0' 'DAT_median_PVC0'}; % update this with fields filled below!
```
HenkMutsaerts, line 556: Why do you skip the initialization here? I see now that we initialize these twice, here and also below. Though note that this is just the end result (e.g., the mean ROI-value), not a mask or input data. I believe it's important to keep initializing this, to be able to differentiate missing values (NaNs in the final `*.tsv` file) with e.g., having no subject.

Also, this is not complete, as in the case you have e.g., `x.S.DAT_sum_PVC0` below (in the case of non-ASL data), this is not initialized. Note that this initialization costs no significant memory (only a matrix of subjects*ROIs).

Perhaps we can discuss a general solution for issues like this: 
1. if processing crashes, do we want a NaN output, such that the remaining pipeline can still continue?
2. if data is incomplete (too small mask or outside FoV), do we want a NaN output?

I believe the second is very important, the first is debatable. If we choose the second only, the initialization should be below, inside the loop (also easier to read the script). If we choose the first, we need to initialize everything here. Thinking of it, I would go for the second option only, removing the initialization here, but moving the initialization in the loop back to its original location

Functions/xASL_stat_GetROIstatistics.m
```diff
@@ -592,13 +597,6 @@
 					x.S.DAT_CoV_PVC0(SubjSess,iROI) = xASL_stat_ComputeSpatialCoV(DataIm, CurrentMask, 0, 0);
 				end
             else
-                % Start with defaults
```
HenkMutsaerts, line 595: So I would put the initialization back to this location, if the masks are empty it should still initialize a NaN, otherwise, there will be zeros for subject-ROI combinations with an empty mask.

Functions/xASL_stat_GetROIstatistics.m
```diff
@@ -553,7 +553,7 @@
         end
         
         % Initialize matrices with NaNs
```
HenkMutsaerts, line 555: To avoid confusion, probably best to replace 

`% Initialize matrices with NaNs`

with

`% Initialize output data matrix ([x y] = [subject/session ROI-statistics]) with NaNs`

HenkMutsaerts (changes requested):

jan-petr:
Last edit - remove further restriction to GM>0.7 inside the xASL_stat_ComputeMean - all thresholding of ROI should be done in the parent function.

Functions/xASL_stat_GetROIstatistics.m
jan-petr: Done

Functions/xASL_stat_GetROIstatistics.m
jan-petr: My main motivation was that for certain cases - like whole-brain or GM/WM-masking turned of completely - it does not make sense to do PVEC. We will therefore skip the calculation. But if NaNs are prefilled, it will save a file full of NaNs. So I am skipping the result for some cases (bSkipPVC), but then I also have to skip the initialization - which I do below, but can't do it here when it is not yet know if this will be skipped.

The easiest solution is to move the NaN initialization to below only - as you say. Then I'm completely happy ;)

Regarding your question - if processing crashes then we are fine - the file is written to disc only after this function is finished, so (1) is not a problem. I agree with (2), so moving the second option below. And for cases where nothing is initialized or computed, the TSV won't be written at all - which will be fine and remove confusion.

Functions/xASL_stat_GetROIstatistics.m
jan-petr: OK

jan-petr:
Cleaned xASL_stat_ComputeSpatialCoV. Updated the function headers and comments. And since calls with PVC==0 and imGM does not do anything, I have removed that from the calls as superfluous.

jan-petr:
Cleaned xASL_stat_ComputeMean. For PVC==0 and -1, it was restricting mask to GM>0.7 if imGM was provided as input parameter. This is unwanted behavior since we want to restrict the mask outside of this function and not inside. However, we were only using this in the QC_Collect function, so I have removed it from inside ComputeMean and added a proper GM>0.5 mask to the QC collect. This change didn't affect PVC==2 option as further narrowing of mask was not done. 

Therefore, I have just cleaned the ComputeMean function, put in comments and removed options that we didn't want and were rather unused. The whole code and all results thus should be unaffected by this change completely. Except that for QC, we use now GM>0.5 instead of GM>0.7.

Overall, I just cleaned the code and removed ambiguity without really affecting any past or future results.

HenkMutsaerts (changes requested):
Nearly there :)

One thing that I believe is important, is that we do not merge this commit in R1.6.0 but in a separate hotfix. The point is that our quantification will change a lot with this small commit, but not so much with all other commits in R1.6. So it makes more sense to separate this commit, then it is easier to see if the effects of the other R1.6 commits on quantification.

so it saves a lot of time if we do the testing for this commit only in a separate release, and then testing the rest (the other commits of R1.6 are probably

jan-petr:
> Nearly there :)
> 
> One thing that I believe is important, is that we do not merge this commit in R1.6.0 but in a separate hotfix. The point is that our quantification will change a lot with this small commit, but not so much with all other commits in R1.6. So it makes more sense to separate this commit, then it is easier to see if the effects of the other R1.6 commits on quantification.
> 
> so it saves a lot of time if we do the testing for this commit only in a separate release, and then testing the rest (the other commits of R1.6 are probably

Copied my answer from the issue: 
As said above. This shouldn't affect the quantification at all. I have removed options that were not really used. Only the QC_CBF will be affected, but none of the Results. And it will skip calculation of PVC2 for some cases, but for cases where it makes no sense.

So we should indeed be careful that it doesn't change the results, but I assume it won't. So lets just run 10-TestDatasets before and after merging it to check if all is good.

Functions/xASL_stat_ComputeMean.m
```diff
@@ -107,64 +123,43 @@
...
-	end
-	if ~isempty(imWM)
-		CBF_WM = xASL_stat_MeanNan(imCBF(imWM>0.7));
-	end
-    case 0
-    %% 3b. No PVC and median
-	if isempty(imGM)
+	case -1
+		%% 3a. No PVC and simple mean
+		CBF_GM = xASL_stat_MeanNan(imCBF);
```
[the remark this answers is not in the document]
jan-petr: OK. Will do that 
PVC = 0,1,2
bParametric = 0,1
I will change this inside xASL_stat_ComputeMean and make sure that the outputs of this function are correctly named.

So I will not change any functionality, but just rearrange the input parameters to make stuff more transparent. The parametric PVC will not yet be there, but it won't be called either.

Functions/xASL_stat_ComputeMean.m
```diff
@@ -69,6 +74,10 @@
 	imWM = [];
 end
 
+if nargout>1 && bPVC~=2
+	warning('Calculates CBF_WM only for PVC==2 option');
```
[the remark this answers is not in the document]
jan-petr: True, we should issue a warning but also set the second input so that it doesn't crash. Will do that.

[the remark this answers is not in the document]
HenkMutsaerts: Why don't we set both output parameters to NaN to start with, for the whole function? Then in any case, these are always initialized as NaN.

Functions/xASL_stat_ComputeMean.m
```diff
@@ -12,6 +12,7 @@
 %            0 - don't do partial volume correction, just calculate a median on imMask
```
HenkMutsaerts, line 12: should be `median or mean`

Functions/xASL_stat_ComputeMean.m
```diff
@@ -12,6 +12,7 @@
...
 %            1 - simple partial volume correction by dividing by the GM mask
```
HenkMutsaerts, line 13: should be `simple, "one component" PVC`

Functions/xASL_stat_ComputeMean.m
```diff
@@ -12,6 +12,7 @@
...
 %            2 - partial volume correction using linear regression and imGM, imWM masks
```
HenkMutsaerts, line 14: should be `two component PVC (otherwise known as PVC)` Perhaps also helpful to move Iris' reference to here (or refer by a * indicator)

Functions/xASL_stat_GetROIstatistics.m
```diff
@@ -647,11 +647,11 @@
                         % mask)
                         fprintf('%s\n', ['Empty CBF mask for subject ' xASL_num2str(iSubject) '_ASL_' xASL_num2str(iSess) ', ROI ' xASL_num2str(iROI)]); % slightly different warning/mask as above for sCoV
                     else
-                        x.S.DAT_mean_PVC0(SubjSess,iROI) = xASL_stat_ComputeMean(DataIm, CurrentMask, MinVoxels, -1);
-                        x.S.DAT_median_PVC0(SubjSess,iROI) = xASL_stat_ComputeMean(DataIm, CurrentMask, MinVoxels, 0);
+                        x.S.DAT_mean_PVC0(SubjSess,iROI) = xASL_stat_ComputeMean(DataIm, CurrentMask, MinVoxels, 0, 1);
```
HenkMutsaerts, line 650: `bParametric` should also be 0 here, right?

Functions/xASL_stat_ComputeSpatialCoV.m
```diff
@@ -1,7 +1,7 @@
-function sCov = xASL_stat_ComputeSpatialCoV(imCBF,imMask,nMinSize,bPVC,imGM,imWM)
+function sCov = xASL_stat_ComputeSpatialCoV(imCBF, imMask, nMinSize, bPVC, imGM, imWM)
```
HenkMutsaerts, line 1: Would be good to give this function also a `bParametric` option, we can do this later

HenkMutsaerts (changes requested):

Functions/xASL_stat_ComputeSpatialCoV.m
jan-petr: Done. Formally prepare and updated function calls.

jan-petr:
All answered - see above. Please make sure to resolve comments you agree with - so that I know you just haven't missed my answers...

Functions/xASL_stat_GetROIstatistics.m
HenkMutsaerts: OK sounds good!

HenkMutsaerts:
> > Nearly there :)
> > One thing that I believe is important, is that we do not merge this commit in R1.6.0 but in a separate hotfix. The point is that our quantification will change a lot with this small commit, but not so much with all other commits in R1.6. So it makes more sense to separate this commit, then it is easier to see if the effects of the other R1.6 commits on quantification.
> > so it saves a lot of time if we do the testing for this commit only in a separate release, and then testing the rest (the other commits of R1.6 are probably
> 
> Copied my answer from the issue:
> As said above. This shouldn't affect the quantification at all. I have removed options that were not really used. Only the QC_CBF will be affected, but none of the Results. And it will skip calculation of PVC2 for some cases, but for cases where it makes no sense.
> 
> So we should indeed be careful that it doesn't change the results, but I assume it won't. So lets just run 10-TestDatasets before and after merging it to check if all is good.

Do you mean that the qCBF doesn't change? I know that, but the QC CBF ROI values do change (because of the different masking right?), and we use those to check if there is not a huge changes between versions. That was my point, this will give a large change between 1.5.1 and 1.6.0 whereas all the other 1000 commits won't really change the ROI values much :)

HenkMutsaerts (approved):

jan-petr:
I talk about the call from the QC module (the only thing that has changed):
Functions/xASL_qc_CollectQC_ASL.m:123:    ASL.CBF_GM_Median_mL100gmin = xASL_stat_ComputeMean(CBFmasked, GMmasked>0.5,[], 0, 0);

The ROI statistics, also those called from xASL_qc_TestExploreASL, haven't changed.

Before I merge this one, I will run the full xASL_qc_TestExploreASL test to verify this...

jan-petr:
{'Data'             }    {'mean_qCBF_TotalGM'}    {'median_qCBF_Tot…'}    {'median_qCBF_Dee…'}    {'CoV_qCBF_TotalGM'}    {'GMvol' }    {'WMvol' }    {'CSFvol'}    {'PipelineCompleted'}    {'TC_ASL_Registra…'}
    {'GE_3Dspiral_T1p…'}    {'65.2658'          }    {'52.4785'          }    {'31.8834'          }    {'0.29875'         }    {[0.6399]}    {[0.5052]}    {[0.4329]}    {[                1]}    {[           0.9005]}
    {'Philips_2DEPI_F…'}    {'39.0707'          }    {'27.7337'          }    {'6.7209'           }    {'0.71364'         }    {[0.5349]}    {[0.4587]}    {[0.4051]}    {[                1]}    {[           0.7297]}
    {'Philips_2DEPI_F…'}    {'41.7674'          }    {'31.8349'          }    {'13.6448'          }    {'0.37681'         }    {[0.5442]}    {[0.3921]}    {[0.3564]}    {[                1]}    {[           0.9019]}
    {'Philips_2DEPI_F…'}    {'47.3394'          }    {'42.4305'          }    {'19.9569'          }    {'0.42897'         }    {[0.6982]}    {[0.5828]}    {[0.2224]}    {[                1]}    {[           0.8162]}
    {'Philips_2DEPI_L…'}    {'empty'            }    {'14.172'           }    {'2.9898'           }    {'0.89531'         }    {'empty' }    {'empty' }    {'empty' }    {[                0]}    {[           0.7593]}
    {'Philips_2DEPI_S…'}    {'45.9493'          }    {'41.1539'          }    {'9.4059'           }    {'0.36499'         }    {[0.8167]}    {[0.4596]}    {[0.1516]}    {[                1]}    {[           0.8543]}
    {'Philips_3DGRASE…'}    {'35.2382'          }    {'26.316'           }    {'7.6074'           }    {'0.66134'         }    {[0.5709]}    {[0.5279]}    {[0.4218]}    {[                1]}    {[           0.8571]}
    {'Philips_3DGRASE…'}    {'32.1459'          }    {'24.3005'          }    {'8.5921'           }    {'0.43058'         }    {[0.7239]}    {[0.5520]}    {[0.3564]}    {[                1]}    {[           0.8413]}
    {'Siemens3DGRASE_…'}    {'37.5377'          }    {'37.2731'          }    {'21.767'           }    {'0.78865'         }    {[0.5191]}    {[0.4658]}    {[0.4054]}    {[                1]}    {[           0.6914]}
    {'Siemens_2DEPI_n…'}    {'67.5267'          }    {'50.6621'          }    {'13.4265'          }    {'0.31694'         }    {[0.6887]}    {[0.6031]}    {[0.3101]}    {[                1]}    {[           0.8758]}

  Column 11

    {'TC_M0_Registrat…'}
    {[           0.9319]}
    {[           0.8139]}
    {[           0.9479]}
    {[           0.8449]}
    {[           0.8910]}
    {[           0.8027]}
    {[           0.7547]}
    {[           0.8952]}
    {[           0.4987]}
    {[           0.9589]}

jan-petr:
Henk, See the xASL_qc_Test - all is good and is equal to 1.5.0. 

Except for one line that crashes on a dilateROI error that was fixed in #483 of course ;) So I need to rebase and retest for that...

Anyway - the numbers look ok and the only change is in the collect_QC as noted before.

If you agree, then I will rebase develop and run the whole test again. And if the numbers will be equal to 1.5.0 then we will  be able to state that this was indeed cosmetic change and no real effect on thresholds and we will merge? To repeat, I have mostly revamped the script and removed confusing options that were never really used.

HenkMutsaerts:
> Henk, See the xASL_qc_Test - all is good and is equal to 1.5.0.
> 
> Except for one line that crashes on a dilateROI error that was fixed in #483 of course ;) So I need to rebase and retest for that...
> 
> Anyway - the numbers look ok and the only change is in the collect_QC as noted before.
> 
> If you agree, then I will rebase develop and run the whole test again. And if the numbers will be equal to 1.5.0 then we will be able to state that this was indeed cosmetic change and no real effect on thresholds and we will merge? To repeat, I have mostly revamped the script and removed confusing options that were never really used.

Yes looks good! Also: I tested #483 and this went well so I guess we don't expect any issues.

HenkMutsaerts (approved):

jan-petr:
{'Data'             }    {'mean_qCBF_TotalGM'}    {'median_qCBF_Tot…'}    {'median_qCBF_Dee…'}    {'CoV_qCBF_TotalGM'}    {'GMvol' }    {'WMvol' }    {'CSFvol'}    {'PipelineCompleted'}    {'TC_ASL_Registra…'}
    {'GE_3Dspiral_T1p…'}    {'65.2658'          }    {'52.4785'          }    {'31.8834'          }    {'0.29875'         }    {[0.6399]}    {[0.5052]}    {[0.4329]}    {[                1]}    {[           0.9005]}
    {'Philips_2DEPI_F…'}    {'39.0707'          }    {'27.7337'          }    {'6.7209'           }    {'0.71364'         }    {[0.5349]}    {[0.4587]}    {[0.4051]}    {[                1]}    {[           0.7297]}
    {'Philips_2DEPI_F…'}    {'41.7674'          }    {'31.8349'          }    {'13.6448'          }    {'0.37681'         }    {[0.5442]}    {[0.3921]}    {[0.3564]}    {[                1]}    {[           0.9019]}
    {'Philips_2DEPI_F…'}    {'47.3394'          }    {'42.4305'          }    {'19.9569'          }    {'0.42897'         }    {[0.6982]}    {[0.5828]}    {[0.2224]}    {[                1]}    {[           0.8162]}
    {'Philips_2DEPI_L…'}    {'empty'            }    {'14.172'           }    {'2.9898'           }    {'0.89531'         }    {[0.7099]}    {[0.5136]}    {[0.1939]}    {[                1]}    {[           0.7593]}
    {'Philips_2DEPI_S…'}    {'45.9493'          }    {'41.1539'          }    {'9.4059'           }    {'0.36499'         }    {[0.8167]}    {[0.4596]}    {[0.1516]}    {[                1]}    {[           0.8543]}
    {'Philips_3DGRASE…'}    {'35.2382'          }    {'26.316'           }    {'7.6074'           }    {'0.66134'         }    {[0.5709]}    {[0.5279]}    {[0.4218]}    {[                1]}    {[           0.8571]}
    {'Philips_3DGRASE…'}    {'32.1459'          }    {'24.3005'          }    {'8.5921'           }    {'0.43058'         }    {[0.7239]}    {[0.5520]}    {[0.3564]}    {[                1]}    {[           0.8413]}
    {'Siemens3DGRASE_…'}    {'37.5377'          }    {'37.2731'          }    {'21.767'           }    {'0.78865'         }    {[0.5191]}    {[0.4658]}    {[0.4054]}    {[                1]}    {[           0.6914]}
    {'Siemens_2DEPI_n…'}    {'67.5267'          }    {'50.6621'          }    {'13.4265'          }    {'0.31694'         }    {[0.6887]}    {[0.6031]}    {[0.3101]}    {[                1]}    {[           0.8758]}

  Column 11

    {'TC_M0_Registrat…'}
    {[           0.9319]}
    {[           0.8139]}
    {[           0.9479]}
    {[           0.8449]}
    {[           0.8910]}
    {[           0.8027]}
    {[           0.7547]}
    {[           0.8952]}
    {[           0.4987]}
    {[           0.9589]}

jan-petr:
All is good. All rows are there and correctly reported. First column of the 5th study is empty, because GMmasking is turned of for that study and then PVC cannot be reported with GMmasking turned of, because GM==WM in that case and PVC is thus ill-conditioned.

PR #476, opened 2021-04-08
#475 Fix bug with space in study folder name

### LINKED ISSUE #475: Study folder cannot have space in the name
If the study folder has a space in the name, then ExploreASL crashes inside the CAT segmentation.

The reason is that cat_main.m at line 954 works with relative rather than absolute paths. Reading NIfTI with a relative path leads to calling a xASL_adm_UnzipNifti that on line 48 gets the current path and escapes special characters - which is fine when used for command-line calls, but not fine as internal Matlab functions won't work with these escaped filenames.

Solution:
1. Remove this Linux escaping as not necessary in xASL_adm_UnzipNifti 
2. Rework the cat_main part of code to work with absolute and not relative paths

I will do both 1 and 2 to avoid future errors, but either of those would be enough.

Or do you think that this could be problematic for other functions? @MichaelStritt @HenkMutsaerts 
I mean - step 2 is definitively helpful and should be done. Step 1 - it actually solves my case when tested on Linux, so I was a bit wondering why it is there in the first place when it was really counter-effective, but maybe you know of an example where this help and you can revert my changes in 1.

### Release notes
Fix error with studies that have a special character in their name.

jan-petr:
Close #475

### DIFF
--- a/External/SPMmodified/toolbox/cat12/cat_main.m
+++ b/External/SPMmodified/toolbox/cat12/cat_main.m
@@ -951,17 +951,15 @@ function [res,job,VT,VT0,pth,nam,vx_vol,d] = cat_main_updatepara(res,tpm,job)
   clear def; 
 
   % EXPLOREASL HACK
-    CurrDir=pwd;
-    if exist('./mri','dir')
-        cd('mri');
-    end
-    if xASL_exist('nT1.nii')
-        tIM = xASL_io_Nifti2Im('nT1.nii');
-        tIM(isnan(tIM)) = 0;
-        xASL_io_SaveNifti('nT1.nii', 'nT1.nii', tIM, [], false);
-    end
-    cd(CurrDir);
-  
+  CurrDir=pwd;
+  if exist(fullfile(CurrDir,'mri'), 'dir')
+	  if xASL_exist(fullfile(CurrDir, 'mri', 'nT1.nii'), 'file')
+		  tIM = xASL_io_Nifti2Im(fullfile(CurrDir, 'mri', 'nT1.nii'));
+		  tIM(isnan(tIM)) = 0;
+		  xASL_io_SaveNifti(fullfile(CurrDir, 'mri', 'nT1.nii'), fullfile(CurrDir, 'mri', 'nT1.nii'), tIM, [], false);
+	  end
+  end
+
   % complete job structure
   defr.ppe = struct(); 
   res = cat_io_checkinopt(res,defr);
--- a/External/SPMmodified/toolbox/cat12/cat_main.m
+++ b/External/SPMmodified/toolbox/cat12/cat_main.m
@@ -952,11 +952,13 @@ function [res,job,VT,VT0,pth,nam,vx_vol,d] = cat_main_updatepara(res,tpm,job)
 
   % EXPLOREASL HACK
   CurrDir=pwd;
-  if exist(fullfile(CurrDir,'mri'), 'dir')
-	  if xASL_exist(fullfile(CurrDir, 'mri', 'nT1.nii'), 'file')
-		  tIM = xASL_io_Nifti2Im(fullfile(CurrDir, 'mri', 'nT1.nii'));
+  pathMRI = fullfile(CurrDir,'mri');
+  if exist(pathMRI, 'dir')
+	  pathNT1 = fullfile(pathMRI, 'nT1.nii');
+	  if xASL_exist(pathNT1, 'file')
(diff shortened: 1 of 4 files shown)


MichaelStritt (approved):

External/SPMmodified/toolbox/cat12/cat_main.m
```diff
@@ -951,17 +951,15 @@
...
-        cd('mri');
-    end
-    if xASL_exist('nT1.nii')
-        tIM = xASL_io_Nifti2Im('nT1.nii');
-        tIM(isnan(tIM)) = 0;
-        xASL_io_SaveNifti('nT1.nii', 'nT1.nii', tIM, [], false);
-    end
-    cd(CurrDir);
-  
+  CurrDir=pwd;
```
HenkMutsaerts, line 954: Yes, I like this. You could only make it a bit more readable and less bug-prone by defining `PathT1` and `PathNT1` rather than repeating the `fullfile()`.

External/SPMmodified/xASL/xASL_adm_UnzipNifti.m
```diff
@@ -45,7 +45,7 @@
 % Checks if the pathIn was given as an absolute path or if as a file-name only
 if isempty(pathstr)
 	% If a file-name only, then add the full current path to avoid ambiguity
-	pathstr = xASL_adm_UnixPath(pwd());
+	pathstr = pwd();
```
HenkMutsaerts, line 48: If this is used internally in Matlab only, then `xASL_adm_UnixPath` should probably indeed go out. If there is still a `system` command somewhere using `pathstr`, then we would have to test it in WSL?

External/SPMmodified/xASL/xASL_io_SaveNifti.m
```diff
@@ -55,13 +55,13 @@ function xASL_io_SaveNifti(pathOrigNifti, pathNewNifti, imNew, nBits, bGZip, cha
 [pathstr, name0, ext0] = fileparts(pathNewNifti);
 if isempty(pathstr)
 	% If a file-name only, then add the full current path to avoid ambiguity
-	pathNewNifti = fullfile(xASL_adm_UnixPath(pwd()),[name0 ext0]);
+	pathNewNifti = fullfile(pwd(),[name0 ext0]);
```
HenkMutsaerts, line 58: I would also check how this behaves on Windows, when you have illegal characters (such as spaces) in the path. But I would generally advocate not using spaces in pathnames, as good practice.

HenkMutsaerts (changes requested):

External/SPMmodified/xASL/xASL_io_SaveNifti.m
MichaelStritt: > But I would generally advocate not using spaces in pathnames, as good practice.

Agree, but I think we should support it :/

External/SPMmodified/xASL/xASL_adm_UnzipNifti.m
jan-petr: Yes, Henk, please try to run a single test dataset on a study called "My study" with a space to verify is WSL works.

External/SPMmodified/xASL/xASL_io_SaveNifti.m
jan-petr: Michael tested this, so Windows should be fine.

Yes - we should not have spaces in pahtnames, but good to be able to deal with users who do that :)

External/SPMmodified/toolbox/cat12/cat_main.m
jan-petr: Good point. Fixed.

jan-petr:
All questions resolved. Henk can test 1 dataset on WSL and then approve.

HenkMutsaerts:
I don't have WSL, but Maurice does. So let's ask Maurice to test before we do the release.

HenkMutsaerts (approved):
I don't have WSL, but Maurice does. So let's ask Maurice to test before we do the release.

jan-petr:
> I don't have WSL, but Maurice does. So let's ask Maurice to test before we do the release.

Good point. I've asked @MauricePasternak per email to test this.

MauricePasternak (approved):
Assuming that ExploreASL checks that WSL is on PATH in Windows, no issues appear to stem from this PR when running on a test batch. It's clear to go from my perspective. 👍

Issue #477, opened 2021-04-08
Incorrect external atlases paths


MDijsselhof:
External Vascular territories atlases paths end with incorrect file extensions and should all end with `.nii.gz`

**RELEASE NOTE**

changed vascular territory atlases file extensions to .nii.gz

PR #478, opened 2021-04-08
#477 Incorrect external atlases paths

MDijsselhof:
External Vascular territories atlases paths end with incorrect file extensions and should all end with `.nii.gz`


MichaelStritt (approved):
👍

PR #487, opened 2021-04-12
Closes #486 Add issue templates

### LINKED ISSUE #486: Add issue template
### Release Notes

Introduce document templates for GitHub QC Workflow

MichaelStritt:
Added the following issue template:

> ### Description
> 
> Required: insert bug or feature description here
> 
> ### Requirements
> 
> Optional: add bug or feature requirements here
> 
> ### How to test
> 
> Optional: insert description about how to test the code changes here
> 
> ### Release notes
> 
> Required: summarize the changes for the release notes here


jan-petr (changes requested):
This is very good. But I don't get one thing - is this going to be evoked automatically when creating a new issue?

I would slightly change one thing:
Requirements -> conditions (because these are conditions under which this bug appears, right? Dataset, parameters, OS? If yes, we might add these examples.

MichaelStritt:
> 
> 
> This is very good. But I don't get one thing - is this going to be evoked automatically when creating a new issue?
> 
> I would slightly change one thing:
> Requirements -> conditions (because these are conditions under which this bug appears, right? Dataset, parameters, OS? If yes, we might add these examples.

It's a bit simpler. It's just a markdown template and every time you create an issue, you can select to use these four titles and the corresponding descriptions. You can see an example [here](https://youtu.be/UPsCXqxxJUA?t=397).

MichaelStritt:
I just wanted to give it a try and if everyone likes it, we can create separate ones for bugs/features/releases etc.

MichaelStritt:
Normally you would write the false behavior, so something like "when does the bug appear" or "on which os" etc. in the **description**. The **requirements** would be something like "if I insert this argument, it shouldnt crash and I get this result". The **how to test** would be something like "run this `script` to see if the requirements are fulfilled".

jan-petr:
> > This is very good. But I don't get one thing - is this going to be evoked automatically when creating a new issue?
> > I would slightly change one thing:
> > Requirements -> conditions (because these are conditions under which this bug appears, right? Dataset, parameters, OS? If yes, we might add these examples.
> 
> It's a bit simpler. It's just a markdown template and every time you create an issue, you can select to use these four titles and the corresponding descriptions. You can see an example [here](https://youtu.be/UPsCXqxxJUA?t=397).

So it does pop-up automatically (or I mean I call *that* behavior automatic :)). That's really awesome!

jan-petr:
> I just wanted to give it a try and if everyone likes it, we can create separate ones for bugs/features/releases etc.

Yes - we need slightly different ones for each - feel free to suggest the 3-4 versions. I'm sure everyone *will* like that :)

jan-petr:
> Normally you would write the false behavior, so something like "when does the bug appear" or "on which os" etc. in the **description**. The **requirements** would be something like "if I insert this argument, it shouldnt crash and I get this result". The **how to test** would be something like "run this `script` to see if the requirements are fulfilled".

Ah ok. Sounds good. But we might want to add a minimal example in each category, as I was a bit confused what each category means.

jan-petr (changes requested):
This is very nice. I want all right now :)
Bug, release, feature would be the basic ones. Perhaps also documentation.

And you want to add HJ so that he modifies these right now before creating a new issue for that :)

HenkMutsaerts:
@MichaelStritt very nice! I would only suggest to add one line, which will help us a lot as well :)

### Summary for CHANGES.md

Don't you agree @jan-petr ?

jan-petr:
> @MichaelStritt very nice! I would only suggest to add one line, which will help us a lot as well :)
> 
> ### Summary for CHANGES.md
> Don't you agree @jan-petr ?

Yes - that's one of the main motivations. But 'Summary for CHANGES.md' == 'Release notes' - I like the second a bit more as it sounds more stylish ;) But I'm fine with both.

jan-petr:
On a second thought. I do not have quick ideas on what to fill for different issue types. So if you do not have any idea either, we can merge right away and fill in later...

jan-petr (approved):
See my and HJ's comments. But feel free to merge at any time.

MichaelStritt:
I'd merge this now and add a PR template in a separate issue 👍 
GitHub creates these branches and PRs automatically, so it's a bit easier this way.
I personally prefer **Release notes** compared to **Summary for changes.md**.

jan-petr:
Agree! Please add release notes to this issue - in this case we probably only need to say that there are no release notes as this doesn't affect the code :)

Issue #494, opened 2021-04-13
ExploreASL_Initialize: Fix cross-platform indentation bug


MichaelStritt:
### Description

The indentation of the settings seems different regarding cross-platform use.

### Requirements

Make it look like this...

```
==============================================================================================
 ________                      __                                 ______    ______   __        
/        |                    /  |                               /      \  /      \ /  |      
########/  __    __   ______  ## |  ______    ______    ______  /######  |/######  |## |      
## |__    /  \  /  | /      \ ## | /      \  /      \  /      \ ## |__## |## \__##/ ## |      
##    |   ##  \/##/ /######  |## |/######  |/######  |/######  |##    ## |##      \ ## |      
#####/     ##  ##<  ## |  ## |## |## |  ## |## |  ##/ ##    ## |######## | ######  |## |      
## |_____  /####  \ ## |__## |## |## \__## |## |      ########/ ## |  ## |/  \__## |## |_____ 
##       |/##/ ##  |##    ##/ ## |##    ##/ ## |      ##       |## |  ## |##    ##/ ##       |
########/ ##/   ##/ #######/  ##/  ######/  ##/        #######/ ##/   ##/  ######/  ########/ 
                    ## |                                                                      
                    ## |                                                                      
                    ##/  

==================================== ExploreASL Settings =====================================
DataParPath			test
Import Modules			test
Process Modules			test
bPause				test
iWorker				test
nWorkers			test
==============================================================================================
```

... on every OS.

### How to test

Run ExploreASL_Initialize with different import settings on different OSs.

### Release notes

Fix cross-platform ExploreASL settings design.

MichaelStritt:
### Test on Windows

```
>> [x] = ExploreASL_Master('M:\SoftwareDevelopment\MATLAB\m.stritt\TestDataSet\analysis\DataParameters_LowQ.json', [1 1 1 1], [1 1 1], 0, 1, 1);
ExploreASL will run the import workflow and will run the processing pipeline...


==============================================================================================
 ________                      __                                 ______    ______   __        
/        |                    /  |                               /      \  /      \ /  |      
########/  __    __   ______  ## |  ______    ______    ______  /######  |/######  |## |      
## |__    /  \  /  | /      \ ## | /      \  /      \  /      \ ## |__## |## \__##/ ## |      
##    |   ##  \/##/ /######  |## |/######  |/######  |/######  |##    ## |##      \ ## |      
[...]

PR #495, opened 2021-04-13
#494 ExploreASL_Initialize: Fix cross-platform settings design

MichaelStritt:
### Linked issue

Check comments and examples in #494.

### How to test

Run tests shown in #494 if necessary.

### Comments

Would be helpful to tell me how you want line 659 in `ExploreASL_Initialize` to look like.

### DIFF
--- a/DataParTemplate.m
+++ b/DataParTemplate.m
@@ -48,7 +48,8 @@ function x = DataParTemplate(x)
 % x.exclusion - cell with list of subjects to exclude (OPTIONAL, DEFAULT = empty)
 %             - example: {'005' '018'}
 % x.SESSIONS  - use this to define sessions (OPTIONAL, DEFAULT = {'ASL_1'})
-%             - example: {'ASL_1' 'ASL_2'}
+%             - example ('.json' file): ["ASL_1","ASL_2"]
+%             - example ('.m' file):    {'ASL_1' 'ASL_2'}
 %             - Specific options: for FEAST: 1=crushed, 2=not crushed. This used to be other way around, 
 %               but the crushed image registers better with the pGM image
 % x.session.options - this is how the sessions will be called (OPTIONAL)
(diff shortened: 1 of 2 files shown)


MichaelStritt:
@HenkMutsaerts: I think we came to an agreement within the issue 👍

DataParTemplate.m
```diff
@@ -48,7 +48,8 @@
 % x.exclusion - cell with list of subjects to exclude (OPTIONAL, DEFAULT = empty)
 %             - example: {'005' '018'}
 % x.SESSIONS  - use this to define sessions (OPTIONAL, DEFAULT = {'ASL_1'})
-%             - example: {'ASL_1' 'ASL_2'}
+%             - example ('.json' file): ["ASL_1","ASL_2"]
+%             - example ('.m' file):    {'ASL_1' 'ASL_2'}
```
HenkMutsaerts, line 52: Nice, yes this should actually fase out slowly but steadily :)

ExploreASL_Initialize.m
```diff
@@ -126,7 +126,7 @@
     end
     
     % Read file
-    if exist(x.DataParPath,'file')
+    if exist(x.DataParPath,'file')==2
```
HenkMutsaerts, line 129: No need to add the `==2`, that is what the `'file'` does already.
I would remove the `==2` but perhaps it doesn't harm?

HenkMutsaerts (approved):

ExploreASL_Initialize.m
MichaelStritt: I prefer the option with `==2` because ...
```matlab
exist(fullfile(server,'Server_xASL','ExploreASL','ExploreASL.m'),'file')==2
```
is `TRUE`, whereas ...
```matlab
exist(fullfile(server,'Server_xASL','ExploreASL','ExploreASL.m'),'file')
```
is `2`. And within `if-else`-statements, it kind of makes sense to my to ask...
```matlab
if (TRUE)
    then do this
else
    do this
end
```
... instead of ...
```matlab
if (2)
    then do this
else
    do this
end
```

PR #498, opened 2021-04-15
Bug #496 xASL_adm_GetPopulationSessions

### LINKED ISSUE #496: Population module calculates ROI's only for the total amount of subjects
**Description**
The Population module calculates ROI values for the total amount of subjects present in the data, but does not include also the total amount of sessions. For example, in a dataset of 90 subjects and 10 sessions each, output of the Population Stats folder CBF values for each ROI is limited to 90 scans, in our case the result is only of subjects 1-9 sessions 1-10 instead of subjects 1-90 sessions 1-10 (see /lood_storage/divi/Projects/ExploreASL/ABBA/NOMARED/derivatives/Stats_not_all_subjects/).

**Requirements**
The Population module should calculate the total amount of subjects and sessions in the dataset or provided in the DataPar.json and calculate the ROIs provided for those subjects and sessions accordingly.

**How to test**
Run the Population module for a single subject with multiple sessions, correct output should provide ROIs for all sessions instead of just one.

The directory I'm working in to test this issue is /lood_storage/divi/Projects/ExploreASL/ABBA/NOMARED/derivatives/Test_vascular_territories/. This directory contains only one subject but also has as output of the population module stats only ASL_1 instead of all 10 sessions.

the original command I run is:

``ExploreASL_Master('DataPar_loc',[0 0 0 0], [0 0 1], 0, 1, 1)``

**Release notes**
Fixed defining correct nSessions for each subject.
Fixed exporting correct nSession ROI statistics for each subject.

MDijsselhof:
closes #496


Functions/xASL_stat_GetROIstatistics.m
```diff
@@ -177,16 +177,12 @@
 end
 
 %% Define number of sessions to use
-nSessions = 0;
 NoSessions = 0;
-SessionList = xASL_adm_GetFileList(x.D.PopDir,['^' x.S.InputDataStr '.*_ASL_' num2str(nSessions+1) '\.nii'], 'FPList', [0 Inf]);
+SessionList = xASL_adm_GetFileList(x.D.PopDir,['^' x.S.InputDataStr '.*_ASL_\d*\.nii'], 'FPList', [0 Inf]);
+MaskedSessions = cellfun('isempty',strfind(SessionList,'masked')); % find qCBF_masked
```
jan-petr, line 182: Code is fine, but you might add comments or change variables as it now looks a bit as if you include CBF_masked and not exclude. 

Also, I would look for '_masked_' and not 'masked' - you want to avoid finding 'masked' in patient name or something. You might probably even look for ['^' x.S.InputDataStr '_masked'] with regexp - as above - this works now, but we want to avoid future errors.

Functions/xASL_stat_GetROIstatistics.m
```diff
@@ -195,6 +191,8 @@
     NoSessions = 1;
 end
 
+nSessions = numel(SessionList_qCBF); % define number of sessions
```
jan-petr, line 194: nSessions can be left here only and removed from line 190

jan-petr (changes requested):
Looks good - check my two minor comments.

Functions/xASL_stat_GetROIstatistics.m
MDijsselhof: I think it needs to stay, otherwise nSessions doesnt get defined because `` nSessions = numel(SessionList_qCBF) `` would be set to 0 and is required for the FOR loop later on. See:

image

jan-petr: Sorry - that was confusing. I can only enter comments on the lines you have modified and not on other lines. This line 194 is OK and it indeed has to stay it this way. But a few lines above - line 190, the nSessions is never used as it is overwritten on line 194 in all cases.

MDijsselhof: Okay I'll change it!

MDijsselhof: I've change it to make it a bit more robust (and testing now):

image

Otherwise SessionsList remains empty and using ``cellfun`` and ``numel`` might give an error.

MDijsselhof: Running into a new problem

image

MDijsselhof: Okay multiple steps are going wrong. To inventarise: 
- Defining sessions by looking at qCBF.nii's is only possible if we differentiate between subject number and sessions number
- Length of x.S.SUBJECTID goes wrong somewhere

MDijsselhof:
@jan-petr I've committed and pushed the latest version, I've included a warning if the amount of sessions is not similar for each subjects. Please have a look!

jan-petr (approved):
OK.

Functions/xASL_stat_PrintStats.m
```diff
@@ -109,15 +109,26 @@
...
+HasSingleSessionOnly = sum(cellfun(@(y) ~isempty(regexp(y,['^' x.S.output_ID])), SingleSessions));
 
 if HasSingleSessionOnly
     nSessions = 1;
 else
-    FileList = xASL_adm_GetFileList(x.D.PopDir,'.*ASL_\d\.nii$');
-    [~, Ffile] = cellfun(@(x) xASL_fileparts(x),FileList, 'UniformOutput',false);
-    SessionsN = unique(cellfun(@(x) x(end-4:end),Ffile, 'UniformOutput',false));
-    nSessions = length(SessionsN);
+    SessionList = xASL_adm_GetFileList(x.D.PopDir,['^' x.S.InputDataStr '.*_ASL_\d*\.nii'], 'FPList', [0 Inf]);
```
HenkMutsaerts, line 117: This is the same code in `xASL_stat_GetROIstatistics` right? That would break the modularity of the code. The idea is that `xASL_stat_PrintStats` only prints stuff that it gets from another function/wrapper, whereas e.g. `xASL_stat_GetROIstatistics` or `xASL_wrp_GetROIstatistics` is responsible for loading the data. Should be clear in the header, otherwise perhaps I need to make this clear ;)

HenkMutsaerts (changes requested):
Very nice suggestions but they break the modularity of the code by hardcoding some stuff that would be nice to keep relative (e.g. allowing different filetypes than only `qCBF` as input). Let me know if we need to discuss this, or otherwise you can think how to change this that it doesn't break the modularity

Functions/xASL_stat_GetROIstatistics.m
```diff
@@ -177,22 +176,30 @@
...
+    bSessionsMissing = 1;
+else % continu with defining sessions from SessionList
+	% find sessions that don't have qCBF_masked, so that _masked can be excluded from SessionList
+    nonMaskedSessions = cellfun('isempty',strfind(SessionList,[ x.S.InputDataStr '_masked'])); 
+    SessionListSUBJECTS_qCBF = SessionList(nonMaskedSessions,:); % include only qCBF in SessionList and exclude _masked
+    for n = 1:size(x.SUBJECTS,2)
+        SessionsSingleSUBJECT = ~cellfun('isempty',strfind(SessionListSUBJECTS_qCBF,[x.SUBJECTS{n}])); % Check how many qCBF files are present of each subject
+        SessionsPerSubject(n,1) = sum(SessionsSingleSUBJECT); % Define number of sessions as amount of qCBF files per subject
+    end
+    nSessions = min(SessionsPerSubject); % Use minimum amount of sessions for processing
```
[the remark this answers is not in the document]
jan-petr: Good point. We wanted a solution that always works - hence the `min`. But if `max` won't crash the code and only deliver a lot of NaNs, then we can use this as a solution before we implement the option that does it individually for each subject. 

@MDijsselhof Can you please test it for max. We will now need it for the Oslo study that have some missing sessions - so MAX solution will save us some emails.

Functions/xASL_stat_GetROIstatistics.m
```diff
@@ -177,22 +176,30 @@
 end
 
 %% Define number of sessions to use
```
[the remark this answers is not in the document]
MDijsselhof: Which sessions may be the technically correct case, however in the following update it will determine the number of sessions per subject individually so it would be change back. I suggest leaving it like this.

Functions/xASL_stat_GetROIstatistics.m
```diff
@@ -177,22 +176,30 @@
...
         fprintf('%s\n','No session or subject files found');
         return;
     end
     nSessions = 1;
-    NoSessions = 1;
+    bSessionsMissing = 1;
+else % continu with defining sessions from SessionList
+	% find sessions that don't have qCBF_masked, so that _masked can be excluded from SessionList
+    nonMaskedSessions = cellfun('isempty',strfind(SessionList,[ x.S.InputDataStr '_masked'])); 
+    SessionListSUBJECTS_qCBF = SessionList(nonMaskedSessions,:); % include only qCBF in SessionList and exclude _masked
```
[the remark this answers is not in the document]
MDijsselhof: Agreed

Functions/xASL_stat_GetROIstatistics.m
```diff
@@ -177,22 +176,30 @@
...
+if isempty(SessionList) % search for subjects instead of sessions
     if isempty(xASL_adm_GetFileList(x.D.PopDir,['^' x.S.InputDataStr '.*\.nii'], 'FPList', [0 Inf]))
         fprintf('%s\n','No session or subject files found');
         return;
     end
     nSessions = 1;
-    NoSessions = 1;
+    bSessionsMissing = 1;
+else % continu with defining sessions from SessionList
+	% find sessions that don't have qCBF_masked, so that _masked can be excluded from SessionList
```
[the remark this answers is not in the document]
MDijsselhof: Agreed

Functions/xASL_stat_GetROIstatistics.m
```diff
@@ -177,22 +176,30 @@
...
+    nonMaskedSessions = cellfun('isempty',strfind(SessionList,[ x.S.InputDataStr '_masked'])); 
+    SessionListSUBJECTS_qCBF = SessionList(nonMaskedSessions,:); % include only qCBF in SessionList and exclude _masked
+    for n = 1:size(x.SUBJECTS,2)
```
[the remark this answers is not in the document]
MDijsselhof: Agreed

Functions/xASL_stat_PrintStats.m
```diff
@@ -109,15 +109,26 @@
 %% 3) Define number of sessions, force to 1 in case of TT or volume metrics
 SingleSessions = {'volume' 'TT' 'PV_pGM' 'PV_pWM' 'pGM' 'pWM' 'mrc1T1' 'mrc2T1'};
 
-HasSingleSessionOnly = sum(cellfun(@(y) ~isempty(findstr(y,x.S.output_ID)), SingleSessions));
+HasSingleSessionOnly = sum(cellfun(@(y) ~isempty(regexp(y,['^' x.S.output_ID])), SingleSessions));
```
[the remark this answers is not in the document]
MDijsselhof: Does this work with previous Matlab versions (e.g. 2016b and before)

Functions/xASL_stat_PrintStats.m
MDijsselhof: Okay so we want to create a new cell in the x structure that feeds into xASL_stat_PrintStats?

Functions/xASL_stat_PrintStats.m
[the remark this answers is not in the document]
MDijsselhof: Nevermind, I see this has been introduced in 2006a

Functions/xASL_stat_GetROIstatistics.m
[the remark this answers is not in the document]
MDijsselhof: It works (I've removed ASL_3 from NMRD_090)! I'll make a commit with all the suggested changes.

image

Functions/xASL_stat_PrintStats.m
HenkMutsaerts: Is all this here only to remove the NIfTIs with suffixes `_masked`? There's an easier solution for that.
Isn't `_masked` used somewhere? `qCBF_masked` have vascular peaks replaced by NaNs, so it will limit the impact of vascular outliers/artifacts on regional CBF estimates. For sCoV you don't want this but for CBF this can be helpful.

Functions/xASL_stat_GetROIstatistics.m
```diff
@@ -177,22 +176,30 @@
...
 %% Define number of sessions to use
-nSessions = 0;
-NoSessions = 0;
-SessionList = xASL_adm_GetFileList(x.D.PopDir,['^' x.S.InputDataStr '.*_ASL_' num2str(nSessions+1) '\.nii'], 'FPList', [0 Inf]);
-
-while ~isempty(SessionList)
-       nSessions = nSessions+1;
-       SessionList = xASL_adm_GetFileList(x.D.PopDir,['^' x.S.InputDataStr '.*_ASL_' num2str(nSessions+1) '\.nii'], 'FPList', [0 Inf]);
-end
+SessionList = xASL_adm_GetFileList(x.D.PopDir,['^' x.S.InputDataStr '.*_ASL_\d*\.nii'], 'FPList', [0 Inf]);
```
HenkMutsaerts, line 179: So if you want to exclude `_masked` you could try something like:

`SessionList = xASL_adm_GetFileList(x.D.PopDir,['^' x.S.InputDataStr '.*(?<!_masked)_ASL_\d*\.nii'], 'FPList', [0 Inf]);`

Per this explanation, check if it should be the lookahead or negative lookahead:

(?! ) is a lookahead (do not match the symbols in the space here that are to the RIGHT)
(?<! ) is a negative lookahead (do not match the symbols in the space here that are to the LEFT)

HenkMutsaerts (changes requested):

Functions/xASL_stat_GetROIstatistics.m
MDijsselhof: Can't seem to get it to work

Functions/xASL_stat_PrintStats.m
MDijsselhof: So we should look for `_masked` instead? If for some reason there is no `_masked` then the code will show NaN's, there should also be a warning then?

Functions/xASL_adm_GetnSessions.m
```diff
@@ -0,0 +1,57 @@
...
+    end
+    nSessions = 1;
+    bSessionsMissing = 1;
+else % If files found, continu with defining sessions from SessionList
+    [IndexStart , ~] = cellfun(@(y) regexp(y,'ASL_\d+\.(nii|nii\.gz)'), SessionList);
+    
+    for n = 1:numel(SessionList)
+        StrLoc = IndexStart(n);
+        SessionsPopFolder = SessionList{n};
+        NewList{n,1} = SessionsPopFolder(1,StrLoc:end); % create a cell containings strings of all session numbers present
```
HenkMutsaerts, line 39: should be cell array` and `chararacters` instead of `cell` and `strings` in the comment.

Functions/xASL_adm_GetnSessions.m
```diff
@@ -0,0 +1,57 @@
...
+    end
+    
+    UniqueSessions = unique(cell2mat(NewList),'rows'); % determine unique session numbers
+    nSessions = size(UniqueSessions,1); % define nSessions as maximum amount of unique session numbers
+    bSessionsMissing = 0;
+    NewList = cellstr(NewList);
+    for o = 1:nSessions
```
HenkMutsaerts, line 46: same here, use an interpretable index instead of `o`

Functions/xASL_adm_GetnSessions.m
```diff
@@ -0,0 +1,57 @@
...
+        CountSessionNumbers(o)= sum(~cellfun('isempty',strfind(NewList,num2str(o))));
```
HenkMutsaerts, line 47: `xASL_num2str` sometimes works better than `num2str` (just saying that it exists, you don't need to use it here)

Functions/xASL_adm_GetnSessions.m
```diff
@@ -0,0 +1,57 @@
...
+    end
+    
+    CompareSessions = ones(1,size(CountSessionNumbers,2)) .* nSessions; % create an array to check differences in sessions per subject with maximum amount of sessions
+    
+    if ~isequal(CountSessionNumbers,CompareSessions) % Check if amount of sessions is similar for each subject and provide warning if not
+        warning('Amount of Sessions per Subject different')
```
HenkMutsaerts, line 53: Can you make this warning easier to understand? E.g., "Amount of sessions differs between subjects"?

And you miss a `;`

Functions/xASL_adm_GetnSessions.m
```diff
@@ -0,0 +1,57 @@
+function [nSessions, bSessionsMissing] = xASL_adm_GetnSessions(x)
```
HenkMutsaerts, line 1: wonder if we should call this `xASL_adm_GetPopulationSessions` to clarify that this is for the population module/folder only

HenkMutsaerts (changes requested):
Nice work! Few minor remarks before we can merge this.
Did you test this?

Functions/xASL_adm_GetnSessions.m
```diff
@@ -0,0 +1,57 @@
...
+end
+if strcmp(currentSubjectRegExp(end), '$')
+    currentSubjectRegExp = currentSubjectRegExp(1:end-1);
+end
+% Look for processed files from which to determine the amount of sessions present
+SessionList = xASL_adm_GetFileList(x.D.PopDir,['^' x.S.InputDataStr '_' currentSubjectRegExp '_ASL_\d*\.nii$'], 'FPList', [0 Inf]);
+
+if isempty(SessionList) % If no files found, search for subjects instead of sessions
+    if isempty(xASL_adm_GetFileList(x.D.PopDir,'^qCBF.*\.nii', 'FPList', [0 Inf]))
+        fprintf('%s\n','No session or subject files found');
```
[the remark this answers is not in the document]
MDijsselhof: I don't follow, do you want to look for subject files outside of this function when identifying the amount of sessions?

Functions/xASL_adm_GetnSessions.m
MDijsselhof: Agreed

HenkMutsaerts (changes requested):
Nice!

You forgot to change the function name throughout, and you didn't delete the `xASL_adm_GetnSessions` function.

Next time, in this example, it would have been nice if you do:
`git mv xASL_adm_GetnSessions xASL_adm_GetPopulationSessions` as this way Github remembers that you have only changed the function name, rather than creating a new function (easier for reviewing).

Functions/xASL_stat_GetROIstatistics.m
```diff
@@ -243,13 +227,14 @@
...
 for iSubject=1:x.nSubjects
+    
 	for iSess=1:nSessions
         
 		% ID (which name, group etc), all for identification
 		% Subject_session definition
 		DataIm = NaN;
 		SubjSess = (iSubject-1)* nSessions +iSess;
-		if NoSessions
+		if bSessionsMissing == 1
```
HenkMutsaerts, line 237: The ` == 1` is superfluous

HenkMutsaerts (approved):
Looks good, minor suggestion. Did you test this already?
Fine for merging then!

Functions/xASL_stat_GetROIstatistics.m
MDijsselhof: I'll change it and merge! I've tested it already and it works.

PR #513, opened 2021-04-29
Closes #466 Fix illegal json

### LINKED ISSUE #466: Create an error when JSON has illegal format
When the DataPat.json has an illegal format, it crashes Matlab instead of giving an error. For example inside the json, if we have "Quality"=1, instead of "Quality":1, Matlab crashes when running ExploreASL_Master.
So a good idea is to create an understandable error if the JSON has some problems, instead of just crashing the program.

MichaelStritt:
Check out #466


jan-petr:
@MichaelStritt Have you tested this a bit? We should check if running the whole thing - DCM->BIDS->Legacy->Processing works on 1-2 datasets. I'll try Linux.

jan-petr (approved):
Full test - import to result works for Linux on two random datasets. Please do a very basic test on Windows as well.

PR #520, opened 2021-04-29
Closes #514 ordering conversion

### LINKED ISSUE #514: Correct ordering of DICOMs during conversion
### Idea

- Add %r to `io_dcm2nii` in line 132 👍 
- dcm2nii_args = '-f "%f_%p_%t_%s_%r"'; 👍 
- Extract the instance number. 👍 
- Sort renamed ASL by instance number 👍 
- return initial instance numbers for all converted NIFTIs 👍 

xASL_imp_DCM2NII_Subject
- Remove functions xASL_imp_DCM2NII_Subject_RealASLContext and xASL_imp_DCM2NII_Subject_SwitchNIIs 👍 

xASL_imp_DCM2NII_Subject_StoreJSON
- Give paths and instance numbers to all creates JSONs
- Make sure that the extracted information is saved to JSON based on instance numbers
- Create the outputs separately for each JSON - change parms to a cell - each cell for each JSON, write information to parms based on the range of instanceNumber.
- We have to do it in one run, so that we read DICOMs only once.
- This completes the JSON information for each separate NIFTI

xASL_imp_DCM2NII_Subject_ShuffleTheDynamics
- We move this and run it only after xASL_imp_DCM2NII_Subject_StoreJSON, that way, we have more information for the merging of files.
- We have to make sure that the JSON fields are correctly as this was all done previously in StoreJSON.

MichaelStritt:
Hey Jan,

I know we originally planned to move `xASL_imp_DCM2NII_Subject_StoreJSON` and modify the part about the InstanceNumbers a bit differently, but this kind of didn't work because of the missing DICOM match parameter. I kind of found another solution, but if it's totally stupid we can still revert this. What I did was:

- Use the correct format to extract the InstanceNumbers in `xASL_io_dcm2nii`.
- I changed the code part regarding the renaming of the NIFTI & JSON files.
- I automatically store the (usually two) InstanceNumbers in the corresponding JSONs.
- Then I reverted my changes in `xASL_bids_Dicom2JSON`, because we don't need these separate ImageTypes and InstanceNumbers arrays.
- I implemented most of the logic within the `xASL_imp_DCM2NII_Subject_ShuffleTheDynamics` subfunction. I create a table called niiTable. In the table we have the InstanceNumbers and our corresponding `deltam` and `m0scan` values. I sort the table and then extract the correct NIFTI order.
- If I give the correct order to `xASL_bids_MergeNifti`, the ordering should be correct later on, right?


jan-petr (approved):
It works perfectly on Linux also with the corrected BIDScomparison function.

You can run a last check on Windows to verify if everything - including GE order - is fine with the new code and new flavors.

MichaelStritt:
Will do first thing in the morning. The windows matlab server is still down right now.

MichaelStritt:
## GE Tests

[logContent.xlsx](https://github.com/ExploreASL/ExploreASL/files/6427630/logContent.xlsx)

### GE_PCASL_2DEPI_volunteer
```
Warning: The following user defined and DICOM fields differ:
  SliceTiming TotalAcquiredPairs
> In xASL_imp_NII2BIDS_SubjectSessionRun (line 125)
```

👍

MichaelStritt:
## Philips

[logContent.xlsx](https://github.com/ExploreASL/ExploreASL/files/6429160/logContent.xlsx)

👍

jan-petr:
> * Replacing the x-struct SliceReadoutTime with the one from
> * TE of ASL and M0 were unequal! Check geometric distortion
> * Overwriting x.Q.LabelingEfficiency=0.6889, with x.LabelingEfficiency=0.83
> * Registration off, spatial CoV detection unreliable
> * WM T1 set to 900 ms for 3T
> * Multiple PLDs detected, selecting maximal value

These warnings are OK for now. Need to be addressed in the future, but OK to merge.

PR #524, opened 2021-04-30
Closes #523 ExploreASL_Master & ExploreASL_ImportConfig

### LINKED ISSUE #523: Add warning & URL
Added warning and URL. Really minor design update. Check out the examples below.

MichaelStritt:
Design update and added warning

Check out #523


ExploreASL_ImportConfig.m
```diff
@@ -21,6 +21,13 @@
 % Initialize ExploreASL
 [x] = ExploreASL_Initialize;
 
+% Print warning
+fprintf('\n========================================== WARNING ===========================================\n');
+fprintf(['\nThe support for this ExploreASL functionality will expire with a future release.\n'...
+         'If you want to be able to use this function within a future release, please pull\n'...
```
HenkMutsaerts, line 27: Nice, though not sure if this is sufficiently clear to the user ;) I would add that this is taken over by the BIDS import support straightly through the `ExploreASL_Master` function. Also, what should the user do with the CustomScripts once pulled? You don't need to put a manual here of course :) but it's good if there some minor guidance

ExploreASL_Master.m
```diff
@@ -160,7 +160,8 @@
 
     % -----------------------------------------------------------------------------    
     %% Finishing touch
-    fprintf('Many thanks for using ExploreASL, please don''t forget to cite https://pubmed.ncbi.nlm.nih.gov/32526385/\n');
+    fprintf('Many thanks for using <a href="https://github.com/ExploreASL" rel="nofollow">ExploreASL</a>, ');
+    fprintf('please don''t forget to cite <a href="https://pubmed.ncbi.nlm.nih.gov/32526385/" rel="nofollow">https://pubmed.ncbi.nlm.nih.gov/32526385/</a>.\n');
```
HenkMutsaerts, line 164: This works very nice!

HenkMutsaerts (changes requested):
I would add some minor explanation to the warning, otherwise great!

ExploreASL_Master.m
MichaelStritt: Thanks 👍  I don't want to add links everywhere, because they can be a bit ugly in the log files, but I thought here it could make sense.

ExploreASL_ImportConfig.m
MichaelStritt: Good idea, but let's change the wording a bit. `ExploreASL_Master` doesn't create the `imPar`, what `ImportConfig` basically does right now. So we have to say that these settings will be part of the BIDS compliant `sourceStructure.json` file, that can be read using `ExploreASL_Master` directly, right?

MichaelStritt: This should say everything:

image

HenkMutsaerts: Very nice!

HenkMutsaerts (approved):

PR #526, opened 2021-04-30
Closes #474 splitting jsons

### LINKED ISSUE #474: Splitting ASL split also JSONs
The function xASL_io_SplitASL back-ups the ASL4D to ASL4D_Source and splits it to ASL and M0, also removing dummy scans from ASL.

The same logic has to be applied also to certain JSON fields of ASL4D that potentially also need to be split. 

The following fields need to be checked:
EchoTime, FlipAngle, RepetitionTimePreparation, PostLabelingDelay, VascularCrushingVENC, LabelingDuration

1. If these are a scalar, just copy the scalar. 
2. If these are a vector, they should always have the same length as the sequence (not having it is a major error in JSON) - we split them to ASL and M0 using the same indexes as done for the NIfTIes. 
3. Note that the fields don't have to be always present.If missing simply skip the field.

It would be good to load the list of fields to check from xASL_bids_Config() - like we did for dataset description. You can create a list called something like FieldsNumberArray - see if you can come up with some reasonable name that fits with the rest of the config file.

Note that for most datasets, we usually provide a scalar only. Still, this is a critical bug for the BIDS->Legacy for some datasets - like for example your DRO, without this, it will have incorrect TR for M0 and ASL and the quantification will be off.

MichaelStritt:
Closes #474

### DIFF
--- a/Functions/xASL_io_SplitASL.m
+++ b/Functions/xASL_io_SplitASL.m
@@ -198,8 +198,8 @@ function xASL_io_SplitASL(inPath, iM0, iDummy)
         xASL_io_SaveNifti(BackupName,ASLname,tIM(:,:,:,ASLindices),[],false);
         
         %% 7. Split relevant JSON parameters/arrays
-        [jsonM0, jsonASL] = xASL_io_SplitASL_SplitJSON(BackupJSONPath);
-
+        [jsonM0, jsonASL] = xASL_io_SplitASL_SplitJSON(BackupJSONPath, iM0, ASLindices);
+        
         %% 8. Copy sidecars
         if exist(BackupMATPath,'file') && bCreateM0
             xASL_Copy(BackupMATPath, M0MATPath);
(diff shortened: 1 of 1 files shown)


Functions/xASL_io_SplitASL.m
```diff
@@ -184,22 +194,81 @@ function xASL_io_SplitASL(inPath, iM0, iDummy)
...
+        
+        % Iterate over JSON fields
+        jsonFields = fieldnames(jsonStruct);
+        for iField = 1:numel(jsonFields)
+            % Check if fieldname is in splitFields list
+            if ismember(jsonFields{iField,1},splitFields)
+                % Check if field has to be split
+                if numel(jsonStruct.(jsonFields{iField,1}))>1
+                    % Compare array length
+                    if (numel(indicesM0)+numel(indicesASL))==numel(jsonStruct.(jsonFields{iField,1}))
```
jan-petr, line 252: Good to have this condition, but we have to adapt it, because we have incides for ASL, M0, and Dummy - these three together give the full-set.

jan-petr: Thinking about how to address this, but adding the Dummy indices (dynamics to be thrown out) to the call of this function and expanding this equation here with numel(indicesDummy) seems like the best solution.

jan-petr (changes requested):
Very good work! It looks complete and correct. One minor thing - there's an option to throw away some dynamics (Dummy dynamics) - we have to take this into account - see my comment in the code.

Functions/xASL_io_SplitASL.m
MichaelStritt: That's an easy fix :)

jan-petr: OK, are you doing it, because I have realized that I can fix it and then quickly test it at the same time :)

jan-petr (approved):
Works perfectly for all cases now.

PR #537, opened 2021-05-03
Closes #536 xASL_adm_DocInitialize: Copy additional template

### LINKED ISSUE #536: QC Walkthrough (markdown)
Create a first version of the QC walkthrough document in markdown

MichaelStritt:
Check out #536


HenkMutsaerts (changes requested):
Nice but shouldn't this function be in the documentation subfolder or repository?

Also you could use the ExploreASL functions for copying if they are more platform independent /efficient

MichaelStritt:
> 
> 
> Nice but shouldn't this function be in the documentation subfolder or repository?
> 
> Also you could use the ExploreASL functions for copying if they are more platform independent /efficient

The script calls the DocCrawler and other ExploreASL functionality. I would keep it as it is right now. It rarely needs an update and it's just nice that it works just by hitting play right now...

MichaelStritt:
> Also you could use the ExploreASL functions for copying if they are more platform independent /efficient

While xASL_Copy may be quicker for NIfTIs and a lot of other stuff, here it really isn't necessary. It's copying markdown files with a few KB. The paths are all **relative**, so this shouldn't be a problem.

HenkMutsaerts (approved):

PR #539, opened 2021-05-03
#538 xASL_im_ReplaceLabel

### LINKED ISSUE #538: xASL_im_ReplaceLabel
### Release Notes
New function for the user to replace label values in a single 3D NIfTI volume, written for the CICERO study analysis

HenkMutsaerts:
Closes #538


Functions/xASL_im_ReplaceLabel.m
```diff
@@ -0,0 +1,100 @@
...
+elseif nargin<1 || isempty(pathNifti)
+    error('pathNifti should be defined');
+elseif ~ischar(pathNifti)
+    error('pathNifti should be char array, a path to the NIfTI file');
+elseif ~xASL_exist(pathNifti)
+    error([pathNifti ' does not exist']);
+end
+
+if nargin<4 || isempty(pathNewNifti)
+    warning(['Overwriting ' pathNifti]);
```
jan-petr, line 60: I would not print that as a warning but as a normal message. We should ideally have minimal warnings and this one only hints to something that the user did on purpose, so he can't get rid of this warning.

Functions/xASL_im_ReplaceLabel.m
```diff
@@ -0,0 +1,100 @@
...
+    pathNewNifti = pathNifti;
+end
+
+if size(LabelNumbersOld, 1)~=1
+    error('LabelNumbersOld should be a row, not a column');
```
jan-petr, line 65: I would not throw an error, but rearrange the vector and throw a warning (and not even that).

Functions/xASL_im_ReplaceLabel.m
```diff
@@ -0,0 +1,100 @@
...
+elseif size(LabelNumbersNew, 1)~=1
```
jan-petr, line 66: Same here

Functions/xASL_im_ReplaceLabel.m
```diff
@@ -0,0 +1,100 @@
...
+
+
+
+%% -------------------
+%% 2. Replace numbers
+for iNumber=1:length(LabelNumbersOld)
+    if sum(Image==LabelNumbersOld(iNumber))==0
+        warning(['Number ' LabelNumbersOld(iNumber) ' could not be found, skipping!'])
+    else
+        Image(Image==LabelNumbersOld(iNumber)) = LabelNumbersNew(iNumber);
```
jan-petr, line 90: You can't do it like this. You need to have ImageOld and ImageNew otherwise you'll be facing potential conflicts. 

For example:
oldLabels = [1 2 3 4 5 6]
newLabels=[2 3 4 5 6 7] 
will, in the current implementation, not increase the label number by 1 but repaint everything to 7...

jan-petr (changes requested):
Several things that need to be improved:
1. Unclear what this function does from the header. It replaces the labes on ROIs or atlas, right? This has to be clarified both in the function name, header, description etc.
2. You can't have the same image for new labels and old labels - see the specific comments in the code.

Functions/xASL_im_ReplaceLabel.m
BeatrizPadrela: so instead of warning it can just be fprintf?

jan-petr (approved):
The additional commits solve my issues.

HenkMutsaerts:
@BeatrizPadrela you can also merge it if you are OK with this PR!

BeatrizPadrela (approved):

PR #540, opened 2021-05-04
Closes #535 update bids comparison

### LINKED ISSUE #535: Skipping missing files in BIDS folder comparison
When comparing BIDS folders, it does not report files/directories that are in Reference but missing in Source. I.e. if the BIDS conversion completely fails and doesn't write anything, it is not reported on screen.

MichaelStritt:
@jan-petr: Can you give this a try and see if it works now?
It doesn't find empty folders right now, so if you know a good regexp that works with these ExploreASL functions then just let me know.


MichaelStritt:
@jan-petr: could you run a really minor test? mediri server is down right now, otherwise I'd test myself.

jan-petr (approved):
Works as it should now.

PR #544, opened 2021-05-06
Closes #543 x asl num2str auto mode

### LINKED ISSUE #543: spm_jsonwrite bug (integer numbers)
I encountered a minor bug in `spm_jsonwrite`.

Running `spm_jsonwrite` with the struct on the left, lead to the following JSON file.

image

The field `"PostLabelingDelay": 2.,` is invalid though. It should either be `2` or `2.0` and not `2.`.

This happens in `spm_jsonwrite` line 61:

```
S = jsonwrite_var(json,~isempty(opts.indent));
```

Before: 

```
json = 
  struct with fields:
             LabelingDuration: 1.8000
           LabelingEfficiency: 0.8500
            PostLabelingDelay: 2.0000
           ...
```

After:

```
S =
    '{"LabelingDuration":1.8,
     "LabelingEfficiency":0.85,
     "PostLabelingDelay":2.,
     ...
```

I think I introduced this :/ 
In line 229 we call `S = xASL_num2str(json, 'auto');`

MichaelStritt:
Check out #543

### DIFF
--- a/Functions/xASL_num2str.m
+++ b/Functions/xASL_num2str.m
@@ -85,7 +85,7 @@ if isnumeric(DataIn)
                 end
             end
             % Fix perfect integers (previously 2.0000 was converted to 2. instead of 2.0)
-            if strcmp(DataOut,'.')
+            if strcmp(DataOut(end),'.')
                 DataOut = [DataOut '0'];
             end
         end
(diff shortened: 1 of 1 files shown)


jan-petr (approved):
Good catch.

PR #550, opened 2021-05-07
Closes #542 ASL context

### LINKED ISSUE #542: ASL split handling of ASLcontext.tsv
ASL4D_aslContext.tsv seems to be ignored during ASL splitting. This is fine as this information is duplicit and should also be present in the JSONs in the Legacy files. However, when the ASL4D is split and saved to ASL4D_Source, then the aslContext.tsv should also be renamed to _source as it matches to the original and not the split version anymore.

Also, for maximum clarity and transparency - after splitting, the field `"M0Type":"Included",` is correctly changed to 
`"M0Type":"Separate"` for ASL, however, it incorrectly stays `"M0Type":"Included",` for M0 where such field shouldn't be defined. Moreover, we have functions that check the format of all fields according to the BIDS format. It is true that this is the legacy format, but there is no reason, why we shouldn't make sure that the JSON is in perfect BIDS already.

1. Make sure that aslContext.tsv is renamed to _Source.
2. Make sure that on re-run, aslContext is taken from the _Source version.
3. Remove the field `M0Type` from the M0.json after splitting.
4. Call functions ` jsonM0Write = xASL_bids_VendorFieldCheck(jsonM0Write);` and `jsonM0Write = xASL_bids_JsonCheck(jsonM0Write,'M0');` before saving the split M0.json to check the fields.
5. Assign the correct IntendedFor field in M0.json referencing the ASL file - again, not need, but good to still BIDS compliant.

### Release note
Split ASL to ASL and M0 - archive aslContext.tsv and correct M0.json - to be merged with the similar issue for release notes.

MichaelStritt:
Check out #542 (& WhatsApp questions)

### DIFF
--- a/Functions/xASL_io_SplitASL.m
+++ b/Functions/xASL_io_SplitASL.m
@@ -80,26 +80,7 @@ function xASL_io_SplitASL(inPath, iM0, iDummy)
 	%% 2. Prepare paths
     [Fpath, Ffile] = xASL_fileparts(inPath);
     [paths] = xASL_io_SplitASL_GetPaths(Fpath, Ffile);
-    
-    ASLlist = xASL_adm_GetFileList(Fpath, ['^' Ffile '(|_.)'  '(|_\d)' '\.nii$'], 'FPList', [0 Inf]);
-  
-    if xASL_exist(paths.ASL_Source,'file')
-        fprintf('Restoring ASL4D.nii from backup NIfTI ASL4D_Source.nii\n');
-        ASLlist{1} = paths.ASL;
-        xASL_Move(paths.ASL_Source, ASLlist{1}, true);
-        % Restore json as well
-        if exist(paths.ASLJSON,'file') && exist(paths.ASL_Source_JSON,'file')
-            xASL_Move(paths.ASL_Source_JSON, paths.ASLJSON, true);
-        end
-        % Delete M0, allowing to restore this from the backup in case M0 is in the ASL timeseries
-        if xASL_exist(paths.M0,'file') && ~isempty(iM0)
-            fprintf('Deleting M0 to restore it from ASL4D\n');
-            xASL_delete(paths.M0);
-            xASL_delete(paths.M0JSON);
-        end
-    elseif isempty(ASLlist)
-        error([paths.ASL ' didnt exist, skipping']);
-    end
+    [ASLlist] = xASL_io_SplitASL_RestoringFromBackup(paths);
     
     if ~xASL_exist(paths.ASL_Source,'file') % otherwise was already split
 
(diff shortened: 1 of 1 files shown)


Functions/xASL_io_SplitASL.m
```diff
@@ -199,15 +203,28 @@ function xASL_io_SplitASL(inPath, iM0, iDummy)
         
         %% 7. Split relevant JSON parameters/arrays
         [jsonM0, jsonASL] = xASL_io_SplitASL_SplitJSON(BackupJSONPath, iM0, ASLindices, iDummy);
+
+        %% 8. Split ASL4D_aslContext.tsv
+        if xASL_exist(ASLTSVContextPath,'file') && (~isempty(iM0) || ~isempty(iDummy))
+            ASLTSVContextPathSource = strrep(ASLTSVContextPath, 'aslcontext', 'aslcontext_Source');
```
jan-petr, line 209: Can you change it to `_Source_aslcontext` - to be in sync with the BackupMATPath - see line 95.

Functions/xASL_io_SplitASL.m
```diff
@@ -218,6 +235,30 @@ function xASL_io_SplitASL(inPath, iM0, iDummy)
...
+        [~, file, extension] = xASL_fileparts(ASLname);
+        jsonM0.IntendedFor = ['perf/' file extension];
+        
+        % Remove M0Type field of M0 JSON if it still exists
+        if isfield(jsonM0, 'M0Type')
+            jsonM0 = rmfield(jsonM0, 'M0Type');
+        end
+
+        % Change M0Type field of ASL JSON to Separate
+        if isfield(jsonASL, 'M0Type') && (~isempty(iM0) || ~isempty(iDummy))
```
jan-petr, line 254: Not sure why it is relevant to have the iDummy - these are about excluding stuff from ASL, not related to M0.

The ~isempty(iM0) probably isn't necessary either. Because if we are running splitting and there's a jsonM0, then it is clear that we have split the files and M0.nii is there - though I'm not 100% sure about this one and it doesn't hurt to have it here (unlike the iDummy though).

jan-petr (changes requested):
Perfect. All is there. Two minor things - see the comments in code.

Functions/xASL_io_SplitASL.m
MichaelStritt: Ah perfect, that is the part where I wasn't 100% sure if it's correct.
In line 78 we check `if isempty(iM0) && isempty(iDummy)` anyway, so this part is redundant.
I just wanted to make sure, that we either split M0 or Dummy scans, so that if someone would call it and there was not M0 within the sequence, that my code section wouldn't lead to unexpected bugs. We can definitely get rid of the iDummy part.

Functions/xASL_io_SplitASL.m
MichaelStritt: Sure, no problem 👍

Functions/xASL_io_SplitASL.m
jan-petr: Yes - without iM0 or iDummy, there's nothing to split. But if there's only iDummy, then we throw some volumes away, but nothing changes on the way M0 is defined. It will mostly be `separate` anyway, but there's still a chance that the M0 will be defined by a single value. So if the only reason for calling the split were dummy scans, we don't want to change the M0Type. So iDummy indeed hasto go

jan-petr (approved):
Super. I haven't check the moving of code to subfunctions, but I assume a simple testing would discover errors there.

PR #552, opened 2021-05-07
 #455 Created reference testing values file

### LINKED ISSUE #455: Reference testing values of xASL_qc_TestExploreASL
Edit by Jan: In order to easily compare results of xASL_qc_TestExploreASL a reference result table should be added to the repository and automatically compared with the current results inside `xASL_qc_TestExploreASL` and a single value pass/no pass outputted at the end to allow a quick assessment if the results are equal (within 0.1%) tolerance to the previous run.

### Release note
Automatically compare results of TestDataSets with a saved reference results.

MDijsselhof:
closes #455


jan-petr (changes requested):
Not really what I had on my mind - sorry ;)

I meant save those values in a easy-to-read format for Matlab - i.e. TSV or MAT file, so that it can be read and parsed by Matlab.

Then in function `xASL_qc_TestExploreASL` in section 8 (line 465), it will read this reference table from a fixed location. And compare with the results table calculated here and then output either 1 or 0 if the difference in all fields is below 0.1% or not. 

The aim is that we don't have to compare the old and new results by eye, but automatically. And that we have a fixed saved reference.

MichaelStritt:
Hey guys, I didn't want to interfere here, but maybe we can close this PR before v1.7.0 because the branch is already behind by a lot of commits.

jan-petr:
> Hey guys, I didn't want to interfere here, but maybe we can close this PR before v1.7.0 because the branch is already behind by a lot of commits.

Great Michael. Have you tested that this works? I will re-run and test everything...

jan-petr:
1. you provide also ExploreASL version and OS in the table. I would like to make it possible to add results from different ExploreASL versions, and it will always compare the results for the current and for all version
2. checking for all lines of the results if they are similar between the current and old version.

MichaelStritt:
The comparison part is still missing, but I can write that this morning. If you want to keep on updating this, maybe a .mat file will be easier than a .tsv file.

jan-petr:
> The comparison part is still missing, but I can write that this morning. If you want to keep on updating this, maybe a .mat file will be easier than a .tsv file.

OK. Please do so. I do like the TSV file actually. Especially the fact that it is human readable. I would keep it, it can be nicely and automatically read, I don't see a huge disadvantage in it...

MichaelStritt:
@jan-petr: It works for me on windows now :)
It only compares to the most up-to-date version, but can read multiple versions. It saves a "comparison table" as well as the reference table.

jan-petr (changes requested):
Perfect. I get the difference table. And prints the xASL version, but can I also get a simple 1/0 passed not passed - i.e. checking if all values are below 0.01% or something. Also if both tables have "n/a" entry, then the error should be 0%, otherwise, with single n/a it should be 100%.

MichaelStritt:
### Test with 3 versions

```matlab
...
Version: xASL_v1_6_0
OS:      Linux_2016b
Version: xASL_v1_6_1
OS:      Linux_2016b
Version: xASL_v1_6_2
OS:      Linux_2016b
...
Version: xASL_v1_6_0
Passed:  false
Version: xASL_v1_6_1
Passed:  false
Version: xASL_v1_6_2
Passed:  false
...
```

jan-petr (approved):
OK. Increased the tolerance to 1%, 0.1% was too optimistic :)
Also ignoring strings...

PR #553, opened 2021-05-07
Closes #479 FME: Initial import version

### LINKED ISSUE #479: Import Multi TE

MichaelStritt:
Hey @jan-petr, I hope this hack is not too dirty ;)
If we can add these lines, then the import at least doesn't crash for the Hadamard encoded images anymore.
This may help @amahroo to play around with the data within ExploreASL a little bit.
Let me know if it is okay to do this or if you would do anything differently.


Modules/SubModule_Import/xASL_imp_NII2BIDS_SubjectSessionRun.m
```diff
@@ -298,6 +298,25 @@
...
     end
+    
+    
+    %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+    % Check for FME Hadamard sequences
+    if isfield(jsonLocal,'SeriesDescription')
+        isHadamardFME = ~isempty(regexp(char(jsonLocal.SeriesDescription),'(Encoded_Images_Had)\d\d(_)\d\d(_TIs_)\d\d(_TEs)', 'once'));
+        if isHadamardFME
+            startDetails = regexp(char(jsonLocal.SeriesDescription),'\d\d(_)\d\d(_TIs_)\d\d(_TEs)', 'once');
+            hadamard.type = xASL_str2num(jsonLocal.SeriesDescription(startDetails:startDetails+1),'auto');
```
jan-petr, line 309: I do like structuring, but they don't understand the concept of it with BIDS ;)
So we better come up now with names that have a chance to be adopted in BIDS - with a bit of luck, we won't have to rename stuff later.

I would thus change hadamard.type to HadamardType and add that directly to JSON.

Similarly NumberOfTEs or even better NumberTEs should be used with full words.Same for TI, except that we use the term PLD. These two actually in theory don't need to be under Hadamard, because they can have other use. In theory. Because in practice, this information is duplicitous to what is in EchoTime and PLD fields. So these two fields will never make it to BIDS. So for now, lets make our life easier and use the fields `HadamardNumberTI` and `HadamardNumberTE`. Later on, we will make sure that the correct values are properly saved to the standard fields and these fields are no longer neceary. What do you think?

MichaelStritt: 4da6aec 👍

MichaelStritt:
@jan-petr: Check out my explanations in the issue :)

jan-petr (changes requested):
Hadamard JSONs look good - and the NII looked good as well on the meeting. 
One question - M0 is single TE or multi-TE? If multi-TE, then this is missing in the JSON. M0 has a separate merging function, so you probably have to port the TE merging code there as well.

I did a minor fix in the testing to make it work for all (a fix in the manual part). 

xASL_bids_MergeNifti_Merge had to be take out as a separate function (as the SiemensMerge was) - I did that, can you please check the header? Same for _Delete and _RenameParms - they all had to be taken out as they were used in the original merge and the siemens merge. But you can also put them all back ;)

Otherwise, I run full test on Flavors and all looks good.

MichaelStritt:
@jan-petr: I think the M0 should have 8 TEs as well. I had to add another regular expression in xASL_bids_MergeNifti_Merge, but other than that, there wasn't much to do. I do not have to repeat the array here, right?
7a48ba0

Functions/xASL_bids_BIDS2Legacy.m
```diff
@@ -284,6 +285,21 @@
 % Add the path to the dataPar.x struct that we return to the Master script
 dataPar.x.DataParPath = fullfile(pathLegacy, 'dataPar.json');
 
+%% 11. Clean up
+try
+    filesCleanUp = xASL_adm_GetFileList(pathStudy,'^import_.+$');
+    if ~isempty(filesCleanUp)
```
HenkMutsaerts, line 291: @jan-petr This line is not needed. iFile=1:0 will automatically be skipped

Functions/xASL_bids_BIDS2Legacy.m
```diff
@@ -284,6 +285,21 @@
 % Add the path to the dataPar.x struct that we return to the Master script
 dataPar.x.DataParPath = fullfile(pathLegacy, 'dataPar.json');
 
+%% 11. Clean up
```
HenkMutsaerts, line 288: `Clean up` is a bit generic, do you mean here that the import-related logs are removed or moved?

Functions/xASL_bids_BIDS2Legacy.m
```diff
@@ -284,6 +285,21 @@
...
+    if ~isempty(filesCleanUp)
+        for iFile = 1:size(filesCleanUp,1)
+            sourceCleanUp = filesCleanUp{iFile};
+            [~, fileCleanUp, extCleanUp] = xASL_fileparts(sourceCleanUp);
+            destCleanUp = fullfile(pathStudy, 'derivatives', 'ExploreASL', [fileCleanUp extCleanUp]);
+            xASL_Move(sourceCleanUp,destCleanUp);
+        end
+    end
+catch
+    fprintf('Clean up failed...\n');
```
HenkMutsaerts, line 300: Here we should also print the error that it crashed with

Functions/xASL_bids_MergeNifti_Delete.m
```diff
@@ -0,0 +1,33 @@
+function xASL_bids_MergeNifti_Delete(NiftiPaths)
```
HenkMutsaerts, line 1: @jan-petr  This function already exists: `xASL_adm_DeleteFilePair` so can be removed

Functions/xASL_bids_MergeNifti_Merge.m
```diff
@@ -0,0 +1,148 @@
+function pathOut = xASL_bids_MergeNifti_Merge(NiftiPaths,indexSortedFile,nameMerged,bAlternatingControlLabel)
```
HenkMutsaerts, line 1: These files are a bit convoluted, can we improve their readability?

Functions/xASL_bids_MergeNifti_Merge.m
```diff
@@ -0,0 +1,148 @@
...
+        % Special treatment for Hadamard encoded files
+        EchoTimes = cell(size(NiftiPaths,2),1);
+        for iFileCheck = 1:size(NiftiPaths,2)
+            % Get JSON
+            [jsonPathX, jsonNameX] = xASL_fileparts(NiftiPaths{iFileCheck});
+            if exist(fullfile(jsonPathX, [jsonNameX '.json']),'file')
+                tmpCheckJSON = spm_jsonread(fullfile(jsonPathX, [jsonNameX '.json']));
+                % Check EchoTimes
+                if isfield(tmpCheckJSON,'SeriesDescription')
+                    isHadamardFME = ~isempty(regexp(char(tmpCheckJSON.SeriesDescription),...
```
HenkMutsaerts, line 104: @jan-petr 
This is only valid for Siemens, needs to be specified here otherwise it will generate errors for Hadamard of other vendors.
Also, what happens if `SeriesDescription` doesn't exist?

Functions/xASL_bids_MergeNifti_RenameParms.m
```diff
@@ -0,0 +1,31 @@
+function xASL_bids_MergeNifti_RenameParms(Fpath,Fname)
```
HenkMutsaerts, line 1: @jan-petr This looks like an admin function that we don't need right, _parms is legacy?

Functions/xASL_bids_MergeNifti_SiemensASLFiles.m
```diff
@@ -0,0 +1,135 @@
+function pathOut = xASL_bids_MergeNifti_SiemensASLFiles(NiftiPaths)
```
HenkMutsaerts, line 1: @jan-petr Also written a bit convoluted :)

Functions/xASL_io_dcm2nii.m
```diff
@@ -312,6 +312,9 @@
                     error('Invalid index: the requested volume (%d) exceeds the actual number of files (%d)',max(parms.Keep),nNifties);
                 end
             end
+            
+            % Fix NIFTI file names
+            [niiEntries, niiNames] = xASL_io_dcm2nii_FixFormat(niiEntries, niiPaths, niiNames);
```
HenkMutsaerts, line 317: @jan-petr 
What does `FixFormat` mean? We need to be specific or before we know it ExploreASL is as unreadable as SPM :)

Functions/xASL_test_BIDSConversion.m
```diff
@@ -81,10 +81,10 @@ function xASL_test_BIDSConversion(baseDirImport, baseDirReference, bImport, bCom
 			case 'Philips_PCASL_3DGRASE_R5.4_TopUp'
                 DirASL = fullfile(baseDirImport, flavorList{iFlavor}, 'analysis', 'Sub1', 'ASL_1');
                 
-				xASL_Move(fullfile(DirASL, 'M0_601_1.nii'), fullfile(DirASL, 'M0.nii'), 1);
-				xASL_Move(fullfile(DirASL, 'M0_601_1.json'), fullfile(DirASL, 'M0.json'), 1);
-				xASL_Move(fullfile(DirASL, 'M0_701_1.nii'), fullfile(DirASL, 'M0PERev.nii'), 1);
-				xASL_Move(fullfile(DirASL, 'M0_701_1.json'), fullfile(DirASL, 'M0PERev.json'), 1);
+				xASL_Move(fullfile(DirASL, 'M0_601_00001.nii'), fullfile(DirASL, 'M0.nii'), 1);
```
HenkMutsaerts, line 84: @jan-petr 
These 601 701 look study-specific, like the 6th or 7th sequence that was scanned? That varies between studies/data sets, so this should probably be more flexible.

Modules/SubModule_Import/xASL_imp_DCM2NII_Subject_ShuffleTheDynamics.m
```diff
@@ -27,7 +27,6 @@
 % 4. Only try shuffling if you dont know the ASL context already
```
HenkMutsaerts, line 27: @jan-petr 
I don't know who came up with this function name, but you shuffle cards to get a random order, I guess that's not what we want here ;)

Modules/SubModule_Import/xASL_imp_DCM2NII_Subject_ShuffleTheDynamics.m
```diff
@@ -113,46 +112,40 @@
...
         end
-    end    
+    end
+    
+    %% FME (Hadamard) Check
+    if numel(nii_files)>=1
+        [resultPath, resultFile] = xASL_fileparts(nii_files{1});
+        if exist(fullfile(resultPath, [resultFile '.json']), 'file')
+            resultJSON = spm_jsonread(fullfile(resultPath, [resultFile '.json']));
+            if isfield(resultJSON,'SeriesDescription')
```
HenkMutsaerts, line 122: @jan-petr Also here same as above, very specific. Is fine, but it should be noted that this only for Siemens etc. If someone now tries to load a Philips Hadamard this can go wrong

Modules/SubModule_Import/xASL_imp_NII2BIDS_SubjectSessionRun.m
```diff
@@ -220,7 +237,9 @@
                 end
             end
             % Prioritize the Phoenix field values in the general case
-            jsonLocal.(fn{1}) = jsonDicom.PhoenixAnalyzed.(fn{1});
+            if ~(strcmp(fn{1},'EchoTime') && isHadamardFME)
```
HenkMutsaerts, line 240: @jan-petr 
Does this change not break other cases? (i.e. Phoenix without `HadamardFME`)

Modules/SubModule_Import/xASL_imp_NII2BIDS_SubjectSessionRun.m
```diff
@@ -298,6 +317,27 @@
...
+    
+    
+    %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+    % Check for FME Hadamard sequences
+    if isfield(jsonLocal,'SeriesDescription')
+        isHadamardFME = ~isempty(regexp(char(jsonLocal.SeriesDescription),'(Encoded_Images_Had)\d\d(_)\d\d(_TIs_)\d\d(_TEs)', 'once'));
+        if isHadamardFME
+            startDetails = regexp(char(jsonLocal.SeriesDescription),'\d\d(_)\d\d(_TIs_)\d\d(_TEs)', 'once');
+            HadamardType = xASL_str2num(jsonLocal.SeriesDescription(startDetails:startDetails+1),'auto');
+            HadamardNumberTI = xASL_str2num(jsonLocal.SeriesDescription(startDetails+3:startDetails+4),'auto');
```
HenkMutsaerts, line 329: @jan-petr Should `Number` be `value`? Otherwise `Number` could be confused with `Count`

Modules/SubModule_Import/xASL_imp_NII2BIDS_SubjectSessionRun.m
```diff
@@ -298,6 +317,27 @@
             end
         end
     end
+    
+    
+    %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+    % Check for FME Hadamard sequences
```
HenkMutsaerts, line 323: @jan-petr Not only `Check` but also other things are edited here, right?

Modules/SubModule_Import/xASL_imp_NII2BIDS_SubjectSessionRun.m
```diff
@@ -309,10 +349,14 @@
     listFieldsRepeat = {'PostLabelingDelay', 'LabelingDuration','VascularCrushingVENC','FlipAngle','RepetitionTimePreparation'};
     for iRepeat = 1:length(listFieldsRepeat)
         if isfield(jsonLocal,(listFieldsRepeat{iRepeat})) && (length(jsonLocal.(listFieldsRepeat{iRepeat})) > 1) && (size(imNii,4) ~= length(jsonLocal.(listFieldsRepeat{iRepeat})))
-            if mod(size(imNii,4),length(jsonLocal.(listFieldsRepeat{iRepeat})))
-                error('Cannot find a match between the %s and the 4th dimension of the NIFTI.\n',listFieldsRepeat{iRepeat});
+            if ~isHadamardFME % This section does not work for FME Hadamard encoded images right now
```
HenkMutsaerts, line 352: @jan-petr But would this work for multi-PLD that is not `HadamardFME`?

Issue #391, opened 2021-03-02
Implementation BASIL


HenkMutsaerts:

HenkMutsaerts:
@jan-petr Can you check in `xASL_quant_SinglePLD` if your `SliceReadoutTime` changes can also be applied to the preparations for Basil? This was changed by Martin before you applied these changes.

HenkMutsaerts:
@mcraig-ibme As you can see I have refactored the code a bit, to have `x.bUseBasilQuantification` in `xASL_quant_SinglePLD` that chooses between preparing BASIL quantification or using the ExploreASL quantification. Then I refactored your main BASIL preparations to `xASL_quant_Basil`. The paths etc should be improved and it works nicely, though I have a few questions left:

* somehow the quantification is 2000* too high, I guess it goes wrong with applying the scale slopes?
* the FSL Python code issues a warning about not being able to find `imcp` (which seems to be a NIfTI copying code), can we somehow disable this warning?
* per your suggestion, I programmed a regular expression to find the final `mean_fitt` NIfTI file, where it will always take the alphabetically latest file (as I understood you have step1 step2 etc)
* BASIL stores a lot of files, which ones would be nice to save for QC purposes?

Thanks again!

jan-petr:
@HenkMutsaerts The x.bUseBasilQuantification has to be initialized if missing (DEFAULT false). Or it will crash in the quantification submodule. I have done that, but please add the bUseBasilQuantification option and explanation to DataParTemplate.m

jan-petr:
> @jan-petr Can you check in `xASL_quant_SinglePLD` if your `SliceReadoutTime` changes can also be applied to the preparations for Basil? This was changed by Martin before you applied these changes. Note that Basil does some things itself, so not all parameters should go to the `ScaleImage` for Basil.

I have updated the code. We can use the fact that SliceReadoutTime is correctly prepared and we can just use SRT(2)-SRT(1). However, it is not a good idea to save this back to x.Q.SliceReadoutTime - it can overwrite the previous values and might be a problem in the future if this is saved somewhere. So I have created a dedicated x.Q.BasilSliceReadoutTime.

Do you agree?

HenkMutsaerts:
I thought we discussed this, and ended up with

1. `SliceTiming`: BIDS-compatible, full vector, absolute values
2. `SliceReadoutTime`: single value (backward compatible), subtracted
[...]

PR #557, opened 2021-05-08
Feature #391 basil implementation

HenkMutsaerts:
Closess #391


Functions/xASL_quant_Basil.m
```diff
@@ -0,0 +1,231 @@
+function [CBF_nocalib] = xASL_quant_Basil(PWI, x)
+%xASL_quant_Basil Perform quantification using FSL BASIL
+% FORMAT: [CBF_nocalib] = xASL_quant_Basil(PWI, x)
+% 
+% INPUT:
+%   PWI             - image matrix of perfusion-weighted image (REQUIRED)
+%   x               - struct containing pipeline environment parameters (REQUIRED)
+%
+% OUTPUT:
+% CBF_nocalib       - Quantified CBF image
```
jan-petr, line 10: Explain here that this means without M0 division.

Functions/xASL_quant_Basil.m
```diff
@@ -0,0 +1,231 @@
...
+% -----------------------------------------------------------------------------------------------------------------------------------------------------
+% DESCRIPTION: This script performs quantification of the PWI using the FSL Basil pipeline. Final calibration to
+%              physiological units is performed by dividing the quantified PWI by the M0 image/value
```
jan-petr, line 13: We have to specify what it can do already. For sure single-TI, multi-TI - as defined in the PLD vector. But what else? It cannot for sure call Hadamard or QUASAR yet?! No velocity or acceleration selective?

Functions/xASL_quant_Basil.m
```diff
@@ -0,0 +1,231 @@
...
+    
+    
+    %% Define paths
+    % NEEDS TO BE MADE A RELATIVE PATH TO BE SPACE-INDEPENDENT
+    pathBasilInput = fullfile(x.SESSIONDIR, 'PWI4D_BasilInput.nii');
+    pathBasilOptions = fullfile(x.SESSIONDIR, 'Basil_ModelOptions.txt');
+    dirBasilOutput = fullfile(x.SESSIONDIR, 'BasilOutput');
+    
+    %% Write the PWI as Nifti file for Basil to read as input
+    % FIXME would be good to have a brain mask too at this point
```
jan-petr, line 32: Make an issue for that already and reference from here?!

Functions/xASL_quant_Basil.m
```diff
@@ -0,0 +1,231 @@
...
+    
+    xASL_io_SaveNifti(x.P.Path_PWI4D, pathBasilInput, PWI, [], 0);
+
+
+    %% Create option_file that contains options which are passed to Fabber
+    % basil_options is a character array containing CLI args for the Basil command
+    bMultiPLD = false; % for now
```
jan-petr, line 39: I see, we go for single-TI only just for testing - that is fine, but should be described upfront.

HenkMutsaerts:
Good idea, we should then indeed in xASL_qc_TestExploreASL after running the full test, remove the quantification lock file and rerun with basil set on.

Notice that we still need to fix the quantification scaling though

MichaelStritt:
Could you test this @MDijsselhof ?

HenkMutsaerts:
I did some final fixes, it works now in all TestDataSets!

@jan-petr would you mind checking the `SliceReadoutTime` scripts, the quantification is off for the Philips & Siemens 2D EPI sequences by a factor ~ 10,000, whereas for 3D sequences of GE, Philips and Siemens the quantification is comparable with ours.

If this is done, we can merge and release 1.6.0!

PR #564, opened 2021-05-10
Closes #563 Minor design fix

### LINKED ISSUE #563: Minor design fix
@HenkMutsaerts requested the x.D.ROOT to be fully visible if possible

MichaelStritt:
Checkout #563


MichaelStritt:
### Example

image

HenkMutsaerts (approved):
Nice! Would be even nicer if it splits at the last folder level but this is great too.

MichaelStritt:
> 
> 
> Nice! Would be even nicer if it splits at the last folder level but this is great too.

Yes, I thought about that, too. Makes the could a bit more bulky though, because you need to start taking into account different os and the path length etc.

Issue #480, opened 2021-04-09
Restructure x fields


MichaelStritt:
Optimize x struct fields. Proposed fields:

- x.opts or x.settings: ExploreASL arguments and derivatives
- x.data: dataset specific fields
- x.vis: colormaps etc.
- x.tools: toolbox specific fields
- x.atlas: atlas specific fields
- x.study: study specific fields
- x.module.(structural/asl/population)
- ...

Add description for each field to documentation website

**Edit 28.05.2021**: The fields I started with are...

- x.opts (input arguments and their derivatives)
- x.dir (directories)
- x.settings (mostly booleans for pipeline settings and some other stuff)
- x.dataset (dataset related fields)
- x.tools (right now there's not much in there besides the spm version)

MichaelStritt:
Since we want to release within the next two weeks, I'd move this to v1.7.0. Changes like these could lead to unexpected bugs within the pipeline.

HenkMutsaerts:
> Optimize x struct fields. Proposed fields:
> 
> * x.opts or x.settings: ExploreASL arguments and derivatives
> * x.data: dataset specific fields
> * x.vis: colormaps etc.
> * x.tools: toolbox specific fields
> * x.atlas: atlas specific fields
> * x.study: study specific fields
> * x.module.(structural/asl/population)
> * ...
> 
> Add description for each field to documentation website
> 
> **Edit 28.05.2021**: The fields I started with are...
> 
> * x.opts (input arguments and their derivatives)
> * x.dir (directories)
> * x.settings (mostly booleans for pipeline settings and some other stuff)
> * x.dataset (dataset related fields)
> * x.tools (right now there's not much in there besides the spm version)

* Is `x.dir` the dataset-related folders (currently in `x.P`)? Or environment/ExploreASL-related folders (currently in `x.D`)?
* Should `x.tools` be `external`?
* What is the difference between `x.opts` and `x.settings`? Input arguments travel through the pipeline right?

PR #587, opened 2021-05-16
#586 xASL_adm_GetPopulationSessions: avoid crashing if no sessions found

### LINKED ISSUE #586: xASL_adm_GetPopulationSessions
### Description

`xASL_adm_GetPopulationSessions` crashes when no sessions are found

### Requirements

Minor change, easy fix

### Release notes

Minor bugfix for `xASL_adm_GetPopulationSessions` to avoid crashing if no sessions are found

HenkMutsaerts:
### Linked issue

Closes #586

### DIFF
--- a/Functions/xASL_adm_GetPopulationSessions.m
+++ b/Functions/xASL_adm_GetPopulationSessions.m
@@ -1,4 +1,4 @@
-function [nSessions, bSessionsMissing] = xASL_adm_GetPopulationSessions(x)
+function [nSessions, bSessionsMissing, SESSIONS] = xASL_adm_GetPopulationSessions(x)
 % xASL_adm_GetPopulationSessions(x) obtain number of Sessions by determining amount of input files present in the Population folder
 % FORMAT: [nSessions, bSessionsMissing] = xASL_adm_GetPopulationSessions(x)
 %
@@ -8,6 +8,7 @@ function [nSessions, bSessionsMissing] = xASL_adm_GetPopulationSessions(x)
 % OUTPUT:
 %   nSessions           - Maximum amount of sessions present in Population folder
 %   bSessionsMissing    - Boolean to show if no sessions can be found in the Population folder
+%   SESSIONS            - x.SESSIONS
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
 % DESCRIPTION: This function looks for the maximum amount of sessions that
 % are present in selected processed files present in the Population folder.
@@ -19,6 +20,12 @@ function [nSessions, bSessionsMissing] = xASL_adm_GetPopulationSessions(x)
 %   5. Check and provide warning of number of sesssions differs per subject
 
 %% 1. Administration
+if ~isfield(x, 'SESSIONS')
+    SESSIONS = 'ASL_1';
+else
+    SESSIONS = x.SESSIONS;
+end
+
 currentSubjectRegExp = x.subject_regexp;
 if strcmp(currentSubjectRegExp(1), '^')
     currentSubjectRegExp = currentSubjectRegExp(2:end);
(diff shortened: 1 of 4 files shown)


Functions/xASL_adm_GetPopulationSessions.m
```diff
@@ -19,6 +20,12 @@
 %   5. Check and provide warning of number of sesssions differs per subject
 
 %% 1. Administration
+if ~isfield(x, 'SESSIONS')
+    SESSIONS = 'ASL_1';
```
MDijsselhof, line 24: But what if some subjects don't have a first session?

MDijsselhof (commented):
Hi Henk-Jan, 

I've left a single comment, curious what you think!

jan-petr (changes requested):
We have to update certain fields similarly as done in #480 . Becausehere, we add new x.nSessions and x.nSubjectSessions - the whole difference file has to be checked and potentially updated... This won't show as a conflict as these are newly added lines. @MichaelStritt should check as he knows what to fix...

MichaelStritt:
> 
> 
> We have to update certain fields similarly as done in #480 . Becausehere, we add new x.nSessions and x.nSubjectSessions - the whole difference file has to be checked and potentially updated... This won't show as a conflict as these are newly added lines. @MichaelStritt should check as he knows what to fix...

In  `Functions/xASL_stat_GetROIstatistics.m`:
- we moved x.WBmask to x.S.masks.WBmask

In `Functions/xASL_stat_PrintStats.m`:
- we moved x.nSessions to x.dataset.nSessions

MichaelStritt:
@jan-petr: I only changed the code sections that were changed by #586. The should work with #480 now. There will still be some conflict resolving.

MichaelStritt (approved):
Should be fine with regard to #480

Functions/xASL_adm_GetPopulationSessions.m
HenkMutsaerts: Good point, should the default be an empty `SESSIONS` output then?

Functions/xASL_stat_GetROIstatistics.m
```diff
@@ -177,7 +177,7 @@
 
 %% Define number of sessions to use
 
-[nSessions, bSessionsMissing] = xASL_adm_GetPopulationSessions(x); % obtain number of Sessions by determining amount of input files present in the Population folder
+[nSessions, bSessionsMissing, x.SESSIONS] = xASL_adm_GetPopulationSessions(x); % obtain number of Sessions by determining amount of input files present in the Population folder
```
HenkMutsaerts, line 180: Is this third new output parameter `x.SESSIONS` used? I'm not sure if we should avoid this, since `x.SESSIONS` is also used in the Structural and ASL modules.

HenkMutsaerts (commented):
Am I actually reviewing my own work? :) I thought I found some parts that I wrote :)

Functions/xASL_adm_GetPopulationSessions.m
MichaelStritt: If you load a dataset that has at least one sessions, you will definitely get a SESSIONS field, even if you did not manually define SESSIONS (`[x] = ExploreASL_Master(my-bids-dataset,0,0);`).

HenkMutsaerts:
@jan-petr @MDijsselhof Yes, I agree. We might soon revamp all this when we move to BIDS derivatives, but for now I believe it is safe to assume that there is an `ASL_1`. If there are no ASL sessions, this works as well.

MichaelStritt:
Great, then only @jan-petr needs to re-approve

Functions/xASL_adm_GetPopulationSessions.m
jan-petr: @MichaelStritt but that would mean that this condition above never really happens. So we can define an empty session there?!

jan-petr (approved):
Fine to merge.

jan-petr:
I am not able to remove Mathijs from the review - he either appears as not-approved-yet or as has-a-comment. So lets keep it as it is and merge. @BeatrizPadrela seems to be assigned for this ;)

Issue #588, opened 2021-05-17
Release v1.7.0: Manage ExploreASL_Master input path


MichaelStritt:
### Description

`ExploreASL_Master` should accept folders for its first argument.

### Requirements

- Change `ExploreASL_Initialize` so that we can input folders.
- Update `ExploreASL_Master` unit test (`xASL_ut_UnitTest_function_ExploreASL_Master`), so that we can take new use-cases into account.
- Update the documentation [here](https://github.com/ExploreASL/Documentation/tree/master/templates)

### How to test

- Clone the [Testing repository](https://github.com/ExploreASL/Testing) into the same directory where your ExploreASL repository is.
- Run `[UnitTests,UnitTestsTable] = xASL_test_UnitTesting`
- Run `BIDS_fullPipelineTest` or `xASL_test_IntegrationTesting` (which is supposed to become its successor)
- Run `xASL_qc_TestExploreASL`

### Release notes

Release v1.7.0: `ExploreASL_Master` can read folders now and automatically searches for `sourceStructure`, `studyPar` and `dataPar` **JSON** files.

jan-petr:
- I propose the following: Input is a JSON or a directory. It takes the provided JSON and if it recognizes that it's a correct match for the step (based on the content) it takes that. Otherwise it looks in the directory for the JSONs with standard names
- that way, you can still call it with sourceStructure.json, sourceStructure2.json or dataPar-hiQ.json for a specific task. Otherwise, it starts looking for the other script or create dummies
- also - we have to work with the option that it looks inside derivatives/ExploreASL, or work directory with the dataPar.json in the root folder (Legacy)
- if we do it this way - the users won't recognize anything - their old way of calling ExploreASL will still work and we can prompt them with a warning that this is deprecated - this is much much better than change stuff completely and give them an error.
- In a similar way - we should recognize that they have called it with the 1.5.0 options and tell them how to update to 1.6.0 options...

MichaelStritt:
### Meeting Notes

- `dataPar.json`, `studyPar.json` and `sourceStructure.json`: fixed prefixes
- DCM2BIDS: requires `sourceStructure.json`
- BIDS2Legacy: requires `dataset_description.json`

MichaelStritt:
### Intermediate testing

```
================================= TEST RESULTS =================================

              name                  unit          tests        passed
    _________________________    __________    ____________    ______

[...]

Issue #583, opened 2021-05-15
Run TestFlavors using ExploreASL_Master


jan-petr:
### Description

We need to test the ExploreASL_Master that it is able to handle correctly all the different inputs from DICOM->BIDS->Legacy->Processing but only BIDS->Legacy->Processing. Currently, we call the subfunctions in testing, but we should call the Master script with different settings.

### How to test
All these combinations should work with different inputs:

- `ExploreASL('/pet/projekte/asl/data/BIDS/TmpConversion/Siemens_PASL_2DEPI_noBsup_AD2/sourceStructure.json',[1 1 0 0],0)`
`ExploreASL('/pet/projekte/asl/data/BIDS/TmpConversion/Siemens_PASL_2DEPI_noBsup_AD2/rawdata/dataset_description.json',[0 0 0 1],1)`

- `ExploreASL('/pet/projekte/asl/data/BIDS/TmpConversion/Siemens_PASL_2DEPI_noBsup_AD2/sourceStructure.json',[1 1 0 1],1)`

- `ExploreASL('/pet/projekte/asl/data/BIDS/TmpConversion/Siemens_PASL_2DEPI_noBsup_AD2/sourceStructure.json',[1 0 0 0],0)`
`ExploreASL('/pet/projekte/asl/data/BIDS/TmpConversion/Siemens_PASL_2DEPI_noBsup_AD2/sourceStructure.json',[0 1 0 1],1)`

- `ExploreASL('/pet/projekte/asl/data/BIDS/TmpConversion/Siemens_PASL_2DEPI_noBsup_AD2/sourceStructure.json',[1 0 0 0],0)`
`ExploreASL('/pet/projekte/asl/data/BIDS/TmpConversion/Siemens_PASL_2DEPI_noBsup_AD2/studyPar.json',[0 1 0 1],1)`

### Release notes

Required: Proper testing of flavors using ExploreASL_Master.

MichaelStritt:
Maybe we should wait with this until we finished #588

MichaelStritt:
Initial idea: d4df0ca

MichaelStritt:
Full pipeline working example: [here](https://github.com/ExploreASL/ExploreASL/blob/feature-%23588_BIDS_DirectoryInput/Testing/xASL_test_IntegrationTesting.m)

MichaelStritt:
[...]

PR #606, opened 2021-05-27
#583 xASL_test_IntegrationTesting: Initial implementation

MichaelStritt:
### Linked issue

Check out issue #583

### How to test

- Create a `testConfig.json` in your ExploreASL/Testing directory
- Run: `[TestResults, CheckResults] = xASL_test_IntegrationTesting;`

### Comments

I know this probably doesn't fulfill all requirements yet, but I thought I could show you my initial implementation and how I want to extract the errors/warning in the future. The x struct includes a logging field now (implemented in #588). This includes both errors from the import and from the processing workflow. Those should be written into a `log.xlsx` file. It doesn't catch the various warnings and stuff right now, but the major things are nicely captured.

### DIFF
--- a/Functions/xASL_adm_RemoveDirectories.m
+++ b/Functions/xASL_adm_RemoveDirectories.m
@@ -21,13 +21,10 @@ function xASL_adm_RemoveDirectories(root)
     end
     
     % ExploreASL
-    try
-        warning('off', 'MATLAB:rmpath:DirNotFound');
-        rmpath(genpath(root));
-        warning('on', 'MATLAB:rmpath:DirNotFound');
-    catch ME
-        fprintf('%s\n', ME.message);
-    end
+    warning('off', 'MATLAB:rmpath:DirNotFound');
+    rmpath(genpath(root));
+    warning('on', 'MATLAB:rmpath:DirNotFound');
+
 
 end
 
--- a/Functions/xASL_adm_RemoveDirectories.m
+++ b/Functions/xASL_adm_RemoveDirectories.m
@@ -22,9 +22,9 @@ function xASL_adm_RemoveDirectories(root)
     
     % ExploreASL
     try
-        warning('off', 'all');
+        warning('off', 'MATLAB:rmpath:DirNotFound');
         rmpath(genpath(root));
-        warning('on', 'all');
+        warning('on', 'MATLAB:rmpath:DirNotFound');
     catch ME
         fprintf('%s\n', ME.message);
     end
--- a/Functions/xASL_adm_RemoveDirectories.m
+++ b/Functions/xASL_adm_RemoveDirectories.m
@@ -1,0 +1,36 @@
+function xASL_adm_RemoveDirectories(root)
+%xASL_adm_RemoveDirectories Script to remove all ExploreASL related paths
+%
+% FORMAT:       xASL_adm_RemoveDirectories(root)
+% 
+% INPUT:        root - Root directory (REQUIRED, CHAR ARRAY)
+%
+% OUTPUT:       n/a
+% 
+% -----------------------------------------------------------------------------------------------------------------------------------------------------
+% DESCRIPTION:  Script to remove all ExploreASL related paths.
+%
+% -----------------------------------------------------------------------------------------------------------------------------------------------------
+% EXAMPLE:      xASL_adm_RemoveDirectories(x.MyPath);
+% __________________________________
+% Copyright 2015-2021 ExploreASL
+
+    %% Check input
+    if nargin < 1
+        error('Please provide a root directory...');
+    end
+    
+    % ExploreASL
+    try
+        warning('off', 'all');
+        rmpath(genpath(root));
+        warning('on', 'all');
+    catch ME
+        fprintf('%s\n', ME.message);
+    end
+
+end
+
+
+
+
--- a/Testing/xASL_test_IntegrationTesting.m
+++ b/Testing/xASL_test_IntegrationTesting.m
@@ -59,7 +59,6 @@ function [TestResults, CheckResults] = xASL_test_IntegrationTesting
 
     % Initialize ExploreASL
     [x] = ExploreASL_Initialize;
-    cd(testConfig.pathTest);
     
     % Clean Up
     clc
@@ -81,12 +80,10 @@ function [TestResults, CheckResults] = xASL_test_IntegrationTesting
     %% Iterate over flavors
     for iFlavor = 1:numel(flavorList)
         % Run individual flavor
-        TestResults(iFlavor) = xASL_test_thisFlavor(testConfig.pathTest, ...
-                                                    fullfile(testConfig.pathTest,'FlavorDatabase'), ...
-                                                    flavorList{iFlavor});
+        TestResults = xASL_test_thisFlavor(testConfig, flavorList{iFlavor});
         % Check logging
-        if isfield(TestResults(iFlavor),'logging')
-            loggingTable = xASL_test_AddLoggingEntryToTable(loggingTable,TestResults(iFlavor).logging);
+        if isfield(TestResults,'logging')
+            loggingTable = xASL_test_AddLoggingEntryToTable(loggingTable,TestResults.logging);
         end
     end
     
@@ -119,8 +116,12 @@ end
 
 
 %% Run individual flavor dataset
-function xFlavor = xASL_test_thisFlavor(testingRoot, databaseRoot, flavorName)
+function xFlavor = xASL_test_thisFlavor(testConfig, flavorName)
 
+    % Get paths
+    testingRoot = testConfig.pathTest;
+    databaseRoot = fullfile(testConfig.pathTest,'FlavorDatabase');
+    
     % Copy flavor
     fprintf('Copy %s...\n', flavorName);
     xASL_Copy(fullfile(databaseRoot,flavorName),fullfile(testingRoot,'FlavorDatabaseTest',flavorName),1);
@@ -136,6 +137,9 @@ function xFlavor = xASL_test_thisFlavor(testingRoot, databaseRoot, flavorName)
     
     % Save JSON
     spm_jsonwrite(fullfile(testingRoot,'FlavorDatabaseTest',flavorName,'dataPar.json'),defaultDataPar);
+    
+    % Remove Matlab paths (independent initialization)
+    xASL_adm_RemoveDirectories(testConfig.pathExploreASL);
 
     % Run test
     try
@@ -218,7 +222,7 @@ function stackText = xASL_test_StackToString(stack)
     end
     
     % Remove initial ' ,'
-    if ~isempty(stackText) && length(stackText>3)
+    if ~isempty(stackText) && length(stackText)>3
         stackText = stackText(3:end);
     end
 
--- a/Testing/xASL_test_IntegrationTesting.m
+++ b/Testing/xASL_test_IntegrationTesting.m
@@ -130,10 +130,8 @@ function xFlavor = xASL_test_thisFlavor(testConfig, flavorName)
     xASL_delete(fullfile(testingRoot,'FlavorDatabaseTest',flavorName,'rawdata'),1);
     
     % Create default dataPar.json
-    defaultDataPar.x.subject_regexp = '^sub-.*$';
     defaultDataPar.x.bUseMNIasDummyStructural = 1; % when no structural data, use ASL-MNI registration
     defaultDataPar.x.Quality = 0; % speed up testing
-    defaultDataPar.x.DELETETEMP = 1;
     
     % Save JSON
     spm_jsonwrite(fullfile(testingRoot,'FlavorDatabaseTest',flavorName,'dataPar.json'),defaultDataPar);
--- a/Testing/xASL_test_IntegrationTesting.m
+++ b/Testing/xASL_test_IntegrationTesting.m
@@ -31,6 +31,9 @@ function [TestResults, CheckResults] = xASL_test_IntegrationTesting
     warning('off','all')                    % Turn of warnings (only add path related)
     addpath(genpath(xASL_dir))              % Add all scripts to path (including testing directory)
     warning('on','all')
+    
+    % Logging table
+    loggingTable = array2table(zeros(0,2), 'VariableNames',{'message','stack'});
 
     %% Get paths
     if exist(fullfile(filepath,'testConfig.json'),'file')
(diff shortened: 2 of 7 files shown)


jan-petr (changes requested):
I think we are nearly there. I agree with your edits. Those weird calls of ExploreASL, we have tested previously. The main thing that was missing is that during the Flavors testing, we were calling directly the ExloreASL submodules and functions like xASL_module_Import or xASL_bids_BIDS2Legacy. This did the job, but wasn't testing if the master script is working. I now replaced all the calls with calls to the master script ExploreASL. All should work and should be better for testing. 

One small thing is missing - it now initializes everything again upon each call of ExploreASL - can we do something about it? Probably not and we have to live with it, right? This would be also OK - more initializations are called because we want every call to be an independent call of ExploreASL, so that's the price we pay for this.

Once we solve this minor thing and once you check if all works (I have tested 1 flavor), then I will approve and I will merge. Note that I did some small restructuring of function names etc.

MichaelStritt:
> One small thing is missing - it now initializes everything again upon each call of ExploreASL - can we do something about it? Probably not and we have to live with it, right? This would be also OK - more initializations are called because we want every call to be an independent call of ExploreASL, so that's the price we pay for this.

If we want the tests to be more independent, we can add a small script to remove the Matlab paths in between. Similar to the call in Henks old test script:

```matlab
% Remove ExploreASL paths
warning('off','MATLAB:rmpath:DirNotFound');
rmpath(genpath(x.MyPath));
warning('on','MATLAB:rmpath:DirNotFound');
```

> Once we solve this minor thing and once you check if all works (I have tested 1 flavor), then I will approve and I will merge. Note that I did some small restructuring of function names etc.

jan-petr:
> > One small thing is missing - it now initializes everything again upon each call of ExploreASL - can we do something about it? Probably not and we have to live with it, right? This would be also OK - more initializations are called because we want every call to be an independent call of ExploreASL, so that's the price we pay for this.
> 
> If we want the tests to be more independent, we can add a small script to remove the Matlab paths in between. Similar to the call in Henks old test script:
> 
> ```matlab
> % Remove ExploreASL paths
> warning('off','MATLAB:rmpath:DirNotFound');
> rmpath(genpath(x.MyPath));
> warning('on','MATLAB:rmpath:DirNotFound');
> ```
> 
> > Once we solve this minor thing and once you check if all works (I have tested 1 flavor), then I will approve and I will merge. Note that I did some small restructuring of function names etc.

Yes - good idea. Can you please add that?

MichaelStritt:
> 
> 
> > > One small thing is missing - it now initializes everything again upon each call of ExploreASL - can we do something about it? Probably not and we have to live with it, right? This would be also OK - more initializations are called because we want every call to be an independent call of ExploreASL, so that's the price we pay for this.
> > 
> > 
> > If we want the tests to be more independent, we can add a small script to remove the Matlab paths in between. Similar to the call in Henks old test script:
> > ```matlab
> > % Remove ExploreASL paths
> > warning('off','MATLAB:rmpath:DirNotFound');
> > rmpath(genpath(x.MyPath));
> > warning('on','MATLAB:rmpath:DirNotFound');
> > ```
> > 
> > > Once we solve this minor thing and once you check if all works (I have tested 1 flavor), then I will approve and I will merge. Note that I did some small restructuring of function names etc.
> 
> Yes - good idea. Can you please add that?

It's a bit difficult, since we have to make sure that we run it in the correct code section and we're still able to reinitialize etc.

The script to remove the paths (920fc28) is pretty simple and I found two code sections where we can easily add this (7ac0165).

HenkMutsaerts:
Looks good, just don't store these in an `XLSX` but rather `TSV` (also according to BIDS). You can use `xASL_tsvWrite`.

Functions/xASL_adm_RemoveDirectories.m
```diff
@@ -0,0 +1,36 @@
...
+% Copyright 2015-2021 ExploreASL
+
+    %% Check input
+    if nargin < 1
+        error('Please provide a root directory...');
+    end
+    
+    % ExploreASL
+    try
+        warning('off', 'all');
```
HenkMutsaerts, line 25: This is tricky, as it disables all warnings (so debugging becomes tricky) and enables all warnings (even those that a user might have disabled!

So this is not good practice. You can disable warnings specifically, google it or check in our current code. You run this script, then when a warning occurs, run `lastwarn` which gives you also the warning ID, and then you can disable this specific warning ID.

Testing/xASL_test_IntegrationTesting.m
```diff
@@ -109,14 +116,30 @@
...
+    
     % Copy flavor
     fprintf('Copy %s...\n', flavorName);
     xASL_Copy(fullfile(databaseRoot,flavorName),fullfile(testingRoot,'FlavorDatabaseTest',flavorName),1);
     
     % Remove rawdata folder from testing dataset
     xASL_delete(fullfile(testingRoot,'FlavorDatabaseTest',flavorName,'rawdata'),1);
+    
+    % Create default dataPar.json
+    defaultDataPar.x.subject_regexp = '^sub-.*$';
```
HenkMutsaerts, line 133: Why did you add these here, when they are defaulted by ExploreASL itself? It seems unnecessary convoluted for me now, you can probably remove many testing lines and make the scripts more efficient and readable?

HenkMutsaerts (changes requested):
Very nice work! Though I wonder if not many of the things that are now stated explicitly in the testing scripts should be stated inside the actual ExploreASL scripts, and the testing iterates them, so simplification. E.g., the logging should be part of ExploreASL as well right, why only logging when testing and not when a user runs the scripts? 

I would make the actual running and testing as equal as possible, so debugging becomes the same if a report comes from our internal testing or from the user.

MichaelStritt:
> [...] logging should be part of ExploreASL as well right, why only logging when testing and not when a user runs the scripts?

Logging is part of ExploreASL already. It can be still improved though. Check out `xASL_Iteration` lines 399-400:             

```
% Catch error in logging field
[x] = xASL_qc_AddLoggingInfo(x, ME);
```

The only thing that the integration testing script does now, is to combine all of the logs from the different flavors into an xlsx and a tsv file. IMHO we don't need this every time somebody runs ExploreASL in general.

Functions/xASL_adm_RemoveDirectories.m
MichaelStritt: 8cdb69b

Testing/xASL_test_IntegrationTesting.m
MichaelStritt: We need ...
```
    defaultDataPar.x.bUseMNIasDummyStructural = 1;
    defaultDataPar.x.Quality = 0;
```
... those are not default.

MichaelStritt: 237f78d

jan-petr (approved):
OK.

HenkMutsaerts:
> > [...] logging should be part of ExploreASL as well right, why only logging when testing and not when a user runs the scripts?
> 
> Logging is part of ExploreASL already. It can be still improved though. Check out `xASL_Iteration` lines 399-400:
> 
> ```
> % Catch error in logging field
> [x] = xASL_qc_AddLoggingInfo(x, ME);
> ```
> 
> The only thing that the integration testing script does now, is to combine all of the logs from the different flavors into an xlsx and a tsv file. IMHO we don't need this every time somebody runs ExploreASL in general.

OK nice!

Testing/xASL_test_IntegrationTesting.m
```diff
@@ -232,12 +232,19 @@
 %% Write logging information to XLSX table
 function xASL_test_writeLoggingTable(filePath,logTable)
 
-    % Write file
+    % Write XLSX file
     try
-        writetable(logTable,filePath,'WriteVariableNames',true);
+        writetable(logTable,[filePath '.xlsx'],'WriteVariableNames',true);
```
HenkMutsaerts, line 237: Why is `xlsx` so useful, can we not keep TSV only? That would be better for multiple reasons: it is more open-access, easier to read, lighter weight.

Functions/xASL_adm_RemoveDirectories.m
```diff
@@ -22,9 +22,9 @@ function xASL_adm_RemoveDirectories(root)
     
     % ExploreASL
     try
-        warning('off', 'all');
+        warning('off', 'MATLAB:rmpath:DirNotFound');
```
HenkMutsaerts, line 25: great, perhaps add some comments about what we do here, and why we use the try-catch construction for this. It seems not required? Matlab never gives an error with `rmpath`, only warnings? I would avoid try-catch constructions if we can before the code becomes too convoluted

Testing/xASL_test_IntegrationTesting.m
```diff
@@ -130,10 +130,8 @@
     xASL_delete(fullfile(testingRoot,'FlavorDatabaseTest',flavorName,'rawdata'),1);
     
     % Create default dataPar.json
-    defaultDataPar.x.subject_regexp = '^sub-.*$';
     defaultDataPar.x.bUseMNIasDummyStructural = 1; % when no structural data, use ASL-MNI registration
```
HenkMutsaerts, line 135: I thought this wasn't the case for all flavors, you mind checking if all flavors don't have an T1.nii, otherwise we should remove any structural NIfTIs (perhaps in another issue)

HenkMutsaerts (changes requested):
Few minor issues but otherwise happy :)

Functions/xASL_adm_RemoveDirectories.m
MichaelStritt: Yeah, but you get like a hundred warnings and I just wanted to make sure that the pipeline is not killed because of this script which is only there to improve test independence.

Testing/xASL_test_IntegrationTesting.m
jan-petr: Agree. TSV is easy to import to Excel. Was there some special reason @MichaelStritt ?

Testing/xASL_test_IntegrationTesting.m
MichaelStritt: I think this isn't the case for all flavors. Jan is the expert here though. I just wanted to create an initial testing script where you can run each flavor from start to finish with less set up and stuff like in Jans version, but his one probably works better.

Testing/xASL_test_IntegrationTesting.m
MichaelStritt: I don't know... I just like it because I can double click open it with excel or libre office and it's nicer too look at than a tsv text editor. This isn't part of the pipeline where we have to be 100% bids compliant. We had the exact same discussion with the previous versions a few months ago as well though, I think we just have a different opinion here.

Functions/xASL_adm_RemoveDirectories.m
jan-petr: I am a bit confused:)
It is OK to switch off warnings for rmpath - makes total sense to avoid too much unnecessary warnings - so just add a comment explaining in the code why you do this.

But the question was mostly the try-catch - why is that needed. Can theoretically `warning` crash or `rmpath` crash - opinion of Henk is that these cannot really crash, or can they?

Testing/xASL_test_IntegrationTesting.m
jan-petr: Good point. This is not really needed. Some datasets have T1. And there were 1-2 datasets where the bUseMNIasDummy option was needed. But I have added the DataPar to the flavors with this option on. So for the purpose of testing flavors - we don't need it.

And we should probably exclude it so that it doesn't use MNI for those data that have T1.

We definitively do **not** want to remove T1.niis from flavors, because we want to test the import of them as well!!!

Testing/xASL_test_IntegrationTesting.m
jan-petr: ah - true. That's a test only. Fine to keep XLSX. Actually - my editor opens Excel with tsvs as well - that's why I don't care that much :D

Functions/xASL_adm_RemoveDirectories.m
MichaelStritt: > Can theoretically warning crash or rmpath crash - opinion of Henk is that these cannot really crash, or can they?

I actually wasn't sure about that, but I thought better safe than sorry. Didn't knew you had so strong opinions on this :D

43f5305 👍

Testing/xASL_test_IntegrationTesting.m
MichaelStritt: > We definitively do not want to remove T1.niis from flavors, because we want to test the import of them as well!!!

Agree, that would make the test database worse.

> And we should probably exclude it so that it doesn't use MNI for those data that have T1.

Oh, sorry, didn't knew that you added a dataPar.json for the ones without T1.


f668525 👍

Testing/xASL_test_IntegrationTesting.m
```diff
@@ -132,7 +132,6 @@
     xASL_delete(fullfile(testingRoot,'FlavorDatabaseTest',flavorName,'rawdata'),1);
     
     % Create default dataPar.json
-    defaultDataPar.x.bUseMNIasDummyStructural = 1; % when no structural data, use ASL-MNI registration
```
HenkMutsaerts: This parameter seems necessary for the flavor database, which hasn't got structural images for many flavors.

HenkMutsaerts (changes requested):

Testing/xASL_test_IntegrationTesting.m
MichaelStritt: > Good point. This is not really needed. Some datasets have T1. And there were 1-2 datasets where the bUseMNIasDummy option was needed. But I have added the DataPar to the flavors with this option on.

Are you sure about that? For example the first one (**GE_PCASL_2DEPI_volunteer**) does not have a T1 scan **and** there is no `dataPar.json` with the `x.bUseMNIasDummyStructural = 1;` setting.

jan-petr: @MichaelStritt very good point :D Sorry for the confusion. I just did remember that setting this parameter wasn't necessary and that everything was running without it. But that's because we already have it in the xASL_test_Flavors.m :) I didn't realize that we need it in that other function because it is another unrelated function :D And that other parameter I was talking about is something else.

So indeed, we need to have this :) Easy solution - I will just remove your last commit and then we can merge everything. I will do that to avoid conflicts :)

Testing/xASL_test_IntegrationTesting.m
jan-petr: You are of course right. My fault of insisting that Michael should remove it... I'm reverting this change.

HenkMutsaerts (approved):

Issue #603, opened 2021-05-27
ExploreASL version in Legacy data


jan-petr:
### Description

Currently, we do no specify which version of ExploreASL was used to do the import.

Edit Michael: Notes about BIDS specification of GeneratedBy [here](https://bids-specification.readthedocs.io/en/stable/03-modality-agnostic-files.html)

### Requirements

Upon BIDS2Legacy conversion, we should add a JSON-field GeneratedBy with subfields Name and Version including ExploreASL and version matching the ExploreASL version - careful about the BETA. Do this both in old and new import.

### Release notes

ExploreASL version specified in JSON files after BIDS to Legacy conversion

MichaelStritt:
[BIDS specification](https://bids-specification.readthedocs.io/en/stable/03-modality-agnostic-files.html): 

> As for any BIDS dataset, a dataset_description.json file MUST be found at the top level of the a derived dataset: `<dataset>/derivatives/<pipeline_name>/dataset_description.json`
> 
> In addition to the keys for raw BIDS datasets, derived BIDS datasets include the following REQUIRED and RECOMMENDED dataset_description.json keys:

> 
> GeneratedBy | REQUIRED | array of objects | Used to specify provenance of the derived dataset. See table below for contents of each object.
> -- | -- | -- | --
> 
> 
> 
> 
>

MichaelStritt:
I'll start to fix this within the #600 branch to avoid unnecessary conflicts 👍

MichaelStritt:
Initial fix:
![image](https://user-images.githubusercontent.com/46593074/120104305-1776b000-c154-11eb-8dde-50abaa547930.png)

PR #610, opened 2021-05-30
#603 xASL_bids_BIDS2Legacy: Add dataset_description.json

MichaelStritt:
### Linked issue

Check out #603

### How to test

Just run the import workflow, you should now get a dataset_description.json within derivatives/ExploreASL

### Comments

None

### DIFF
--- a/Functions/xASL_bids_BIDS2Legacy.m
+++ b/Functions/xASL_bids_BIDS2Legacy.m
@@ -309,8 +309,12 @@ try
     if size(jsonFiles,1)>0
         % Iterate over files
         for iFile = 1:size(jsonFiles,1)
-            thisPath = jsonFiles{iFile};
-            xASL_bids_AddGeneratedByField(x, thisPath);
+            % Check if file should be excluded first (exclude participants.json)
+            [~,fileName] = xASL_fileparts(jsonFiles{iFile});
+            if ~strcmp(fileName,'participants')
+                thisPath = jsonFiles{iFile};
+                xASL_bids_AddGeneratedByField(x, thisPath);
+            end
         end
     end
 catch
--- a/Functions/xASL_bids_BIDS2Legacy.m
+++ b/Functions/xASL_bids_BIDS2Legacy.m
@@ -311,7 +311,7 @@ try
         for iFile = 1:size(jsonFiles,1)
             % Check if file should be excluded first (exclude participants.json)
             [~,fileName] = xASL_fileparts(jsonFiles{iFile});
-            if ~strcmp(fileName,'participants')
+            if ~strcmpi(fileName,'participants')
                 thisPath = jsonFiles{iFile};
                 xASL_bids_AddGeneratedByField(x, thisPath);
             end
--- a/Functions/xASL_bids_BIDS2Legacy.m
+++ b/Functions/xASL_bids_BIDS2Legacy.m
@@ -297,22 +297,24 @@ elseif xASL_exist(fullfile(pathStudy, 'Participants.tsv'),'file')==2
     xASL_Copy(fullfile(pathStudy, 'Participants.tsv'),fullfile(pathLegacy, 'Participants.tsv'));
 end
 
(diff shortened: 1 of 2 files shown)


Functions/xASL_bids_BIDS2Legacy.m
```diff
@@ -295,7 +297,25 @@
...
+
+% Check rawdata directory
+if xASL_exist(fullfile(pathStudy, 'rawdata'),'dir')==7
+    % Check dataset_description.json
+    if xASL_exist(fullfile(pathStudy, 'rawdata', 'dataset_description.json'),'file')==2
+        % Load dataset_description.json
+        datasetDescription = spm_jsonread(fullfile(pathStudy, 'rawdata', 'dataset_description.json'));
+        % Add "GeneratedBy" field
+        datasetDescription.GeneratedBy = struct;
+        datasetDescription.GeneratedBy.Name = 'xASL-BIDS';
```
jan-petr, line 310: Thinking if `xASL-BIDS` is the name we want. It might sound as `conversionASL-BIDS` not making the link to ExploreASL. Plain `ExploreASL` might be better. @HenkMutsaerts what do you think?

Functions/xASL_bids_BIDS2Legacy.m
```diff
@@ -32,7 +33,8 @@
 % 9. Parse M0
 % 10. Create DataPar.json
 % 11. Copy participants.tsv
-% 12. Clean up
+% 12. Add dataset_description.json
```
jan-petr, line 36: Theoretically, enough to have it there. But practically, we mostly need it on the subject level - written to all ASL and M0 JSONs, because at the individual file level, we will always check what ExploreASL version was used.

Our motivation here is that we currently remove and apply the Philips scalings already when doing DCM2BIDS conversion. But we leave the SIemens and GE scalings only in the quantification module, because they are not really stored directly in DICOM. But for ASL-BIDS, we wanted to remove any scalings, independent of what is their origin. So once we implement that, we want to be sure what version of ExploreASL was used to create data - data generated by the old version will apply these scalings still in quantification, data generated by the newer version will apply the scalings already in import - as promised in ASL-BIDS.

jan-petr (changes requested):
Good. But we want the same thing also in ASL.json and M0.json of derivatives - see explanation in the comments. Should be an easy fix.

Functions/xASL_bids_BIDS2Legacy.m
MichaelStritt: Sure, changed to ExploreASL in e7ee311 👍

Functions/xASL_bids_BIDS2Legacy.m
MichaelStritt: Sounds good. It's just added to all JSONs now in e7ee311 👍 
I can remove it from the dataPar.json if you want, but it probably doesn't hurt either.

jan-petr: That's an elegant way to add it. We indeed want it in all JSON data sidecars. We want it in `dataset_description`, it doesn't hurt in `dataPar.json`. We probably don't want it in `participants.json` as this should have fields describing columns and an extra field that doesn't match the structure can be confusing. 
See the [BIDS-modality-agnostic-files.html](https://bids-specification.readthedocs.io/en/stable/03-modality-agnostic-files.html). 

What do you think?

jan-petr (changes requested):
One minor discussion point.

Functions/xASL_bids_BIDS2Legacy.m
MichaelStritt: Agree, just read about it. I exclude it in 3afe366  👍

jan-petr (approved):
OK to merge. I made a minor commit. Note that I changed the commit message so I had to remove the commit and force-pushed it again - so hopefully you didn't pull inbetween...

Anyway - extra caution during merging :)

Issue #602, opened 2021-05-27
Remove the bClone2Source option in Import


MichaelStritt:
### Description

Do we still need `xASL_imp_DCM2NII_Subject_CopyTempDir`? Copying to a `source` directory does not seem to make sense anymore, but maybe I'm wrong. @HenkMutsaerts found this during the ABBA meeting. Can you help here @jan-petr ?

### Requirements

Remove the bClone2Source option and all related functions.

### How to test

Re-run Flavors

### Release notes

Remove option for cloning the NIfTI output after import as BIDS directory fulfills the function of a read-only archive.

jan-petr:
I have checked that - I have created that function but I just copy pasted code that was there already. I do not think I was the one creating the original. It seems to be all related to the bClone2Source parameter that clones the NIfTIs after import. I think that the original function was to clone all after import as the analysis folder was then modified by processing and we wanted to have a clean copy right after import. This doesn't make any sense anymore as everything is stored in BIDS which acts as a read-only archive of freshly imported data by itself. 

So I propose to remove this bClone2Source option and all related functions. Any further comments to this @MichaelStritt @HenkMutsaerts ?

MichaelStritt:
> This doesn't make any sense anymore as everything is stored in BIDS which acts as a read-only archive of freshly imported data by itself.

Agree. That's what Henk and I found out as well.

jan-petr:
@MichaelStritt I have also renamed the import submodule 'Anonymize' to 'Deface'. Historically, we wanted to have both anonymization and defacing, but only defacing is implemented now. And any other anonymization really does not make much sense because private fields are removed already when going to BIDS.

I have tried to fix all the function names, messages, and comments. Any other idea what to fix? We should also port all these to the documentation - @MichaelStritt can you please do that?

MichaelStritt:
> I have tried to fix all the function names, messages, and comments. Any other idea what to fix? We should also port all these to the documentation - @MichaelStritt can you please do that?

The documentation should mostly adapt automatically. We need to change a few lines in the tutorials, but that's not that much work and does not need to be done within the ExploreASL repository.

jan-petr:
> > I have tried to fix all the function names, messages, and comments. Any other idea what to fix? We should also port all these to the documentation - @MichaelStritt can you please do that?
> 
> The documentation should mostly adapt automatically. We need to change a few lines in the tutorials, but that's not that much work and does not need to be done within the ExploreASL repository.

[...]

PR #617, opened 2021-06-01
Optimize #602 remove b clone2 source

jan-petr:
### Linked issue

#602 

### How to test

Test by running flavors - enough to run DCM2BIDS - further testing not needed and might work only after merging all the other current branches that all work with the import.


MichaelStritt (approved):
Seems fine to me 👍

MichaelStritt:
### Tests

I tested using two GE flavors, everything seems fine in the comparison step of the BIDS full pipeline:

```
                    ## |                                                                      
                    ## |                                                                      
                    ##/  

Dataset Root        
Import Modules      
Process Modules     
bPause              False
iWorker             1
nWorkers            1
ExploreASL v1.7.0_BETA initialized ... 
Dataset: GE_PCASL_3Dspiral_Product_pharma
Dataset: GE_PCASL_3Dspiral_volunteer
```

Functions/xASL_init_printSettings.m
```diff
@@ -23,7 +23,7 @@ function xASL_init_printSettings(x)
     %% Fallbacks
```
HenkMutsaerts, line 23: Are these fallbacks the printed lines? Then I would make this clear in the variable name.
Also, the printed lines can be made a bit more reader friendly. `NII` = `NIfTI`, and `NIfTI` is actually `legacy` right? Why is `DEFACE` big caps and `Structural` not? The `NIfTI` phase should be fased out soon anyway, so I wouldn't care too much about that naming :)

Modules/SubModule_Import/xASL_imp_Deface.m
```diff
@@ -17,22 +17,20 @@ function xASL_imp_Anonymize(imPar)
...
-% EXAMPLE:     xASL_imp_Anonymize(imPar);
+% EXAMPLE:     xASL_imp_Deface(imPar);
 % __________________________________
 % Copyright 2015-2021 ExploreASL
 
-    %% Anonymize
-    
-    % 1. Iterate over list of subjects
+    %% 1. Iterate over list of subjects
     listSubjects = xASL_adm_GetFileList(imPar.TempRoot,[],false,[],true);
```
HenkMutsaerts, line 25: Not important for now as I guess nobody uses this function (yet), but this doesn't give any result, it looks for folders without a name in the `TempRoot` folder :) Or does the second input argument default to something (it shouldn't I believe).

HenkMutsaerts (changes requested):
Few minor issues but looks good overall!

Modules/SubModule_Import/xASL_imp_Deface.m
MichaelStritt: There really seems to be a bug related to all the renaming and stuff. Another issue here is that we have no nice way to test if the defacing happened or not, yet.

MichaelStritt: Fix of the first problem: 490c59c

Functions/xASL_init_printSettings.m
jan-petr: You mean NII in DCM2NII - this is not legacy, this is the middle step of DCM->NII->BIDS. Although the Legacy format is used there, we still need this step. In the future, we will probably try to do this smarter, but we will step have it.

My proposed solution:
- Put also `STRUCTURAL` and `POPULATION` in capitals.
- Keep NII.

Agreed?

Modules/SubModule_Import/xASL_imp_Deface.m
MichaelStritt: Fix of the bug: e94cab3

jan-petr: Weird. It should of course work. But maybe something happened with the directories during all the revamping. This is not a function we were testing. I will fix this.

jan-petr: Seems that Michael already did that :D

jan-petr: Tested and all works. Good work Michael.

Functions/xASL_init_printSettings.m
jan-petr: Done - do you agree?

MichaelStritt: ### Test

```
                    ## |                                                                      
                    ## |                                                                      
                    ##/  

Dataset Root        ...ATLAB\m.stritt\Server_xASL\New_TestData\Philips_PCASL_3DGRASE_volunteer
Import Modules      DCM2NII NII2BIDS DEFACE BIDS2LEGACY
Process Modules     STRUCTURAL ASL POPULATION 
bPause              False
iWorker             1
nWorkers            1
```

👍

HenkMutsaerts (changes requested):
Cannot find the `fallback` variable anymore, but where I see the word `fallback` this should be `default` instead right? Just to make it understandable. Comments should explain what is happening, not only from a debugger's perspective :)

So let's just pay a bit more attention to commenting, otherwise this thing is really going to bite our back soon. Otherwise it looks good!

jan-petr:
> Cannot find the `fallback` variable anymore, but where I see the word `fallback` this should be `default` instead right? Just to make it understandable. Comments should explain what is happening, not only from a debugger's perspective :)
> 
> So let's just pay a bit more attention to commenting, otherwise this thing is really going to bite our back soon. Otherwise it looks good!

I have removed `fallbacks` completely :) Because the default value was "", so I just reorganized the script a bit that we don't even need those.

HenkMutsaerts:
> > Cannot find the `fallback` variable anymore, but where I see the word `fallback` this should be `default` instead right? Just to make it understandable. Comments should explain what is happening, not only from a debugger's perspective :)
> > So let's just pay a bit more attention to commenting, otherwise this thing is really going to bite our back soon. Otherwise it looks good!
> 
> I have removed `fallbacks` completely :) Because the default value was "", so I just reorganized the script a bit that we don't even need those.

It said "fallback" somewhere in the comment ;) If you change that to "default" then I am happy right?

jan-petr:
> It said "fallback" somewhere in the comment ;) If you change that to "default" then I am happy right?

Yes exactly. Actually the 'fallback'/'default' setting is now completely removed. As we directly print stuff, without even setting the variable to something - so that comment was implemented and no more relevant.

Read to approve?!

HenkMutsaerts (approved):

Issue #618, opened 2021-06-01
ASLCheck- and M0check-files missing session name


jan-petr:
### Description

Certain files in the ASLCheck and M0check directories have only subject name and not session name in their filename. 

### Requirements

This is true for all M0check files and only for a half of ASLcheck files. This can be easily fixed by copying what is done for the ASLcheck files to all of them. See the screenshots below to see this.

### How to test

Run on a single dataset to see that this is fixed.

### Release notes

Add session name to all M0Check and ASLCheck QC files in the Population folder.

jan-petr:
![image](https://user-images.githubusercontent.com/29886537/120294199-42811100-c2c6-11eb-8414-15a2462ddc42.png)

![image](https://user-images.githubusercontent.com/29886537/120294269-52005a00-c2c6-11eb-8d55-f1c306286c3d.png)

HenkMutsaerts:
Yes it looks like this should be consistent. Not sure if the `ASL_1` is omitted if there is only one session though, or if it is always missing in some examples?

jan-petr:
See my fix - it is rather simple, but effective. It makes the filename longer, so we could go later for a more complex strategy...

jan-petr:
Also - I made a small fix that the master script will work with `DataPar.json` and not only `DataPar.json` as this would be bad for many people to notice. And I have improved the warning that now explains this better.

PR #622, opened 2021-06-01
Bug #618 as land m0check filenames

jan-petr:
### Linked issue

#618 

### How to test

Simple running of the pipeline should be enough to test.

### DIFF
--- a/Functions/xASL_vis_CreateVisualFig.m
+++ b/Functions/xASL_vis_CreateVisualFig.m
@@ -345,10 +345,7 @@ if numel(DirOut)>1 && numel(ImOut)>1 % when DirOut or ImOut==NaN, skip this
     xASL_adm_CreateDir(DirOut);
 	FileName = NamePrefix;
     for iC=1:length(Ffile)
-		if iC>1
-			FileName = [FileName '_'];
-		end
-        FileName = [FileName Ffile{iC}];
+        FileName = [FileName '_' Ffile{iC}];
     end
     OutputFile = fullfile(DirOut,[FileName '.jpg']);
     xASL_vis_Imwrite(ImOut, OutputFile);
--- a/Functions/xASL_vis_CreateVisualFig.m
+++ b/Functions/xASL_vis_CreateVisualFig.m
@@ -345,7 +345,13 @@ if numel(DirOut)>1 && numel(ImOut)>1 % when DirOut or ImOut==NaN, skip this
     xASL_adm_CreateDir(DirOut);
 	FileName = NamePrefix;
     for iC=1:length(Ffile)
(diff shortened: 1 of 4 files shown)


MichaelStritt (approved):

Functions/xASL_vis_CreateVisualFig.m
```diff
@@ -343,8 +343,12 @@
 % requested. If DirOut is provided, this image is also saved as jpg file
 if numel(DirOut)>1 && numel(ImOut)>1 % when DirOut or ImOut==NaN, skip this
     xASL_adm_CreateDir(DirOut);
+	FileName = NamePrefix;
```
HenkMutsaerts, line 346: How does this work when `pWM` is overlaid over `qCBF`, this would become `qCBF_pWM_sub-001_ASL_1.jpg` right? That sounds good, though I wonder how this works with other functions that may expect the previous naming convention, did we test this?

HenkMutsaerts (changes requested):
Small question, otherwise fine

Functions/xASL_vis_CreateVisualFig.m
jan-petr: it gives pWM_sub_001_qCBF_sub_001_ASL_1.jpg - so there is some redundancy.

We can try to be a bit more smart to remove this redundancy, but I don't see it as a disadvantage to have full names of both. Because previously we have only the name of the last file and that was not really helpful sometimes.

Previous naming convention? We only save those to disk, we don't reload them. I don't see a conflict there. YOu can test and see if you like it...

HenkMutsaerts: No the prefix dealth with this right? The prefix would be `pWM` in this case. I like your approach, which is smarter, but you cannot do this easily without testing... If 1.7.0 needs to be out soon I would move this to 1.8.0

jan-petr: The prefix dealt with it, but it didn't have the patient session. So I kept the prefix and added the additional names. 

So it works, but it is a but superfluous. But that's not an issue, because these files are written and nowhere read again?! Right, so it can't cause any conflict, or can it?

I actually really want it in 1.7.0 for Wibeke. They have multi-session ASL, and we can't check the registration of each session, because it always overwrites it with the latest session. So rather annoying bug

jan-petr: Consensus solution (To be implemented by Jan). We keep names of all files. We check where this function is used and remove the pWM part of the prefix - on all occurrences. Make sure that all the underscores are correct. Then we can merge.

HenkMutsaerts (approved):

Issue #632, opened 2021-06-09
Improve untouched nifti comparison


MichaelStritt:
### Description

Move the code for the nifti comparison from `xASL_bids_CompareStructures` to a separate script. Improve code quality and generalize the functionality.

### Requirements

Comparison script for untouched nifti comparison based on `xASL_Copy & _Delete`.

### How to test

Run `[identical,RMSE,minDiff,maxDiff,dimCheck] = xASL_im_CompareNiftis(pathA,pathB,bVerbose)` on two niftis.

- If the images are not 100% identical, identical should be false
- If the matrix dimensions agree, RMSE should contain the root mean square error
- If the matrix dimensions agree, minDiff and maxDiff should contain double values
- If the matrix dimensions do not agree, dimCheck should be false

### Release notes

Added comparison script for untouched nifti comparison.

PR #633, opened 2021-06-09
#632 Compare niftis

MichaelStritt:
### Linked issue

Check out the detailed explanations in #632

### DIFF
--- a/Functions/xASL_bids_CompareStructures.m
+++ b/Functions/xASL_bids_CompareStructures.m
@@ -198,7 +198,7 @@ function [identical,results] = xASL_bids_CompareStructures(pathDatasetA,pathData
     
     
     % Compare file content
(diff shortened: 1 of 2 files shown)


jan-petr (approved):
Nice change. But I didn't like that it printed the RMSE also within flavor test when there was no difference. So I have activated that only when `detailedOutput` is on. Because I like to check directly on the screen if the import went fine and there are no errors - without going to the tsv or xls. But if you still want to have that - no problem ;)

Otherwise - if you agree with the change, we can merge...

MichaelStritt:
> 
> 
> Nice change. But I didn't like that it printed the RMSE also within flavor test when there was no difference. So I have activated that only when `detailedOutput` is on. Because I like to check directly on the screen if the import went fine and there are no errors - without going to the tsv or xls. But if you still want to have that - no problem ;)
> 
> Otherwise - if you agree with the change, we can merge...

No, that's perfect 👍 
I just wanted to put it in a separate script because I saw that @BeatrizPadrela recently needed a script which basically does exactly this :)

Issue #631, opened 2021-06-09
Warning for M0 repeated multiple times for multiple runs


BeatrizPadrela:
### Description

The warning ""Caution when usgin contol as M0, background supression was applied" is given multiple times, when multiple runs are being converted from BIDs to Legacy.

### Requirements
If the warning is given once, we don't want it to be repeated for the other runs.

jan-petr:
This is both in `xASL_bids_parseM0` and in `bids.layout>parse_perf` so I have removed it from the xASL function.

MichaelStritt:
consensus abba meeting: @BeatrizPadrela could check this again and find out if it behaves correctly now. could we check if the warning appeared already using the warning identifier or is it a custom warning?

PR #635, opened 2021-06-09
#631 Removed M0 Bsup warning from xASL_bids_parseM0

jan-petr:
### Linked issue

#631 

### How to test

Run Flavors to see if this doesn't appear twice anymore - tested by Jan.

### DIFF
--- a/Functions/xASL_bids_parseM0.m
+++ b/Functions/xASL_bids_parseM0.m
@@ -78,9 +78,6 @@ if isfield(JSON, 'M0Type')
         % which happens in xASL_wrp_Resample
         case {'use_control_as_m0', 'Absent'}
             if isfield(JSON, 'BackgroundSuppression')
-                if JSON.BackgroundSuppression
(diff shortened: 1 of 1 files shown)


HenkMutsaerts:
This looks good, I asked @BeatrizPadrela to check if this warning is not repeated anymore. In fact, this may be a general problem, what to do if the same warning is mentioned 100 times when you have 100 subjects. Perhaps we could think of a general transparent solution for this?

HenkMutsaerts:
I see now that I'm confusing this issue with the other... So this comment should be moved then (typing this in my bed from my phone... Tried the laptop in bed but Kyra gets mad (that rhymes)

BeatrizPadrela:
> I see now that I'm confusing this issue with the other... So this comment should be moved then (typing this in my bed from my phone... Tried the laptop in bed but Kyra gets mad (that rhymes)

No this is the right issue, but Jan rmeoved the wrong warning

HenkMutsaerts (approved):

PR #642, opened 2021-06-12
#627 xASL_bids_BIDS2Legacy: print warnings when catched

### LINKED ISSUE #627: BIDS2Legacy incomplete information when crashing
### Description

If a BIDS dataset is invalid, `BIDS2Legacy` may crash. However, the reason why it crashes - e.g., "`BackgroundSuppression` field is missing" - is catched and not shown to the user . It is helpful to the user to show the reason why this function crashed. Also, when this part crashes, it's useful to show a message if the BIDS validator was run.

### How to test

Remove a BIDS fiels and BIDS2Legacy should crash and show you why it crashed.

HenkMutsaerts:
### Linked issue

Closes #627

### DIFF
--- a/External/bids-matlab/+bids/layout.m
+++ b/External/bids-matlab/+bids/layout.m
@@ -159,6 +159,8 @@ function BIDS = layout(root, tolerant)
     end
   end
   fprintf('\n');
+  fprintf('Note that any warnings may have only printed once if they were repeated for multiple scans\n');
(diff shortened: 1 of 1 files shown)


HenkMutsaerts:
@MichaelStritt this is all that was missing:) does that make sense to you?

MichaelStritt (approved):
Easy fix 👍

MichaelStritt:
I think something went wrong here :/

PR #645, opened 2021-06-12
Feature #643 bids.layout

### LINKED ISSUE #643: bids.layout Only print unique warnings
### Description

When you currently do the parsing of BIDS datasets with bids.layout, it will issue warnings for each subject/ASL scan. If each ASL scan has the same problem, then it will issue the same warning many times. This is likely to happen, since when working with populations, that often have the same or similar ASL scans, you add BIDS fields in batch.

### Requirements

For each ASL warning to be issued, check if this warning was already given. Don't throw the warning again if the warning was already provided.

Also, provide user feedback by tracking progress.

### Release notes

bids.layout: avoid printing the same warning repetitively, in case multiple scans in a data set have the same issue.

HenkMutsaerts:
### Linked issue

Closes #643

### DIFF
--- a/External/bids-matlab/+bids/layout.m
+++ b/External/bids-matlab/+bids/layout.m
@@ -159,6 +159,8 @@ function BIDS = layout(root, tolerant)
     end
   end
   fprintf('\n');
+  fprintf('Note that any warnings may have only printed once if they were repeated for multiple scans\n');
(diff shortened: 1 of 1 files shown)


MichaelStritt (approved):
I really like the fix, we can use that in other places for ExploreASL in general 👍 

Did you test it already?

PR #647, opened 2021-06-12
Bug #646 load failure

### LINKED ISSUE #646: Failure to load data
### Description

When running the following on `develop` branch:
`ExploreASL('/home/hjmutsaerts/lood_storage/divi/Projects/ExploreASL/ABBA/NOMARED',0,0);`
it initializes ExploreASL, but incorrectly skips loading the study in the folder that is provided. Why is this?
Also, how does it detect what to load, it should search for the `dataPar` in the `/derivatives/ExploreASL` subfolder right?

When I then run:
`ExploreASL('/home/hjmutsaerts/lood_storage/divi/Projects/ExploreASL/ABBA/NOMARED',0,1);`

It gives an incorrect warning:

`Neither the sourceStructure.json, dataset_description.json nor dataPar.json exist`

where this should (per modularity) only give a warning about the dataPar.json right?

The following warning states: `Use the study root directory instead...`
This should be `data set root directory...`

HenkMutsaerts:
### Linked issue

Closes #646

### DIFF
--- a/Functions/xASL_init_DetermineRequiredPaths.m
+++ b/Functions/xASL_init_DetermineRequiredPaths.m
@@ -21,22 +21,32 @@ function [x] = xASL_init_DetermineRequiredPaths(x)
 
     %% Check the BIDS dataset root for the metadata JSON files
 
+    % Regular expressions used below, explained:
+    % ^ = expression must start with this
+    % $ = expression must end with this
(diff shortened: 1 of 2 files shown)


MichaelStritt (approved):
Ahh, I understand. So the case sensitivity was the problem?

Issue #655, opened 2021-06-15
xASL_adm_GetPopulationSessions gives incorrect warning


HenkMutsaerts:
### Description

`xASL_adm_GetPopulationSessions` was nicely written, but the check if the number of sessions differs between subjects is not working properly. It gave me a warning when I had two sessions. I did some code improvement (see first commit), and @MDijsselhof up to you to explain to me how this should work. My example was with a single subject, with two sessions, where `CountSessionNumbers` == [1 1] and `CompareSessions` = [2 2]. `nSessions` == 2, and `NewList` was a 2x1 cell array.

BeatrizPadrela:
This an example composed of 1 sub, 3 ses and 1 run for each sessions (BIDS notation), 
![image](https://user-images.githubusercontent.com/73699072/122352629-93903680-cf4f-11eb-95c6-0c709b010483.png)

MDijsselhof:
The function creates an array with the amount of sessions present in total. So in HJ's case the following should happen:
`CountSessionNumbers` = [1 1]
`nSessions'= 2
'NewList' = 2
'CompareSessions' = [2 2]

The problem originates from multiplying with 'nSessions' whereas this should be 'x.nSubjects'

PR #658, opened 2021-06-17
Bug #655 get population sessions warning

MDijsselhof:
#655 

### How to test

Run Population module for dataset with all subjects the same amount of sessions and subjects with different amount of sessions


jan-petr (approved):
OK

PR #671, opened 2021-06-19
#187 xASL_io_Nifti2Im: Load as uint8 int16 or single

### LINKED ISSUE #187: If xASL_io_Nifti2Im loads a mask, don’t use single() but logical()
This is just to speed things up, xASL_io_Nifti2Im will always by default convert an image matrix loaded from a NIfTI to a single-precision format. This is a trade-off between precision and memory usage (assuming that the quantification differences between single and double-precision formats are negligible).

### Release notes
Load NIfTI as UINT8 or INT16 if not floating point.

HenkMutsaerts:
### Linked issue

Closes #187


jan-petr (approved):
OK

jan-petr (approved):
OK.

Issue #666, opened 2021-06-19
Initialization should issue a warning if there are multiple dataPar*.json


HenkMutsaerts:
### Description

Initialization should issue a warning if there are multiple dataPar*.json in `/derivatives/ExploreASL`, and print which one it is using.
Or should it only issue the warning and request to move the other dataPar file out?

### Testing
/External/TestDataSet has this, when you skip the import and run the processing only.

### Release
Issue a warning when multiple dataPar*.json or studyPar*.json or sourcestructure*.json are present.

jan-petr:
> ### Description
> Initialization should issue a warning if there are multiple dataPar*.json in `/derivatives/ExploreASL`, and print which one it is using.
> Or should it only issue the warning and request to move the other dataPar file out?
> 
> ### Testing
> /External/TestDataSet has this, when you skip the import and run the processing only.

PR #673, opened 2021-06-19
#666 Warning on multiple DataPar.jsons

jan-petr:
### Linked issue

#666

### How to test

Add multiple dataPars or studyPars or sourcestructures and see that a warning is printed with clarifying what file is actually used.


MichaelStritt (approved):
Works 👍 

image

A bit weird how Matlab sorts the files, but the warning tells the user which one is used, so I think that's okay 👍 

image

Issue #687, opened 2021-07-01
Update Reference TSV File


MichaelStritt:
### Description

Update the reference TSV file after the release.

### Release Notes

Update reference values for pipeline testing.

PR #688, opened 2021-07-01
#687 ReferenceValues: Update

MichaelStritt:
### Linked issue

Check out #687

### Comments

Only a minor improvement. We can swap the Windows results with linux results or adapt the comparison later if necessary.


jan-petr (approved):
Very good.

Issue #566, opened 2021-05-10
Hadamard_MotionCorrection


BeatrizPadrela:
### Description

Rework on the motion correction module for Hadamard encoded data

### Requirements

Add modifications to xASL_wrp_RealignASL to estimate the motion of every first TE and apply the same motion for the other TEs.

PR #689, opened 2021-07-01
Feature #566 hadamard motion correction

MichaelStritt:
### Linked issue

Check out #566

### DIFF
--- a/Development/xASL_im_Move3DVolume.m
+++ b/Development/xASL_im_Move3DVolume.m
@@ -1,7 +1,7 @@
 function xASL_im_Move3DVolume(pathNifti, pathResult, dim, volNum, numVoxel)
 %xASL_im_MoveImage Move zero padded image in spatial direction.
 %
-% FORMAT:       xASL_im_Move3DVolume(pathNifti, pathResult, dim, numVoxel)
+% FORMAT:       xASL_im_Move3DVolume(pathNifti, pathResult, dim, volNum, numVoxel)
 % 
 % INPUT:        pathNifti (CHAR ARRAY, REQUIRED)
 %               pathResult (CHAR ARRAY, REQUIRED)
@@ -40,11 +40,11 @@ function xASL_im_Move3DVolume(pathNifti, pathResult, dim, volNum, numVoxel)
     %% Move 3D volume
     switch dim
         case 'x'
-            newVolume(1:numVoxel,:,:) = originalVolume(1:dimX-numVoxel,:,:);
+            newVolume(1+numVoxel:dimX,:,:) = originalVolume(1:dimX-numVoxel,:,:);
         case 'y'
-            newVolume(:,1:numVoxel,:) = originalVolume(:,1:dimX-numVoxel,:);
+            newVolume(:,1+numVoxel:dimY,:) = originalVolume(:,1:dimY-numVoxel,:);
         case 'z'
-            newVolume(:,:,1:numVoxel) = originalVolume(:,:,1:dimX-numVoxel);
+            newVolume(:,:,1+numVoxel:dimZ) = originalVolume(:,:,1:dimZ-numVoxel);
         otherwise
             error('Unknown dimension...');
     end
--- a/Development/xASL_im_Move3DVolume.m
+++ b/Development/xASL_im_Move3DVolume.m
@@ -1,0 +1,58 @@
+function xASL_im_Move3DVolume(pathNifti, pathResult, dim, volNum, numVoxel)
+%xASL_im_MoveImage Move zero padded image in spatial direction.
+%
+% FORMAT:       xASL_im_Move3DVolume(pathNifti, pathResult, dim, numVoxel)
+% 
+% INPUT:        pathNifti (CHAR ARRAY, REQUIRED)
+%               pathResult (CHAR ARRAY, REQUIRED)
+%               dim (CHAR, REQUIRED)
+%               volNum (INTEGER, REQUIRED)
+%               numVoxel (INTEGER, REQUIRED)
+%
+% OUTPUT:       NewList  - vertical list of things
+% 
+% -----------------------------------------------------------------------------------------------------------------------------------------------------
+% DESCRIPTION:  Move zero padded image in spatial direction. Was written to help with motion correction testing.
+%
+% -----------------------------------------------------------------------------------------------------------------------------------------------------
+% EXAMPLE:      n/a
+% __________________________________
+% Copyright 2015-2021 ExploreASL
+
+    %% Checks
+    if nargin<5
+        error('Missing input arguments...');
+    end
+
+    %% Load image
+    imageNifti = xASL_io_Nifti2Im(pathNifti);
+    originalVolume = imageNifti(:,:,:,volNum);
+
+    %% Create new volume
+    newVolume = zeros(size(originalVolume));
+
+    % Get image dimensions
+    dimensions = size(originalVolume);
+    dimX = dimensions(1);
+    dimY = dimensions(2);
+    dimZ = dimensions(3);
+
+    %% Move 3D volume
+    switch dim
+        case 'x'
+            newVolume(1:numVoxel,:,:) = originalVolume(1:dimX-numVoxel,:,:);
+        case 'y'
+            newVolume(:,1:numVoxel,:) = originalVolume(:,1:dimX-numVoxel,:);
+        case 'z'
+            newVolume(:,:,1:numVoxel) = originalVolume(:,:,1:dimX-numVoxel);
+        otherwise
+            error('Unknown dimension...');
+    end
+
+    %% Store volume
+    imageNifti(:,:,:,volNum) = newVolume;
+    xASL_io_SaveNifti(pathNifti, pathResult, imageNifti);
+    
+
+end
+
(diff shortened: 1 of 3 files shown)


BeatrizPadrela:
Ok, when inducing motion to the first volume, to clearly see if some motion correction was being done, it was not working at all. No motion correction was being applied even if the rp_file was okay.
So I tried to run the motion correction without this specific motion correction for multi-tE, so the normal motion correction approach, only with a single line that defines zigzag=false to multiPLD data, and it worked!!

(difference between volume 0 and volume 8 - before - ASL4D.nii)
image

(difference between volume 0 and volume 8 - after - rASL4D.nii)
image

[this was the way I found to show the difference between vol0 and vol8, but if we open the images with fsleyes is really clear that there's no motion after the normal motion correction]

Modules/xASL_module_ASL.m
```diff
@@ -81,6 +81,13 @@
     end
 end
 
+% Check if Hadamard data
+if isfield(x,'HadamardType')
```
jan-petr, line 85: This was a misunderstanding :) I mean that you simply define x.HadamardType = 0 if missing, not a new HadamardType variable :)

Modules/SubModule_ASL/xASL_wrp_RealignASL.m
```diff
@@ -94,8 +97,37 @@ function xASL_wrp_RealignASL(x,bSubtraction)
...
+%         save('rp_ASL4D.txt','MotionAllTEs','-ascii') %saves the rpfile again into .txt
+%         
+%     elseif numel(unique(x.Q.Initial_PLD))>1 % multiTE + multiPLD=Hadamard
+%         spm_realign(ImInfoFirstTEs,flags,false);
+%         MotionFirstTEs=load(rpfile);
+%         MotionAllTEs=repelem(MotionFirstTEs(:,:),NumTEs,1); 
+%         save('rp_ASL4D.txt','MotionAllTEs','-ascii')
+%     end  
+    
+if nFrames>2 && bSubtraction && (numel(unique(x.Q.Initial_PLD))>1 || numel(unique(x.EchoTime))>1 || HadamardType~=0) %multiPLD or multiTE or Hadamard
```
jan-petr, line 125: and then here x.HadamardType as well, because we have previously made sure that this is always set.

Modules/xASL_module_ASL.m
jan-petr: Having a duplicitous variable would make stuff confusing, and will also not be visible elsewhere ;)

jan-petr (changes requested):
Good. Except the HadamardType...

jan-petr (approved):
OK.

Development/xASL_im_Move3DVolume.m
```diff
@@ -0,0 +1,58 @@
+function xASL_im_Move3DVolume(pathNifti, pathResult, dim, volNum, numVoxel)
+%xASL_im_MoveImage Move zero padded image in spatial direction.
+%
+% FORMAT:       xASL_im_Move3DVolume(pathNifti, pathResult, dim, volNum, numVoxel)
+% 
+% INPUT:        pathNifti (CHAR ARRAY, REQUIRED)
```
HenkMutsaerts, line 6: @BeatrizPadrela You need to explain a bit here what the input arguments mean, not only their format.

Modules/SubModule_ASL/xASL_wrp_RealignASL.m
```diff
@@ -69,6 +69,9 @@ function xASL_wrp_RealignASL(x,bSubtraction)
 % Define realignment settings
 tempnii = xASL_io_ReadNifti(InputPath);
 nFrames = double(tempnii.hdr.dim(5));
+if length(x.EchoTime)>1
```
HenkMutsaerts, line 72: @BeatrizPadrela This could go wrong for non-multi-TE sequences. The `EchoTime` can also be (incorrectly) a vector in enhanced DICOM. So we need to add a boolean like `if bMultiEcho`

Modules/SubModule_ASL/xASL_wrp_RealignASL.m
```diff
@@ -94,8 +97,37 @@ function xASL_wrp_RealignASL(x,bSubtraction)
 xASL_delete(rpfile);
 
 % Run SPM
-if nFrames>2 && bSubtraction
+
+%####### Calculating the motion for every first TE of each PLD and
```
HenkMutsaerts, line 101: @BeatrizPadrela `####` is for languages like Bash, `%%%%` for Matlab :)

Modules/SubModule_ASL/xASL_wrp_RealignASL.m
```diff
@@ -94,8 +97,37 @@ function xASL_wrp_RealignASL(x,bSubtraction)
 xASL_delete(rpfile);
 
 % Run SPM
-if nFrames>2 && bSubtraction
+
+%####### Calculating the motion for every first TE of each PLD and
+% considering it the same for the other TEs from that PLD.
+%
+% if nFrames>2 && bSubtraction && length(x.EchoTime)>1  %Multi TE
```
HenkMutsaerts, line 104: @BeatrizPadrela So this should also start with something like `if bMultiEcho`. Makes it less prone to error and also easier to read

Modules/SubModule_ASL/xASL_wrp_RealignASL.m
```diff
@@ -94,8 +97,37 @@ function xASL_wrp_RealignASL(x,bSubtraction)
 xASL_delete(rpfile);
 
 % Run SPM
-if nFrames>2 && bSubtraction
+
+%####### Calculating the motion for every first TE of each PLD and
+% considering it the same for the other TEs from that PLD.
+%
+% if nFrames>2 && bSubtraction && length(x.EchoTime)>1  %Multi TE 
+%     uniqueTE=uniquetol(x.EchoTime); %gives the number of unique TEs
```
HenkMutsaerts, line 105: @BeatrizPadrela Should we really use `uniquetol` here or would `unique` be better? We haven't specified the tolerance here...

Modules/SubModule_ASL/xASL_wrp_RealignASL.m
```diff
@@ -94,8 +97,37 @@ function xASL_wrp_RealignASL(x,bSubtraction)
...
-if nFrames>2 && bSubtraction
+
+%####### Calculating the motion for every first TE of each PLD and
+% considering it the same for the other TEs from that PLD.
+%
+% if nFrames>2 && bSubtraction && length(x.EchoTime)>1  %Multi TE 
+%     uniqueTE=uniquetol(x.EchoTime); %gives the number of unique TEs
+%     NumTEs=numel(uniqueTE);
+%     minTE=min(uniqueTE); 
+%     positionMinTE=find(x.EchoTime == minTE); %positions that have the min TE
```
HenkMutsaerts, line 108: @BeatrizPadrela At some locations in this script we use `TE` at other locations `EchoTime`, though they mean the same right? Would be good to homogenize for readability.

Modules/SubModule_ASL/xASL_wrp_RealignASL.m
```diff
@@ -94,8 +97,37 @@ function xASL_wrp_RealignASL(x,bSubtraction)
...
+%     ImInfo=spm_vol(InputPath);
+%     ImInfoFirstTEs=ImInfo(positionMinTE);
+%     
+%     if numel(unique(x.Q.Initial_PLD))==1 %multiTE + single PLD
+%         spm_realign(ImInfoFirstTEs,flags,true);
+%         MotionFirstTEs=load(rpfile);
+%         MotionAllTEs=repelem(MotionFirstTEs(:,:),NumTEs,1); %repeats each row NumTEs times
```
HenkMutsaerts, line 115: @BeatrizPadrela Is this also possible with `repmat`? Is a bit more backward compatible.

Modules/SubModule_ASL/xASL_wrp_RealignASL.m
```diff
@@ -94,8 +97,37 @@ function xASL_wrp_RealignASL(x,bSubtraction)
...
+%         save('rp_ASL4D.txt','MotionAllTEs','-ascii') %saves the rpfile again into .txt
+%         
+%     elseif numel(unique(x.Q.Initial_PLD))>1 % multiTE + multiPLD=Hadamard
+%         spm_realign(ImInfoFirstTEs,flags,false);
+%         MotionFirstTEs=load(rpfile);
+%         MotionAllTEs=repelem(MotionFirstTEs(:,:),NumTEs,1); 
+%         save('rp_ASL4D.txt','MotionAllTEs','-ascii')
```
HenkMutsaerts, line 122: `;` missing

Modules/SubModule_ASL/xASL_wrp_RealignASL.m
```diff
@@ -109,9 +141,15 @@ function xASL_wrp_RealignASL(x,bSubtraction)
 MeanRadius = 50; % typical distance center head to cerebral cortex (Power et al., NeuroImage 2012)
```
HenkMutsaerts, line 141: @BeatrizPadrela Apparently, this needs adaptation in some cases. Something for the future :)

Issue #690, opened 2021-07-01
Add new unit tests


MichaelStritt:
### Description

Add new unit tests for some basic scripts.

### Release notes

Improved unit testing.

Issue #574, opened 2021-05-13
Unit test backwards compatibility


MichaelStritt:
### Description

Update unit tests regarding backwards compatibility. Functions like `isfolder` are new and should be replaced by e.g. exist('', 'dir') for backward compatibility and increased reliability (newer functions could theoretically be more error prone as they are less tested).

### Requirements

Optional: add feature requirements here

### How to test

Run `xASL_ut_UnitTesting.m` on the AMC FLUX server. Check the results and if some of the tests contain bugs.

### Release notes

Update unit tests regarding backwards compatibility

MichaelStritt:
I think we already fixed this, but we should re-evaluate.

MichaelStritt:
Run it once at AMC using 2016a or b. Should be fixed already.

MichaelStritt:
Fixed in #690

PR #691, opened 2021-07-01
Test #690 unit tests

MichaelStritt:
### Linked issue

Check out #690

### Comments

Added some small unit tests.


MichaelStritt:
### QC Meeting (Henk & Michael)

- standardize unit test names
- get test name from script name
- get test file type form script name as well

- `xASL_spm_reslice`

MichaelStritt:
### Quick test run

```matlab
====================================== TEST RESULTS ======================================

                        name                           unit          tests        passed
    ____________________________________________    __________    ____________    ______

    'xASL_adm_CatchNumbersFromString'               'function'    [1×3 struct]    true  
    'xASL_adm_CheckFileCount'                       'function'    [1×2 struct]    true  
    'xASL_adm_CompareLists'                         'function'    [1×2 struct]    true  
    'xASL_adm_CorrectName'                          'function'    [1×2 struct]    true  
    'xASL_adm_OrderFields'                          'function'    [1×1 struct]    true  
    'xASL_bids_BIDS2Legacy'                         'function'    [1×2 struct]    true  
    'xASL_bids_Config'                              'function'    [1×1 struct]    true  
    'xASL_bids_CreateDatasetDescriptionTemplate'    'function'    [1×2 struct]    true  
    'xASL_bids_JsonCheck'                           'function'    [1×3 struct]    true  
    'xASL_bids_VendorFieldCheck'                    'function'    [1×1 struct]    true  
    'xASL_io_ExportVTK'                             'function'    [1×5 struct]    true  
    'xASL_io_Nifti2Im'                              'function'    [1×1 struct]    true  
    'xASL_stat_PSNR'                                'function'    [1×2 struct]    true  
    'xASL_stat_QuantileNan'                         'function'    [1×1 struct]    true  
    'xASL_stat_StdNan'                              'function'    [1×1 struct]    true  
    'xASL_str2num'                                  'function'    [1×2 struct]    true  
    'xASL_test_GetLogContent'                       'function'    [1×2 struct]    true  
    'xASL_tsvRead'                                  'function'    [1×2 struct]    true  
    'xASL_tsvWrite'                                 'function'    [1×2 struct]    true  

==========================================================================================
```

MichaelStritt:
### Test AMC (for backwards compatibility #574 )

image

Testing/UnitTests/xASL_ut_function_xASL_adm_CheckFileCount.m
```diff
@@ -30,6 +30,12 @@
 fid2 = fopen(fullfile(TestRepository,'UnitTesting','working_directory','test_2.txt'), 'wt');
 fid3 = fopen(fullfile(TestRepository,'UnitTesting','working_directory','test_3.txt'), 'wt');
 
+try
+    fclose('all');
+catch
```
HenkMutsaerts, line 35: also print message if you catch error

Testing/UnitTests/xASL_ut_function_xASL_bids_BIDS2Legacy.m
```diff
@@ -44,7 +44,7 @@
 if ~exist(fullfile(droTestPatient,'derivatives'),'dir')
     testCondition = false; % Test failed
 end
-if ~exist(fullfile(droTestPatient,'derivatives','ExploreASL','DataPar.json'),'file')
+if ~exist(fullfile(droTestPatient,'derivatives','ExploreASL','dataPar.json'),'file')
```
HenkMutsaerts, line 47: I guess here it strictly cannot be `dataPar_somethingelse.json`?

Testing/UnitTests/xASL_ut_function_xASL_adm_CheckFileCount.m
```diff
@@ -26,12 +26,9 @@
 testTime = tic;
 
 % Create three test files in test directory
-fid = fopen(fullfile(TestRepository,'UnitTesting','working_directory','test_1.txt'), 'wt');
-fclose(fid);
-fid = fopen(fullfile(TestRepository,'UnitTesting','working_directory','test_2.txt'), 'wt');
-fclose(fid);
-fid = fopen(fullfile(TestRepository,'UnitTesting','working_directory','test_3.txt'), 'wt');
-fclose(fid);
+fid1 = fopen(fullfile(TestRepository,'UnitTesting','working_directory','test_1.txt'), 'wt');
```
HenkMutsaerts, line 29: Please add the function name in the commit title ;)
Weird why this doesn't work. In Linux there is a much easier way to create an empty file, which is called "touching" a file": `system('touch test_1.txt');`

Testing/xASL_test_UnitTesting.m
```diff
@@ -89,9 +99,9 @@
     
     %% Print test results
     clc
-    fprintf('================================= TEST RESULTS =================================\n\n')
+    fprintf('====================================== TEST RESULTS ======================================\n\n')
     disp(UnitTestsTable);
```
HenkMutsaerts, line 103: Replace `disp` or `display` by `fprintf('%s\n', MessageToPrint);`

Testing/UnitTests/xASL_ut_UnitTest_function_StdNan.m
```diff
@@ -0,0 +1,66 @@
...
+
+% Give your individual subtest a name
+UnitTest.tests(1).testname = 'Multiple test calls with different arrays';
+
+% Start the test
+testTime = tic;
+
+% Run your test here
+y1 = xASL_stat_StdNan([0, 0, 0, 0, 0]);
+y2 = xASL_stat_StdNan([0, 1, 2, 3, 4]);
```
HenkMutsaerts: @MichaelStritt 
I know that we are not ready to sweep across a number of parameters, but this would be the desired unit test, here we only test a limited amount of values. What do you think of creating a function `xASL_ut_CreateInputValues`, where you provide the upper and lower limit as well as the "resolution/number bins" of the values to test. So e.g., upper limit of 1,000,000 and lower limit of -1,000,000 with resolution 10 would create as input parameters for the unit test:
>> [-1000000:2000000/10:1000000]

ans =

  Columns 1 through 3

    -1000000     -800000     -600000

  Columns 4 through 6

     -400000     -200000           0

  Columns 7 through 9

      200000      400000      600000

  Columns 10 through 11

      800000     1000000

Also, we now do this unit test by assuming that another unit (==Matlab's `std`) is perfect. Whereas, this unit `xASL_stat_StdNaN` probably uses Matlab's `std` internally :) And whereas Matlab's std may differ between versions (theoretically)

Testing/UnitTests/xASL_ut_UnitTest_function_str2num.m
```diff
@@ -0,0 +1,102 @@
...
+% Give your individual subtest a name
+UnitTest.tests(1).testname = 'Number types: integer, double, positive and negative';
+
+% Start the test
+testTime = tic;
+
+% Run your test here
+numOutA = xASL_str2num('123');
+numOutB = xASL_str2num('-123');
+numOutC = xASL_str2num('123.456');
```
HenkMutsaerts: Same suggestion here as for the other unit test below, why do you test `123456`, why not `1234567890`? But that's just my theoretical thinking probably :)

HenkMutsaerts (changes requested):
Very nice work, few minor suggestions

Testing/UnitTests/xASL_ut_function_xASL_bids_BIDS2Legacy.m
MichaelStritt: Not really, since the default case creates a `dataPar.json `

Testing/UnitTests/xASL_ut_function_xASL_adm_CheckFileCount.m
MichaelStritt: Yeah, I always try to avoid system statements wherever possible, because then we have to take these Windows/Unix differences into account.

Testing/xASL_test_UnitTesting.m
MichaelStritt: But the UnitTestsTable isn't a string, it's a table. You can't fprintf a table.

Testing/UnitTests/xASL_ut_UnitTest_function_str2num.m
MichaelStritt: From the scientific point of view this is definitely a valid argument, but on the other hand we can't test for every integer/double value. In my opinion testing with one positive integer, one negative interger and a double should be enough. We can write something like...

```
index = 1;
for iTest = -1000:+1000
    testResult(index) = xASL_str2num(num2str(iTest));
    index = index +1;
end
```

... though.

MichaelStritt: I don't expect relevant testing results here though. What could be more interesting is to test the cases where we have problems regarding rounding a bit better. So tests like:

```
numOutX = xASL_str2num('0.000000000001');
numOutY = xASL_str2num('-0.000000000001');
```

Testing/UnitTests/xASL_ut_function_xASL_adm_CheckFileCount.m
MichaelStritt: I just removed it, this shouldn't fail anyway. Fix in bfbadb7

Testing/UnitTests/xASL_ut_UnitTest_function_StdNan.m
MichaelStritt: > Also, we now do this unit test by assuming that another unit (==Matlab's std) is perfect

I don't understand what you would do differently. Do you want to open an old math book and use reference values? 

> And whereas Matlab's std may differ between versions (theoretically)

Same here. Matlab is tested and approved software. If you don't trust Matlab to this level, you shouldn't use it at all.

MichaelStritt: > I know that we are not ready to sweep across a number of parameters, but this would be the desired unit test, here we only test a limited amount of values. 

Check out my comments for `xASL_ut_UnitTest_function_str2num`. We can implement this, but I don't think it's needed right now. Let's start as simple as possible. We should focus on testing more scripts and not on testing the few ones more thoroughly.

Testing/UnitTests/xASL_ut_function_xASL_adm_CorrectName.m
```diff
@@ -0,0 +1,80 @@
...
+%% Test run 1
+
+% Give your individual subtest a name
+UnitTest.tests(1).testname = 'Simple text example';
+
+% Start the test
+testTime = tic;
+
+% Run your test here
+strOut = xASL_adm_CorrectName('abc$%&def');
```
jan-petr, line 29: you want to have some backslashes and umlauts - these were a horror

Testing/UnitTests/xASL_ut_function_xASL_stat_StdNan.m
```diff
@@ -0,0 +1,59 @@
...
+%% Test run 1
+
+% Give your individual subtest a name
+UnitTest.tests(1).testname = 'Multiple test calls with different arrays';
+
+% Start the test
+testTime = tic;
+
+% Run your test here
+y1 = xASL_stat_StdNan([0, 0, 0, 0, 0]);
```
jan-petr, line 29: you want to test also 3D or 4D files...

Testing/UnitTests/xASL_ut_function_xASL_str2num.m
```diff
@@ -0,0 +1,95 @@
...
+%% Test run 1
+
+% Give your individual subtest a name
+UnitTest.tests(1).testname = 'Number types: integer, double, positive and negative';
+
+% Start the test
+testTime = tic;
+
+% Run your test here
+numOutA = xASL_str2num('123');
```
jan-petr, line 29: also stuff like 10e6 , 5.3e-6, .12, and some nonsense stuff - to see if it doesn't crash if a non-number is provided.

jan-petr (approved):
Really great. We should start making stats on how many errors the UT discovered :)

I made some minor comments on what unusual cases to test - it's usually these that crash.

Testing/UnitTests/xASL_ut_function_xASL_adm_CorrectName.m
MichaelStritt: 1b1f95f

Testing/UnitTests/xASL_ut_function_xASL_stat_StdNan.m
MichaelStritt: 67a2c79

Testing/UnitTests/xASL_ut_function_xASL_str2num.m
MichaelStritt: 01dfe87

MichaelStritt:
@HenkMutsaerts: I added everything you and @jan-petr asked for. If I'm not completely wrong, then the accuracy of `xASL_stat_StdNan` is really quite low. Especially if we look at the third unit test (_Multidimensional test with NaN values (2D)_). Can you re-approve so that we can merge the current state? Would be easier to add unit tests step by step.

HenkMutsaerts (approved):
Looks good. 

The `std` inaccuracy could be explained by the way of taking the mean difference. They should both do a parametric mean difference (i.e., "standard" deviation) as opposed to a non-parametric one (i.e., mean absolute difference). But one thing that we should check, is if they do `sum(diff)/n` or `sum(diff)/(n-1)`. Matlab's std probably does the latter by default, which accounts for the loss of one degree of freedom (from calculating the mean diff), but I usually find `sum(diff)/n` easier to interpret.

I like these unit tests a lot. Would be good to still think of ways to generalize them.

Issue #692, opened 2021-07-01
xASL_fsl_TopUp: Minor error


HenkMutsaerts:
### Description

xASL_fsl_TopUp reused a temporary variable which shouldn't have caused errors but it stopped the TopUp processing.

### Release notes

Required: summarize the changes for the release notes here

jan-petr:
Your fix have probably solved the issue, but the real error was on line 206, where you were adding to a variable that might have been initialized differently. I have fixed that because that was prone to future errors - I have submitted to the EPAD branch as well.

PR #693, opened 2021-07-01
#692 xASL_fsl_TopUp: Reused variable corrected

HenkMutsaerts:
### Linked issue

#692


jan-petr (approved):
OK. But check my further small edit with the explanation in the issue. If you agree, then we can merge.

Issue #702, opened 2021-07-06
Move discontinued code to dedicated directory


MichaelStritt:
### Description

We mentioned in **two previous versions** that the support for some scripts will be **discontinued**. We should start by moving them to a **dedicated directory.**

### Release notes

Move discontinued code to a dedicated directory.

PR #703, opened 2021-07-06
#702 Move discontinued scripts to dedicated directory

MichaelStritt:
### Linked issue

Check out #702


jan-petr (changes requested):
Agree that this is a good thing to do - but an issue for discussion. 

The files are now 'hidden' in the outdated directory, which is good. But we shall perhaps add this directory to the path, because otherwise, ExploreASL won't find them there and the users neither. So moving them to this directory is equal to deleting them - which we want to do only later.

So I would indeed move them there. Make sure that a big warning is printed if these functions are executed. But I would still put this directory to the initialized paths for ExploreASL.

MichaelStritt:
@jan-petr: I fixed the paths. It was a bit weird that we already had the full Developments folder within the search path. I removed that and added the Outdated scripts in 
c957a48 .

The warnings are already quite vivid:

image

image

In the warnings we also mention that we will move the ConfigFiles to CustomScripts within v1.8.0. Do you want to do this within this PR as well?

jan-petr (approved):
xASL_qc_CheckValidityJSON.m seems to be used by ExploreQC - we can move that to functions. All the rest is good.

Issue #481, opened 2021-04-09
Custom Atlases


MichaelStritt:
`xASL_init_FileSystem`: 
- Add customAtlases folder with automatic behavior like the main Atlases folder

Documentation:
- Add description about the behavior of both folders to the documentation website

### Release Notes

CLOSED, NOT MERGED.

PR #704, opened 2021-07-06
#481 Custom atlas directory

MichaelStritt:
### Linked issue

Check out #481

### DIFF
--- a/External/CustomAtlases/README.md
+++ b/External/CustomAtlases/README.md
@@ -1,0 +1,7 @@
+
+# Custom Atlases
+
+You can add your custom atlas NIfTI & TSV label file in this directory. Use the same file name for both NIfTI and & TSV file.
+
+
+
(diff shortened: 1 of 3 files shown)


jan-petr (approved):
OK

HenkMutsaerts (changes requested):
Nice idea, but I vote against this, for multiple reasons:
* the pipeline should not be changed by the user, the pipeline should be read-only
* the user can already add custom `ROI_T1_1.nii` or `ROI_FLAIR_1.nii` or `ROI_FLAIR_2.nii` , etc (depending on the image the ROI(s) are registered with and the number of ROIs)
* atlases should be together, why do we want different folders? This suggests a different property of the atlases (e.g., different license). If the users need to add an atlas, it's easier for them to copy the current behaviour. So they see `MNIstructural.nii.gz` `MNIstructural.tsv` and `MNIstructural.license` and they know that they need to add these three files in the same folder, for their custom atlas (rather than having to check a separate folder)
* If the user wants an extra atlas, and this makes sense, we can add it ourselves as well

MichaelStritt:
Consensus in the developer group was to close this PR. Users should rather add to ExploreASL instead of using custom local atlases etc.

Issue #717, opened 2021-07-08
Revamp x structure fields of import & processing


HenkMutsaerts:
### Description

* `x.opts` contains general parameters for initialization or running ExploreASL, not specific to import or processing
* `x.settings` contains settings specific to the processing (should not be used in import) Does it help to rename this to `x.Processing` or should these go to `x.modules.processing.settings`?

* `x.S` (statistics) should be initialized when starting the processing (begin of `ExploreASL_ProcessMaster`)
* `x.Q` (quantification) should be initialized when starting the processing (begin of `ExploreASL_ProcessMaster`)
* `x.P` (processing paths of the dataset) should be initialized when starting the processing (begin of `ExploreASL_ProcessMaster`)

* `x.dataset` should not be initialized when only initialising ExploreASL? Also, this should probably go under `x.opts`. 
* `x.dir` should be a subtract of `x.opts` to clarify that these directories are general (non-import/processing specific)
* `x.MyPath` should go under `x.opts` as this is a general option
* `x.subject_regexp`, `x.Quality`` and `x.DELETETEMP` are specific to processing, so should go to `x.module.process` and should not exist before data loading & processing

* `x.name` this is the same as `x.dataset` and can be removed I think
* `x.nSubjects` `x.nSessions` etc should go under `x.module.process`
* `x.SESSIONS` can go to x.module.process.SESSIONS, same for `x.SUBJECTS` but perhaps better to wait with this as this would be a huge change

MichaelStritt:
> * `x.opts` contains general parameters for initialization or running ExploreASL, not specific to import or processing
> 
> * `x.settings` contains settings specific to the processing (should not be used in import) Does it help to rename this to `x.Processing` or should these go to `x.modules.processing.settings`?

I don't think we should have something like `x.modules.processing`. I'd rather just have:

- `x.module.import`
- `x.module.structural`
- `x.module.asl`
- `x.module.population`

> * `x.S` (statistics) should be initialized when starting the processing (begin of `ExploreASL_ProcessMaster`)
> * `x.Q` (quantification) should be initialized when starting the processing (begin of `ExploreASL_ProcessMaster`)
> * `x.P` (processing paths of the dataset) should be initialized when starting the processing (begin of `ExploreASL_ProcessMaster`)

Hm... I'm not sure if this works. We can make sure that they are only initialized within the data loading though. Within the data loading we load the dataPar.json, which can contain x.S or x.Q fields as far as I know.

[...]

Issue #311, opened 2021-02-08
Use of xASL_tsvWrite in Population module


MichaelStritt:
In #56 we saw that we don't use `xASL_tsvWrite` to write the **tsv atlas files** in the population stats directory. If there are errors in the atlas label file, we would detect them if we would use `xASL_tsvWrite`.

MichaelStritt:
Done in #717

PR #720, opened 2021-07-09
Revamp #717 x

MichaelStritt:
### Linked issue

Check out #717

### How to test

I ran the following tests ...

### Comments

I'd recommend to check each commit one by one.

Overview:

* Move `x.MyPath` to `x.opts.MyPath` 👍 
* Move `x.Quality` to `x.settings.Quality` 👍 
* Move `x.DELETETEMP` to `x.settings.DELETETEMP` 👍 
* Move `x.SESSIONDIR` -> `x.dir.SESSIONDIR` 👍 
* Move `x.T1BiasFieldRegularization` -> `x.modules.structural.T1BiasFieldRegularization`  👍 
* Move `x.bNativeSpaceAnalysis` -> `x.modules.population.bNativeSpaceAnalysis` 👍 
* Move `x.bHammersCAT12` -> `x.modules.structural.bHammersCAT12` 👍 
* Move `x.bFixResolution` -> `x.modules.structural.bFixResolution` 👍 
* Move `x.SegmentSPM12` -> `x.modules.structural.SegmentSPM12` 👍 
* Move `x.bRunModule_DARTEL` -> `x.modules.bRunDARTEL` 👍 
* Move `x.bRunModule_LongReg` -> `x.modules.bRunLongReg` 👍 
* Move `x.SkipIf...` -> `x.settings.SkipIf...` 👍 
* Move `x.ForceInclusionList` -> `x.dataset.ForceInclusionList` 👍 
* Move `x.bAutomaticallyDetectFSL` -> `x.external.bAutomaticallyDetectFSL` 👍 
* Move `x.MakeNIfTI4DICOM` -> `x.settings.MakeNIfTI4DICOM` 👍 
* Move `x.subject_regexp` to `x.dataset.subjectRegexp` 👍 
* Move `x.Vendor` to `x.Q.Vendor` 👍 
* Move `x.readout_dim` to `x.Q.readoutDim` 👍 
* Move `x.name` to `x.dataset.name` 👍 
* Move `x.Sequence` to `x.Q.Sequence` 👍


HenkMutsaerts:
Nice work! Few minor suggestions:

* `DARTEL` and `LongReg` are separate non-standard modules, so `bRunDARTEL` and `bRunLongReg` should not go under `x.modules.structural` (but perhaps under `x.modules`?)

* Can we add the b for Boolean:
`SegmentSPM12` -> `bSegmentSPM12`
`MakeNIfTI4DICOM` -> `bMakeNIfTI4DICOM`

* This change is not only a name change but also a change of an internal legacy format to BIDS format:
`Vendor` -> `Manufacturer`. While this is a good change, I would consider doing this in a future BIDS-specific revamp.

* `x.SESSIONDIR` is a very specific path used in the ASL module (like `SUBJECTDIR` in structural module)
that is dataset-specific, an absolute path (not relative so not fully “dataset”). `x.dir` on the other hand, contains exclusively paths used for initialization. I wouldn’t mix these up.

* By the way: it would be good to add these new categories of subfields (`dir`, `opts`, `settings`, etc)
to `DataParTemplate.m`

HenkMutsaerts (changes requested):
See comments

MichaelStritt:
>     * `DARTEL` and `LongReg` are separate non-standard modules, so `bRunDARTEL` and `bRunLongReg` should not go under `x.modules.structural` (but perhaps under `x.modules`?)

Sure, I can do that. I just did that based on the DataParTemplate.m, where we have these modules within the structural section.

>     * Can we add the b for Boolean:
>       `SegmentSPM12` -> `bSegmentSPM12`
>       `MakeNIfTI4DICOM` -> `bMakeNIfTI4DICOM`

Good idea 👍 

>     * This change is not only a name change but also a change of an internal legacy format to BIDS format:
>       `Vendor` -> `Manufacturer`. While this is a good change, I would consider doing this in a future BIDS-specific revamp.

Yeah... but why do this multiple times if we can just do it now? The old dataPar.json's should still work, since we added this backwards compatibility feature to the dataPar reader.

>     * `x.SESSIONDIR` is a very specific path used in the ASL module (like `SUBJECTDIR` in structural module)
>       that is dataset-specific, an absolute path (not relative so not fully “dataset”). `x.dir` on the other hand, contains exclusively paths used for initialization. I wouldn’t mix these up.

I know, we use it as one of these symbolic fields. I think I moved SUBJECTDIR to there as well, so in my head this change kind of makes sense. Does this bug you? 😄 

>     * By the way: it would be good to add these new categories of subfields (`dir`, `opts`, `settings`, etc)
>       to `DataParTemplate.m`

Yeah, I started writing a new DataParTemplate.md, where we can have both .m and .json examples. We can integrate it there if you want.

HenkMutsaerts (approved):
Very nice

MichaelStritt:
@jan-petr: Became a really big change after all. I'll run some tests now. You can start to check if you're happy with the changes. I'm not sure about how to update `xASL_adm_LoadX_DeprecatedNamesTable` within `xASL_adm_LoadX`, because for some of the new fields we need to go into a substructure of a substructure. Other than that it looks pretty good so far. Only one merge conflict with the DataParTemplate.m, but we said we would remove it, so that shouldn't be a bit problem. The other thing was about the Vendor field. I got some warnings/errors in previous tests because it couldn't find x.Q.Vendor. Maybe I'm doing something wrong there?

MichaelStritt:
# Testing 

## Initialization

The initialization seems to work perfectly fine. We didn't change that much there though.

```console
>> x = ExploreASL;
ExploreASL will run the initialization...
                    ## |                                                                      
                    ## |                                                                      
                    ##/  

Dataset Root        
Import Modules      
Process Modules     
bPause              False
iWorker             1
nWorkers            1
ExploreASL v1.8.0_BETA initialized ... 

>> x
x = 
  struct with fields:

        opts: [1×1 struct]
           S: [1×1 struct]
           D: [1×1 struct]
           P: [1×1 struct]
           Q: [1×1 struct]
     modules: [1×1 struct]
     dataset: [1×1 struct]
    settings: [1×1 struct]
    external: [1×1 struct]
         dir: [1×1 struct]
     Version: '1.8.0_BETA'
```

## BIDS 2 Legacy

BIDS 2 Legacy doesn't crash and seems to behave normally.

```console
>> pathTest = '.\TestDataSet';
>> x = ExploreASL(pathTest,[0 0 0 1],0);
ExploreASL will run the import workflow and will run the initialization...
                    ## |                                                                      
                    ## |                                                                      
                    ##/  

Dataset Root        ...\TestDataSet
Import Modules      BIDS2LEGACY
Process Modules     
bPause              False
iWorker             1
nWorkers            1
ExploreASL v1.8.0_BETA initialized ... 
Parsing BIDS scans: 100%
Note that any warnings may have only printed once if they were repeated for multiple scans
Always run your rawdata folder through the BIDS validator first and aim to avoid warnings
Converting from BIDS to Legacy: .\TestDataS100%
M0 parsed for .\TestDataSet\derivatives\ExploreASL\sub-Sub1\ASL_1\ASL4D.nii.gz
Creating default dataPar.json since file was not found...
Overwriting x.dir.dataPar...

>> x
x = 
  struct with fields:
        opts: [1×1 struct]
           S: [1×1 struct]
           D: [1×1 struct]
           P: [1×1 struct]
           Q: [1×1 struct]
     modules: [1×1 struct]
     dataset: [1×1 struct]
    settings: [1×1 struct]
    external: [1×1 struct]
         dir: [1×1 struct]
     Version: '1.8.0_BETA'
```

## Processing

After the BIDS 2 Legacy conversion from above, I added the `x.Q.Vendor` field within the `dataPar.json`.

```json
{
    "x":
    {
            "dataset":
            {
                "name":"ExampleDataSet",
                "subjectRegexp":"^sub-Sub\\d{1}$"
            },
            "settings":
            {
                "Quality":0,
                "DELETETEMP":1
            }
            },
            "Q":
            {
                "Vendor": "Philips"
            },
            "GeneratedBy":
            {
                "Name":"ExploreASL",
                "Version":"1.8.0_BETA"
            }      
}
```

...

Functions/xASL_io_ReadDataPar.m
```diff
@@ -140,36 +140,167 @@
...
-		x = rmfield(x,'exclusion');
-	end
 	
 else
 	error('Unknown file extension');
 end
 
 end
 
+
```
jan-petr, line 152: The example at lines 32-38 has to be updated

Functions/xASL_quant_SliceTiming.m
```diff
@@ -91,19 +91,19 @@
 end
 
 % The readout_dim needs to be provided
```
jan-petr, line 93: Update the commentary

Modules/xASL_module_ASL.m
```diff
@@ -421,7 +421,7 @@
 if ~x.mutex.HasState(StateName{iState}) && x.mutex.HasState(StateName{iState-4})
 
     fprintf('%s\n','Quantifying ASL:   ');
-    if isfield(x, 'bUseBasilQuantification') && x.bUseBasilQuantification
+    if isfield(x, 'bUseBasilQuantification') && x.Q.bUseBasilQuantification
```
jan-petr, line 424: Q missing in the first term.

Modules/xASL_module_ASL.m
```diff
@@ -38,28 +38,28 @@
...
+x.P.Path_ASL4D = fullfile(x.dir.SESSIONDIR, 'ASL4D.nii');
+x.P.Path_ASL4D_json = fullfile(x.dir.SESSIONDIR, 'ASL4D.json');
+x.P.Path_ASL4D_parms_mat = fullfile(x.dir.SESSIONDIR, 'ASL4D_parms.mat');
 
 if ~xASL_exist(x.P.Path_ASL4D, 'file')
     % First try to find one with a more BIDS-compatible name & rename it (QUICK & DIRTY FIX)
-    FileList = xASL_adm_GetFileList(x.SESSIONDIR, 'ASL4D.*\.(nii|nii\.gz)');
+    FileList = xASL_adm_GetFileList(x.dir.SESSIONDIR, 'ASL4D.*\.(nii|nii\.gz)');
 
     if ~isempty(FileList) && isfield(x,'M0PositionInASL4D')
```
jan-petr, line 64: module.asl missing?

Functions/xASL_init_FileSystem.m
```diff
@@ -53,7 +53,7 @@
     x.iSubject = find(cellfun(@(y) strcmp(y, x.P.SubjectID), x.SUBJECTS));
 end
 if isfield(x,'SESSIONDIR')
```
jan-petr, line 55: dir missing.

Functions/xASL_init_DefineStudyData.m
```diff
@@ -57,32 +57,35 @@
 
 % ------------------------------------------------------------------------------------------------
 %% Admin
-if ~isfield(x,'name')
-    x.name = ''; 
+if ~isfield(x,'dataset')
+    x.dataset = struct; 
+end
+if ~isfield(x.dataset,'name')
+    x.dataset.name = '';
```
jan-petr, line 64: Check  xternal/SPMmodified/spm_atlas.m, line 77 - is this related?

Functions/xASL_io_ReadDataPar.m
```diff
@@ -140,36 +140,167 @@
...
-	end
-	
-	if isfield(x,'M0_conventionalProcessing')
-		warning('Deprecated field. Please use x.settings.M0_conventionalProcessing instead of x.M0_conventionalProcessing');
-		if ~isfield(x,'settings') || ~isfield(x.settings,'M0_conventionalProcessing')
-			x.settings.M0_conventionalProcessing = x.M0_conventionalProcessing;
-		end
-		x = rmfield(x,'M0_conventionalProcessing');
-	end
+    x = xASL_io_ReadDataParCheckDeprecatedFields(x);
```
jan-petr, line 143: Fix line 33 of this file - the comment of subject_regexp -> subjectRegexp

ExploreASL_ProcessMaster.m
```diff
@@ -54,10 +54,10 @@
 
     % Optional modules
     % The following are optional extensions of the structural module and not required to run, normally they can be ignored:
-    if isfield(x, 'bRunModule_LongReg') && x.bRunModule_LongReg
+    if isfield(x.modules, 'bRunLongReg') && x.modules.bRunLongReg
```
jan-petr, line 57: This is ok not to have it inside structural - Henk agreed?

jan-petr (changes requested):
Great that you have thought about automatically upgrading the parameters like x.modules.asl.bUseMNIasDummyStructural
But the following seems not to be updated - it might be nice to make this more automatically within readDataPar.m - make a list of those and not to enter all of them by hand like the bUseMNI... :
bNativeSpaceAnalysis, bPVCNativeSpace, SpikeRemovalThreshold, bRegisterM02ASL, bPVCGaussianMM

We might need to be careful with some of the x-struct internal stuff as well. For processing in progress - people experimenting with different versions:
x.opts.MyPath,

And finally, some fields inside the Q need to be upgraded in the xASL_adm_LoadParms. See my comment there and please do the same for:
bUseBasilQuantification, ApplyQuantification, Vendor

And these need to be somehow considered as well - for renaming and for moving inside the structure - otherwise, this will be a big mess:
motionCorrection, bRunLongReg, bRunDARTEL

Functions/xASL_fsl_RunFSL.m
```diff
@@ -51,11 +51,11 @@
...
 
 %% Find FSL directory
-if ~isfield(x,'bAutomaticallyDetectFSL')
-    x.bAutomaticallyDetectFSL = 0;
+if ~isfield(x.external,'bAutomaticallyDetectFSL')
+    x.external.bAutomaticallyDetectFSL = 0;
 end
 
-[FSLdir, x, RootFSLdir] = xASL_fsl_SetFSLdir(x, x.bAutomaticallyDetectFSL);
+[FSLdir, x, RootFSLdir] = xASL_fsl_SetFSLdir(x, x.external.bAutomaticallyDetectFSL);
```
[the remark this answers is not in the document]
MichaelStritt: I don't know, I think Henk wrote that script. I can change that easily though 👍

ExploreASL_ProcessMaster.m
MichaelStritt: Henk actually requested this, I had it the other way before.

Henk:
> `DARTEL` and `LongReg` are separate non-standard modules, so `bRunDARTEL` and `bRunLongReg` should not go under `x.modules.structural` (but perhaps under `x.modules`?)

MichaelStritt:
@jan-petr 

> Great that you have thought about automatically upgrading the parameters [...]
> But the following seems not to be updated - it might be nice to make this more automatically within readDataPar.m - make a list of those and not to enter all of them by hand like the bUseMNI... : bNativeSpaceAnalysis, bPVCNativeSpace, SpikeRemovalThreshold, bRegisterM02ASL, bPVCGaussianMM

I thought about that too. I started by adding the code to a separate function. My only problem was that it's not that easy to automate, since some of the fields get assigned to a field of x whereas other ones are within a sub or subsubstructure 😄 I'll try to think of a better way to do this though.

> We might need to be careful with some of the x-struct internal stuff as well. For processing in progress - people experimenting with different versions: x.opts.MyPath

Yep, I was a bit afraid of changing this. It was actually Henk who mentioned that if we revamp the other fields, we should also do MyPath, SESSIONS and so on. We could implement a simple warning and we should also document/communicate these changes as transparent as possible.

> And finally, some fields inside the Q need to be upgraded in the xASL_adm_LoadParms. See my comment there and please do the same for: bUseBasilQuantification, ApplyQuantification, Vendor

I'm on it. Sorry, I hope my coding doesn't seem too sloppy, it's just hard to keep track of all of these changes.

> And these need to be somehow considered as well - for renaming and for moving inside the structure - otherwise, this will be a big mess: motionCorrection, bRunLongReg, bRunDARTEL

In one of these code comments I already answered the bRunLongReg and bRunDARTEL thing. Henk specifically asked for them to be within x.modules instead of x.modules.structural. I tried to keep as close as possible to the different sections of the old DataParTemplate.m, where we had Sequence parameters, Structural parameters, Environment parameters, etc.

ExploreASL_Initialize.m
```diff
@@ -81,51 +81,51 @@
     % Check if the current directory is the ExploreASL directory
     CurrCD = pwd;
     if exist(fullfile(CurrCD, 'ExploreASL_Master.m'), 'file')
-        x.MyPath = CurrCD;
+        x.opts.MyPath = CurrCD;
     end
 
     % Check whether MyPath is correct, otherwise obtain correct folder
     if ~isfield(x, 'MyPath')
```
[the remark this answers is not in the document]
MichaelStritt: 45ca0c1

Functions/xASL_io_ReadDataPar.m
```diff
@@ -140,36 +140,167 @@
...
 
 end
 
+
+%% Check deprecated fields
+function x = xASL_io_ReadDataParCheckDeprecatedFields(x)
+
+    if isfield(x,'SegmentSPM12')
+        warning('Deprecated field. Please use x.modules.structural.bSegmentSPM12 instead of x.SegmentSPM12');
+        if ~isfield(x,'modules') || ~isfield(x.modules,'structural') || ~isfield(x.modules,'SegmentSPM12')
```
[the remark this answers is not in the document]
MichaelStritt: 2f0b4f4

Functions/xASL_adm_LoadX.m
```diff
@@ -128,7 +128,7 @@
 					   'bOverwrite'         'settings' 'bOverwrite';...
 					   'dryRun'             'settings' 'dryRun';...
 					   'M0_conventionalProcessing' 'settings' 'M0_conventionalProcessing';...
-					   'SegmentSPM12'       'settings' 'SegmentSPM12';...
+					   'SegmentSPM12'       'settings' 'bSegmentSPM12';...
```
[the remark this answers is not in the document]
MichaelStritt: Fix 1: fec1fa7

Functions/xASL_adm_LoadX.m
```diff
@@ -65,21 +65,40 @@
...
+        end
+        % Remove old field
+        OldX.x = rmfield(OldX.x,conversionTable{iField,1});
+    end
 end
 
 if bOldFieldsDetected
-	warning('Detected deprecated fields in x.mat. Older version of ExploreASL was used to process this data previously. Field names fixed');
+    % Write fields back
+    x = OldX.x;
```
MichaelStritt, line 96: @jan-petr: Why didn't we write the OldX.x to x before? Otherwise all of this doesn't work at all 😄

Functions/xASL_adm_LoadX.m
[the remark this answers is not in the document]
MichaelStritt: Fix 2: ee7d7d1

Functions/xASL_fsl_RunFSL.m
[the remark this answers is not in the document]
MichaelStritt: @jan-petr: I think this should fix it ffa603c

Functions/xASL_init_PrintCheckSettings.m
```diff
@@ -29,13 +29,13 @@
 
     %% -----------------------------------------------------------------------
     %% 1) Set default settings if not defined
-    if ~isfield(x,'Quality') || (x.Quality~=0 && x.Quality~=1)
-        x.Quality = 1;
+    if ~isfield(x,'Quality') || (x.settings.Quality~=0 && x.settings.Quality~=1)
```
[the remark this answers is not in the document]
MichaelStritt: aaa1577

Functions/xASL_spm_coreg.m
```diff
@@ -40,15 +40,15 @@ function xASL_spm_coreg(refPath, srcPath, OtherList, x, sep, FastReg)
 refFile = [refFile refExt];
 srcFile = [srcFile srcExt];
 
-if nargin<4 || isempty(x) || ~isfield(x,'Quality') || isempty(x.Quality)
-    x.Quality = true; % default quality is high
-    x.DELETETEMP = true;
+if nargin<4 || isempty(x) || ~isfield(x,'Quality') || isempty(x.settings.Quality)
```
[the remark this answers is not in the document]
MichaelStritt: aa2cf8c

Functions/xASL_bids_BIDS2Legacy.m
```diff
@@ -80,17 +80,27 @@
...
+if isfield(dataPar.x,'settings')
+    if ~isfield(dataPar.x.settings,'Quality')
+        dataPar.x.settings.Quality = 1;
+    end
+else
+    % Default
+    dataPar.x.settings.Quality = 1;
 end
 
 if ~isfield(dataPar.x,'DELETETEMP')
```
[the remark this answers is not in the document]
MichaelStritt: b4658aa

Functions/xASL_init_VisualizationSettings.m
```diff
@@ -142,10 +142,10 @@
         x.S.masks.WBmask = logical(ImageWB);
     end
 
-    if ~isfield(x,'Quality') && isfield(x,'QUALITY')
-        x.Quality = x.QUALITY; % backward compatibility
-    elseif ~isfield(x,'Quality')
-        x.Quality = 1; % default is normal quality
+    if ~isfield(x.settings,'Quality') && ~isfield(x,'Quality') && isfield(x,'QUALITY')
```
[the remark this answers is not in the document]
MichaelStritt: Oh, good question. Maybe that was there already or I just was a bit tired at that point in time 😆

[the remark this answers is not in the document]
MichaelStritt: I think we should just ensure the backwards compatibility in the dataParReader and the loadX script. Here I'd actually just remove it. 7fc6d10

Modules/xASL_module_Population.m
```diff
@@ -85,7 +85,7 @@
     FoVPath = xASL_adm_GetFileList(x.D.TemplatesStudyDir, ['^FoV_n' xASL_num2str(x.dataset.nSubjectsSessions) '_bs-mean_Unmasked\.nii$'], 'FPList');
     SusceptPath = fullfile(x.D.TemplatesStudyDir,['MaskSusceptibility_n' xASL_num2str(x.dataset.nSubjectsSessions) '_bs-mean.nii']);
 
-    if strcmpi(x.Sequence,'3d_spiral') && ~isempty(FoVPath)
+    if strcmpi(x.Q.Sequence,'3d_spiral') && ~isempty(FoVPath)
```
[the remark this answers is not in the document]
MichaelStritt: I think I got most of them here ee7d7d1, but I can go through the list another time.

Functions/xASL_io_ReadDataPar.m
MichaelStritt: 6b34e0a

Functions/xASL_quant_SliceTiming.m
MichaelStritt: d68121e

Modules/xASL_module_ASL.m
MichaelStritt: 1aafc86

Modules/xASL_module_ASL.m
MichaelStritt: 31ce308

Functions/xASL_init_FileSystem.m
MichaelStritt: 458277d

Functions/xASL_init_DefineStudyData.m
MichaelStritt: The name in spm_atlas isn't the name of the subject I think. And there we don't work with the exploreasl x struct as far as I can see it, but I never modified that much spm code.

Functions/xASL_io_ReadDataPar.m
MichaelStritt: 6b34e0a

MichaelStritt:
I'm rerunning the 10 Test datasets right now. Let's see if everything works tomorrow 😄

Functions/xASL_fsl_SetFSLdir.m
```diff
@@ -34,7 +34,11 @@
     x = struct;
 end
 if nargin<2 || isempty(bAutomaticallyDetectFSL)
-    bAutomaticallyDetectFSL = false;
+    if isfield(x,'external') && isfield(x.external,'bAutomaticallyDetectFSL')
```
jan-petr, line 37: This might need the attention of @HenkMutsaerts as well. Not sure what is the best solution - removing the bAutomatically parameter completely from the function call and only using inside the x-struct version... But this version of default vs x-struct vs implicitly passed might not be bad either... TBD

Functions/xASL_fsl_RunFSL.m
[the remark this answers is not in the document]
jan-petr: yes. totally fixed - but we should discuss with Henk if there should be a possibility to even give the option in the function call or rely only on x-struct.

Functions/xASL_init_VisualizationSettings.m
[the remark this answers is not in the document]
jan-petr: Haha. Yes - that is of course a major issue with late night edits :D Agree to remove it here.

Functions/xASL_init_DefineStudyData.m
jan-petr: You are right - not related...

Functions/xASL_adm_LoadX.m
jan-petr: :D

jan-petr (changes requested):
Looks all very good - except one minor thing in calling FSL - we have to discuss with Henk what was actually intended by this and if we can do it with x-struct only or not...

MichaelStritt:
The test dataset `dataPar.json`s should be fixed within https://github.com/ExploreASL/TestDataSets/commit/4b00f066c8c0755802044de2ea9691ad434fb022

MichaelStritt:
I rebased the commits for testing purposes like Jan suggested...

MichaelStritt:
### Testing results

There are still some bugs in the pipeline. Will have a look this morning 👍 

image

MichaelStritt:
### GE_3Dspiral_T1prefilled_Cyst_M0_noFLAIR_2ndRun_LoQ

There are these warnings...

```
xASL_adm_DefineASLSequence, line 27
x.Q.readoutDim parameter missing, skipping determining ASL sequence
x.Q.Vendor missing, skipping determining ASL sequence
No x.Q.Sequence defined
```

... which of course lead to an error later on ...

```
ERROR: ASL module terminated for subject 1: 001DM_1
ans = 'Reference to non-existent field 'Vendor'
```

MichaelStritt:
### Philips_2DEPI_FEAST_Longitudinal_FLAIR_LGA_ASL_PartlyDone_HiQ

Here there are no warnings, but then there's this ...

```
xASL_wrp_ResampleASL, line 159
ERROR: ASL module terminated for subject 1: 151001_1
ans = 'Reference to non-existent field 'M0'
```

MichaelStritt:
### Philips_2DEPI_FLAIR_DisabledQuantification_M0_LoQ

Warnings are similar to GE_3Dspiral_T1prefilled_Cyst_M0_noFLAIR_2ndRun_LoQ, but the error is a bit different ...

```
ERROR: ASL module terminated for subject 1: Sub-001
ans = 'Unable to perform assignment because dot indexing is not supported for variables of this type.
'xASL_adm_LoadX, line 90
```

MichaelStritt:
@jan-petr: explanation to commits 091db85 and b5e630a: The major problem was that `xASL_adm_LoadParms `just dumps all fields of the JSON into the x struct. This is not really helpful and maybe we should do this in a more controlled way in a future release. For now it should mostly be fixed. We use a standardized script called `xASL_io_CheckDeprecatedFieldsX `for this now. We use the same script for `xASL_io_ReadDataPar `and we could also integrate it in the `xASL_adm_LoadX `function, we just have to use OldX.x for x there I think.

MichaelStritt:
### Test dataset

Finally one more or less successful run:

```
>> x = ExploreASL(pathTest,0,1);
ExploreASL will run the processing pipeline...
                    ## |                                                                      
                    ## |                                                                      
                    ##/  

Dataset Root        .\TestDataSet
Import Modules      
Process Modules     STRUCTURAL ASL POPULATION 
bPause              False
iWorker             1
nWorkers            1
ExploreASL v1.8.0_BETA initialized ... 


Automatically defining sessions...
==================================== Additional Settings =====================================
1 scans - 0 exclusions, resulting in 1 scans of: 
Longitudinal timePoint 1 = 1 scans - 0 exclusions = 1 scans
ASL sessions: 1

Ancillary data, sets: 4 sets are defined for 1 "SubjectsSessions"
Set 1 = "session" options "ASL_1", codes for paired data
Set 2 = "LongitudinalTimePoint" options "TimePoint_1", codes for paired data
Set 3 = "SubjectNList" options "SubjectNList", codes for paired data
Set 4 = "Site" options "SingleSite", codes for two-sample data
x.D.ROOT            .\TestDataSet\derivatives\ExploreASL
x.settings.DELETETEMP 1 (delete temporary files)
x.settings.Quality    0 (0 = fast try-out; 1 = normal high quality)

Running xASL_module_Structural ...   
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
=== Subject: sub-Sub1, Module: xASL_module_Structural,  14-Jul-2021 13:35:55 ===
...
Collecting QC parameters...
Printing ExploreASL PDF report:   
Job-iteration 1 stopped at 14-Jul-2021 13:45:42 and took 587 seconds
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++100%
xASL_module_Structural completed

====================================================================================
Printing file reports... Check this for a summary of missing files
Detected: FLAIR : MoCo results : M0 
Checking native space: anat files:100%, ASL files:100%
Checking common space: anat files:100%, ASL files:100%, checking lock files:100%
Please check missing MNI_session files: .\derivatives\ExploreASL\Missing_MNI_session_files.csv
Please check missing Lock files: .\derivatives\ExploreASL\Missing_Lock_files.csv
Done generating file reports
====================================================================================
Running xASL_module_ASL ...   
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
=== Subject: sub-Sub1, Session: ASL_1, Module: xASL_module_ASL_<SESSION>,  14-Jul-2021 13:45:42 ===
...
080_SortBySpatialCoV was perform11%Deleting temporary files failed because: Undefined function or variable 'File2Search'.
G-zZzZipping 54 files
Gzipping Niftis Successful...
090_DeleteAndZip was performed
Job-iteration 1 stopped at 14-Jul-2021 13:51:37 and took 28 seconds
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++100%
xASL_module_Population completed
Many thanks for using ExploreASL, please don't forget to cite https://pubmed.ncbi.nlm.nih.gov/32526385/.
Note that ExploreASL is a collaborative effort.
Therefore, please don't hesitate to contribute by feedback, adding code snippets, or clinical experience!
```

MichaelStritt:
@jan-petr: Minor fix in d9dde69 should make most of it work now. I'm running the 10 test datasets now. I don't get where this bug from b821f81 comes from. I'm always getting `Undefined function or variable 'File2Search'`. Was that there before?

HenkMutsaerts (changes requested):
Looks great! One question & one suggested change

* Does SESSIONDIR fit in x.dir? (it belongs to the processing modules only)

* M0 is not a general processing setting, but specific to the quantification (x.Q), so I would move it to x.Q

Also, I agree with you that it would be beneficial to initiate many empty structs/x-fields at the beginning of a processing module.

Very nice work!

MichaelStritt:
### Testing

Weirdly the testing script failed shortly before it was done. It seems that one of the csv files wasn't formatted correctly:

```
Error using spm_load>dsvread (line 166)
Varying number of delimiters per line.

Error in spm_load (line 46)
    x = dsvread(f,'\t');

Error in xASL_tsvRead (line 42)
ReadCell = spm_load(PathTSV);

Error in xASL_bids_csv2tsvReadWrite (line 70)
    CellContents = xASL_tsvRead(PathIn);

Error in xASL_qc_TestExploreASL (line 416)
    [~, TempTable] = xASL_bids_csv2tsvReadWrite(ResultFile{iFile}{end});
```

MichaelStritt:
> * Does SESSIONDIR fit in x.dir? (it belongs to the processing modules only)

For me it makes sense to keep it in `x.dir`. The reason is that it's a symbolic field which is interpreted in `xASL_Iteration`. Substructs of x with these fields have to be scanned and I don't want to scan these huge substructs `x.D` or `x.P`. In addition, we do the same with SUBJECTDIR right now, so if it doesn't bug you, then I'd keep it there.

> * M0 is not a general processing setting, but specific to the quantification (x.Q), so I would move it to x.Q

Sure, I need to re-test because of this TSV/CSV issue anyway. We should have talked about that before I made the change.

MichaelStritt:
> @MichaelStritt Could be read-write permission issue

It really shouldn't. Check out the code:

```matlab
Files2Del   = {'ATT_BiasField.nii' 'Mask_Template.nii' 'Mean_CBF_Template.nii' 'PseudoCBF.nii' 'RawTemplate.nii' 'VascularArtifact_Template.nii' 'mean_PWI_Clipped.nii' 'SliceGradient_extrapolated.nii' 'FoV.nii'};

try
    for iP=1:length(Files2Del)
        xASL_TrackProgress(iP,length(Files2Del));
        [~, Ffile, Fext]    = xASL_fileparts(Files2Del{iP});

        PathList    = xASL_adm_GetFileList(x.D.ROOT,['^' File2Search Fext '$'],'FPListRec',[0 Inf]);
        for iL=1:length(PathList)
            xASL_delete(PathList{iL});
        end

    end
catch ME
    fprintf('Deleting temporary files failed because: %s', ME.message);
end
```

The variable `File2Search` isn't initialized. I think `File2Search` should be `Ffile` instead.

MichaelStritt:
@jan-petr & @HenkMutsaerts: I know this is probably not something we would usually fix within another issue, but it was kind of necessary since it results in so many bugs during testing otherwise.
I think I fixed the writing of xASL_stat_PrintStats in d3f25f2. The original plan was to do this within #311. The only part I didn't really know how to test is the code in line 120 and following. Other than that it seems to work pretty nicely.

HenkMutsaerts (approved):
> It seems as if we really should use xASL_tsvWrite everywhere 😉

Good idea! Some of these TSV/CSVs may not be written by us but by external packages such as SPM.

The `SUBJECTDIR` and `SESSIONDIR` are data-specific fields created within a module that is run by xASL_Iteration, whereas `x.dir` seems to be initialization folders, so their meaning is different. The `x.P` definitely should be improved, this was a temporary fix by me to create some standardization in generating the filestructure. `x.D` should contain environmental directories only, not data-specific fields/folders. Perhaps we can move `SUBJECTDIR` and `SESSIONDIR` to x.modules. then, in `x.dir` they seem arbitrary. Overview of these fields to the user/debugger is more important than parsing speed in general.

One other comment by me went missing, hope you can still find it!

Testing/xASL_qc_TestExploreASL.m
```diff
@@ -408,37 +415,42 @@
...
-			else
-				ResultsTable{1+iList,2+iFile} = 'n/a';
-			end
-			if ~isempty(IndexCSF)
-				ResultsTable{1+iList,3+iFile} = TempTable{2, IndexCSF};
-			else
-				ResultsTable{1+iList,2+iFile} = 'n/a';
-			end
-		end
+        % Make sure one individual file does not crash the table generation
```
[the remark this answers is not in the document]
MichaelStritt: Totally get what you mean. The thing right here is that we run the testing for all 10 cases and after processing all of them you don't even get 50% of the results, you get nothing, because one single TSV file was damaged.

Functions/xASL_init_SubStructs.m
```diff
@@ -0,0 +1,47 @@
...
+%   x       - ExploreASL x struct (STRUCT, REQUIRED)
+%
+% OUTPUT:
+%   x       - ExploreASL x struct
+%
+% -----------------------------------------------------------------------------------------------------------------------------------------------------
+% DESCRIPTION:  Initialize the ExploreASL x structure substructs/fields.
+%               Only fields which do not exist so far are added.
+%               This script is supposed to help with the overall modularity of ExploreASL.
+%               This script is identical to the function ExploreASL_Initialize_SubStructs within
```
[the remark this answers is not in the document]
MichaelStritt: > Wouldn't create two scripts that are supposed to be identical. Instead, I would then remove the subfunction from ExploreASL_Initialize, and if needed initialize the path at this point. Keep it clean and simple ;)

Agree. Let's improve this later on though, since it's only an optimization and not required for the revamp. I don't like changing the order of the ExploreASL_Initialize code that randomly. We have to be careful here. Maybe we can move an xASL_init_SubStructs call to the part after the addpath in ExploreASL_Initialize and remove the subfunction there, but then we have to be sure that all the lines & subfunctions before have all the substructs they need.

Functions/xASL_stat_PrintStats.m
```diff
@@ -142,13 +117,13 @@
                 warning(['Missing data, skipping printing data for ' x.SUBJECTS{iSubject} '_' x.SESSIONS{iSession}]);
             else
                 % print subject name
-                fprintf(x.S.FID,'%s\t', x.S.SubjectSessionID{iSubjectSession, 1});
+                x.modules.population.(thisFile){iSubjectSession+2,1} = x.S.SubjectSessionID{iSubjectSession, 1}; % I'm not 100% sure this is correct, how do I test this?
```
MichaelStritt, line 120: @jan-petr: I haven't properly tested this part, yet. Other than that it seems to work just fine.

MichaelStritt:
```matlab
====================================== TEST RESULTS ======================================

                        name                           unit           tests        passed
    ____________________________________________    __________    _____________    ______

    'xASL_adm_CatchNumbersFromString'               'function'    [1×3  struct]    true  
    'xASL_adm_CheckFileCount'                       'function'    [1×2  struct]    true  
    'xASL_adm_CompareLists'                         'function'    [1×2  struct]    true  
    'xASL_adm_CorrectName'                          'function'    [1×3  struct]    true  
    'xASL_adm_OrderFields'                          'function'    [1×1  struct]    true  
    'xASL_bids_BIDS2Legacy'                         'function'    [1×2  struct]    true  
    'xASL_bids_Config'                              'function'    [1×1  struct]    true  
    'xASL_bids_CreateDatasetDescriptionTemplate'    'function'    [1×2  struct]    true  
    'xASL_bids_JsonCheck'                           'function'    [1×3  struct]    true  
    'xASL_bids_VendorFieldCheck'                    'function'    [1×1  struct]    true  
    'xASL_io_ExportVTK'                             'function'    [1×5  struct]    true  
    'xASL_io_Nifti2Im'                              'function'    [1×1  struct]    true  
    'xASL_stat_PSNR'                                'function'    [1×2  struct]    true  
    'xASL_stat_QuantileNan'                         'function'    [1×1  struct]    true  
    'xASL_stat_StdNan'                              'function'    [1×3  struct]    true  
    'xASL_str2num'                                  'function'    [1×3  struct]    true  
    'xASL_test_GetLogContent'                       'function'    [1×2  struct]    true  
    'xASL_tsvRead'                                  'function'    [1×2  struct]    true  
    'xASL_tsvWrite'                                 'function'    [1×2  struct]    true  
    'xASL_ExploreASL_Master'                        'master'      [1×10 struct]    true  
    'xASL_module_ASL'                               'module'      [1×1  struct]    true  
    'xASL_module_Population'                        'module'      [1×1  struct]    true  
    'xASL_module_Structural'                        'module'      [1×1  struct]    true  

==========================================================================================
```

MichaelStritt:
### ADNI Case 

I ran one ADNI case to work on the import in https://github.com/ExploreASL/CustomScripts/issues/7, there I got this weird bug.

```
ERROR: ASL module terminated for subject 1: sub-sub001
ans =
    'Error using xASL_quant_M0 (line 166)
     Unknown x.Q.readoutDim specified
```

At the end of the pipeline, `x.Q.readoutDim` seems fine, so either it's not fine during the ASL module or we're doing something wrong somewhere else.

```
>> x.Q
ans = 
  struct with fields:
        LabelingType: 'PASL'
         Initial_PLD: 1900
    LabelingDuration: 700
          readoutDim: '3D'
              Vendor: 'Siemens'
            Sequence: '3D_spiral'
```

MichaelStritt:
### 10 Test datasets

Okay, so it seems there are still some minor bugs in there:

image

image

There are still some issues with those TSV files (`Philips_2DEPI_FEAST_Longitudinal_FLAIR_LGA_ASL_PartlyDone_HiQ\Population\Stats\CoV_qCBF_StandardSpace_DeepWM_n=2_15-Jul-2021_PVC0.tsv`):

```
SUBJECT	session	LongitudinalTimePoint	SubjectNList	Site	AcquisitionTime	GM_vol	WM_vol	CSF_vol	GM_ICVRatio	GMWM_ICVRatio	WMH_vol	WMH_count	MeanMotion	DeepWM_B	DeepWM_L	DeepWM_R
StudyID	...	integer	integer	integer	hhmmss	Liter	Liter	Liter	ratio GM/ICV	ratio (GM+WM)/ICV	mL	n lesions (integer)	mm	mL/100g/min	mL/100g/min	mL/100g/min
151001_1	151001_1	TimePoint_1	1	1	111637.04	0.56226	0.45775	0.36724	0.40531	0.73527	0.12	6	0.10501	1.8042	1.7556	1.8489
151001_1	151001_1	TimePoint_1	1	1	112154.26	n/a	n/a	n/a	n/a	n/a	n/a	n/a	0.15063	1.5887	1.4323	1.7774
151001_2	151001_2															
151001_2	151001_2															
```


```
=>QuantificationParameters_ASL4D	Varying number of delimiters per line.
=>QuantificationParameters_M0		Varying number of delimiters per line.
=>QuantificationParameters_ASL4D	Varying number of delimiters per line.
=>CoV_qCBF_StandardSpace_DeepWM_n=2_15-Jul-2021_PVC0		Varying number of delimiters per line.
=>CoV_qCBF_StandardSpace_DeepWM_n=2_15-Jul-2021_PVC2		Varying number of delimiters per line.
=>CoV_qCBF_StandardSpace_TotalGM_n=2_15-Jul-2021_PVC0		Varying number of delimiters per line.
=>CoV_qCBF_StandardSpace_TotalGM_n=2_15-Jul-2021_PVC2		Varying number of delimiters per line.
=>mean_qCBF_StandardSpace_DeepWM_n=2_15-Jul-2021_PVC0		Varying number of delimiters per line.
=>mean_qCBF_StandardSpace_DeepWM_n=2_15-Jul-2021_PVC2		Varying number of delimiters per line.
=>mean_qCBF_StandardSpace_TotalGM_n=2_15-Jul-2021_PVC0		Varying number of delimiters per line.
=>mean_qCBF_StandardSpace_TotalGM_n=2_15-Jul-2021_PVC2		Varying number of delimiters per line.
=>median_qCBF_StandardSpace_DeepWM_n=2_15-Jul-2021_PVC0		Varying number of delimiters per line.
=>median_qCBF_StandardSpace_TotalGM_n=2_15-Jul-2021_PVC0	Varying number of delimiters per line.
...
```

MichaelStritt:
### 10 Test datasets

After fixing the TSV writing in cfcaae1, the results table already seems to be improved by a lot.

image

For some of these comparison values we don't expect an exact match, but there are still some bugs I think 😞 

image

MichaelStritt:
After the fix cccff25 I get values for dataset 7 e.g., but they are a bit further away from the reference values.

image

HenkMutsaerts:
> After the fix [cccff25](https://github.com/ExploreASL/ExploreASL/commit/cccff2574f5cb7088cdc54d8b59cacb58c38cc0c) I get values for dataset 7 e.g., but they are a bit further away from the reference values.
> 
> image

That's fine, you didn't change quantification

MichaelStritt:
@jan-petr: So I just tested dataset 7 another time. We definitely run the correct SPM routine there. I don't fully understand why the quantification is a bit off for some of the datasets. Does that really have something to do with my changes? Or can we merge the PR like this? Should I rerun the 10 test datasets or do you have an idea which log files I should check another time?

MichaelStritt:
### Current results: revamp/feature branch

image

image

image

MichaelStritt:
### Current results: develop

image

image

MichaelStritt:
- Datasets 3, 7, 8, 10 are basically equal to the reference
- ApplyQuantification, Vendor, Sequence, readOutDim, bUseBasil seem to be missing in the pdf

jan-petr:
Comparing Siemens_2DEPI_noBsup_FLAIR_LoQ before and after...
- Native space T1, GM, WMH look the same
- Standard space structural differ in resolution. PV_pGM, rT1, rcT1, mrc1T1 - they all appear to be slightly differently smoothed. This can be caused by the fact that ASL effective resolution was estimated differently for these two. That can happen if the vendor/sequence was incorrectly assigned during the resampling - I can find out where in the code, but having readoutDim, sequence, vendor correctly at all time is the key to fix this.

jan-petr:
xASL_spm_deformations. Line 54 :D Quality wrong...

jan-petr (changes requested):
ApplyQuantification, Vendor, Sequence, readOutDim, bUseBasil seem to be missing in the pdf

MichaelStritt:
image

image

image

MichaelStritt:
```
ERROR: Structural module terminated for subject 1: cruise001
ans =
     Error using xASL_Move line 48
     Source doesnt exist: 
     .\Philips_2DEPI_FLAIR_Segm_PWI4D_Quant_Parameters_LoQ\cruise001\y_T1.nii
     
     Error in xASL_wrp_SegmentT1w line 284
```


```
ERROR: Structural module terminated for subject 1: 101
ans =
    'Unable to perform assignment because dot indexing is not supported for variables of this type.
     Error in xASL_module_Structural line 349
```

jan-petr:
ataset Root        ...lips_2DEPI_FEAST_Longitudinal_FLAIR_LGA_ASL_PartlyDone_HiQ/dataPar.json
Import Modules      
Process Modules     STRUCTURAL ASL POPULATION 
bPause              False
iWorker             1
nWorkers            1
ExploreASL v1.8.0_BETA initialized ... 
Detected deprecated fields in dataPar.json ...
Deprecated field: name
Deprecated field: subject_regexp
Deprecated field: Vendor
Deprecated field: readout_dim


Warning: Dataset can not be loaded, there is no derivatives directory, try to run the import first... 
> In xASL_init_DataLoading (line 76)
  In ExploreASL_Master (line 74)
  In xASL_qc_TestExploreASL (line 340) 
Automatically defining sessions...

MichaelStritt:
> ```
> Dataset Root        ...lips_2DEPI_FEAST_Longitudinal_FLAIR_LGA_ASL_PartlyDone_HiQ/dataPar.json
> ==============================================================================================
> Detected deprecated fields in dataPar.json ...
> Deprecated field: name
> Deprecated field: subject_regexp
> Deprecated field: Vendor
> Deprecated field: readout_dim
> 
> 
> Warning: Dataset can not be loaded, there is no derivatives directory, try to run the import first... 
> > In xASL_init_DataLoading (line 76)
>   In ExploreASL_Master (line 74)
>   In xASL_qc_TestExploreASL (line 340) 
> Automatically defining sessions...
> ```

Ah sorry, I think we can ignore this warning. It's just there because the dataset root is a json in this case. The deprecated fields are just shown because we use the old dataPar.json files, which should work for backwards compatibility.

jan-petr:
> > ```
> > Dataset Root        ...lips_2DEPI_FEAST_Longitudinal_FLAIR_LGA_ASL_PartlyDone_HiQ/dataPar.json
> > ==============================================================================================
> > Detected deprecated fields in dataPar.json ...
> > Deprecated field: name
> > Deprecated field: subject_regexp
> > Deprecated field: Vendor
> > Deprecated field: readout_dim
> > 
> > 
> > Warning: Dataset can not be loaded, there is no derivatives directory, try to run the import first... 
> > > In xASL_init_DataLoading (line 76)
> >   In ExploreASL_Master (line 74)
> >   In xASL_qc_TestExploreASL (line 340) 
> > Automatically defining sessions...
> > ```
> 
> Ah sorry, I think we can ignore this warning. It's just there because the dataset root is a json in this case. The deprecated fields are just shown because we use the old dataPar.json files, which should work for backwards compatibility.

that's what I thought. But we should fix the testDataSet, or leave it intentionally in?! :)

MichaelStritt:
> 
> 
> > > ```
> > > Dataset Root        ...lips_2DEPI_FEAST_Longitudinal_FLAIR_LGA_ASL_PartlyDone_HiQ/dataPar.json
> > > ==============================================================================================
> > > Detected deprecated fields in dataPar.json ...
> > > Deprecated field: name
> > > Deprecated field: subject_regexp
> > > Deprecated field: Vendor
> > > Deprecated field: readout_dim
> > > 
> > > 
> > > Warning: Dataset can not be loaded, there is no derivatives directory, try to run the import first... 
> > > > In xASL_init_DataLoading (line 76)
> > >   In ExploreASL_Master (line 74)
> > >   In xASL_qc_TestExploreASL (line 340) 
> > > Automatically defining sessions...
> > > ```
> > 
> > 
> > Ah sorry, I think we can ignore this warning. It's just there because the dataset root is a json in this case. The deprecated fields are just shown because we use the old dataPar.json files, which should work for backwards compatibility.
> 
> that's what I thought. But we should fix the testDataSet, or leave it intentionally in?! :)

I planned on creating a BIDS structure for the test datasets, easiest thing to do.

jan-petr:
CAT12 indeed crashes here because of bad contrast:
Philips_2DEPI_FLAIR_Segm_PWI4D_Quant_Parameters_LoQ

jan-petr:
> ### Philips_2DEPI_FLAIR_Segm_PWI4D_Quant_Parameters_LoQ

The issue is `SegmentSPM12`. 
In the code, you've renamed that to 
`x.modules.structural.bSegmentSPM12`
In dataPar.json, you have
x.settings.SegmentSPM12 = 1

Please make sure that dataPar.json is correct (x.modules.structural.bSegmentSPM12 or x.SegmentSPM12) and that x.SegmentSPM12 is correctly converted to bSegmentSPM12 in the correct position.

> ### Philips_2DEPI_Segmentation_2D_T2_SPM_HiQ
Same issue, except that there it is at x.SegmentSPM12.

So two components
1. Correct name in dataPar in the first dataset
2. Old name in the second dataset, but making sure that it is correctly converted.

MichaelStritt:
> 
> 
> > ### Philips_2DEPI_FLAIR_Segm_PWI4D_Quant_Parameters_LoQ
> 
> The issue is `SegmentSPM12`. In the code, you've renamed that to
> `x.modules.structural.bSegmentSPM12` In dataPar.json, you have
> `x.settings.SegmentSPM12 = 1`

Ahhh... sorry, I first moved it to settings, but then we discussed it another time and realized that it would make more sense to have it in the Structural module.

> Please make sure that dataPar.json is correct (x.modules.structural.bSegmentSPM12 or x.SegmentSPM12) and that x.SegmentSPM12 is correctly converted to bSegmentSPM12 in the correct position.

Yeah, sorry 😞 

> Philips_2DEPI_Segmentation_2D_T2_SPM_HiQ
> Same issue, except that there it is at x.SegmentSPM12.
> 
> So two components
> 
>     1. Correct name in dataPar in the first dataset
>     2. Old name in the second dataset, but making sure that it is correctly converted.

I'm on it 👍

jan-petr:
We are getting somewhere - two datasets differ <0.1-0.5% which is probably Linux vs Windows difference.

But `Philips_2DEPI_FLAIR_Segm_PWI4D_Quant_Parameters_LoQ` differs 4% and `Philips_2DEPI_FLAIR_DisabledQuantification_M0_LoQ` differs a LOT.

@MichaelStritt - I will try to investigate the data and code a bit, but it would be great if you could again run old data+old code vs new data+new code and send me the full directories - then we can for sure quickly discover the issue.

MichaelStritt:
### Test results

grafik

grafik

grafik

### Evaluation

@jan-petr: Awesome, I think it works a lot better now 😄 

In dataset 4 there still seems to be a minor bug. The printing of "ApplyQuantification" seems to work correctly, but the values aren't correct.

grafik

I think I just forgot to add it in xASL_adm_GetDeprecatedFields 😞 

grafik

Would be an easy fix though 👍 

For dataset 5 there's still a minor bug with Segment_SPM12 I think. After the processing both versions are still there, so the old one wasn't removed, but both have the correct state:

grafik

MichaelStritt:
### ApplyQuantification

```matlab
>> pathTest = '.\New_TestData\Philips_2DEPI_FLAIR_DisabledQuantification_M0_LoQ\dataPar.json';
>> x = xASL_io_ReadDataPar(pathTest);
Detected deprecated fields in dataPar.json ...
Deprecated field: name
Deprecated field: subject_regexp
Deprecated field: ApplyQuantification
Deprecated field: Quality

>> x
x = 
  struct with fields:
           S: [1×1 struct]
           D: [1×1 struct]
           P: [1×1 struct]
           Q: [1×1 struct]
     modules: [1×1 struct]
     dataset: [1×1 struct]
    settings: [1×1 struct]
    external: [1×1 struct]
         dir: [1×1 struct]
        opts: [1×1 struct]

>> x.Q
ans = 
  struct with fields:
    ApplyQuantification: [5×1 double]
```

MichaelStritt:
### Philips_2DEPI_FLAIR_DisabledQuantification_M0_LoQ

So in https://github.com/ExploreASL/TestDataSets/commit/0c9a36d627e90b5b1e0ad9c749f435aa9c95bb7f I fixed the JSON file, because ` [0 0 0 0 0] ` isn't correct **JSON** notation. I think we actually always read it wrong. I think the correct values are used now, but we should be careful with incorrect **JSON** notation and reading these arrays as columns/rows. At least the PDF only seems to print the first value:

grafik

I did a rerun of the ASL and Population module for this dataset. The values are a lot better now. They're not that overestimated anymore, but still not acceptable.

MichaelStritt:
### Philips_2DEPI_FLAIR_Segm_PWI4D_Quant_Parameters_LoQ

Here the backwards compatibility script seems to do what it is supposed to. The fields are read and moved to the correct substructures. The actual bug 🐛  must be in the pipeline.

```matlab
>> pathTest = '.\New_TestData\Philips_2DEPI_FLAIR_Segm_PWI4D_Quant_Parameters_LoQ\dataPar.json';
>> x = xASL_io_ReadDataPar(pathTest);
Detected deprecated fields in dataPar.json ...
Deprecated field: name
Deprecated field: subject_regexp
Deprecated field: Quality
Deprecated field: SegmentSPM12
Deprecated field: bAffineRegistration

>> x
x = 
  struct with fields:
           Q: [1×1 struct]
           S: [1×1 struct]
           D: [1×1 struct]
           P: [1×1 struct]
     modules: [1×1 struct]
     dataset: [1×1 struct]
    settings: [1×1 struct]
    external: [1×1 struct]
         dir: [1×1 struct]
        opts: [1×1 struct]

>> x.Q
ans = 
  struct with fields:
    SliceReadoutTime: 'shortestTR'

>> x.dataset
ans = 
  struct with fields:
             name: 'Philips_2DEPI_FLAIR_Segm_PWI4D_Quant_Parameters_LoQ'
    subjectRegexp: '^cruise\d{3}$'

>> x.modules.asl.bAffineRegistration
ans =
     2

>> x.modules.structural.bSegmentSPM12
ans =
     1
```

MichaelStritt:
@jan-petr: do you think this part in `xASL_bids_parms2BIDS` could lead to problems regarding the `SliceReadoutTime`? Especially if `SliceReadoutTime` isn't in the base level of `x` anymore.

```matlab
% For SliceTime - compare the difference of the first two fields
if strcmpi(FieldsA{iA},'SliceTiming') || strcmpi(FieldsA{iA},'SliceReadoutTime') || strcmpi(FieldsA{iA},'SiemensSliceTime')
valueCheck = inXasl.(FieldsA{iA});
if length(valueCheck) > 1
    valueCheck = valueCheck(2) - valueCheck(1);
end
else
     valueCheck = inXasl.(FieldsA{iA});
end
```

jan-petr:
> @jan-petr: do you think this part in `xASL_bids_parms2BIDS` could lead to problems regarding the `SliceReadoutTime`? Especially if `SliceReadoutTime` isn't in the base level of `x` anymore.
> 
> ```matlab
> % For SliceTime - compare the difference of the first two fields
> if strcmpi(FieldsA{iA},'SliceTiming') || strcmpi(FieldsA{iA},'SliceReadoutTime') || strcmpi(FieldsA{iA},'SiemensSliceTime')
> valueCheck = inXasl.(FieldsA{iA});
> if length(valueCheck) > 1
>     valueCheck = valueCheck(2) - valueCheck(1);
> end
> else
>      valueCheck = inXasl.(FieldsA{iA});
> end
> ```
> 
> **Edit:** At least for the usage of `xASL_bids_parms2BIDS` within `xASL_adm_LoadParms` this shouldn't be a problem, since the deprecated field check happens afterwards.
> 
> And this part in `xASL_stat_GetDICOMStatistics` doesn't work as intended anymore either I think 
> 
> ```matlab
> matFields = {'RepetitionTime' 'EchoTime' 'NumberOfTemporalPositions' 'MRScaleSlope' 'RescaleSlopeOriginal' 'RescaleIntercept' 'Scanner' 'SliceReadoutTime'};
> nFields = length(matFields);
> PathTSV = fullfile(x.D.DICOMparameterDir, ['QuantificationParameters_' ScanType '.tsv']);
> ```

Not really sure that this isn't a problem. xASL_bids_parms2BIDS is potentially called on many places in the pipeline (potentially), so it should correctly convert between BIDS and parms and back. And between old and new parms as well... 

So all parameters that are in the new format not in the root of X should be correctly moved to the root before the conversion. And then moved back if the output is parms and not BIDS. See how this is done for the x.Q field. Not sure where the new fields are, but it should be able to handle both new fields and deprecated fields and correctly place it.

MichaelStritt:
> 
> Not really sure that this isn't a problem. xASL_bids_parms2BIDS is potentially called on many places in the pipeline (potentially), so it should correctly convert between BIDS and parms and back. And between old and new parms as well...
> 
> So all parameters that are in the new format not in the root of X should be correctly moved to the root before the conversion. And then moved back if the output is parms and not BIDS. See how this is done for the x.Q field. Not sure where the new fields are, but it should be able to handle both new fields and deprecated fields and correctly place it.


But for moving them back, if we move them based on the table at the end of the script (b26cc2f), this should work correctly, right?

So for moving them to the base level for the comparison at the beginning of the script, I maybe also need to check of x.modules.(...)?

MichaelStritt:
### Philips_2DEPI_FLAIR_DisabledQuantification_M0_LoQ

@jan-petr: It's really weird. So in `xASL_module_ASL` line 42-50 we have:

```matlab
K>> x.Q.ApplyQuantification
ans =
     0
     0
     0
     0
     0
```

But when we reach `xASL_quant_M0` line 65, there's:

```matlab
K>> x.Q.ApplyQuantification
ans =
     1
     1
     1
     1
     0
```

Is the code supposed to do that? That's a bit in-transparent or wrong 😄

MichaelStritt:
@jan-petr: I think we're pretty close

image

image

image

Functions/xASL_adm_GetDeprecatedFields.m
```diff
@@ -85,11 +86,14 @@
     'bSkipIfNoM0'                   'settings'  'bSkipIfNoM0'                        '';...
     'bSkipIfNoFlair'                'settings'  'bSkipIfNoFlair'                     '';...
     'bRunModule_DARTEL'             'modules'   'bRunDARTEL'                         '';...
+    'bRunDARTEL'                    'modules'   'bRunDARTEL'                         '';...
```
jan-petr, line 89: That's modules, not modules.structural? Just checking.

Functions/xASL_adm_GetDeprecatedFields.m
```diff
@@ -85,11 +86,14 @@
...
     'bRunModule_LongReg'            'modules'   'bRunLongReg'                        '';...
+    'bRunLongReg'                   'modules'   'bRunLongReg'                        '';...
     'ForceInclusionList'            'dataset'   'ForceInclusionList'                 '';...
     'bAutomaticallyDetectFS'        'external'  'bAutomaticallyDetectFS'             '';...
     'MakeNIfTI4DICOM'               'settings'  'MakeNIfTI4DICOM'                    '';...
     'SegmentSPM12'                  'modules'   'structural'                         'bSegmentSPM12';...
```
jan-petr, line 95: Wasn't it Segment_SPM12?

Function `Functions/xASL_adm_GetDeprecatedFields.m` has it together, function `Functions/xASL_init_DefineDataDependentSettings.m` has it with the underscore...

Functions/xASL_bids_parms2BIDS.m
```diff
@@ -327,8 +321,53 @@
...
+        if isfield(inXasl,firstLevelStructs{iStruct})
+            FieldsQ = fields(inXasl.(firstLevelStructs{iStruct}));
+            for iField=1:length(FieldsQ)
+                inXasl.(FieldsQ{iField}) = inXasl.(firstLevelStructs{iStruct}).(FieldsQ{iField});
+            end
+            inXasl = rmfield(inXasl,firstLevelStructs{iStruct});
+        end
+    end
+    
+    % Iterate over second level structs
```
jan-petr, line 355: Above is an error! The first-level struct `modules` will be checked for Q-Fields and deleted at the end of the previous cycle. There will never be a chance that this reaches this point of code with anything left in x.modules...

So modules have to be excluded from the first level stuff - or not deleted...

Functions/xASL_bids_parms2BIDS.m
```diff
@@ -327,8 +321,53 @@
...
+    if isfield(inXasl,'modules')
+        for iStruct = 1:numel(secondLevelStructs)
+            if isfield(inXasl.modules,secondLevelStructs{iStruct})
+                FieldsQ = fields(inXasl.modules.(secondLevelStructs{iStruct}));
+                for iField=1:length(FieldsQ)
+                    inXasl.(FieldsQ{iField}) = inXasl.modules.(secondLevelStructs{iStruct}).(FieldsQ{iField});
+                end
+                inXasl = rmfield(inXasl.modules,secondLevelStructs{iStruct});
```
jan-petr, line 363: error!! `inxAasl.modules = rmfield(inXasl.modules,....)`

Or even better - move this one cycle below and delete the whole `modules` stuff. 
`inXasl = rmfield(inXasl,'modules')`

Functions/xASL_bids_parms2BIDS.m
```diff
@@ -327,8 +321,53 @@
...
+    % Move x.Q, x.settings, x.dataset, x.modules, x.modules.structural,
+    % x.modules.asl, and x.modules.population parameters to main level of x
+    
+    firstLevelStructs = {'Q', 'settings', 'dataset', 'modules'};
+    secondLevelStructs = {'structural', 'asl', 'population'};
+    
+    % Iterate over first level structs
+    for iStruct = 1:numel(firstLevelStructs)
+        if isfield(inXasl,firstLevelStructs{iStruct})
+            FieldsQ = fields(inXasl.(firstLevelStructs{iStruct}));
```
jan-petr, line 347: `FieldsQ` is a bit confusing name. I know where it comes from, but it's not always `Q` anymore. What about 'subFields' or something like that?

jan-petr (changes requested):
See inline comments for some bugs...

Functions/xASL_bids_parms2BIDS.m
MichaelStritt: Oh damn... I wasn't thinking of that. I'm not even sure if we have to move fields from there to the base level anyway, since it's mostly processing settings so far. Let's talk later in our coding session.

Functions/xASL_adm_GetDeprecatedFields.m
MichaelStritt: I added that line to make sure that all the fields are moved correctly within `Functions/xASL_bids_parms2BIDS`. So If we move `x.modules.bRunDARTEL` to the base level of x e.g., then we can't move bRunModule_DARTEL back. I could have done a better job describing this though.

MichaelStritt:
@jan-petr: I don't think there are any bids relevant fields in x.modules.(...) anyway. Maybe we should only check Q, settings and dataset: 147a173

Functions/xASL_bids_parms2BIDS.m
MichaelStritt: 147a173

Functions/xASL_bids_parms2BIDS.m
MichaelStritt: 147a173

Functions/xASL_bids_parms2BIDS.m
MichaelStritt: 147a173

MichaelStritt:
@jan-petr: `Functions/xASL_bids_parms2BIDS.m` lines 263-265, does this still work correctly if we add the `GeneratedBy` field e.g.? Because we can't check `isnan()` of a struct.

```matlab
if (sum(isnan(inBids.(FieldsA{iA})))==0) && (sum(inBids.(FieldsA{iA}) ~= 0) > 0) && (sum(~isinf(inBids.(FieldsA{iA})))>0) && (~isempty(inBids.(FieldsA{iA})))
    outParms.(FieldsA{iA}) = inBids.(FieldsA{iA});
end
```

Functions/xASL_adm_GetDeprecatedFields.m
jan-petr: Totally fine. You can resolve. I was just double-checking...

jan-petr:
Thanks Michael - all above looks good! segment_SPM12...

Functions/xASL_adm_GetDeprecatedFields.m
MichaelStritt: I think this is the easiest fix: e1ae03f 
I also updated the table 👍 
This way we use the same script to check the backwards compatibility everywhere 👍

MichaelStritt:
### Current revamp branch results

image

jan-petr:
Some thoughts on Philips_2DEPI_Lesion_NoFLAIR_NoM0_HiQ: Report identical, native T1 identical, native ASL identical, standard space all identical. But also the TSV tables in stats are totally equal - both equal to the new version and neither equal to the reference table.

jan-petr (approved):
FInally OK.

MichaelStritt:
## Final check of stats files

### Philips_2DEPI_FLAIR_DisabledQuantification_M0_LoQ

Everything identical for revamp vs. develop 👍 

### Philips_2DEPI_FLAIR_Segm_PWI4D_Quant_Parameters_LoQ

More or less minor deviations

CoV_qCBF_StandardSpace_DeepWM_n=1_26-Jul-2021_PVC0
`0.53514	0.50813	0.56235`	vs.  `0.51967	0.4923	0.54709`
CoV_qCBF_StandardSpace_DeepWM_n=1_26-Jul-2021_PVC2
`11.012	9.5511	13.0241`	vs.  `10.6944	9.2535	12.6714`
...

### Philips_2DEPI_Lesion_NoFLAIR_NoM0_HiQ

Everything identical for revamp vs. develop 👍

MichaelStritt:
### Develop
```
0.53514	0.50813	0.56235	
11.012	9.5511	13.0241	
0.43237	0.43763	0.42662	
3.5452	3.6436	3.4441	
...
```

### Affine Registration 0
```
0.52546	0.49493	0.55606
10.7931	9.3276	12.7621
0.42595	0.43039	0.4211
3.4958	3.5904	3.3991
...
```

### Affine Registration 1
```
0.52546	0.49493	0.55606
10.7931	9.3276	12.7621
0.42595	0.43039	0.4211
3.4958	3.5904	3.3991
...
```

### Affine Registration 2
```
0.51967	0.4923	0.54709
10.6944	9.2535	12.6714
0.41847	0.42215	0.41443
3.4388	3.5225	3.3529
...
```

MichaelStritt:
@jan-petr: I'll give it another try with these changes in place: https://github.com/ExploreASL/TestDataSets/commit/0a7cfc696f36c6e49ff11e9a13e4d29dc9fb7522

MichaelStritt:
### Revamp with changes in Testdatasets

I think the problem was really in the test dataset. Previously the prioritization of the metadata fields didn't work correctly. Now it should and that is why we get better results with: https://github.com/ExploreASL/TestDataSets/commit/0a7cfc696f36c6e49ff11e9a13e4d29dc9fb7522

```
0.53514	0.50813	0.56235
11.012	9.5511	13.0241
0.43237	0.43763	0.42662
3.5452	3.6436	3.4441
```

jan-petr (approved):
OK to merge, but we should revert the TestDataSet - good we discovered the error, but the parameters are correctly there.

Edit: Can you make also a new issue to compare the develop with 1.7.0 in the last two datasets?

MichaelStritt:
> 
> 
> OK to merge, but we should revert the TestDataSet - good we discovered the error, but the parameters are correctly there.
> 
> Edit: Can you make also a new issue to compare the develop with 1.7.0 in the last two datasets?

I've already reset the Testdatasets 👍 
Creating a the issue you talked about right now and will mention #717 there.

PR #733, opened 2021-07-20
Fixes #717 revamp

### LINKED ISSUE #717: Revamp x structure fields of import & processing

MichaelStritt:
### Linked issue

Check out #717

### Comments

I moved all the relevant code changes, which do not actually move fields but optimize the code, from the original #717 issue to this branch and PR. I'll test these changes tonight and then we can move forward step by step. Hopefully it is easier to revamp the parameters this way. The original PR can later be closed without a merge if this works.


MichaelStritt:
### Test results

image

image

PR #741, opened 2021-07-26
Fixes #575 cicero requests

### LINKED ISSUE #575: CICERO requests
### Description

* Create Flip function, that will flip a NIfTI left-right:
IN = 4V_ICA_L.nii (allow list)
OUT = 4V_ICA_L_flippedLeftRight.nii
Function = xASL_im_FlipNiftiLeftRight
Uses = xASL_im_Flip
* Allow custom ROIs (e.g. vascular territory 4V_ICA_L.nii) as ROI input to obtain CBF or sCoV, also function to allow combining regions (perhaps exists already?)

### Requirements

### How to test

Optional: add description about how to test here

### Release notes
#575 Simplification/revamp of some scripts that check image flips and report on them

HenkMutsaerts:
### Linked issue

Fixes #575


Functions/xASL_qc_CollectQC_ASL.m
```diff
@@ -52,7 +52,11 @@
     % The determinant of the current matrix and old matrix should be the same,
     % otherwise this is suspicious of a left-right flip.
     PathOrientationResults = fullfile(x.SESSIONDIR,'xASL_qc_PrintOrientation_RigidRegASL.tsv');
-    ASL = xASL_im_DetermineFlip(x, iSubject, PathOrientationResults, ASL);
+    ASL.LR_flip_YesNo = uint8(xASL_im_DetermineFlip(PathOrientationResults));
```
MichaelStritt, line 55: We could add this parameter to the new sub-structs of x while we're at it. So Why not use `x.modules.asl.LR_flip_YesNo` instead 😃

Functions/xASL_qc_CollectQC_Structural.m
```diff
@@ -35,10 +35,13 @@
     Struct.ID = x.SUBJECTS{iSubject};
 
     PathOrientationResults = fullfile(x.dir.SUBJECTDIR,'xASL_qc_PrintOrientation_RigidRegT1.tsv');
-    TempS = xASL_im_DetermineFlip(x, iSubject, PathOrientationResults, Struct);
-    Struct.T1w_LR_flip_YesNo = TempS.LR_flip_YesNo;
+    Struct.T1w_LR_flip_YesNo = uint8(xASL_im_DetermineFlip(PathOrientationResults));
```
MichaelStritt, line 38: Same here. We could use `x.modules.structural.T1w_LR_flip_YesNo`, this way you can still find it in the resulting x struct 😃

Modules/SubModule_Structural/xASL_wrp_VisualQC_Structural.m
```diff
@@ -105,7 +105,7 @@ function xASL_wrp_VisualQC_Structural(x)
 
 %% -----------------------------------------------------------------------------------
 %% 5) Final QCs
-xASL_qc_PrintOrientation(x.dir.SUBJECTDIR, x.P.Path_T1, x.dir.SUBJECTDIR, 'RigidRegT1');
+xASL_qc_PrintOrientation(x.P.Path_T1, x.dir.SUBJECTDIR, 'RigidRegT1');
```
MichaelStritt, line 108: So this is probably the one where we have to be careful with #717, but looks good so far.

xASL_qc_ReportLeftRightFlips.m
```diff
@@ -0,0 +1,21 @@
+% Check for left-right flips for all NIfTIs inside a folder
```
MichaelStritt, line 1: This file shouldn't be in the main directory of ExploreASL

MichaelStritt (changes requested):
We should definitely move `xASL_qc_ReportLeftRightFlips.m` and check out my other comments about the new usage of x struct fields. Could be helpful. I also added @jan-petr for the review.

Functions/xASL_qc_CollectQC_Structural.m
MichaelStritt: Ah sorry, I saw that this `ASL` struct is stored in x.Output later on. Then my request doesn't really make sense 👍 I'll mark it as resolved.

Functions/xASL_qc_CollectQC_ASL.m
MichaelStritt: Ah sorry, I saw that this ASL struct is stored in x.Output later on. Then my request doesn't really make sense 👍 I'll mark it as resolved.

xASL_qc_ReportLeftRightFlips.m
MichaelStritt: Had another look. It seems as if there is also another function with the same name under `Functions`. Maybe move this to `CustomScripts` because of the fixed paths etc.?

Modules/SubModule_Structural/xASL_wrp_VisualQC_Structural.m
HenkMutsaerts: Did you change the input arguments of `xASL_qc_PrintOrientation` in #717? That is confusing, I though #717 was about moving structures only... Are you sure here?

xASL_qc_ReportLeftRightFlips.m
HenkMutsaerts: Yes you're right, fixed this now in last commits

Modules/SubModule_Structural/xASL_wrp_VisualQC_Structural.m
HenkMutsaerts: OK

Functions/xASL_qc_CollectQC_ASL.m
HenkMutsaerts: Yes the `x.output` carries all QC stuff, we could resort this later.

MichaelStritt (approved):
👍

jan-petr (approved):
OK

PR #749, opened 2021-07-29
Fixes #611 Participants.tsv -> participants.tsv

### LINKED ISSUE #611: Participants.tsv in BIDS format
### Description

The current participants.tsv matches ASL and not BIDS format.

### Requirements

- The filename should start with minuscule `p` - this needs to be corrected everywhere - also in ExploreASL, not only in import. 
- Also, the subject notation must follow BIDS: [bids definition](https://bids-specification.readthedocs.io/en/stable/03-modality-agnostic-files.html#participants-file)
- Change also the current notation to BIDS and make sure that the reading is also backwards compatible.
- What about `Participant_id` given and session not? Treat it like the same for all sessions?!

### Release notes

Adapt participants.tsv to BIDS format

MichaelStritt:
### Linked issue

Check out #611

### How to test

Run test dataset in BIDS format with old and new `participants.tsv` / `Participants.tsv`


MichaelStritt:
@jan-petr: The way I wrote this in cb286d9 it basically works for that individual script, but that's not really how we should do this. Maybe we should rather revert cb286d9 and do this regexp check ...
```matlab
    subjectNameOld = subjectName;
    if isempty(regexp(subjectNameOld,'sub-')==1)
        subjectName = ['sub-' subjectNameOld];
        fprintf('Change subject name: %s -> %s\n',subjectNameOld,subjectName);
    end
``` 
... at the beginning of the pipeline and use it to rename the overall dataset. Otherwise you will always suffer from subsequent faults like these ones ...
```console
Printing csv-files with volume statistics...100%
Change subject name: 109 -> sub-109
Change subject name: 109 -> sub-109
Change subject name: 109 -> sub-109
Change subject name: 109 -> sub-109
Change subject name: 109 -> sub-109
Change subject name: 109 -> sub-109
Change subject name: 109 -> sub-109
050_GetVolumeStatistics was performed
Collecting motion metadata with motion statistics:100%
Saving motion plot to .\Siemens_2DEPI_noBsup_FLAIR_LoQ\Population\MotionASL\Overview_motion_pair-exclusion.jpg
Change subject name: 109 -> sub-109
060_GetMotionStatistics was performed
Loading & saving individual parameter files...100%
065_GetRegistrationStatistics was performed
participants.tsv (BIDS) detected, loading...
Loading AcquisitionTime:100%
Variable AcquisitionTime not included because all subjects were missing
Loading GM_vol:100%
Variable GM_vol not included because all subjects were missing
Loading WM_vol:100%
Variable WM_vol not included because all subjects were missing
Loading CSF_vol:100%
Variable CSF_vol not included because all subjects were missing
Loading GM_ICVRatio:100%
Variable GM_ICVRatio not included because all subjects were missing
Loading GMWM_ICVRatio:100%
Variable GMWM_ICVRatio not included because all subjects were missing
Loading WMH_vol:100%
Variable WMH_vol not included because all subjects were missing
Loading WMH_count:100%
Variable WMH_count not included because all subjects were missing
Loading MeanMotion:100%
Variable MeanMotion not included because all subjects were missing
```

MichaelStritt:
Move this ...

```matlab
    if isempty(regexp(subjectName ,'sub-')==1)
        subjectName = ['sub-' subjectName ];
        fprintf('Change subject name: %s -> %s\n',subjectNameOld,subjectName);
    end
```

... to the import (in a separate issue?).

Functions/xASL_init_LoadMetadata.m
```diff
@@ -170,10 +170,20 @@
 %xASL_bids_LoadParticipantTSV 
 
     %% Load participants.tsv
-    PathTSV = fullfile(x.D.ROOT, 'Participants.tsv');
+    PathTSV = fullfile(x.D.ROOT, 'participants.tsv');
+    PathTSVold = fullfile(x.D.ROOT, 'Participants.tsv');
+    bParticipantsTSV = false;
```
jan-petr, line 175: to me, it is better to rename it if it detects the old one...

Functions/xASL_init_LoadMetadata.m
```diff
@@ -170,10 +170,20 @@
...
     if exist(PathTSV, 'file')
-        fprintf('Participants.tsv (BIDS) detected, loading...\n');
         CellArray = xASL_tsvRead(PathTSV);
+        bParticipantsTSV = true;
+    elseif exist(PathTSVold, 'file')
+        CellArray = xASL_tsvRead(PathTSV);
+        bParticipantsTSV = true;
+    end
+
+    if bParticipantsTSV
```
jan-petr, line 185: especially since the 'participatns.tsv' is reported here.

Modules/SubModule_Population/xASL_wrp_CreatePopulationTemplates.m
```diff
@@ -33,7 +33,7 @@ function xASL_wrp_CreatePopulationTemplates(x, bSaveUnmasked, bCompute4Sets, Spe
 %                       string)
 %   bUpdateMetadata   - boolean specifying if we reload the metadata, e.g.
 %                       for potentially other defined cohorts etc in the
-%                       Participants.tsv. This can some time though.  Only relevant when computing multiple sets. 
+%                       participants.tsv. This can some time though.  Only relevant when computing multiple sets.
```
jan-petr, line 36: "this can _take_ some time" :D

jan-petr (changes requested):
Minor fixes needed. otherwise looks good.

Modules/SubModule_Population/xASL_wrp_CreatePopulationTemplates.m
MichaelStritt: b30ab8f 👍

Functions/xASL_init_LoadMetadata.m
MichaelStritt: You're right. That's probably the more straight forward way to do this. 
2b189c8 👍

Functions/xASL_init_LoadMetadata.m
MichaelStritt: 2b189c8 👍

MichaelStritt:
Okay, minor problem... [file names are not case sensitive on windows...](https://www.mathworks.com/matlabcentral/answers/270711-when-matlab-read-file-file-name-is-case-insenstive)

MichaelStritt:
### Test with existing (old) `Participants.tsv` within the BIDS root directory

It perfectly renamed, moved and filled it:

image

Functions/xASL_bids_BIDS2Legacy.m
```diff
@@ -306,11 +306,22 @@
...
+% Define file names
+pathParticipantsTSV = fullfile(pathStudy, 'participants.tsv');
+pathParticipantsTSVxASL = fullfile(pathLegacy, 'participants.tsv');
+
+% Backwards compatibility: rename to lowercase
+fileListParticipantsTSVold = xASL_adm_GetFileList(pathStudy,'^Participants.tsv$',false);
+if ~isempty(fileListParticipantsTSVold)
+    pathParticipantsTSVold = fullfile(pathStudy, 'Participants.tsv');
+    pathParticipantsTSVoldTemp = fullfile(pathStudy, 'Participants_temp.tsv');
+    xASL_Move(pathParticipantsTSVold,pathParticipantsTSVoldTemp);
```
jan-petr, line 318: Sorry, but I don't understand why this middle file is needed when you do two consecutive moves. Do I miss something?

Functions/xASL_bids_BIDS2Legacy.m
```diff
@@ -306,11 +306,22 @@
...
+    xASL_Move(pathParticipantsTSVoldTemp,pathParticipantsTSV);
```
jan-petr, line 319: Also - if both old and new exist - you might decide to leave the old one, do nothing with it and report a warning. 

No likely that people have both these files, but something can go wrong during the move etc.

Functions/xASL_init_LoadMetadata.m
jan-petr: thanks for the edit. Same as above - do we need the temp? Should we check for existence of both

jan-petr (changes requested):
Almost there :)

Functions/xASL_bids_BIDS2Legacy.m
MichaelStritt: I actually explain this quite detailed here d47a947 ...

> The previous code did not work because windows will complain if you move Participants.tsv to participants.tsv, since it regards both as the same file. The easiest fix is to rename it to another filename temporarily in between. Additionally we had the problem that you can not find the lowercase/uppercase differences using exist or xASL_exist, but xASL_adm_GetFileList uses regular expressions and worked nicely, so I implemented a fix with that. Makes the code a bit clunky though.

... and here ...

https://github.com/ExploreASL/ExploreASL/pull/749#issuecomment-889699476

Functions/xASL_bids_BIDS2Legacy.m
MichaelStritt: At least on windows this is not even possible TBH

MichaelStritt: And I'm not using xASL_Move with overwrite on (`xASL_Move(SrcPath, DstPath, bOverwrite, bVerbose)`), so it shouldn't overwrite the new version on default.

Functions/xASL_bids_BIDS2Legacy.m
jan-petr: Sorry. Didn't check the commit itself. Makes perfect sense of course!! Can you please add a comment in the code that explains this. Or some guy like me will fix it in the future :D

Functions/xASL_bids_BIDS2Legacy.m
jan-petr: Makes total sense - please add a comment in the code to avoid future 'fixes'

jan-petr (approved):
OK to merge after the 2 comments are added.

PR #753, opened 2021-08-02
Fixes #732 Multi-Session support

### LINKED ISSUE #732: NII2BIDS Convert visits to BIDS sessions
### Description

@jan-petr you incorrectly removed an underscore in the `DCM2NII` part ('_'), I removed this, so visits (legacy structure) are correctly stored as:

`sub-001_1`
`sub-001_2`

for visits 1 and 2.

In the NII2BIDS part, these two visits need to be converted to the BIDS sessions:
`sub-001/session-001`
`sub-001/session-002`

(whereas now this becomes `sub-0011` and `sub-0012`).
@jan-petr can you apply the latter? I can help but difficult to debug what you wrote already :)
So in NII (legacy format), you iterate over `iSubjectVisit` instead of `iSubject`, and then it should split `iSubjectVisit` to `iSubject` and `iSession` in BIDS.

### Release notes

Improve overall subject/visit import behavior for both anatomical and perfusion files.

MichaelStritt:
### Linked issue

Check out #732

### Major changes

- Move `BIDS_fullPipelineTest.m` to `xASL_test_FullPipelineTest.m`
- Add studyPar option to `xASL_io_ReadDataPar.m`
- Move `xASL_bids_BIDS2Legacy.m` to import module
- Add underscore `xASL_imp_DCM2NII_Subject.m`
- Rename `xASL_imp_NII2BIDS_SubjectSession.m` to `xASL_imp_NII2BIDS_Session.m`
- Rename `xASL_imp_NII2BIDS_SubjectSessionRun.m` to `xASL_imp_NII2BIDS_Run.m`
- ... split functionality to `...RunAnat.m`  &  `...RunPerf.m` 

### Release notes

Cleaning code of DCM to BIDS import


HenkMutsaerts:
I would do `ses-001` instead of` ses-1`, like with `sub-001`

MichaelStritt:
> 
> 
> I would do `ses-001` instead of` ses-1`, like with `sub-001`

That 1 comes directly from the aliases. We can do aliases that do this. According to BIDS things like plain text would be possible too though: https://github.com/ExploreASL/ExploreASL/pull/753

image

Functions/xASL_io_ReadDataPar.m
```diff
@@ -82,6 +84,9 @@
 if nargin < 1 || isempty(pathDataPar) || ~exist(pathDataPar, 'file')
```
HenkMutsaerts, line 84: If the first argument is not provided, it gives an error that the data par doesn't exist

Functions/xASL_io_ReadDataPar.m
```diff
@@ -1,10 +1,12 @@
-function [x] = xASL_io_ReadDataPar(pathDataPar)
+function [x] = xASL_io_ReadDataPar(pathDataPar,bDataPar)
 % xASL_io_ReadDataPar This function reads data-parameter .json or .m file, which contains data and processing settings, and creates the x structure.
 %
 % FORMAT:   [x] = xASL_io_ReadDataPar(pathDataPar)
 %
 % INPUT:
 %   pathDataPar     - Filename of the input parameter file with either .m or .json extension
+%   bDataPar        - Read JSON as dataPar.json (BOOLEAN, OPTIONAL, DEFAULT = true)
```
HenkMutsaerts, line 8: I would turn it around. If DataPar is the default, then the new option is `bStudyPar`. Default is off, and special case is to turn it on.

Functions/xASL_io_ReadDataPar.m
```diff
@@ -166,7 +171,9 @@
 	x = xASL_io_ReadDataPar_FixFields(x);
 	
 	%% Check deprecated fields
-    x = xASL_io_CheckDeprecatedFieldsX(x,true);
+    if bDataPar
```
HenkMutsaerts, line 174: Seems as if this was done by default previously, is this needed for the studypar as well? Or did it previously just do nothing inside this function?

Modules/SubModule_Import/xASL_imp_NII2BIDS.m
```diff
@@ -63,14 +61,13 @@ function xASL_imp_NII2BIDS(imPar, studyPath, studyParPath)
...
 	
-	%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 	%% 3. Go through all subjects and check all the M0 and ASLs and modify the JSONs
 	% This step should be completely automatic, just taking the info filled above and using it to convert to full BIDS.
 	
 	% Go through all subjects
-	listSubjects = xASL_adm_GetFileList(imPar.TempRoot,[],false,[],true);
-    for iSubject = 1:length(listSubjects)
-        xASL_imp_NII2BIDS_Subject(imPar,bidsPar,studyPar,listSubjects{iSubject});
+	listSubjectsSessions = xASL_adm_GetFileList(imPar.TempRoot,[],false,[],true);
```
HenkMutsaerts, line 68: I would add a few folder names of which we're sure that they shouldn't be subjects, like `population`, but we can do that later. Also, the folder name should have at least one letter or at least one digit. Now, folders named like `*&` or `temp` can become a subject as well

Modules/SubModule_Import/xASL_imp_NII2BIDS_Session.m
```diff
@@ -0,0 +1,71 @@
+function xASL_imp_NII2BIDS_Session(imPar, bidsPar, studyPar, listSessions, nameSubjectSession, bidsLabel, iSession)
+%xASL_imp_NII2BIDS_SubjectSession NII2BIDS conversion for a single sessions.
```
HenkMutsaerts, line 2: a single sessions -> a single session.
Also, add that this is a BIDS session and legacy visit.

Modules/SubModule_Import/xASL_imp_NII2BIDS_Session.m
```diff
@@ -0,0 +1,71 @@
...
+
+    %% 1. Make a subject directory
+    if length(listSessions)>1 || ~isempty(visitLabel)
+        if ~isempty(visitLabel)
+            sessionLabel = ['ses-' visitLabel];
+        else
+            sessionLabel = ['ses-' listSessions{iSession}(5:end)];
+        end
+
+        if ~exist(fullfile(imPar.BidsRoot,['sub-' subjectLabel],sessionLabel),'dir')
```
HenkMutsaerts, line 41: You can remove this if `~exist` line, and directly add the full folder with 'perf' at the end, `xASL_adm_CreateDir` always checks for each folder layer if it exists and create it.

Modules/SubModule_Import/xASL_imp_NII2BIDS_Session.m
```diff
@@ -0,0 +1,71 @@
...
+    %% Unpack bidsLabel struct
+    subjectLabel = bidsLabel.subject;
+    visitLabel = bidsLabel.visit;
+
+    %% 1. Make a subject directory
+    if length(listSessions)>1 || ~isempty(visitLabel)
+        if ~isempty(visitLabel)
+            sessionLabel = ['ses-' visitLabel];
+        else
+            sessionLabel = ['ses-' listSessions{iSession}(5:end)];
```
HenkMutsaerts, line 38: I don't get this fallback, why the `5:end`? Does that always work? I'm not sure if BIDS allows a single session as folder structure, I would test this in the BIDS validator, but it should be fine to always add a session folder by default.

Modules/SubModule_Import/xASL_imp_NII2BIDS_Session.m
```diff
@@ -0,0 +1,71 @@
...
+        end
+
+        if ~exist(fullfile(imPar.BidsRoot,['sub-' subjectLabel],sessionLabel),'dir')
+            xASL_adm_CreateDir(fullfile(imPar.BidsRoot,['sub-' subjectLabel],sessionLabel));
+            xASL_adm_CreateDir(fullfile(imPar.BidsRoot,['sub-' subjectLabel],sessionLabel,'perf'));
+        end
+        inSessionPath = fullfile(imPar.TempRoot,nameSubjectSession,listSessions{iSession});
+        outSessionPath = fullfile(imPar.BidsRoot,['sub-' subjectLabel],sessionLabel);
```
HenkMutsaerts, line 46: Adding `sub-` as suffix is fine as long as we check if this is not double (in case this suffix `sub-` was already in the DICOM folder). Same for `ses-`.

Modules/SubModule_Import/xASL_imp_NII2BIDS_Session.m
```diff
@@ -0,0 +1,71 @@
...
+
+        % Need to add the underscore so that it doesn't need to be added automatically and can be skipped for empty session
+        sessionLabel = ['_' sessionLabel];
+    else
+        % Session label is skipped
+        sessionLabel = '';
```
HenkMutsaerts, line 52: Aren't we making things difficult, if we don't create an additional session folder for a single session here?

Modules/SubModule_Import/xASL_imp_NII2BIDS_Session.m
```diff
@@ -0,0 +1,71 @@
...
+
+        % Only one session - no session labeling
+        if ~exist(fullfile(imPar.BidsRoot,['sub-' subjectLabel]),'dir')
```
HenkMutsaerts, line 55: Same as above. But I would always create a folder for sessions, that makes our life easier

Modules/SubModule_Import/xASL_imp_NII2BIDS_Subject.m
```diff
@@ -1,13 +1,13 @@
...
 % 
 % INPUT:
-%   imPar           - JSON file with structure with import parameter (REQUIRED, STRUCT)
-%   bidsPar         - Output of xASL_imp_Config (REQUIRED, STRUCT)
-%   studyPar        - JSON file with the BIDS parameters relevant for the whole study (REQUIRED, STRUCT)
-%   nameSubject     - name of the subject (REQUIRED, LIST)
+%   imPar               - JSON file with structure with import parameter (REQUIRED, STRUCT)
+%   bidsPar             - Output of xASL_imp_Config (REQUIRED, STRUCT)
+%   studyPar            - JSON file with the BIDS parameters relevant for the whole study (REQUIRED, STRUCT)
+%   nameSubjectSession  - name of the subject (REQUIRED, LIST)
```
HenkMutsaerts, line 10: `LIST` should be a `cell structure` right?

Modules/SubModule_Import/xASL_imp_NII2BIDS_Subject.m
```diff
@@ -16,68 +16,62 @@ function xASL_imp_NII2BIDS_Subject(imPar, bidsPar, studyPar, nameSubject)
...
 % EXAMPLE:     xASL_imp_NII2BIDS_Subject(imPar, bidsPar, studyPar, nameSubject);
 % __________________________________
 % Copyright 2015-2021 ExploreASL
 
     %% 1. Initialize
+    bidsLabel = xASL_imp_CheckForAliasInVisit(imPar,nameSubjectSession);
     
-    subjectLabel = xASL_adm_CorrectName(nameSubject,2);
+    % Make a subject directory
+    if ~exist(fullfile(imPar.BidsRoot,['sub-' bidsLabel.subject]),'dir')
```
HenkMutsaerts, line 33: Same as above

Modules/SubModule_Import/xASL_imp_NII2BIDS_Subject.m
```diff
@@ -16,68 +16,62 @@ function xASL_imp_NII2BIDS_Subject(imPar, bidsPar, studyPar, nameSubject)
...
+        xASL_adm_CreateDir(fullfile(imPar.BidsRoot,['sub-' bidsLabel.subject]));
+    end
+    
+    %% 2. Process the anat & perfusion files
+    listSessions = xASL_adm_GetFileList(fullfile(imPar.TempRoot,nameSubjectSession),'^ASL.+$',false,[],true);
```
HenkMutsaerts, line 38: Are we not mixing up sessions (visits) and runs (sessions) here? A list of sessions should in legacy be based on `_1 ` `_2` `_3` etc., not on `ASL.+`

Modules/SubModule_Import/xASL_imp_NII2BIDS_Subject.m
```diff
@@ -16,68 +16,62 @@ function xASL_imp_NII2BIDS_Subject(imPar, bidsPar, studyPar, nameSubject)
...
+    % Iterate over aliases
+    for iAlias = 1:numel(visitAliases)
+        checkExpression = regexp(nameSubjectSession, [separator visitAliases{iAlias,1} '$'], 'once');
+        if ~isempty(checkExpression) % nameSubject should end in the visit alias
+            visitName = nameSubjectSession(checkExpression:end);
+            subjectName = nameSubjectSession(1:checkExpression-1);
         end
     end
+    
+    bidsLabel.subject = xASL_adm_CorrectName(subjectName,2);
```
HenkMutsaerts, line 71: Are we sure here that we correct only the 'value' (for the keys `sub-` and `ses-`)? Or should this be done somewhere else to ensure that this is always the value?

Please also add a bit more comments, it's quite difficult to follow these scripts. I know that you're in a rush but this will backfire.

HenkMutsaerts (changes requested):
Very nice work! Only some minor questions, and the general request to add comments. One hour of typing additional comments will help us (including you ;) a lot in the near future.

Functions/xASL_io_ReadDataPar.m
MichaelStritt: Do you remember the studyPar.json that Beatriz showed us yesterday? It had some of these basic fields that shouldn't be there and that were in a sub-structure. That's partially caused by `xASL_io_CheckDeprecatedFieldsX`, which was only recently added to `xASL_io_ReadDataPar` (#717).

Modules/SubModule_Import/xASL_imp_NII2BIDS.m
MichaelStritt: Yeah, we can do things like these. I didn't change this line though, I only moved it. Maybe keep it like it is for now and then optimize later on?

Modules/SubModule_Import/xASL_imp_NII2BIDS_Session.m
MichaelStritt: That's true, but DCM2NII & NII2BIDS heavily depend on the sourceStructure.json, where we define the tokens. With the tokens we controlled that the subject ids and session ids are only the id without the prefix so far. I didn't change that part though, like the other code this was only moved from the original xASL_imp_NII2BIDS_SubjectSession.

Modules/SubModule_Import/xASL_imp_NII2BIDS_Session.m
MichaelStritt: Agree, that would make more sense for me too. I didn't change that part though. If we would change this now, we would need to update **all** flavors, too (otherwise the reference will always be off). Do you want me to create an issue for that and then we talk it through with @jan-petr?

Modules/SubModule_Import/xASL_imp_NII2BIDS_Session.m
MichaelStritt: Definitely agree, I answered above.

Modules/SubModule_Import/xASL_imp_NII2BIDS_Subject.m
MichaelStritt: Yeah, this is a bit mixed up almost everywhere. It would be a lot easier if we had written down both notations for each function/section. This should be legacy visits I think (which is BIDS sessions) and there we have ASL_1 for visit 1 etc., right? I mean... it worked so far 😆

Modules/SubModule_Import/xASL_imp_NII2BIDS_Subject.m
MichaelStritt: Like with a few other comments I have to say that this wasn't necessarily created in a rush. This was like this before and just moved. Before it was called `nameSubject`, but other than that we used the same correction function and everything. Adding more comments to code in general can be helpful though 😉 
Regarding the specific functionality: I had to figure this out first too, but a simple debug made it super clear. We extract subject/visit from the name, so you could have something like `sub-001_ses-1` or whatever and there we basically only want the 001 and the 1. We need to get rid of the hyphens and underscores, which is why we use xASL_adm_CorrectName. We did this exactly like this before though and was probably originally written by @jan-petr.

Functions/xASL_io_ReadDataPar.m
MichaelStritt: Sure, easy fix: 4dbefa8 👍

Functions/xASL_io_ReadDataPar.m
MichaelStritt: Sure, easy fix: 4dbefa8 👍

Modules/SubModule_Import/xASL_imp_NII2BIDS_Session.m
MichaelStritt: Ah yes, fixed the other comments in the header too 9c36709 👍

Modules/SubModule_Import/xASL_imp_NII2BIDS_Session.m
MichaelStritt: Sure, I changed it from mkdir to xASL_adm_CreateDir, so it kind of makes sense to remove the exist 9c36709 👍

Modules/SubModule_Import/xASL_imp_NII2BIDS_Subject.m
MichaelStritt: I think so, fixed it 088eadb 👍

Modules/SubModule_Import/xASL_imp_NII2BIDS_Subject.m
MichaelStritt: Same as above 088eadb 😄

Modules/SubModule_Import/xASL_imp_NII2BIDS_Session.m
MichaelStritt: TBH I don't know. We can ask @jan-petr about that part when he's back from vacation. I found a small bug anyway and will debug now.

MichaelStritt:
### Full BIDS Pipeline Test

I just tested these six datasets and everything seems fine to me 😄 

```
Dataset: GE_PCASL_3Dspiral_Product_pharma
Dataset: GE_PCASL_3Dspiral_volunteer
Dataset: Philips_PCASL_2DEPI_Bsup_AD1
Dataset: Philips_PCASL_3DGRASE_functional
Dataset: Siemens_PASL_2DEPI_ADNI2_B15_941_S_4764
Dataset: Siemens_PCASL_3DGRASE_AD
```

MichaelStritt:
@HenkMutsaerts: I encountered some problems in the ASL module

```
ERROR: ASL module terminated for subject 1: sub-11
ans =
    'Error using xASL_qc_PrintOrientation
     Too many input arguments.
```

I think this is related to your cherry-picked changes from #575 because the old version was `xASL_qc_PrintOrientation(DIR, reg_exp_INPUT,OUTPUT_DIR,Name)` and the new version is `xASL_qc_PrintOrientation(niftiList, outputDir, outputFile)`. Can you fix this in a separate issue?

MichaelStritt:
@HenkMutsaerts: I created the follow-up issues #754 & #755. Otherwise everything should be fine. We should focus on fixing #754 quickly.

Functions/xASL_bids_BIDS2Legacy.m
```diff
@@ -284,13 +284,14 @@
...
-    fprintf('%s\n', ['M0 parsed for ' ListASL4D{iList}]);
 else
     warning(['No ASL4D file found in ' pathLegacy]);
 end
 
-%% 10. Create DataPar.json
+%% 10. Create dataPar.json
 
 % Write DataParFile if it does not exist already
 if ~(xASL_exist(fullfile(pathLegacy, 'dataPar.json'),'file')==2)
```
HenkMutsaerts, line 297: You can change this into:

`if ~xASL_exist(fullfile(pathLegacy, 'dataPar.json'), 'file')`

so in Matlab you can omit the outer () brackets, and with 'file' you don't need to ask if it's equal to 2.

Modules/SubModule_Import/xASL_imp_BIDS2Legacy.m
```diff
@@ -26,6 +26,7 @@
 
 
     %% 1. Input check
+    fprintf('================================== BIDS to LEGACY CONVERSION =================================\n');
```
HenkMutsaerts, line 29: `LEGACY` is our internal nomenclature, this has no meaning to the users, if you call it "ExploreASL legacy" this should be fine

Modules/SubModule_Import/xASL_imp_NII2BIDS_Session.m
HenkMutsaerts: OK, if it's indeed iterating over sessions and not SubjectSession (what is very confusing as throughout the processing modules, we refer to SubjectSession as what is SubjectRun in BIDS... :)

Modules/SubModule_Import/xASL_imp_NII2BIDS_Session.m
```diff
@@ -37,11 +37,10 @@ function xASL_imp_NII2BIDS_Session(imPar, bidsPar, studyPar, listSessions, nameS
         else
             sessionLabel = ['ses-' listSessions{iSession}(5:end)];
         end
-
-        if ~exist(fullfile(imPar.BidsRoot,['sub-' subjectLabel],sessionLabel),'dir')
-            xASL_adm_CreateDir(fullfile(imPar.BidsRoot,['sub-' subjectLabel],sessionLabel));
-            xASL_adm_CreateDir(fullfile(imPar.BidsRoot,['sub-' subjectLabel],sessionLabel,'perf'));
-        end
+        
+        xASL_adm_CreateDir(fullfile(imPar.BidsRoot,['sub-' subjectLabel],sessionLabel));
```
HenkMutsaerts, line 41: Yes so you can even remove this folder, it is a folder layer above the next folder that you create, so `xASL_adm_CreateDir` will create it automatically!

Modules/SubModule_Import/xASL_imp_NII2BIDS_Session.m
HenkMutsaerts: Sure, but we don't get to see what the users do. So we only have to add a check later that it doesn't have twice `sub-` in the name, same for `ses-`.

Modules/SubModule_Import/xASL_imp_NII2BIDS_Subject.m
HenkMutsaerts: Nope, in legacy we have sessions (BIDS runs) ASL_1 ASL_2 ASL_3.
Also, in legacy we have visits (BIDS sessions) sub-001_1 sub-001_2 sub-001_3.

Modules/SubModule_Import/xASL_imp_NII2BIDS_Subject.m
```diff
@@ -50,7 +49,11 @@ function xASL_imp_NII2BIDS_Subject(imPar, bidsPar, studyPar, nameSubjectSession)
 function bidsLabel = xASL_imp_CheckForAliasInVisit(imPar,nameSubjectSession)
 
     % Get visitAliases from imPar
-    visitAliases = imPar.tokenVisitAliases(:,2);
+    if isfield(imPar,'tokenVisitAliases') && ~isempty(imPar.tokenVisitAliases) && size(imPar.tokenVisitAliases,2)>1
```
HenkMutsaerts, line 52: You can omit the `~isempty` if you check for `(size( , 2)>1`

Testing/xASL_test_FullPipelineTest.m
```diff
@@ -0,0 +1,86 @@
+%% xASL_test_FullPipelineTest: BIDS testing script
```
HenkMutsaerts, line 1: This looks familiar :), what is the difference between this script and the BIDS full pipeline test that was in `/Development/BIDS`? If this is the same one, please use `git mv` so it is "recorded" that this is not a new function but an adaption of an existing function

HenkMutsaerts (changes requested):
Minor stuff left, nearly there!

Testing/xASL_test_FullPipelineTest.m
MichaelStritt: I moved it, but the problem was the moving and renaming I guess. You don't really need me to revert all the changes because of that, right?

Modules/SubModule_Import/xASL_imp_NII2BIDS_Session.m
MichaelStritt: Yeah, but we did it like this before and I didn't want to change even more of Jans code. You always insist on me making smaller but more gradual changes and now we are starting to optimize each line of code again...

Modules/SubModule_Import/xASL_imp_BIDS2Legacy.m
MichaelStritt: 0d367c3

Functions/xASL_bids_BIDS2Legacy.m
MichaelStritt: 0d367c3 👍

Modules/SubModule_Import/xASL_imp_NII2BIDS_Subject.m
MichaelStritt: Doesn't hurt though...

Testing/xASL_test_FullPipelineTest.m
HenkMutsaerts: Only for this file, yes. I would copy it, revert this comment, and use `git mv` and then `git add` for the changes. That should keep it nice. If it doesn't work, no big deal, but it is clearer for history (and for Jan and me to know what's happening :)

Modules/SubModule_Import/xASL_imp_NII2BIDS_Subject.m
HenkMutsaerts: it doesn't add anything. If it is empty, size()>1 will also return 0. So it is just extra time reading, would remove this.

Modules/SubModule_Import/xASL_imp_NII2BIDS_Session.m
HenkMutsaerts: That's true :) But this line is not used, as the next line does exactly the same. I would remove it.

MichaelStritt:
## Current version

sourcedata...

image

### DCM2NII

```
>> pathTest = 'M:\SoftwareDevelopment\MATLAB\m.stritt\Server_xASL\test_ADNI\test_out\011_S_4105';
>> x = ExploreASL(pathTest,[1 0 0 0],0);
ExploreASL will run the import workflow and will run the initialization...
                    ## |                                                                      
                    ## |                                                                      
                    ##/  

Dataset Root        ...reDevelopment\MATLAB\m.stritt\Server_xASL\test_ADNI\test_out\011_S_4105
Import Modules      DCM2NII 
Process Modules     
bPause              False
iWorker             1
nWorkers            1
ExploreASL v1.8.0_BETA initialized ... 
================================== DICOM to NIFTI CONVERSION =================================

Matching files:
    'sub-001\session_1_2013-07-26_15_10_24.0\ASL\S196762'
    'sub-001\session_1_2013-07-26_15_10_24.0\FLAIR\S196758'
    'sub-001\session_1_2013-07-26_15_10_24.0\T1w\S196752'
    'sub-001\session_2_2015-07-29_12_17_39.0\ASL\S267385'
    'sub-001\session_2_2015-07-29_12_17_39.0\FLAIR\S267396'
    'sub-001\session_2_2015-07-29_12_17_39.0\T1w\S267391'

#=6
Running import (i.e. dcm2niiX)
...
```

temp data...

image

image

### NII2BIDS

```
>> x = ExploreASL(pathTest,[0 1 0 0],0);
ExploreASL will run the import workflow and will run the initialization...
                    ## |                                                                      
                    ## |                                                                      
                    ##/  

Dataset Root        ...reDevelopment\MATLAB\m.stritt\Server_xASL\test_ADNI\test_out\011_S_4105
Import Modules      NII2BIDS 
Process Modules     
bPause              False
iWorker             1
nWorkers            1
ExploreASL v1.8.0_BETA initialized ... 
================================== NIFTI to BIDS CONVERSION ==================================
================================== dataset_description.json ==================================
Missing recommended fields:           HEDVersion

Warning: The following user-defined/DICOM fields and DICOM-Phoenix fields differ:
  SoftwareVersions
================================== BIDS JSON Check Results ===================================
Missing recommended ASL fields:       SliceEncodingDirection, AcquisitionVoxelSize, LabelingOrientation
The field ArterialSpinLabelingType is PASL, please check the dependencies below:
The following fields should be empty: LabelingDuration

Warning: The following user-defined/DICOM fields and DICOM-Phoenix fields differ:
  SoftwareVersions
================================== BIDS JSON Check Results ===================================
Missing recommended ASL fields:       SliceEncodingDirection, AcquisitionVoxelSize, LabelingOrientation
The field ArterialSpinLabelingType is PASL, please check the dependencies below:
The following fields should be empty: LabelingDuration
```

rawdata...

image

image

### BIDS2LEGACY

```
>> x = ExploreASL(pathTest,[0 0 0 1],0);
ExploreASL will run the import workflow and will run the initialization...
                    ## |                                                                      
                    ## |                                                                      
                    ##/  

Dataset Root        ...reDevelopment\MATLAB\m.stritt\Server_xASL\test_ADNI\test_out\011_S_4105
Import Modules      BIDS2LEGACY
Process Modules     
bPause              False
iWorker             1
nWorkers            1
ExploreASL v1.8.0_BETA initialized ... 
============================ BIDS to ExploreASL LEGACY CONVERSION ============================
Parsing BIDS scans: 100%
Note that any warnings may have only printed once if they were repeated for multiple scans
Always run your rawdata folder through the BIDS validator first and aim to avoid warnings
Converting from BIDS to Legacy: 011_S_4105 1100%
M0 parsed for 011_S_4105 ASL4D.nii ...
M0 parsed for 011_S_4105 ASL4D.nii ...
Creating default dataPar.json since file was not found...
Overwriting x.dir.dataPar...
```

derivatives...

image

image

Modules/SubModule_Import/xASL_imp_NII2BIDS_Session.m
jan-petr: Don't remember the history and we have to revamp this whole import to really use correctly sessions and not visits. 

Here, obviously, if the visitLabel is missing, it extracts the visit from the session label by removing the first four characters (=="ses-") from the session label.

Again - the sessions/visit run methodology has to be unified for BIDS - I have created new issues for that in 1.9.0. #775

Modules/SubModule_Import/xASL_imp_NII2BIDS_Session.m
jan-petr: Good to make minimal changes only. But agree to remove the first line here as not needed anymore. I did that already in a recent commit

Modules/SubModule_Import/xASL_imp_NII2BIDS_Session.m
jan-petr: Good point - let's make a new issue for that. Or rather lets glue that to #775 (I've just done that).

Modules/SubModule_Import/xASL_imp_NII2BIDS_Session.m
jan-petr: Not really understand the issue here. Yes - let's add to #775.

Modules/SubModule_Import/xASL_imp_NII2BIDS_Session.m
jan-petr: Now I understand. But disagree. My understanding was that BIDS encourages you **not** to create a `ses` subfolder when there is only a single session. So it makes your life easier, but goes against BIDS philosophy. We have this here, I would keep it.

Modules/SubModule_Import/xASL_imp_NII2BIDS_Subject.m
jan-petr: Yes. Agree. This is mixed up. But only the terminology is mixed up. It functions correctly...
We have used here in BIDS function the Legacy terminology. It is indeed incorrect to call it sessions, but it is treating it as runs, so we only need to change the names. I would do that in issue #775 - we already mention it there.

jan-petr (approved):
See the commits I made - they should all be minor.

See my answers. All can be resolved and/or moved to #775. So as far as I am concerned, we can merge.

Modules/SubModule_Import/xASL_imp_NII2BIDS_Session.m
MichaelStritt: Agree. BIDS philosophy makes it easier for users here. Sadly that comes with a little bit more work for the developers.

Anindita94-bit:
Hello ExploreASL experts,

Previously, I have performed ExploreASL on OASIS datasets by keeping manually prepared subjects folder inside the analysis folder using version 1.5.1. I was able to run 5-6 subjects simultaneously. For the ADNI dataset, can you please inform me what major changes I need to make to run the ExploreASL on a group of subjects? Will the convert to source options work if I place several subjects with label files inside the ADNI folders?

Thanks for your support and consideration.

Regards,
Anindita

PR #771, opened 2021-08-16
Fixes #769 Load data bug

### LINKED ISSUE #769: Dataset loading doesn't work
### Description

When a dataset root (or analysis folder in legacy mode) is given to ExploreASL, `x.opts.LoadData` should always be true, e.g. as in `x = ExploreASL('Path2Data', 0, 0);`

When the path doesn't contain data without import or after running import, this should then provide an error or warning in `xASL_init_DataLoading`.

### How to test

[See this usecase](https://docs.google.com/document/d/1-b272yYb6zyS4QoB1BciBs4-kuBS1PkmTi7LJ1MC1u0/edit) in which ExploreASL has already been performed, and only the data needs to be loaded to run additional features.

### Release notes

Fix data loading of processed datasets.

MichaelStritt:
### Linked issue

#769

### DIFF
--- a/ExploreASL_Initialize.m
+++ b/ExploreASL_Initialize.m
@@ -332,12 +332,11 @@ function [x] = ExploreASL_Initialize_GetBooleansImportProcess(x)
         x.opts.bProcessData = 0;
     end
     
-    % On default we do not load the data
-    x.opts.bLoadData = false;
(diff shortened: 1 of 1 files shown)


jan-petr (approved):
OK

PR #773, opened 2021-08-16
Fixes #746 Record ExploreASL version on Import

### LINKED ISSUE #746: Record ExploreASL version on Import
### Description

- Save ExploreASL version into GeneratedBy in Legacy BIDS-derived data. :+1:
- Save ExploreASL version in Acknowledgement into dataset_description for all BIDS functions ("Imported with ExploreASL 1.8.0") :+1: 
- Write a function that takes dataset_description (already loaded from JSON on the input) and looks into the acknowledgment and returns the following information: :+1: 
    - bImportedExploreASL - was the dataset imported with ExploreASL or not (== did we find the string "Imported with ExploreASL XXX") :+1: 
    - bImportedSameVersion - true if the version matches exactly the current version (full match with BETA) :+1: 
    - versionExploreASLBIDS - empty (if above == false), or string with version (excluding the BETA) :+1: 
    - bImportedBETA - true if BETA version was used -> see above, otherwise false :+1: 
    - That's a lot of outputs, but the aim is to use them only locally and not save them to x and therefore we will have a simply and nice function the prepares us for all different testing scenarios later. The order of the outputs should be something like proposed above - because we might not need to extract all (we can check with nargout if all need to be extracted). I anticipate that all these above can be useful in some situations. But you can correct me if needed. :+1: 
- All this above will be used for possible warnings when working with BIDS data of any origin. Now currently. Working with BIDS data in ExploreASL limits to doing BIDS->Legacy conversion (right?!). :+1: 
- So we want to issue the following warnings in BIDS2Legacy - I guess it is pretty obvious which of the outputs above will be used: :+1: 
     - "BIDS created externally, or with an unknown ExploreASL version" :+1: 
     - "BIDS created with different ExploreASL version than BIDS->Legacy conversion" :+1: 
     - "BETA version is being used for data conversion" :+1: 
     - The question is what to do (what to save to derivatives/GeneratedBy) when a different version is used for DCM->BIDS and BIDS->Legacy. Because then the derivatives might not be perfect and the GeneratedBy field might not be completely reliable. Open question to be probably discussed with Henk as well to see if he has some ideas. @HenkMutsaerts might want to comment on this one...
     - The last parameter - the actual version used for the import - this will be only used later on when we will move some of the quantification steps (scalings) to import. The during the quantification, we will want to know what was the version used for import to either do the quantification the old way (with scalings in quantification), or the new way (with scalings in import). :+1: 

### Release notes

Save ExploreASL version in both BIDS and legacy-derivatives imported data.

MichaelStritt:
### Linked issue

#746

### DIFF
--- a/Functions/xASL_bids_CheckDatasetDescription.m
+++ b/Functions/xASL_bids_CheckDatasetDescription.m
@@ -51,7 +51,15 @@ function [bImportedExploreASL, bImportedSameVersion, versionExploreASLBIDS, bImp
     % True if the version matches exactly the current version
     if bImportedExploreASL
         indexVersionString = regexp(datasetDescription.Acknowledgements,'Imported with ExploreASL', 'once')+length('Imported with ExploreASL');
-        fileVersion = datasetDescription.Acknowledgements(indexVersionString+1:end);
+        fileVersionToEnd = datasetDescription.Acknowledgements(indexVersionString+1:end);
+        if ~isempty(regexp(fileVersionToEnd,'_BETA', 'once'))
+            lastElement = regexp(fileVersionToEnd,'_BETA')+length('_BETA')-1;
+            fileVersion = fileVersionToEnd(1:lastElement);
+        else
+            versionDots = regexp(fileVersionToEnd,'[.]');
+            lastElement = versionDots(3)-1;
+            fileVersion = fileVersionToEnd(1:lastElement);
+        end
         if strcmp(fileVersion,versionExploreASL)
             bImportedSameVersion = true;
         end
@@ -59,12 +67,7 @@ function [bImportedExploreASL, bImportedSameVersion, versionExploreASLBIDS, bImp
     
     % Empty (if above == false), or string with version
     if bImportedExploreASL
-        indexUnderscore = regexp(fileVersion,'_');
-        if ~isempty(indexUnderscore)
-            versionExploreASLBIDS = fileVersion(1:indexUnderscore-1);
-        else
-            versionExploreASLBIDS = fileVersion;
-        end
+        versionExploreASLBIDS = fileVersion;
     end
     
     % True if BETA version was used -> see above, otherwise false
--- a/Functions/xASL_bids_CheckDatasetDescription.m
+++ b/Functions/xASL_bids_CheckDatasetDescription.m
@@ -1,11 +1,12 @@
-function [bImportedExploreASL, bImportedSameVersion, versionExploreASLBIDS, bImportedBETA] = xASL_bids_CheckDatasetDescription(datasetDescription,versionExploreASL)
+function [bImportedExploreASL, bImportedSameVersion, versionExploreASLBIDS, bImportedBETA] = xASL_bids_CheckDatasetDescription(datasetDescription, versionExploreASL)
 %xASL_bids_CheckDatasetDescription Check the dataset_description.json field
-% contents with special regard to the import version.
+% contents with special regard to the import version
 %
-% FORMAT: [bImportedExploreASL, bImportedSameVersion, versionExploreASLBIDS, bImportedBETA] = xASL_bids_CheckDatasetDescription(datasetDescription,versionExploreASL)
+% FORMAT: [bImportedExploreASL, bImportedSameVersion, versionExploreASLBIDS, bImportedBETA] = xASL_bids_CheckDatasetDescription(datasetDescription, versionExploreASL)
 % 
 % INPUT:
 %   datasetDescription - Structure which defines the dataset_description fields (STRUCT, REQUIRED)
+%   versionExploreASL  - Version of the current ExploreASL (STRING, REQUIRED)
 %
 % OUTPUT:
 %   bImportedExploreASL   - Was the dataset imported with ExploreASL or not (BOOLEAN)
@@ -17,7 +18,7 @@ function [bImportedExploreASL, bImportedSameVersion, versionExploreASLBIDS, bImp
 % DESCRIPTION:  Check the dataset_description.json field contents with special regard to the import version.
 %
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
-% EXAMPLE:      [bImportedExploreASL, bImportedSameVersion, versionExploreASLBIDS, bImportedBETA] = xASL_bids_CheckDatasetDescription(datasetDescription,x.Version);
+% EXAMPLE:      [bImportedExploreASL, bImportedSameVersion, versionExploreASLBIDS, bImportedBETA] = xASL_bids_CheckDatasetDescription(datasetDescription, x.Version);
 %               
 % __________________________________
 % Copyright 2015-2021 ExploreASL
--- a/Functions/xASL_bids_CheckDatasetDescription.m
+++ b/Functions/xASL_bids_CheckDatasetDescription.m
@@ -1,0 +1,86 @@
+function [bImportedExploreASL, bImportedSameVersion, versionExploreASLBIDS, bImportedBETA] = xASL_bids_CheckDatasetDescription(datasetDescription,versionExploreASL)
+%xASL_bids_CheckDatasetDescription Check the dataset_description.json field
+% contents with special regard to the import version.
+%
+% FORMAT: [bImportedExploreASL, bImportedSameVersion, versionExploreASLBIDS, bImportedBETA] = xASL_bids_CheckDatasetDescription(datasetDescription,versionExploreASL)
+% 
+% INPUT:
+%   datasetDescription - Structure which defines the dataset_description fields (STRUCT, REQUIRED)
+%
+% OUTPUT:
+%   bImportedExploreASL   - Was the dataset imported with ExploreASL or not (BOOLEAN)
+%   bImportedSameVersion  - True if the version matches exactly the current version (full match with BETA) (BOOLEAN)
+%   versionExploreASLBIDS - Empty (if above == false), or string with version (excluding the BETA) (STRING)
+%   bImportedBETA         - True if BETA version was used -> see above, otherwise false (BOOLEAN)
+%                         
+% -----------------------------------------------------------------------------------------------------------------------------------------------------
+% DESCRIPTION:  Check the dataset_description.json field contents with special regard to the import version.
+%
+% -----------------------------------------------------------------------------------------------------------------------------------------------------
+% EXAMPLE:      [bImportedExploreASL, bImportedSameVersion, versionExploreASLBIDS, bImportedBETA] = xASL_bids_CheckDatasetDescription(datasetDescription,x.Version);
+%               
+% __________________________________
+% Copyright 2015-2021 ExploreASL
+
+
+    %% Input check
+    if nargin<1 || isempty(datasetDescription)
+        error('Please provide a struct based on the dataset_description.json BIDS file...');
+    end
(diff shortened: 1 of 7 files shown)


jan-petr (changes requested):
This is good (I made minor changes only - all committed already). But not all I wanted - see the issue description. I wrote some explanations as well so that you get the big picture of why we need that ;) Cause we will need to implement a lot in 1.9.0 which is related to these versions.

MichaelStritt:
Note to myself: also fix the "RandomText" things here 👍

MichaelStritt:
You can run `UnitTest = xASL_ut_function_xASL_bids_CheckDatasetDescription(TestRepository)` to test the `xASL_bids_CheckDatasetDescription` script.

jan-petr:
> Note to myself: also fix the "RandomText" things here 
> 
> Edit: don't really think this is a problem anymore. If I remove those fields from the original studyPar, they don't get added as that.
> 
> ```
> ...
>                 Name: '027_S_5079'
>          BIDSVersion: '1.6.0'
>          DatasetType: 'raw'
>     Acknowledgements: 'Imported with ExploreASL 1.8.0_BETA'
>     HowToAcknowledge: 'Please cite this paper: https://www.ncbi.nlm.nih.gov/pubmed/001012092119281'
> ...
> ```

Yes of course you are right. RandomText has been removed long time ago - it is only in studyPar.jsons and we can keep it there, no need to remove it from there and from reference in all cases.

Functions/xASL_bids_CheckDatasetDescription.m
```diff
@@ -0,0 +1,86 @@
...
+    % Was the dataset imported with ExploreASL or not
+    if isfield(datasetDescription,'Acknowledgements') && ...
+       ~isempty(regexp(datasetDescription.Acknowledgements,'Imported with ExploreASL', 'once'))
+        bImportedExploreASL = true;
+    end
+    
+    % True if the version matches exactly the current version
+    if bImportedExploreASL
+        indexVersionString = regexp(datasetDescription.Acknowledgements,'Imported with ExploreASL', 'once')+length('Imported with ExploreASL');
+        fileVersion = datasetDescription.Acknowledgements(indexVersionString+1:end);
```
jan-petr, line 53: Sorry. A wrong assumption that we have the only acknowledgement is from ExploreASL.

We have to go for "Imported with ExploreASL XX.YY.ZZ(_BETA)." Anything can come after the last "."

Functions/xASL_bids_CreateDatasetDescriptionTemplate.m
```diff
@@ -77,14 +78,19 @@
...
-		if isfield(draft,bidsPar.datasetDescription.Optional{1,iCell})
-			json.(bidsPar.datasetDescription.Optional{1,iCell}) = draft.(bidsPar.datasetDescription.Optional{1,iCell});
-		end
+        % Only copy if defined
+        if isfield(draft,bidsPar.datasetDescription.Optional{1,iCell})
+            json.(bidsPar.datasetDescription.Optional{1,iCell}) = draft.(bidsPar.datasetDescription.Optional{1,iCell});
+        end
+        % Add version which was used for the import to the Acknowledge field (Imported with xASL v1.x.x)
+        if strcmp(bidsPar.datasetDescription.Optional{1,iCell},'Acknowledgements')
+            json.Acknowledgements = ['Imported with ExploreASL ' versionExploreASL];
```
jan-petr, line 87: lets end this statement with a fullstop - see my comment later.

Functions/xASL_bids_CheckDatasetDescription.m
jan-petr: Note that we previously didn't end "our" statement with a fullstop, but we probably should in case there are more to add there.

jan-petr: then be careful to detect this last fullstop, but then remove it from the fileVersion

Functions/xASL_bids_CheckDatasetDescription.m
```diff
@@ -0,0 +1,86 @@
...
+        if strcmp(fileVersion,versionExploreASL)
+            bImportedSameVersion = true;
+        end
+    end
+    
+    % Empty (if above == false), or string with version
+    if bImportedExploreASL
+        indexUnderscore = regexp(fileVersion,'_');
+        if ~isempty(indexUnderscore)
+            versionExploreASLBIDS = fileVersion(1:indexUnderscore-1);
```
jan-petr, line 63: I would keep the "BETA" inside. We want to have the bBETA output as well, but it is probably a bit more transparent and error-prone not to remove the BETA from out. Sorry if this was previously unclear.

Testing/UnitTests/xASL_ut_function_xASL_bids_CheckDatasetDescription.m
```diff
@@ -0,0 +1,100 @@
...
+testTime = tic;
+
+% Run your test here
+datasetDescription.Name = 'TestPatient';
+datasetDescription.BIDSVersion = '1.6.0';
+datasetDescription.DatasetType = 'raw';
+datasetDescription.HowToAcknowledge = 'Please cite this paper: https://www.ncbi.nlm.nih.gov/pubmed/001012092119281';
+
+% Example 1: correct version
+datasetDescription.Acknowledgements = 'Imported with ExploreASL 1.8.0_BETA';
```
jan-petr, line 35: Add fullstop - see above

Testing/UnitTests/xASL_ut_function_xASL_bids_CheckDatasetDescription.m
```diff
@@ -0,0 +1,100 @@
...
+xVersion = '1.8.0_BETA';
+[bImportedExploreASL1, bImportedSameVersion1, versionExploreASLBIDS1, bImportedBETA1] = xASL_bids_CheckDatasetDescription(datasetDescription,xVersion);
+
+% Example 2: different version
+datasetDescription.Acknowledgements = 'Imported with ExploreASL 1.7.0';
```
jan-petr, line 40: Add fullstop, and random text after it...

jan-petr: And probably also before.

jan-petr (changes requested):
All looks very good. Including the test! One thing we need to improve. The "Imported with ExploreASL" is potentially not the only text in the Acknowledgement - we need to be able to deal with extra text - see comments inside the code.

Functions/xASL_bids_CheckDatasetDescription.m
MichaelStritt: Makes sense. I think this should work with 773a8c2 👍

Functions/xASL_bids_CreateDatasetDescriptionTemplate.m
MichaelStritt: 773a8c2 👍

Functions/xASL_bids_CheckDatasetDescription.m
MichaelStritt: No worries, easy fix 773a8c2 👍

Testing/UnitTests/xASL_ut_function_xASL_bids_CheckDatasetDescription.m
MichaelStritt: 773a8c2 👍

Testing/UnitTests/xASL_ut_function_xASL_bids_CheckDatasetDescription.m
MichaelStritt: I added an extra test case 773a8c2 👍

Functions/xASL_bids_CheckDatasetDescription.m
MichaelStritt: If it's a beta version, I search for that and use it to find the last version character. If not I just look for the third dot. That should work pretty robustly, as long as people don't write wrong version formats in there manually.

jan-petr (changes requested):
Thanks for the fixes - it looks good. Two more comments, that weren't somehow submitted before...

First - we don't create the whole acknowledgement - we load it from studyPar and add our ExploreASL part at the end.
Second - now that we export the version, it will mess up the Flavors - i.e. the reference and new import will always differ because everything was done with the older version - shall we ignore the acknowledgement field in the comparison script?!

Functions/xASL_bids_CreateDatasetDescriptionTemplate.m
```diff
@@ -77,14 +78,19 @@
...
-		% Only copy if defined
-		if isfield(draft,bidsPar.datasetDescription.Optional{1,iCell})
-			json.(bidsPar.datasetDescription.Optional{1,iCell}) = draft.(bidsPar.datasetDescription.Optional{1,iCell});
-		end
+        % Only copy if defined
+        if isfield(draft,bidsPar.datasetDescription.Optional{1,iCell})
+            json.(bidsPar.datasetDescription.Optional{1,iCell}) = draft.(bidsPar.datasetDescription.Optional{1,iCell});
+        end
+        % Add version which was used for the import to the Acknowledge field (Imported with xASL v1.x.x)
+        if strcmp(bidsPar.datasetDescription.Optional{1,iCell},'Acknowledgements')
```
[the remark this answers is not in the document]
MichaelStritt: 62c75db & 586f97a should solve this 😃 👍

MichaelStritt:
### Example before changes 9712786 

```
>> [identical,results] = xASL_bids_CompareStructures(pathA,pathB,1,[],1,0);
====================================================================================================
Dataset:   x027_S_5079
           No missing files
====================================================================================================
Dataset:   x027_S_5079_b
Missing:   sourceStructure.json
====================================================================================================
Identical: false
RMSE:      2.13
Min.diff.: 118.00
Max.diff.: 1428.00
File:      ASL4D.nii
           RMSE of NIFTIs above threshold.
Text file: Neues Textdokument.txt
           Different file content.
File:      \rawdata\dataset_description.json
           Different value: Acknowledgements (Imported with ExploreASL 1.8.0_BETA. vs Imported with ExploreASL 1.9.0_BETA.)
```

### Example after changes 9712786 

```
>> [identical,results] = xASL_bids_CompareStructures(pathA,pathB,1,[],1,0);
====================================================================================================
Dataset:   x027_S_5079
           No missing files
====================================================================================================
Dataset:   x027_S_5079_b
Missing:   sourceStructure.json
====================================================================================================
Identical: false
RMSE:      2.13
Min.diff.: 118.00
Max.diff.: 1428.00
File:      ASL4D.nii
           RMSE of NIFTIs above threshold.
Text file: Neues Textdokument.txt
           Different file content.
```

Functions/xASL_bids_CheckDatasetDescription.m
```diff
@@ -0,0 +1,87 @@
+function [bImportedExploreASL, bImportedSameVersion, versionExploreASLBIDS, bImportedBETA] = xASL_bids_CheckDatasetDescription(datasetDescription, versionExploreASL)
```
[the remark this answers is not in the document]
MichaelStritt: I added an example below. Did you mean like this? 9712786

jan-petr (approved):
OK

Issue #442, opened 2021-03-23
Quantification at 7T


jan-petr:
Add all the necessary options and default values to make quantification at 7T possible:

- Set default T1blood for 7T, and also for 1.5T based on field strength (previously it was only for 3T at default). :+1: 
- Calculate T1blood based on Hct using B0 - previously default 3T was used for all. :+1:
- Create warnings when non-3T scanner is used and we don't have the non-3T defaults in the code - can create a lot of warnings, but it would have been wrong to use 3T without a warning. We might only check that some fields can be unused - so no need to warn on them. :+1:
- Tissue T1GM of 1240 has a wrong reference in DataParTemplate.md and is possibly incorrect for 1.5T, though this default is not used in the code - has to be updated for other B0 as well.

#### Release notes
Added default T1-time values for different field- strengths.

HenkMutsaerts:
> * Calculate T1blood based on Hct using B0 - previously default 3T was used for all. 👍
I wonder if this is what we need. We should assume 3T and issue a warning when not 3T.
> * Create warnings when non-3T scanner is used and we don't have the non-3T defaults in the code - can create a lot of warnings, but it would have been wrong to use 3T without a warning. We might only check that some fields can be unused - so no need to warn on them. 👍
Also here, the warning should be when non-3T, not for the non-3T defaults.
> * Tissue T1GM of 1240 has a wrong reference in DataParTemplate.md and is possibly incorrect for 1.5T, though this default is not used in the code - has to be updated for other B0 as well.
With B0 you mean field strength here?

PR #782, opened 2021-08-16
Feature #442 Quantify 7T

jan-petr:
### Linked issue

#442 

### How to test

Does not change anything for 3T. Runs rather fine - TestDataSet should be a good test. 

### Comments

But the values have to be proved and also added for other fields - where currently missing.


DataParTemplate.md
```diff
@@ -77,8 +77,8 @@ This is to allow for valid JSONs. The conversion is carried out internally.
 | ------------------------------------- |:---------------------------------------------:|:------------------:|
 | x.Q.bUseBasilQuantification           | True for using BASIL quantification in addition to ExploreASL's quantification. |  |
 | x.Q.Lambda                            | Brain/blood water coefficient (mL 1H/ mL blood). Example: `0.32` (for GSP phantom). | OPTIONAL, DEFAULT = 0.9 |
-| x.Q.T2art                             | `T2*` of arterial blood at 3T, only used when no M0 image (ms). | OPTIONAL, DEFAULT = 50 |
-| x.Q.BloodT1                           | T1 relaxation time of arterial blood (ms). Defaults (Alsop MRM 2014), 1800 for GSP phantom. | OPTIONAL, DEFAULT = 1650 @ 3T, 1350 @ 1.5 T |
+| x.Q.T2art                             | `T2*` of arterial blood, only used when no M0 image (ms). | OPTIONAL, DEFAULT = 50 @ 3T|
+| x.Q.BloodT1                           | T1 relaxation time of arterial blood (ms). Defaults (Alsop MRM 2014), 1800 for GSP phantom. | OPTIONAL, DEFAULT = 2100 @ 7T, 1650 @ 3T, 1350 @ 1.5 T |
```
HenkMutsaerts, line 81: Should we add references here, now that you can still remember them?

Modules/SubModule_ASL/xASL_wrp_Quantify.m
```diff
@@ -68,11 +68,25 @@ function xASL_wrp_Quantify(x, PWI_Path, OutputPath, M0Path, SliceGradientPath, b
 if ~isfield(x,'Q')
     x.Q = struct;
 end
+
+if ~isfield(x,'MagneticFieldStrength') || isempty(x.MagneticFieldStrength)
```
HenkMutsaerts, line 72: This parameter should probably be in `x.Q` together with the other quantification parameters?

HenkMutsaerts: Also, I wonder if this location alone is sufficient. MagneticFieldStrength is a BIDS parameter, so if it is missing then the user should be warned (also) earlier?

Modules/SubModule_ASL/xASL_wrp_Quantify.m
```diff
@@ -168,9 +182,9 @@ function xASL_wrp_Quantify(x, PWI_Path, OutputPath, M0Path, SliceGradientPath, b
 %% ------------------------------------------------------------------------------------------------
 %% 3.   Hematocrit & blood T1 correction
 if isfield(x,'Hematocrit')
-        x.Q.BloodT1 = xASL_quant_Hct2BloodT1(x.Hematocrit);
+        x.Q.BloodT1 = xASL_quant_Hct2BloodT1(x.Hematocrit, [], x.MagneticFieldStrength);
```
HenkMutsaerts, line 185: While we're at it, `x.Hematocrit` should probably also go to `x.Q.Hematocrit`. However, it is also a "covariate" (`x.S.SetsName`, `x.S.SetsID`, etc.) from `participants.tsv`

Modules/SubModule_ASL/xASL_wrp_Quantify.m
```diff
@@ -225,19 +239,42 @@ function xASL_wrp_Quantify(x, PWI_Path, OutputPath, M0Path, SliceGradientPath, b
...
+			otherwise
+				fprintf('%s\n',['Warning: Unknown T1-blood for ' num2str(x.MagneticFieldStrength) 'T scanners']);
+		end
+	end
     if ~isfield(x.Q,'T2art')
-        x.Q.T2art = 50; % T2* of arterial blood, only used when no M0 image
+		if x.MagneticFieldStrength == 3
+			x.Q.T2art = 50; % T2* of arterial blood, only used when no M0 image
+		else
+			fprintf('%s\n','Warning: Unknown T2-art for non-3T scanners');
```
HenkMutsaerts, line 276: We can consider only issuing this warning when needed. `x.Q.T2art` is only used with an M0 value, or when the T2 differs between deltaM and M0.

Modules/SubModule_ASL/xASL_wrp_Quantify.m
```diff
@@ -225,19 +239,42 @@ function xASL_wrp_Quantify(x, PWI_Path, OutputPath, M0Path, SliceGradientPath, b
...
+		switch(x.MagneticFieldStrength)
+			case 3
+				x.Q.TissueT1 = 1240; % T1 GM tissue @ 3T
+			case 1
+				x.Q.TissueT1 = 1000; 
+			case 4
+				x.Q.TissueT1 = 1350;
+			otherwise
+				x.Q.TissueT1 = 1240;
+				fprintf('%s\n',['Warning: Unknown T1 GM for ' num2str(x.MagneticFieldStrength) 'T scanners']);
```
HenkMutsaerts, line 257: Warning is perhaps not needed when not using TissueT1. `x.Q.TissueT1` is used for two-compartment model or for M0 correction.

HenkMutsaerts (changes requested):

DataParTemplate.md
jan-petr: That's the point - I don't have those - I took all from OSIPI TF4.1, but some other references were a bit unclear.

Modules/SubModule_ASL/xASL_wrp_Quantify.m
jan-petr: This one is never used, so I have remove the warning.

Modules/SubModule_ASL/xASL_wrp_Quantify.m
jan-petr: We can keep warning since Wibeke gave me a reference for almost all B0s - check the new values + references

Modules/SubModule_ASL/xASL_wrp_Quantify.m
jan-petr: We can make that change later in a separate issue if you want. Do you want to create that issue?

Modules/SubModule_ASL/xASL_wrp_Quantify.m
jan-petr: Yes. User will be warned before, but for completeness and modularity, I kept it here as well. Also, I wanted to be minimally invasive - all previous defaults were set to 3T, so I made 3T the default here also to avoid changes in the results

DataParTemplate.md
jan-petr: We have too many values now - I list here only the 3T default and I give the correct values and all references in the Quantify-wrapper where I really used them

jan-petr (commented):
I have fixed all the comments and added more values for different field-strength, plus supplemented all the references.

jan-petr:
@HenkMutsaerts Can you please check the code and comments. And resolve them. Also, can you try to find some references for the T2?

DataParTemplate.md
HenkMutsaerts: OK, I'll make a separate issue on adding references to our scripts.

Modules/SubModule_ASL/xASL_wrp_Quantify.m
HenkMutsaerts: Yes good idea

HenkMutsaerts (approved):

PR #788, opened 2021-08-18
Fixes #787 Move bImportData & bProcessData

### LINKED ISSUE #787: Move bImport and bProcess to ExploreASL_Master
### Description

Henk requested to move the bImport and bProcess checks to the ExploreASL_Master script, so that we don't switch scripts if we only initialize.

### Release notes

Move import & processing checks to the master script.

MichaelStritt:
### Linked issue

#787

### DIFF
--- a/ExploreASL_ImportMaster.m
+++ b/ExploreASL_ImportMaster.m
@@ -19,84 +19,85 @@ function [x] = ExploreASL_ImportMaster(x)
 
     %% Import Workflow
     
-    if x.opts.bImportData
-        % We expect x.opts.DatasetRoot to be the study root directory, but if it is not defined, 
-        % then the user probably used a path to a descriptive JSON file instead
-        if isfield(x, 'dir') && isfield(x.dir, 'DatasetRoot') && isempty(x.dir.DatasetRoot)
-            x.dir.DatasetRoot = xASL_fileparts(x.opts.DatasetRoot);
-        end
+    % We expect x.opts.DatasetRoot to be the study root directory, but if it is not defined,
+    % then the user probably used a path to a descriptive JSON file instead
+    if isfield(x, 'dir') && isfield(x.dir, 'DatasetRoot') && isempty(x.dir.DatasetRoot)
+        x.dir.DatasetRoot = xASL_fileparts(x.opts.DatasetRoot);
+    end
+    
+    % On default we expect the JSON files to be there
+    missingJSON = false;
 
-        % Check if at least one of the three steps should be performed
-        missingJSON = false; % Fallback
-        if sum(x.opts.ImportModules)>0
-            % DICOM TO NII
-            if x.opts.ImportModules(1)==1
-                if ~isempty(x.dir.sourceStructure)
-                    try
-                        xASL_module_Import(x.dir.DatasetRoot, x.dir.sourceStructure, x.dir.studyPar, [1 0 0 0], false, true, false, x);
-                    catch loggingEntry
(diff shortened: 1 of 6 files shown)


MichaelStritt:
### Testing

Everything looks okay so far 👍 

grafik

ExploreASL_ImportMaster.m
```diff
@@ -19,85 +19,86 @@
...
-                        [x] = xASL_qc_AddLoggingInfo(x, loggingEntry);
-                    end
-                else
-                    missingJSON = true;
-                end
+    % DICOM TO NII
+    if x.opts.ImportModules(1)==1
+        if ~isempty(x.dir.sourceStructure)
+            try
+                xASL_module_Import(x.dir.DatasetRoot, x.dir.sourceStructure, x.dir.studyPar, [1 0 0 0], false, true, false, x);
```
jan-petr, line 35: Isn't that a bit weird that you call the import module 1 by 1? I mean, I am fine with this strategy, but shouldn't you then rather have 1,2,3,4 instead of [1 0 0 0], [0 1 0 0] etc, when it is never run as a whole?!

jan-petr: Because as it is now, I don't see any added value of the ImportMaster script...

jan-petr (changes requested):
Very nice for ExploreASL_Master, but somehow weird in ImportMaster - see my comments...

ExploreASL_ImportMaster.m
```diff
@@ -19,85 +19,86 @@
...
+    % Print warning about missing JSON files
+    if missingJSON
+        warning('Import workflow is turned on, but at least one required JSON file is missing...');
+    end
+    
+    % Reset the import parameters
+    x.opts.bImportData = 0;
+    x.opts.ImportModules = [0 0 0 0];
+    
+    % We are running the import here, but if BIDS to Legacy did not run, we can not load the data
```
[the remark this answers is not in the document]
MichaelStritt: Oh sorry, you're definitely right. I kind of overlooked that. We can move it to the start, then we should be safe 42ea2be 👍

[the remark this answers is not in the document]
MichaelStritt: We can also use the try catch to stop the pipeline from printing multiple errors with this 50ffa06 👍

ExploreASL_ImportMaster.m
MichaelStritt: Hm... I think it got like because of how we used to do things before. We had the testing import scripts where we did it like this and I kind of copied that behavior from there and applied it to the former ExploreASL_Import which is xASL_module_Import now.

MichaelStritt: What you're saying makes sense. The added value of the ImportMaster script is only the try-catch basically.

MichaelStritt: Using 1,2,3,4 doesn't sound that clever to me. We agreed on using binary notation and we already changed the ExploreASL parameters for that. We shouldn't change those things back.

MichaelStritt: @jan-petr: what you were asking for required some bigger changes. I had to change some things for the dataPar creation in derivatives and also some of the error handling. looks pretty good now I think. The ImportMaster is more like a small wrapper now, but still helps with the try-catch and so on. Running the ExploreASL_Master unit test ATM. This is the commit 
7fdaef3  👍

MichaelStritt: @jan-petr good news 👍 

grafik

jan-petr (approved):
Looks perfect now!

Issue #790, opened 2021-08-18
Import file ordering by SeriesNumber


jan-petr:
### Description

In the Mood study, ASL files are imported by single repetition and they are ordered by SeriesNumber instead of InstanceNumber.

### How to test
It should work on Siemens_PCASL_3DGRASE_VB17A_TopUp_2 and on none other. There it seems to work and now testing if it doesn't brake the other flavors.

### Release notes

Import ASL ordering by SeriesNumber

PR #792, opened 2021-08-18
#790 Merge ASL files in import by SeriesNumber

jan-petr:
### Linked issue

#790

### DIFF
--- a/Functions/xASL_bids_CompareStructures.m
+++ b/Functions/xASL_bids_CompareStructures.m
@@ -476,6 +476,9 @@ end
 %% Compare JSON files
 function [differences,identical,dn] = compareJSON(differences,identical,bPrintReport,allFiles,iFile,dn,currentFileA,currentFileB)
 
+    % Ignore Acknowledgements (setting to ignore version differences (ExploreASL v1.x.x) there)
+    ignoreAcknowledgements = true;
+
     jsonErrorReport='';
 
     % Compare JSON files on field basis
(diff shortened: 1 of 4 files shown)


MichaelStritt:
@jan-petr: will run the flavors tonight and I'll post the results tomorrow.

jan-petr:
@MichaelStritt see my first results
- Plenty of datasets differ in acknowledgement. Despite the ignoring. My guess is that if both reference and source have the field, then this is ignored, but if only one of the compared sets have it, then it complained about the difference. So it is not about the content but about a difference in existence.
- Siemens_PCASL_2DEPI_VB17A_BSupN_1 has different asl.nii. By coincidence, this is a dataset from the same center as the one I was adding for the newest tests. My guess is that previously, we were ordering this dataset wrong and this now got fixed :) You can see that in the reference - the motion across repetitions is sketchy. And in the new version, this is much more smooth. Do you agree with this? If yes, then we have to replace the reference asl.nii in the flavors with this newly converted asl.nii
- The only other issue is with Hadamard - but Beatriz knows about this and is fixing it in some other issue, so we can ignore that.

@MichaelStritt Can you please upload the new asl.nii to flavors and fix the acknowledgement issue in this issue?

MichaelStritt:
@jan-petr: here is the print out of the bids comparison: [bids_comparison.txt](https://github.com/ExploreASL/ExploreASL/files/7012580/bids_comparison.txt)

> Plenty of datasets differ in acknowledgement. Despite the ignoring. My guess is that if both reference and source have the field, then this is ignored, but if only one of the compared sets have it, then it complained about the difference. So it is not about the content but about a difference in existence.

I saw the acknowledgement thing. will fix it immediately, should be easy 👍 

> Siemens_PCASL_2DEPI_VB17A_BSupN_1 has different asl.nii. By coincidence, this is a dataset from the same center as the one I was adding for the newest tests. My guess is that previously, we were ordering this dataset wrong and this now got fixed :) You can see that in the reference - the motion across repetitions is sketchy. And in the new version, this is much more smooth. Do you agree with this? If yes, then we have to replace the reference asl.nii in the flavors with this newly converted asl.nii

I saw that too. That's really weird.

```
Dataset: Siemens_PCASL_2DEPI_VB17A_BSupN_1
File:      \sub-Sub1\perf\sub-Sub1_asl.nii.gz
           RMSE of NIFTIs above threshold.
Warning: RMSE of NIFTIs above threshold: \sub-Sub1\perf\sub-Sub1_asl.nii.gz    
> In xASL_bids_CompareStructures>printDifferencesAsWarnings (line 217)
  In xASL_bids_CompareStructures (line 206)
  In xASL_test_Flavors (line 137) 
```

I checked out both rawdata NIfTIs in ImageJ and it seems like the new version has less "motion". It really seems like there's something wrong with the ordering in the old one. I personally would say the new one is correct.

> The only other issue is with Hadamard - but Beatriz knows about this and is fixing it in some other issue, so we can ignore 

Agree, we shouldn't fix this here and rather wait for @BeatrizPadrela to fix this 😛

MichaelStritt (approved):
Should be fine after we fixed the Acknowledgement issue and updated the Siemens dataset 👍

PR #804, opened 2021-08-30
#801 xASL_*_Dicom2(Parms|JSON): Avoid reporting defaults missing

### LINKED ISSUE #801: Remove warnings about default values in import
### Description

Not all defaults need to be reported if they are missing, as they are not needed for all sequences. We currently have warnings and checks in place in the conversion to BIDS, so can remove them in the dicom2nii import stage.

No need to report missing fields and replacing defaults in dicom2nii. But instead, we put in an empty or NaN to these missing fields. And then later in NII2BIDS, when a value is needed and is missing, we solve it properly.

### Requirements

Optional: add bug requirements here

### How to test

### Release notes

HenkMutsaerts:
### Linked issue

Closes #801

### How to test

### Comments


MichaelStritt (approved):
Seems fine to me. You only removed two fprintf statements as far as I can see. I did some minor cosmetics and got rid of one if statement (commented it out) 👍

jan-petr (approved):
OK.

Issue #810, opened 2021-09-01
BIDS2Legacy incorrect conversion of multi-session data


jan-petr:
### Description

If there are multiple sessions, but not all subjects have multiple sessions, then the conversion does not go correctly.

### Requirements

The error is in BIDS.layout - there, it extracts the total number of sessions/session names from the last subject and not from all subjects. Therefore, if the last subject has only one session, it converts all others as if they would have had one session only.

### Release notes

Correctly convert multi-session BIDS2Legacy even if not all subjects have multiple sessions.

PR #811, opened 2021-09-01
#810 bids.layout: Fix session name extraction

jan-petr:
### Linked issue

#810

### DIFF
--- a/Modules/SubModule_Import/xASL_imp_Initialize.m
+++ b/Modules/SubModule_Import/xASL_imp_Initialize.m
@@ -33,23 +33,23 @@ function imPar = xASL_imp_Initialize(studyPath, imParPath)
     imPar = spm_jsonread(imParPath);
 
     %% 2. Specify paths
-    if ~isfield(imPar,'studyID') || isempty(imPar.studyID)
+    if ~isfield(imPar, 'studyID') || isempty(imPar.studyID)
         imPar.studyID = [fname fext];
     end
-    if ~isfield(imPar,'TempRoot') || isempty(imPar.TempRoot) 
+    if ~isfield(imPar, 'TempRoot') || isempty(imPar.TempRoot) 
         imPar.TempRoot = fpath;
     end
-    if ~isfield(imPar,'RawRoot') || isempty(imPar.RawRoot)
+    if ~isfield(imPar, 'RawRoot') || isempty(imPar.RawRoot)
         imPar.RawRoot = fpath;
     end
-    if ~isfield(imPar,'BidsRoot') || isempty(imPar.BidsRoot)
+    if ~isfield(imPar, 'BidsRoot') || isempty(imPar.BidsRoot)
         imPar.BidsRoot = fpath;
     end
 
     %% 3. Finalize the directories
-    imPar.RawRoot = fullfile(imPar.RawRoot,imPar.studyID, 'sourcedata'); % default name
-    imPar.TempRoot = fullfile(imPar.TempRoot,imPar.studyID,'temp');
-    imPar.BidsRoot = fullfile(imPar.BidsRoot,imPar.studyID,'rawdata');
+    imPar.RawRoot = fullfile(imPar.RawRoot, imPar.studyID, 'sourcedata'); % default name
+    imPar.TempRoot = fullfile(imPar.TempRoot, imPar.studyID, 'temp');
+    imPar.BidsRoot = fullfile(imPar.BidsRoot, imPar.studyID, 'rawdata');
 
     %% 4. Specify the tokens
     
@@ -60,58 +60,62 @@ function imPar = xASL_imp_Initialize(studyPath, imParPath)
     xASL_imp_InitializeCheckTokens(imPar);
 
     %% 5. Specify the additional details of the conversion
-    if ~isfield(imPar,'bVerbose') || isempty(imPar.bVerbose)
+    if ~isfield(imPar, 'bVerbose') || isempty(imPar.bVerbose)
         imPar.bVerbose = true;
     end
-    if ~isfield(imPar,'bOverwrite') || isempty(imPar.bOverwrite)
+    if ~isfield(imPar, 'bOverwrite') || isempty(imPar.bOverwrite)
         imPar.bOverwrite  = false; % NB, the summary file will be recreated anyway and dicom conversion in temp is always done, even if dest. exists
     end
-    if ~isfield(imPar,'visitNames') || isempty(imPar.visitNames)
+    if ~isfield(imPar ,'visitNames') || isempty(imPar.visitNames)
         imPar.visitNames = {};
     end
     if ~isfield(imPar,'sessionNames') || isempty(imPar.sessionNames)
         imPar.sessionNames = {};
     end
+    if ~isfield(imPar, 'sessionNames') || isempty(imPar.sessionNames)
+        imPar.sessionNames = {};
+    end
+
 
 end
 
 %% Specify the imPar struct
 function imPar = xASL_imp_InitializeBuildImPar(imPar)
 
-    if ~isfield(imPar,'folderHierarchy')
+    if ~isfield(imPar, 'folderHierarchy')
         imPar.folderHierarchy = {}; % must define this per study; use a cell array of regular expressions. One cell per directory level.
     end
-    if ~isfield(imPar,'tokenOrdering')
+    if ~isfield(imPar, 'tokenOrdering')
         imPar.tokenOrdering = []; % must match imPar.folderHierarchy: 1==subject, 2=visit, 3==session, 4==scan (if visit or session are omitted, they will be skipped)
     else
         imPar.tokenOrdering = imPar.tokenOrdering(:)';
     end
-    if ~isfield(imPar,'tokenScanAliases')
+    if ~isfield(imPar, 'tokenScanAliases')
         imPar.tokenScanAliases = [];
     else
         if (size(imPar.tokenScanAliases,2) > 2) || (size(imPar.tokenScanAliases,2) == 1)
             tokenScanAliasesOld = imPar.tokenScanAliases;
-			imPar = rmfield(imPar,'tokenScanAliases');
+			imPar = rmfield(imPar, 'tokenScanAliases');
             imPar.tokenScanAliases(:,1) = tokenScanAliasesOld(1:2:end);
             imPar.tokenScanAliases(:,2) = tokenScanAliasesOld(2:2:end);
         end
     end
-    if ~isfield(imPar,'tokenVisitAliases')
+    if ~isfield(imPar, 'tokenVisitAliases')
         imPar.tokenVisitAliases = [];
     else
         if (size(imPar.tokenVisitAliases,2) > 2) || (size(imPar.tokenVisitAliases,2) == 1)
             tokenVisitAliasesOld = imPar.tokenVisitAliases;
-			imPar = rmfield(imPar,'tokenVisitAliases');
+			imPar = rmfield(imPar, 'tokenVisitAliases');
             imPar.tokenVisitAliases(:,1) = tokenVisitAliasesOld(1:2:end);
             imPar.tokenVisitAliases(:,2) = tokenVisitAliasesOld(2:2:end);
         end
     end
-    if ~isfield(imPar,'tokenSessionAliases')
+    if ~isfield(imPar, 'tokenSessionAliases')
         imPar.tokenSessionAliases = [];
     else
         if (size(imPar.tokenSessionAliases,2) > 2) || (size(imPar.tokenSessionAliases,2) == 1)
             tokenSessionAliasesOld = imPar.tokenSessionAliases;
-			imPar = rmfield(imPar,'tokenSessionAliases');
+			imPar = rmfield(imPar, 'tokenSessionAliases');
             imPar.tokenSessionAliases(:,1) = tokenSessionAliasesOld(1:2:end);
             imPar.tokenSessionAliases(:,2) = tokenSessionAliasesOld(2:2:end);
         end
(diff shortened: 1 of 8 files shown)


MichaelStritt (approved):
Hm... I partially get what you're doing there 😄 
I can quickly test it on a dataset with varying modalities for the sessions like you described.
The question is, if we should note this in our changes.md or if we should create the same PR in the bids-matlab repo instead.

MichaelStritt:
Damn... I think there's still something wrong.

### My multi-session test

- session 1: t1w & asl
- session 2: t1w, flair, & asl
- session 3: only asl

- Log for release branch: [testLogRelease.txt](https://github.com/ExploreASL/ExploreASL/files/7097759/testLogRelease.txt)
- Log for bugfix branch: [testLogBugFix.txt](https://github.com/ExploreASL/ExploreASL/files/7097758/testLogBugFix.txt)

### Result in bugfix branch

The folders sub-001_1, sub-001_2, & sub-001_3 are there. In sub-001_1 there is only the asl, in sub-001_2 there is the asl and the flair, & in sub-001_3 there is only the asl again... :thinking: 

Within DCM2NII we seem to find the correct sessions & modalities:

```
Matching files:
    'sub-001/session_1_abc/ASL/S811203'
    'sub-001/session_1_abc/T1w/S811205'
    'sub-001/session_2_abc/ASL/S811203'
    'sub-001/session_2_abc/FLAIR/S811207'
    'sub-001/session_2_abc/T1w/S811205'
    'sub-001/session_3_abc/ASL/S811203'
```

MichaelStritt:
@jan-petr

### Test set-up

image

image

image

### Test DCM2NII

Here everything seems to work fine. There are a few warnings about missing scans, but the results are perfectly fine.

image

image

image

image

image

### Test NII2BIDS

Here it already seems to go wrong. The T1 images are missing.

image

image

image

image

image

MichaelStritt (approved):
Added a bit more user feedback and some cosmetics. For me everything seems to work fine. Can you check if you're okay with 26f08c34c95eaee3b11d948b3a5cfdf2064063bd ?

Modules/SubModule_Import/xASL_imp_Initialize.m
```diff
@@ -52,6 +52,38 @@
...
+    xASL_imp_InitializeCheckTokens(imPar);
+
+    %% 5. Specify the additional details of the conversion
+    if ~isfield(imPar,'bVerbose') || isempty(imPar.bVerbose)
+        imPar.bVerbose = true;
+    end
+    if ~isfield(imPar,'bOverwrite') || isempty(imPar.bOverwrite)
+        imPar.bOverwrite  = false; % NB, the summary file will be recreated anyway and dicom conversion in temp is always done, even if dest. exists
+    end
+    if ~isfield(imPar,'visitNames') || isempty(imPar.visitNames)
```
HenkMutsaerts, line 69: Why do we define these here, these are not needed for `imPar` (dcm2nii), these are needed for BIDS2legacy, right?

Modules/SubModule_Import/xASL_imp_Initialize.m
```diff
@@ -94,26 +126,30 @@
...
+%% Warn the user if token aliases are invalid according to our pipeline/BIDS
+function xASL_imp_InitializeCheckTokens(imPar)
+
+    % Check tokenScanAliases
+    if isfield(imPar,'tokenScanAliases')
+        if size(imPar.tokenScanAliases,2)==2
+            for iToken = 1:size(imPar.tokenScanAliases,1)
+                currentToken = imPar.tokenScanAliases{iToken,2};
+                % Warn if the token contains ASL but is not ASL4D
+                if ~isempty(regexp(currentToken,'ASL', 'once')) && ~strcmp(currentToken,'ASL4D')
```
HenkMutsaerts, line 141: `regexp` should be `regexpi` here
`'once'` can be omitted
(same for T1w below)

Modules/SubModule_Import/xASL_imp_Initialize.m
```diff
@@ -94,26 +126,30 @@
...
+                    warning('Please use ASL4D for your ASL tokenScanAlias instead...');
```
HenkMutsaerts, line 142: Now we only issue a warning, shouldn't we also automatically convert to ASL4D and T1w?

Modules/SubModule_Import/xASL_imp_NII2BIDS_RunAnat.m
```diff
@@ -56,6 +56,12 @@ function xASL_imp_NII2BIDS_RunAnat(imPar, bidsPar, studyPar, subjectSessionLabel
             % Create the anatomical directory
             xASL_adm_CreateDir(anatOutDirLabel);
 
+            % Current scan name
+            [~, scanName] = xASL_fileparts([anatOutLabel,'_' iAnatType{1}]);
```
HenkMutsaerts, line 60: The `,` here is confusing

Modules/SubModule_Import/xASL_imp_NII2BIDS_RunAnat.m
```diff
@@ -56,6 +56,12 @@ function xASL_imp_NII2BIDS_RunAnat(imPar, bidsPar, studyPar, subjectSessionLabel
...
+            
+            % Print anat name
+            fprintf('Scan %s ...\n', scanName);
+            
             % Move the NiFTI file
             anatNiiPath = [anatOutLabel,'_' iAnatType{1} '.nii.gz'];
```
HenkMutsaerts, line 66: Also here the `,` is confusing. Are you sure both shouldn't have `fullfile`?

Modules/SubModule_Import/xASL_imp_NII2BIDS_RunAnat.m
```diff
@@ -56,6 +56,12 @@ function xASL_imp_NII2BIDS_RunAnat(imPar, bidsPar, studyPar, subjectSessionLabel
...
             xASL_Move([anatPath '.nii'],anatNiiPath,1);
```
HenkMutsaerts, line 67: This is incorrect behavior, ExploreASL functions such as `xASL_Move` should be ignoring .nii vs .nii.gz. Instead, you don't want to move but zip a file here right?

Modules/SubModule_Import/xASL_imp_NII2BIDS_Session.m
```diff
@@ -30,6 +30,10 @@
     %% Unpack bidsLabel struct
     subjectLabel = bidsLabel.subject;
```
HenkMutsaerts, line 31: Why do we "unpack"? This only means that we have twice the number of temporary variables right?
Can't we simply use `bidsLabel.subject` throughout? Same for the visit one.

Modules/SubModule_Import/xASL_imp_NII2BIDS_Session.m
```diff
@@ -30,6 +30,10 @@
...
     visitLabel = bidsLabel.visit;
+    
+    %% Print current session which is being converted
```
HenkMutsaerts, line 34: `%% Print current session which is being converted`

should be

`%% Print the session that is being converted`

(learn English grammar ;)

HenkMutsaerts (changes requested):

Modules/SubModule_Import/xASL_imp_Initialize.m
MichaelStritt: That has been there forever and was just moved here...

Modules/SubModule_Import/xASL_imp_Initialize.m
MichaelStritt: Once is recommended by matlab, agree on the regexpi

Modules/SubModule_Import/xASL_imp_Initialize.m
jan-petr: yes, but what if the guy intended to really use 'ASL' and 'T1'?! Warning is probably enough.

Modules/SubModule_Import/xASL_imp_Initialize.m
jan-petr: Agree - fixed.

Modules/SubModule_Import/xASL_imp_Initialize.m
MichaelStritt: We can set them to those values at this point on default if you want.

Modules/SubModule_Import/xASL_imp_NII2BIDS_RunAnat.m
MichaelStritt: Why?

Modules/SubModule_Import/xASL_imp_NII2BIDS_RunAnat.m
MichaelStritt: We can optimize this, but that's really just cosmetics IMHO

Modules/SubModule_Import/xASL_imp_Initialize.m
jan-petr: Yes - we can make an issue for reconstructing this function if needed...

Modules/SubModule_Import/xASL_imp_NII2BIDS_RunAnat.m
MichaelStritt: Nope, we move from temp to rawdata

Modules/SubModule_Import/xASL_imp_Initialize.m
jan-petr: Unless Henk has a strong opinion on it (because I don't), I would go the way of warning, but giving confidence to the user :)

Modules/SubModule_Import/xASL_imp_NII2BIDS_Session.m
MichaelStritt: Yeah, this is there because we were working on reducing the number of io arguments while keeping the code stable. This was a nice temporary solution. Of course we can remove the "unpacking" and just use bidsLabel throughout the code. We just haven't had time to do this before and there was a time when the code was super fragile 😄

Modules/SubModule_Import/xASL_imp_NII2BIDS_RunAnat.m
jan-petr: yes - here I agree that you should either have `','` or `' '` everywhere. No `fullfile` that's just filename being constructed, not path

Modules/SubModule_Import/xASL_imp_NII2BIDS_Session.m
jan-petr: Fixed.

Modules/SubModule_Import/xASL_imp_NII2BIDS_RunAnat.m
jan-petr: on the right hand side - some delimiters are `' '`, some are `','`... Fixing...

Modules/SubModule_Import/xASL_imp_NII2BIDS_RunAnat.m
jan-petr: Check line 37 of that function - the most of the path is already constructed using `fullfile`, so this is just a filename postfix - `fullfile` cannot be used here anymore.

jan-petr: `,` fixed throughout the code

Modules/SubModule_Import/xASL_imp_Initialize.m
jan-petr: Fixed...

HenkMutsaerts:
Looks good. My only question is why we set max sessions /visits to 0 by default, shouldn't this default to 1?

jan-petr:
> Looks good. My only question is why we set max sessions /visits to 0 by default, shouldn't this default to 1?

nMaxVisits and nMaxSessions - not something we changed in this commit. And I can't find this parameter to be used/described anywhere else in the code. AFAIK, we can remove it completely - perhaps make an issue for that and do that in the next release (to be on the safe side).

> Also, i see that in your test case, you miss having a session with Anat but without ASL. I would test this as well.

Yes. Not totally related, but doesn't hurt testing as well. @MichaelStritt - can you run your ADNI test again with a non-ASL session as well?

HenkMutsaerts:
> > Looks good. My only question is why we set max sessions /visits to 0 by default, shouldn't this default to 1?
> 
> nMaxVisits and nMaxSessions - not something we changed in this commit. And I can't find this parameter to be used/described anywhere else in the code. AFAIK, we can remove it completely - perhaps make an issue for that and do that in the next release (to be on the safe side).
> 
> > Also, i see that in your test case, you miss having a session with Anat but without ASL. I would test this as well.
> 
> Yes. Not totally related, but doesn't hurt testing as well. @MichaelStritt - can you run your ADNI test again with a non-ASL session as well?

`nMaxVisits` and `nMaxSessions` were added by Michael I believe in a commit in this PR.
"not totally related" -> I just noticed that Michael left out this test-case:)

jan-petr:
> `nMaxVisits` and `nMaxSessions` were added by Michael I believe in a commit in this PR.

I think he just moved stuff around to a subfunction - he did not touch any functionality - just added checks and warnings. Can you confirm Michael - and then create an issue to 1.9.0 to remove the variables...

> "not totally related" -> I just noticed that Michael left out this test-case:)

Yes - I agree, useful to check it to avoid future issues.
 
> Very happy otherwise. Most important thing is to keep stuff modular, the code should be as independent as pragmatically possible.

Yes - agree!!

jan-petr:
@HenkMutsaerts Can you approve? Because we don't plan changes, just Michael will test stuff and make a new issue...

MichaelStritt:
Give me 10 more minutes, I've been improving the unit test this morning 😄

jan-petr:
Agree. Henk, please click the button. And we'll run the full-tests after merging everything (no risk with merging to release branch without all tests).

Issue #858, opened 2021-09-30
xASL_wrp_Realign has outdated HadamardType variable


BeatrizPadrela:
### Description

![image](https://user-images.githubusercontent.com/73699072/135414094-589ac016-1fc8-4e3f-8575-9ae4cd3a2020.png)

### Requirements

Instead of
`x.modules.asl.HadamardType~=0`

I'm using 
`isfield(x.modules.asl,'TimeEncodedMatrixType')`

### How to test

Optional: insert description about how to test the code changes here

### Release notes

Solved in issue #850

BeatrizPadrela:
@MichaelStritt @jan-petr this is a very small thing. Please review so we can merge this

BeatrizPadrela:
One of the issues present in #850. Corrected there already

Issue #850, opened 2021-09-24
Hadamard or TimeEncoded field check required


MichaelStritt:
### Description

While processing ADNI case 003_S_6264 using develop, I got this error:

```
====================================================================================
Running xASL_module_ASL ...   
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
=== Subject: sub-003S6264_1, Session: ASL_1, Module: xASL_module_ASL_<SESSION>,  24-Sep-2021 10:53:25 ===

Cleaning data to rerun subject sub-003S6264_1
nRemoving lock folders:100%
Restoring backupped _ORI (original) files:   
Deleting native space CAT12 temporary folders:   
Deleting native space files for module 2:100%
Deleting all standard space files for module 2

Cleaning up x-struct:100%
Cleaning up QC_Collection json:100%
3D_GRASE sequence detected
No TopUp scans available, skipping...
Automatic alignment with center of mass: skipping, was too close
x.modules.asl.SpikeRemovalThreshold was not defined yet, default setting = 0.01 used
SPM motion estimation
ERROR: ASL module terminated for subject 1: sub-003S6264_1

ans =

    'Reference to non-existent field 'HadamardType'.
     
     Error in xASL_wrp_RealignASL (line 125)
     if nFrames>2 && bSubtraction && (numel(unique(x.Q.Initial_PLD))>1 
     || numel(unique(x.EchoTime))>1 || x.modules.asl.HadamardType~=0) %multiPLD or multiTE or Hadamard
     
     Error in xASL_module_ASL (line 306)
                 xASL_wrp_RealignASL(x);
```

I've been rerunning it using main, to check if the last release is fine. There the bug does not occur.

### Requirements

Optional: add bug requirements here

### How to test

Optional: insert description about how to test the code changes here

### Release notes

Required: summarize the changes for the release notes here

PR #851, opened 2021-09-24
Fixes #850 xASL_wrp_RealingASL: field checks

MichaelStritt:
### Linked issue

Check out #850

### How to test

Develop without this fix should basically crash for almost every non-Hadamard dataset. Just do some basic testing of the ASL module using the test dataset. Alternatively you can always run the 10 test datasets, but that's probably overkill for such a small change.

### Comments

@BeatrizPadrela: Okay, same here as with https://github.com/ExploreASL/ExploreASL/pull/849, please check if everything works correctly. If yes, then approve & move it to @jan-petr's bucket.

PR #859, opened 2021-09-30
Feature #858 bug xASL_wrp_RealignASL

BeatrizPadrela:
### Linked issue

See #858
Very small bug, @MichaelStritt @jan-petr please review so I can merge this

### How to test

### Comments

Solved in issue #850

### DIFF
--- a/Modules/SubModule_ASL/xASL_wrp_RealignASL.m
+++ b/Modules/SubModule_ASL/xASL_wrp_RealignASL.m
@@ -122,7 +122,7 @@ xASL_delete(rpfile);
 %         save('rp_ASL4D.txt','MotionAllTEs','-ascii')
 %     end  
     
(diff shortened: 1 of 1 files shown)


MichaelStritt (commented):
We have to be careful here, since I saw the same bug and tried to fix it in #850 (PR #851). Do we need both? We could easily adapt #850 so that we don't have to resolve conflicts later on. Can you check this @jan-petr?

BeatrizPadrela:
> We have to be careful here, since I saw the same bug and tried to fix it in #850 (PR #851). Do we need both? We could easily adapt #850 so that we don't have to resolve conflicts later on. Can you check this @jan-petr?

You're right. I just replaced HadamardType by TimeEncodedMatrixType in #850

jan-petr (changes requested):
Please do this change in #850 and remove PR+branch and label this issue as duplicate.

PR #871, opened 2021-10-04
Fixes #759 Adapting xASL to multi-PLD

### LINKED ISSUE #759: Adapting ExploreASL pipeline to multi-PLD data
### Description
**Import module:**
Merging of multiple single-PLD files is done automatically according to file-placement in directories. 
Should provide option to disable merge if preferred, see issue #917.

Output: ASL4D (consisting of according to BIDS structure: PLD1(C-L,C-L,C-L)PLD2(C-L,C-L,C-L))

**ASL module:**
At the beginning of the module (after LoadParams in line 206): recognition of multi-TE, multi-PLD, time encoded (Hadamard)

- TopUp: no changes 👍
   -> check if it can be applied to more dimensions ( see issue #918)

- Motion correction:  👍 (ASL module - step 2) @BeatrizPadrela 
Default is `on`.
-> zigzag=false 
-> xASL_wrp_Realign_ASL: 
  -- bENABLE instead of bExecuteENABLE 👍
  -- bENABLE = 0 at the beginning of the function if multiPLD. Default bENABLE=1 👍
  --bZigZag defined at the beginning of the Realign function 👍

Check what others do in multi-PLD motion correction (see issue #919)

- Registration  (ASL module - step 3) @BeatrizPadrela 
-> xASL_wrp_RegisterASL:
  -- Step H: There's this Pairwise subtraction that creates this PWI_Clipped image. Jan checked that this PWI_Clipped image is only used in this function and in the Realign [this Clipped_ORI can be deleted in the future because is not used anywhere else]. So, since in this function we want to obtain a single volume:
  + if multi-PLD, we can run the Pairwise subtraction (since we have control and label images for each PLD) and take an average of all the PLDs (->later on we might want to do some extra-check because instead of average with all of the PLDs, we probably don't want to include the early PLDs but only the later ones that look like CBF) -> #914 
-We don't need to change this step, just test if it works
  
And after this, the PWI_Clipped is going to work irrespectively how the Clipped was created (bRegistrationCBF loop in the step H)

- Resampling  (ASL module - step 4) 
->xASL_wrp_ResampleASL:
  We currently use PWI (a 4 dimensional matrix), however we can also use PWI4D (containing all repeats). For now, insert warning that we use only PWI
  -- check if BASIL uses multiple repeats to optimise quantification (PVC) and implement that in future issue #920

  -- Step8: pair-wise subtraction: (discuss with @HenkMutsaerts)
  + for multiPLD: stays the same - subtracting controls and labels (Paired subtraction part). For the PWI, we need to calculate the mean across the repetitions and keep the PLDs separate (Average PWI part).
BASIL-> 4D PWI file, containing the repeats and PLD's. Basil inputs a txt file with the order of the volumes (PLD, Repetitions - check BIDS structure)

- Process M0 (ASL module - step 6) @MDijsselhof 
-> Acquired M0
-> Non-background suppressed mean control of the latest (or only) PLD 
     - include `fprintf` that the last average control image is used.
     - New issue to improve this (issue #921)
-> Pseudo M0: Mean control of the latest PLD.

- Quantification (ASL module - step 8) @MDijsselhof 
->xASL_wrp_Quantify -> Perform Quantification (step 8 of xASL_wrp_Quantify) -> xASL_quant_SinglePLD, line 131 -> xASL_quant_Basil
-> Check which outputs we want from Basil
-> Create config file

### How to test
With the multiPLD dataset we have

### Requirements

MDijsselhof:
### Linked issue

closes #759 

### How to test

### Comments

### DIFF
--- a/Functions/xASL_qc_CollectQC_ASL.m
+++ b/Functions/xASL_qc_CollectQC_ASL.m
@@ -98,13 +98,7 @@ function [x] = xASL_qc_CollectQC_ASL(x, iSubject)
     % Read CBF NIfTI
     pGM = xASL_io_Nifti2Im(Path_pGM);
     pWM = xASL_io_Nifti2Im(Path_pWM);
-    if isfield(x.Q,'bUseBasilQuantification') && x.Q.bUseBasilQuantification==1
-        % use BASIL CBF image paths
-        imCBF = xASL_io_Nifti2Im(x.P.Path_CBF_Basil);
-    else
-        % use xASL CBF image paths
-        imCBF = xASL_io_Nifti2Im(x.P.Path_CBF);
-    end
+    imCBF = xASL_io_Nifti2Im(x.P.Path_CBF);
         
     if xASL_stat_SumNan(pGM(:))==0
         warning(['Empty image, invalid ' Path_pGM]);
--- a/Functions/xASL_qc_CollectQC_ASL.m
+++ b/Functions/xASL_qc_CollectQC_ASL.m
@@ -37,12 +37,11 @@ function [x] = xASL_qc_CollectQC_ASL(x, iSubject)
 %
 % EXAMPLE: x = xASL_qc_CollectQC_ASL(x, 10);
 % __________________________________
-% Copyright (C) 2015-2019 ExploreASL
+% Copyright (c) 2015-2021 ExploreASL
 
 
     %% Admin
     ASL = struct;
-
     SubjectID = x.SUBJECTS{iSubject};
     SessionID = x.SESSIONS{1};
     ASL_ID = [SubjectID '_' SessionID];
@@ -84,8 +83,10 @@ function [x] = xASL_qc_CollectQC_ASL(x, iSubject)
     else
         x = xASL_adm_DefineASLResolution(x);
 		warning('T1w-related files missing, computing ASL data using MNI templates!!!');
-        xASL_im_PreSmooth(x.P.Path_CBF,fullfile(x.D.TemplateDir,'rc1T1_ASL_res.nii'),x.P.Path_rc1T1,x.S.optimFWHM_Res_mm,[],x.P.Path_mean_PWI_Clipped_sn_mat, 1);
-        xASL_im_PreSmooth(x.P.Path_CBF,fullfile(x.D.TemplateDir,'rc2T1_ASL_res.nii'),x.P.Path_rc2T1,x.S.optimFWHM_Res_mm,[],x.P.Path_mean_PWI_Clipped_sn_mat, 1);
+        xASL_im_PreSmooth(x.P.Path_CBF,fullfile(x.D.TemplateDir,'rc1T1_ASL_res.nii'),...
+            x.P.Path_rc1T1,x.S.optimFWHM_Res_mm,[],x.P.Path_mean_PWI_Clipped_sn_mat, 1);
+        xASL_im_PreSmooth(x.P.Path_CBF,fullfile(x.D.TemplateDir,'rc2T1_ASL_res.nii'),...
+            x.P.Path_rc2T1,x.S.optimFWHM_Res_mm,[],x.P.Path_mean_PWI_Clipped_sn_mat, 1);
 	
 		xASL_spm_reslice(x.P.Path_CBF, x.P.Path_rc1T1, x.P.Path_mean_PWI_Clipped_sn_mat, 1, x.settings.Quality, x.P.Path_rc1T1);
 		xASL_spm_reslice(x.P.Path_CBF, x.P.Path_rc2T1, x.P.Path_mean_PWI_Clipped_sn_mat, 1, x.settings.Quality, x.P.Path_rc2T1);
@@ -94,12 +95,15 @@ function [x] = xASL_qc_CollectQC_ASL(x, iSubject)
 		Path_pWM = x.P.Path_rc2T1;
 	end
 	
+    % Read CBF NIfTI
     pGM = xASL_io_Nifti2Im(Path_pGM);
     pWM = xASL_io_Nifti2Im(Path_pWM);
-    if isfield(x.Q,'bUseBasilQuantification') && x.Q.bUseBasilQuantification == 1 % use BASIL CBF image paths
+    if isfield(x.Q,'bUseBasilQuantification') && x.Q.bUseBasilQuantification==1
+        % use BASIL CBF image paths
         imCBF = xASL_io_Nifti2Im(x.P.Path_CBF_Basil);
     else
-        imCBF = xASL_io_Nifti2Im(x.P.Path_CBF); % use xASL CBF image paths
+        % use xASL CBF image paths
+        imCBF = xASL_io_Nifti2Im(x.P.Path_CBF);
     end
         
     if xASL_stat_SumNan(pGM(:))==0
@@ -195,8 +199,9 @@ function [x] = xASL_qc_CollectQC_ASL(x, iSubject)
 
 end
 
+
+%% Fill fields
 function [OutputFields] = xASL_qc_FillFields(OutputFields, InputFields)
-%xASL_qc_FillFields
         
 FieldsI = fields(InputFields);
 
@@ -209,3 +214,4 @@ for iO=1:length(FieldsI)
 end    
 
 end
+
--- a/Functions/xASL_qc_CollectQC_ASL.m
+++ b/Functions/xASL_qc_CollectQC_ASL.m
@@ -96,7 +96,13 @@ function [x] = xASL_qc_CollectQC_ASL(x, iSubject)
 	
     pGM = xASL_io_Nifti2Im(Path_pGM);
     pWM = xASL_io_Nifti2Im(Path_pWM);
-    imCBF = xASL_io_Nifti2Im(x.P.Path_CBF);
+    if x.Q.bUseBasilQuantification
+        imCBF = xASL_io_Nifti2Im(x.P.Path_CBF_Basil);
+    else
+        imCBF = xASL_io_Nifti2Im(x.P.Path_CBF);
+    end
+        
+    
 
     if xASL_stat_SumNan(pGM(:))==0
         warning(['Empty image, invalid ' Path_pGM]);
--- a/Functions/xASL_qc_CollectQC_ASL.m
+++ b/Functions/xASL_qc_CollectQC_ASL.m
@@ -96,14 +96,12 @@ function [x] = xASL_qc_CollectQC_ASL(x, iSubject)
 	
     pGM = xASL_io_Nifti2Im(Path_pGM);
     pWM = xASL_io_Nifti2Im(Path_pWM);
-    if x.Q.bUseBasilQuantification
+    if isfield(x.Q,'bUseBasilQuantification') && x.Q.bUseBasilQuantification == 1 % use BASIL CBF image paths
         imCBF = xASL_io_Nifti2Im(x.P.Path_CBF_Basil);
     else
-        imCBF = xASL_io_Nifti2Im(x.P.Path_CBF);
+        imCBF = xASL_io_Nifti2Im(x.P.Path_CBF); % use xASL CBF image paths
     end
         
-    
-
     if xASL_stat_SumNan(pGM(:))==0
         warning(['Empty image, invalid ' Path_pGM]);
     end
--- a/Functions/xASL_quant_BSupCalculation.m
+++ b/Functions/xASL_quant_BSupCalculation.m
@@ -18,6 +18,7 @@ function SignalPercentage = xASL_quant_BSupCalculation(BackgroundSuppressionPuls
 % OUTPUT:
 %   SignalPercentage - signal percentage (0-1) of the remaining signal after background suppression.
 %                      It can be a scalar or a vector of the same dimension as sliceTime
+%
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
 % DESCRIPTION: This function computes the tissue signal percentage that
 %              remains after background suppression pulses are played in the ASL
@@ -28,8 +29,9 @@ function SignalPercentage = xASL_quant_BSupCalculation(BackgroundSuppressionPuls
 %
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
 % EXAMPLE: signalPercentage = xASL_quant_BSupCalculation([2200 2600], 2700, 1, 1240, [0 30 50 70], '/home/test/graph.jpg')
+%
 % __________________________________
-% Copyright 2015-2020 ExploreASL
+% Copyright (c) 2015-2021 ExploreASL
  
 % Admin
 if nargin < 1 || isempty(BackgroundSuppressionPulseTime)
@@ -88,18 +90,16 @@ SignalPercentageInitial = zeros(1, numel(ReadoutTime));
 if PresaturationTime
     % Recovery from 0 to 1 at T1 for the period between PresaturationTime and the first BSup pulse
     SignalPercentageInitial(1) = 1 - (1 - 0)*exp(-(BackgroundSuppressionPulseTime(1)-PresaturationTime)/T1Time);
-    PresaturationTimeDummy = PresaturationTime;
 else
     % Without saturation, full signal is preset
     SignalPercentageInitial(1) = 1;
     % For the purpose of visualization, in the absence of presaturation we set the presaturation time to 1 
-    PresaturationTimeDummy = 1;
 end
  
 % First phase calculation for a graph visual output
 if ~isempty(PathGraph)
-    % Preallocate
-	SignalPercentageVector = zeros(round(sum(ReadoutTime + max(SliceTime))), 1); % sum is required if multiPLD and therefore multiple ReadoutTime is present
+    % Preallocate (sum is required if multiPLD and therefore multiple ReadoutTime is present)
+	SignalPercentageVector = zeros(round(sum(ReadoutTime + max(SliceTime))), 1);
 	
 	if PresaturationTime
 		SignalPercentageVector(1:PresaturationTime,1) = 1;
@@ -219,3 +219,4 @@ end
  
 
 end
+
--- a/Functions/xASL_quant_BSupCalculation.m
+++ b/Functions/xASL_quant_BSupCalculation.m
@@ -99,7 +99,7 @@ end
 % First phase calculation for a graph visual output
 if ~isempty(PathGraph)
     % Preallocate
-	SignalPercentageVector = zeros(round(ReadoutTime + max(SliceTime)), 1);
+	SignalPercentageVector = zeros(round(sum(ReadoutTime + max(SliceTime))), 1); % sum is required if multiPLD and therefore multiple ReadoutTime is present
 	
 	if PresaturationTime
 		SignalPercentageVector(1:PresaturationTime,1) = 1;
--- a/Functions/xASL_quant_BSupCalculation.m
+++ b/Functions/xASL_quant_BSupCalculation.m
@@ -99,7 +99,7 @@ end
 % First phase calculation for a graph visual output
 if ~isempty(PathGraph)
     % Preallocate (sum is required if multiPLD and therefore multiple ReadoutTime is present)
-	SignalPercentageVector = zeros(round(sum(ReadoutTime + max(SliceTime))), 1);
+	SignalPercentageVector = zeros(round(ReadoutTime + max(SliceTime)), 1);
 	
 	if PresaturationTime
 		SignalPercentageVector(1:PresaturationTime,1) = 1;
--- a/Functions/xASL_quant_Basil.m
+++ b/Functions/xASL_quant_Basil.m
@@ -59,12 +59,20 @@ function [CBF_nocalib, resultFSL] = xASL_quant_Basil(PWI, x)
     % dummy input image here
     PWI(isnan(PWI)) = 0;
     
-    xASL_io_SaveNifti(x.P.Path_PWI, pathBasilInput, PWI, [], 0);
+    if size(PWI,4) == 1 % singlePLD
+        xASL_io_SaveNifti(x.P.Path_PWI, pathBasilInput, PWI, [], 0); % use PWI path
+    elseif size(PWI,4) > 1 %multiPLD
+        xASL_io_SaveNifti(x.P.Path_PWI4D, pathBasilInput, PWI, [], 0); % use PWI4D path
+    end
 
 
     %% 4. Create option_file that contains options which are passed to Fabber
     % basil_options is a character array containing CLI args for the Basil command
-    bMultiPLD = false; % for now
+    if size(PWI,4) < 2
+        bMultiPLD = false; 
+    else
+        bMultiPLD = true;
+    end
     
     BasilOptions = xASL_quant_Basil_Options(pathBasilOptions, x, PWI, bMultiPLD);
 
@@ -86,7 +94,7 @@ function [CBF_nocalib, resultFSL] = xASL_quant_Basil(PWI, x)
     
     pathBasilMean = xASL_adm_GetFileList(dirBasilOutput, '^mean_ftiss\.nii$', 'FPListRec');
     pathBasilMean = pathBasilMean{end}; % we assume the latest iteration (alphabetically) is optimal. also converting cell to char array
-    
+       
     CBF_nocalib = xASL_io_Nifti2Im(pathBasilMean);
     
     
@@ -132,13 +140,22 @@ function [BasilOptions] = xASL_quant_Basil_Options(pathBasilOptions, x, PWI, bMu
     
     
     %% Basic acquisition and tissue parameters
-    fprintf(FIDoptionFile, '# Basil options written by ExploreASL\n');
-    fprintf(FIDoptionFile, '--repeats=%i\n', size(PWI, 4));
-    fprintf(FIDoptionFile, '--ti1=%f\n', (x.Q.LabelingDuration + x.Q.Initial_PLD)/1000);
-    fprintf(FIDoptionFile, '--t1b=%f\n', x.Q.BloodT1/1000);
-    fprintf(FIDoptionFile, '--tau=%f\n', x.Q.LabelingDuration/1000); % FIXME tau could be list
-    fprintf(FIDoptionFile, '--slicedt=%f\n', x.Q.BasilSliceReadoutTime/1000);
-    fprintf(FIDoptionFile, '--save-model-fit\n');
+     % create CSV style text for TIs input
+
+     % print parameters
+     fprintf(FIDoptionFile, '# Basil options written by ExploreASL\n');
+     fprintf(FIDoptionFile, '--iaf=diff\n'); % as input is PWI
+     
+     TIs = (x.Q.LabelingDuration + x.Q.Initial_PLD)'/1000;
+     PLDAmount = length(TIs);
+     for TIsingle = 1:PLDAmount
+         fprintf(FIDoptionFile, ['--ti' num2str(TIsingle) '=%.2f\n'], TIs(TIsingle));
+     end
+     fprintf(FIDoptionFile, '--repeats=%i\n', size(PWI, 4)/PLDAmount);
+     fprintf(FIDoptionFile, '--t1b=%f\n', x.Q.BloodT1/1000);
+     fprintf(FIDoptionFile, '--tau=%f\n', x.Q.LabelingDuration/1000); % FIXME tau could be list
+     fprintf(FIDoptionFile, '--slicedt=%f\n', x.Q.BasilSliceReadoutTime/1000);
+     fprintf(FIDoptionFile, '--save-model-fit\n');
 
 
     %% FIXME Aquisition options we might be able to use in the future
--- a/Functions/xASL_quant_Basil.m
+++ b/Functions/xASL_quant_Basil.m
@@ -1,5 +1,6 @@
 function [CBF_nocalib, resultFSL] = xASL_quant_Basil(PWI, x)
 %xASL_quant_Basil Perform quantification using FSL BASIL
+
 % FORMAT: [CBF_nocalib] = xASL_quant_Basil(PWI, x)
 % 
 % INPUT:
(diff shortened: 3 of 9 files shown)


Functions/xASL_quant_BSupCalculation.m
```diff
@@ -88,18 +90,16 @@
 if PresaturationTime
     % Recovery from 0 to 1 at T1 for the period between PresaturationTime and the first BSup pulse
     SignalPercentageInitial(1) = 1 - (1 - 0)*exp(-(BackgroundSuppressionPulseTime(1)-PresaturationTime)/T1Time);
-    PresaturationTimeDummy = PresaturationTime;
 else
     % Without saturation, full signal is preset
     SignalPercentageInitial(1) = 1;
     % For the purpose of visualization, in the absence of presaturation we set the presaturation time to 1 
-    PresaturationTimeDummy = 1;
 end
```
MichaelStritt, line 97: I removed PresaturationTimeDummy, since I didn't see where it is used. Hope that was correct. Please re-check 😉

Functions/xASL_quant_Basil.m
```diff
@@ -46,28 +46,33 @@
...
     xASL_delete(pathBasilOptions);
     xASL_delete(pathBasilInput);
-    xASL_adm_DeleteFileList(x.dir.SESSIONDIR, '(?i)^.*basil.*$', 1, [0 Inf]); % remove residual BASIL-related files
+    xASL_adm_DeleteFileList(x.dir.SESSIONDIR, '(?i)^.*basil.*$', 1, [0 Inf]);
     
     %% 3. Write the PWI as Nifti file for Basil to read as input
-    % FIXME would be good to have a brain mask at this point
-    % PM: if this would be a brainmask as well, we can skip creating a
-    % dummy input image here
+    % FIXME would be good to have a brain mask at this point -> PM: if this would be a brainmask as well, we can skip creating a dummy input image here
```
MichaelStritt, line 55: Is that comment still relevant? I personally don't mind it, but I wasn't sure if it's still up-to-date.

MichaelStritt (approved):
I only found minor stuff. Please check my comments and wait for @jan-petr's review. Nice work 👍

Modules/xASL_module_ASL.m
```diff
@@ -421,21 +459,25 @@
...
-            xASL_wrp_Quantify(x, x.P.Path_PWI4D, x.P.Path_CBF, x.P.Path_rM0, x.P.Path_SliceGradient, 1);
+            x = xASL_wrp_Quantify(x, x.P.Path_PWI4D, x.P.Path_CBF, x.P.Path_rM0, x.P.Path_SliceGradient, 1); 
         else
-            xASL_wrp_Quantify(x, x.P.Path_PWI, x.P.Path_CBF, x.P.Path_rM0, x.P.Path_SliceGradient, 1);
-        end        
-    end
-    
+            x = xASL_wrp_Quantify(x, x.P.Path_PWI, x.P.Path_CBF, x.P.Path_rM0, x.P.Path_SliceGradient, 1);
+        end   
+    else
```
jan-petr, line 472: So until now, when we had BASIL set for single-PLD, it outputted both CBF.nii and CBF_BASIL.nii.
I am not against this - if you decide to go for BASIL, you go for BASIL. But - we need to make changes in the TestDataSet - because currently, it is running 1-2 BASIL quantification, but really evaluating the non-BASIL output.

So I propose to do the following:
1. Keep here your change of running either normal or BASIL quantification.
2. But make sure that it always outputs only the BASIL output as CBF.nii
3. Revert the BASIL option in TestDataSet, or make sure that BASIL and no-BASIL CBF are equal, or we will have issues with it later.

Functions/xASL_quant_MultiPLD.m
```diff
@@ -0,0 +1,295 @@
...
+%
+% -----------------------------------------------------------------------------------------------------------------------------------------------------
+% EXAMPLE: [ScaleImage, CBF] = xASL_quant_MultiPLD(PWI, M0_im, imSliceNumber, x, bUseBasilQuantification);
+% __________________________________
+% Copyright 2015-2019 ExploreASL
+
+
+%% Admin
+fprintf('%s\n','Quantification CBF multi-PLD:');
+
```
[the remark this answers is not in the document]
MichaelStritt: Should be fixed now 😉

Functions/xASL_quant_Basil.m
```diff
@@ -56,18 +56,24 @@
     % dummy input image here
     PWI(isnan(PWI)) = 0;
     
-    xASL_io_SaveNifti(x.P.Path_PWI, pathBasilInput, PWI, [], 0);
+    if size(PWI,4) == 1 % singlePLD
+        xASL_io_SaveNifti(x.P.Path_PWI, pathBasilInput, PWI, [], 0); % use PWI path
+    elseif size(PWI,4) > 1 %multiPLD
```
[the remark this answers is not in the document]
MichaelStritt: Fixed in ad89bfaad2dbab7d97923b069226bdf5f6967e41 👍

Functions/xASL_quant_Basil.m
```diff
@@ -118,17 +123,23 @@
...
+     % create CSV style text for TIs input
+
+     % print parameters
+     fprintf(FIDoptionFile, '# Basil options written by ExploreASL\n');
+     fprintf(FIDoptionFile, '--iaf=diff\n'); % as input is PWI
+     
+     TIs = (x.Q.LabelingDuration + x.Q.Initial_PLD)'/1000;
+     PLDAmount = length(TIs);
+     for TIsingle = 1:PLDAmount
+         fprintf(FIDoptionFile, ['--ti' num2str(TIsingle) '=%.2f\n'], TIs(TIsingle));
```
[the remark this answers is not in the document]
MichaelStritt: Fixed in ad89bfa 👍

Functions/xASL_quant_MultiPLD.m
```diff
@@ -0,0 +1,295 @@
+function [ScaleImage, CBF] = xASL_quant_MultiPLD(PWI, M0_im, imSliceNumber, x, bUseBasilQuantification)
+%xASL_quant_MultiPLD % Perform a multi-step quantification
+% FORMAT: [ScaleImage[, CBF]] = xASL_quant_MultiPLD(PWI, M0_im, imSliceNumber, x, bUseBasilQuantification)
```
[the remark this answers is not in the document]
MichaelStritt: Fixed in ad89bfa 👍

Functions/xASL_quant_MultiPLD.m
```diff
@@ -0,0 +1,295 @@
...
+%              PWI stage)
+%
+%              Note that BASIL is also implemented, but it doesn't allow a
+%              standard space quantification yet (it would need to use
+%              imSliceNumber)
+%
+% -----------------------------------------------------------------------------------------------------------------------------------------------------
+% EXAMPLE: [ScaleImage, CBF] = xASL_quant_MultiPLD(PWI, M0_im, imSliceNumber, x, bUseBasilQuantification);
+% __________________________________
+% Copyright 2015-2019 ExploreASL
```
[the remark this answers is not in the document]
MichaelStritt: Fixed in ad89bfa 👍

Modules/SubModule_ASL/xASL_wrp_Quantify.m
```diff
@@ -1,4 +1,4 @@
-function xASL_wrp_Quantify(x, PWI_Path, OutputPath, M0Path, SliceGradientPath, bUseBasilQuantification)
+function x = xASL_wrp_Quantify(x, PWI_Path, OutputPath, M0Path, SliceGradientPath, bUseBasilQuantification)
 %xASL_wrp_Quantify Submodule of ExploreASL ASL Module, that performs quantfication
 %
 % FORMAT: xASL_wrp_Quantify(x)
```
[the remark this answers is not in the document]
MichaelStritt: Fixed in ad89bfa 👍

Modules/SubModule_ASL/xASL_wrp_Quantify.m
```diff
@@ -291,7 +291,15 @@ function xASL_wrp_Quantify(x, PWI_Path, OutputPath, M0Path, SliceGradientPath, b
...
 %% ------------------------------------------------------------------------------------------------
 %% 8.   Perform Quantification
-[~, CBF] = xASL_quant_SinglePLD(PWI, M0_im, SliceGradient, x, bUseBasilQuantification); % also runs BASIL, but only in native space!
+if size(PWI,4) == 1 % single PLD quantification
+    fprintf('%s\n',['Performing single PLD quantification']);
+    [~, CBF] = xASL_quant_SinglePLD(PWI, M0_im, SliceGradient, x, bUseBasilQuantification); % also runs BASIL, but only in native space!
+elseif size(PWI,4) > 1 && x.Q.bUseBasilQuantification % perform BASIL multi-PLD quantification
+    fprintf('%s\n',['Performing multi PLD quantification using BASIL']);
+    [~, CBF] = xASL_quant_MultiPLD(PWI, M0_im, SliceGradient, x, bUseBasilQuantification); % also runs multi-PLD BASIL, but only in native space!
+else size(PWI,4) > 1 && ~x.Q.bUseBasilQuantification % perform BASIL multi-PLD quantification
```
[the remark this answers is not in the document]
MichaelStritt: I fixed the elseif / else thing in ad89bfa 👍 Jan is right, we probably can get rid of the `size(PWI,4) > 1` statement. I wasn't sure about all of those `bUseBasilQuantification` vs. `x.Q.bUseBasilQuantification` lines. I tried to make it a bit more easy to read by using `x.Q.bUseBasilQuantification` throughout. We could probably also update the general function statement and get rid of the input argument, if we update all scripts where this is used as well, this minor optimization can be done later on though.

Modules/SubModule_ASL/xASL_wrp_RealignASL.m
```diff
@@ -121,15 +128,13 @@ function xASL_wrp_RealignASL(x,bSubtraction)
...
-if nFrames>2 && bSubtraction && (numel(unique(x.Q.Initial_PLD))>1 || numel(unique(x.EchoTime))>1 || x.modules.asl.HadamardType~=0) %multiPLD or multiTE or Hadamard
-    spm_realign(spm_vol(InputPath),flags,false);
-    
-elseif nFrames>2 && bSubtraction
-    spm_realign(spm_vol(InputPath),flags,true);
-    
-elseif nFrames>1
-    spm_realign(spm_vol(InputPath),flags,false);
+
+if ~bSubtraction || nFrames<2
```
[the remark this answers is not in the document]
MichaelStritt: Fixed in ad89bfa 👍

Functions/xASL_qc_CollectQC_ASL.m
```diff
@@ -94,10 +95,17 @@
 		Path_pWM = x.P.Path_rc2T1;
 	end
 	
+    % Read CBF NIfTI
     pGM = xASL_io_Nifti2Im(Path_pGM);
     pWM = xASL_io_Nifti2Im(Path_pWM);
-    imCBF = xASL_io_Nifti2Im(x.P.Path_CBF);
-
+    if isfield(x.Q,'bUseBasilQuantification') && x.Q.bUseBasilQuantification==1
```
HenkMutsaerts, line 101: @MDijsselhof The way it was, the output has two different paths, `CBF` and `CBF_Basil`. We may want to think this through before changing everything. If with `bUseBasilQuantification`, we want to do the whole pipeline based on the BASIL images, we might as well rename the Basil output to `qCBF` rather than `qCBF_BASIL`, to avoid unnecessary complexity?

Also, best to define as much as these paths etc at the beginning of the ASL module

Functions/xASL_quant_BSupCalculation.m
HenkMutsaerts: Best to ask Jan!

Functions/xASL_quant_BSupCalculation.m
```diff
@@ -88,18 +90,16 @@
...
  
 % First phase calculation for a graph visual output
 if ~isempty(PathGraph)
-    % Preallocate
-	SignalPercentageVector = zeros(round(ReadoutTime + max(SliceTime)), 1);
+    % Preallocate (sum is required if multiPLD and therefore multiple ReadoutTime is present)
+	SignalPercentageVector = zeros(round(sum(ReadoutTime + max(SliceTime))), 1);
```
HenkMutsaerts, line 102: Are you sure it should be a sum? Is the sum on the correct location? Sum seems weird here... @jan-petr

Functions/xASL_quant_Basil.m
```diff
@@ -1,5 +1,5 @@
 function [CBF_nocalib] = xASL_quant_Basil(PWI, x)
-%xASL_quant_Basil Perform quantification using FSL BASIL
+%xASL_quant_Basil Perform quantification (either single- or multi-PLD) using FSL BASIL
```
HenkMutsaerts, line 2: nice!

Functions/xASL_quant_Basil.m
HenkMutsaerts: BASIL requires a brain mask, and we create specific mask(s) in ExploreASL, so this comment is there to remind us to see if this can be synchronised better, for better comparability between the results of our quantification, and BASILs quantification.

Functions/xASL_quant_Basil.m
```diff
@@ -46,28 +46,35 @@
...
     
     %% 3. Write the PWI as Nifti file for Basil to read as input
-    % FIXME would be good to have a brain mask at this point
-    % PM: if this would be a brainmask as well, we can skip creating a
-    % dummy input image here
+    % FIXME would be good to have a brain mask at this point -> PM: if this would be a brainmask as well, we can skip creating a dummy input image here
     PWI(isnan(PWI)) = 0;
     
-    xASL_io_SaveNifti(x.P.Path_PWI, pathBasilInput, PWI, [], 0);
+    if size(PWI,4) == 1
```
HenkMutsaerts, line 58: @MDijsselhof I believe BASIL always requires the PWI4D, because even with single PLD, it can then still do some modelling. Also, we should check if PWI4D exists if you have a single PLD scan with only 1 control-label subtraction (it should). Let's make sure that PWI4D always exists, even if it is the same as PWI.

Functions/xASL_quant_Basil.m
```diff
@@ -56,18 +56,24 @@
...
+        xASL_io_SaveNifti(x.P.Path_PWI, pathBasilInput, PWI, [], 0); % use PWI path
+    elseif size(PWI,4) > 1 %multiPLD
+        xASL_io_SaveNifti(x.P.Path_PWI4D, pathBasilInput, PWI, [], 0); % use PWI4D path
+    end
 
 
     %% 4. Create option_file that contains options which are passed to Fabber
     % basil_options is a character array containing CLI args for the Basil command
-    bMultiPLD = false; % for now
+    if size(PWI,4) < 2
```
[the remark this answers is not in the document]
HenkMutsaerts: Agreed, then it can be inside the same if statement instead of needing multiple if statements.

Functions/xASL_quant_Basil.m
```diff
@@ -211,9 +223,9 @@
     end
 
 
-    %% Noise specification. For small numbers of time points we need an informative
-    % noise prior. The user can specify an assumed SNR for this, or give noise std.dev
-    % directly.
+    %% Noise specification
+    % For small numbers of time points we need an informative noise prior. 
+    % The user can specify an assumed SNR for this, or give noise standard deviation directly.
```
HenkMutsaerts, line 228: How can the user give a SD if we only use the SNR option here?

Functions/xASL_quant_MultiPLD.m
```diff
@@ -0,0 +1,299 @@
+function [ScaleImage, CBF] = xASL_quant_MultiPLD(PWI, M0_im, imSliceNumber, x, bUseBasilQuantification)
```
HenkMutsaerts, line 1: Good to discuss next week if we should have separate singlePLD and MultiPLD functions, perhaps a lot is in common between the two. Always good to reduce copies of code, as this can potentially lead to bias (if only one of the copies is changed).

Modules/SubModule_ASL/xASL_wrp_Quantify.m
```diff
@@ -1,19 +1,20 @@
-function xASL_wrp_Quantify(x, PWI_Path, OutputPath, M0Path, SliceGradientPath, bUseBasilQuantification)
+function x = xASL_wrp_Quantify(x, PWI_Path, OutputPath, M0Path, SliceGradientPath, bUseBasilQuantification)
```
HenkMutsaerts, line 1: Why the `x` output? This quantify wrapper is a submodule of `xASL_module_ASL`, and its outputs are quantified CBF NIfTI files, not `x`.

Modules/SubModule_ASL/xASL_wrp_Quantify.m
```diff
@@ -77,7 +80,7 @@ function xASL_wrp_Quantify(x, PWI_Path, OutputPath, M0Path, SliceGradientPath, b
 
 if ~xASL_exist(PWI_Path, 'file')
     warning('Skipped xASL_wrp_Quantify: files missing, please rerun step 4: xASL_wrp_ResampleASL');
-	if bUseBasilQuantification
+	if x.Q.bUseBasilQuantification
```
HenkMutsaerts, line 83: At the header it is without `x.Q`, see `xASL_module_ASL` that calls this wrapper.

Modules/SubModule_ASL/xASL_wrp_Quantify.m
```diff
@@ -291,7 +294,17 @@ function xASL_wrp_Quantify(x, PWI_Path, OutputPath, M0Path, SliceGradientPath, b
...
+if size(PWI,4) == 1 % single PLD quantification
+    fprintf('%s\n',['Performing single PLD quantification']);
+    [~, CBF] = xASL_quant_SinglePLD(PWI, M0_im, SliceGradient, x, x.Q.bUseBasilQuantification); % also runs BASIL, but only in native space!
+elseif size(PWI,4) > 1 && x.Q.bUseBasilQuantification
+    % perform BASIL multi-PLD quantification
+    fprintf('%s\n',['Performing multi PLD quantification using BASIL']);
+    [~, CBF] = xASL_quant_MultiPLD(PWI, M0_im, SliceGradient, x, x.Q.bUseBasilQuantification); % also runs multi-PLD BASIL, but only in native space!
+elseif size(PWI,4) > 1 && ~x.Q.bUseBasilQuantification 
+    % perform BASIL multi-PLD quantification
+    fprintf('%s\n',['Multi PLD quantification without BASIL not implemented yet']);
```
HenkMutsaerts, line 306: Nice! Should perhaps be a warning.

Modules/SubModule_ASL/xASL_wrp_RealignASL.m
```diff
@@ -61,6 +61,15 @@ function xASL_wrp_RealignASL(x,bSubtraction)
 PercExcl = NaN;
 MinimumtValue = NaN;
 
+if (isfield(x.modules.asl,'bMultiPLD') && x.modules.asl.bMultiPLD) || ...
```
HenkMutsaerts, line 64: Would be good to define this at the start of `xASL_module_ASL`, so you don't have to do the `isfield` at each instance.

Modules/SubModule_ASL/xASL_wrp_RealignASL.m
```diff
@@ -216,20 +223,20 @@ function xASL_wrp_RealignASL(x,bSubtraction)
 %% ----------------------------------------------------------------------------------------
 %% 3) Threshold-free spike definition (based on ENABLE, but with t-stats rather than the threshold p<0.05)
 
-bExecutedENABLE = 0;
 
 if bSubtraction && nFrames<=10
```
HenkMutsaerts, line 227: Now at multiple locations in this wrapper, we check for a condition to set bENABLE. Would be good to put all these conditions on the top at a single location, for more overview. E.g., if bSubtraction is false, you would never run bENABLE or bZigZag, and then you don't have to check bMultiTE or bMultiPLD. So that could go to the top. Good to keep this in a logic order, so it's easy to read and debug.

Modules/SubModule_ASL/xASL_wrp_RealignASL.m
```diff
@@ -334,96 +343,94 @@ function xASL_wrp_RealignASL(x,bSubtraction)
 end
 
 %only run if ENABLE is on
-if bExecutedENABLE == 1;
-    if bSubtraction && nFrames>10 % if we performed outlier exclusion
-        tValue(1:3) = tValue(4); % for nicer plotting
+if bENABLE && bSubtraction && nFrames>10 % if we performed outlier exclusion
```
HenkMutsaerts, line 346: What happened here? Please don't only add a tab, unless it is in a separate commit (I didn't check that). But now I cannot see what the difference is, so I hope the difference is only a tab indentation ;)

Modules/SubModule_ASL/xASL_wrp_RegisterASL.m
```diff
@@ -296,14 +296,20 @@
 %% H. Here we create a temporary dummy ASL image of which the image contrast is curated,
 % for registration only
 xASL_io_PairwiseSubtraction(x.P.Path_despiked_ASL4D, x.P.Path_mean_PWI_Clipped, 0, 0); % create PWI & mean_control
+% MultiPLD: we can keep the PairwiseSubtraction function->because we have controls and labels for each PLD and take average of all PLDs
+% (later on maybe we only need to use the later PLDs)
+
 if bRegistrationCBF
+    % This clipped_ORI is not used in the rest of the pinpeline
```
HenkMutsaerts, line 303: What is a pinpeline? haha

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
```diff
@@ -222,18 +223,38 @@ function xASL_wrp_ResampleASL(x)
...
+    else
+        
         % Paired subtraction
         [ControlIm, LabelIm] = xASL_quant_GetControlLabelOrder(ASL_im);
         ASL_im = ControlIm - LabelIm;
-        % Save PWI4D
-        fprintf('%s', [PathPWI4D{iSpace} ', ']);
-        xASL_io_SaveNifti(PathASL{iSpace}, PathPWI4D{iSpace}, ASL_im, 32, false);
+        % preallocate PWI
+        PWI = zeros(size(ASL_im,1),size(ASL_im,2),size(ASL_im,3));
```
HenkMutsaerts, line 232: What happens here if `size(ASL_im,4)~=1`?

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
```diff
@@ -222,18 +223,38 @@ function xASL_wrp_ResampleASL(x)
...
+        
         % Average PWI
-        PWI = xASL_stat_MeanNan(ASL_im, 4);
-        % Save PWI
-        fprintf('%s\n', PathPWI{iSpace});
-        xASL_io_SaveNifti(PathASL{iSpace}, PathPWI{iSpace}, PWI, 32, false);
+        if isfield(x.modules.asl,'bMultiPLD') && x.modules.asl.bMultiPLD 
+            % multiPLD PWI
+            RepeatsInPLD = size(ASL_im,4)/length(x.Q.Initial_PLD); % we still need to create a single Repeats in X struct
+            for PLDnmbr = 1:length(x.Q.Initial_PLD)
```
HenkMutsaerts, line 238: What does `nmbr` stand for? `nPLD` would be a more understood abbreviated. Always either write it out fully, or use something that is used a lot. also `i` for Index (not `loc`)

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
```diff
@@ -222,18 +223,38 @@ function xASL_wrp_ResampleASL(x)
...
+                PWI(:,:,:,PLDnmbr) = xASL_stat_MeanNan(ASL_im(:,:,:,RepeatsLocStartPerPLD:RepeatsLocEndPerPLD), 4);
+            end
+            % Save PWI
+            fprintf('%s\n', PathPWI4D{iSpace});
+            xASL_io_SaveNifti(PathASL{iSpace}, PathPWI4D{iSpace}, PWI, 32, false);
+            
+            PWIsingle = xASL_stat_MeanNan(ASL_im, 4); % create single PWI for further steps in ASL module
+            % Save single PWI
+            fprintf('%s\n', PathPWI{iSpace});
+            xASL_io_SaveNifti(PathASL{iSpace}, PathPWI{iSpace}, PWIsingle, 32, false);
```
HenkMutsaerts, line 250: Nice work! What does iSpace stand for? Should it be iDim (dimension)? Or iPLD? Just avoid confusion

Modules/xASL_module_ASL.m
```diff
@@ -213,6 +209,62 @@
 %% Define sequence (educated guess)
 x = xASL_adm_DefineASLSequence(x);
 
+%% MultiPLD parsing
+
+if isfield(x.Q,'Initial_PLD') && numel(unique(x.Q.Initial_PLD))>1  % 2 because the PLD is being repeated for control and label -> check this: 1 or 2?
```
HenkMutsaerts, line 214: Great work, add more comments for the next guy that will read this!

Modules/xASL_module_ASL.m
HenkMutsaerts: Yes I thought so as well. Makes the code simpler as well, as you don't need if `bUseBasilQuantification` in subsequent wrappers. Would be good then to state at several locations a clear message (not a warning) that Basil quantification is used.

HenkMutsaerts (changes requested):
Sorry this is a big change so will get lots of comments. Hoge bomen vangen veel wind ;)

Modules/SubModule_ASL/xASL_wrp_Quantify.m
jan-petr: Yes - you need to be very careful to distinguish between local variables and global ones saved in x-struct. Especially since you now return the x-struct.

Modules/SubModule_ASL/xASL_wrp_Quantify.m
jan-petr: in some cases, x-struct is modified to save certain variables for QC etc. But here, you have to be supercareful. This is for example connected with the bUseBasilQuantification. In the main xASL-module, this function is called once with BASIL=1 and once with BASIL=0. If you save it to x-struct and return, this will totally mess up things.

So bUseBasilQuantification should stay local, and you should prove if there's anything else saved in the x-struct that needs to be passed as an output.

Modules/xASL_module_ASL.m
jan-petr: the numel+unique is good now, but the comment needs to be rewritten. We don't need to check for 1/2 - this has either ==1 or >1, irrespective of control and label - so please put an explanation that we go for unique PLDs.

jan-petr (changes requested):

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
MDijsselhof: iSpace is either native or common space, this part runs the analysis for both if inputted

Functions/xASL_quant_Basil.m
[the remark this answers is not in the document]
MDijsselhof: Done

Modules/xASL_module_ASL.m
```diff
@@ -213,6 +209,48 @@
 %% Define sequence (educated guess)
 x = xASL_adm_DefineASLSequence(x);
 
+%% MultiPLD parsing
+
+if length(x.Q.Initial_PLD)>1  % 2 because the PLD is being repeated for control and label -> check this: 1 or 2?
```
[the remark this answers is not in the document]
MDijsselhof: Agreed and done.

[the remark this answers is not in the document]
MichaelStritt: This was fixed in ad89bfaad2dbab7d97923b069226bdf5f6967e41 👍

Modules/xASL_module_ASL.m
```diff
@@ -213,6 +209,48 @@
...
+    x.modules.asl.bMultiPLD = 1;
+    fprintf('%s\n','Multiple PLDs detected. Make sure that this is a valid multiPLD dataset');
```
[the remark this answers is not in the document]
MichaelStritt: This was fixed in ad89bfa 👍

Modules/xASL_module_ASL.m
```diff
@@ -213,6 +209,48 @@
...
+    
+else
+    x.modules.asl.bTimeEncoded = 0;
+end
+
+%% MultiTE parsing
+
+if length(x.EchoTime)>1
+    x.modules.asl.bMultiTE = 1;
+    fprintf('%s\n','Multiple echo times detected. Make sure that this is a valid multiTE dataset');
```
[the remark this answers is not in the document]
MichaelStritt: This was fixed in ad89bfa 👍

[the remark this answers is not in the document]
MDijsselhof: Done

Modules/xASL_module_ASL.m
MDijsselhof: Okay I will change all BASIL output to CBF.nii, instead of the current CBF_BASIL.nii. Will add the warnings as well.

Modules/SubModule_ASL/xASL_wrp_Quantify.m
MichaelStritt: Ah damn, okay, so the idea was to make it possible to run some sessions with and some without BASIL? Didn't knew that this was a use-case.

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
MDijsselhof: Done

Modules/xASL_module_ASL.m
MDijsselhof: Done

MDijsselhof:
@jan-petr Apparently I can't comment if someone requests changes so I'll just write it down here. 
Regarding the x.Q.bUseBasilQuantification: is this fine the way it is now? I don't follow.
Regarding the x struct: this was required to output qCBF_Basil.nii and relatable paths, however if we change all paths back to the original qCBF.nii this is not required anymore. Does that work for you?

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
MDijsselhof: I've tested it, it works fine. The result would be that we average the ASL_im over one repeat and that works, the PWI that is create is a single 3D volume

Modules/SubModule_ASL/xASL_wrp_Quantify.m
MDijsselhof: @jan-petr make it a warning?

HenkMutsaerts (approved):

PR #894, opened 2021-11-11
Closes #892 ExploreASL_Initialize: rm paths before addpath

### LINKED ISSUE #892: Existing toolbox Matlab paths conflict with ExploreASL toolboxes
### Description

If ExploreASL contains a toolbox (e.g., SPM) that was already initialized — i.e., its Matlab path was already included — then there may be conflicts between the toolbox version within ExploreASL and the one that was pre-existing

### Requirements

The implemented solution is to first detect ExploreASL toolboxes that were already added to the Matlab path outside of ExploreASL — i.e., not within `ExploreASL/External`.

### How to test

### Release notes

Bugfix: detect and remove pre-existing toolbox paths that might conflict with toolboxes included in ExploreASL

HenkMutsaerts:
### Linked issue

Closes #892

### How to test

### Comments

### DIFF
--- a/ExploreASL_Initialize.m
+++ b/ExploreASL_Initialize.m
@@ -200,6 +200,39 @@ end
 %% Add ExploreASL Directory
 function addExploreASLDirectory(MyPath)
 
+    fprintf('Managing Matlab paths:   ');
+
+    % First remove existing toolbox initializations
+    % This could contain other toolbox versions and create conflicts
+    currentPathList = path;
+    if ischar(currentPathList)
+        % convert to cell
+        indicesAre = [0 strfind(currentPathList, pathsep)];
+        
+        for iIndex=1:numel(indicesAre)-1
+            pathList{iIndex, 1} = currentPathList(indicesAre(iIndex)+1:indicesAre(iIndex+1)-1);
+        end
+    else
+        pathList = currentPathList;
+    end
+        
+    for iPath=1:numel(pathList)
+        if ~isempty(regexpi(pathList{iPath}, '(spm|cat12|lst|fsl)'))
+            % this part is useful for debugging
+%                 pathList{iPath,2} = 1;
+%             else
+%                 pathList{iPath,2} = 0;
+            if isempty(regexp(pathList{iPath}, fullfile('ExploreASL', 'External')))
+                % if this path is not an ExploreASL-contained toolbox
+                rmpath(pathList{iPath});
+                warning(['Removed Matlab path to avoid conflicts: ' pathList{iPath}]);
+            end
+        end
+    end
+    
+    fprintf('\n');
+
+
     % Define paths (should be equal when loading data or only initializing)
     addpath(MyPath); % ExploreASL
     subfoldersToAdd = {...
--- a/ExploreASL_Initialize.m
+++ b/ExploreASL_Initialize.m
@@ -200,37 +200,43 @@ end
 %% Add ExploreASL Directory
 function addExploreASLDirectory(MyPath)
 
-    fprintf('Managing Matlab paths:   ');
+    % Vebose mode for debugging
+    verboseMode = false;
+    if verboseMode
+        fprintf('Managing Matlab paths:\n');
+    end
 
     % First remove existing toolbox initializations
     % This could contain other toolbox versions and create conflicts
     currentPathList = path;
     if ischar(currentPathList)
-        % convert to cell
+        % Convert to cell
         indicesAre = [0 strfind(currentPathList, pathsep)];
-        
+        % Add to pathList
         for iIndex=1:numel(indicesAre)-1
             pathList{iIndex, 1} = currentPathList(indicesAre(iIndex)+1:indicesAre(iIndex+1)-1);
         end
     else
         pathList = currentPathList;
     end
-        
+    
+    % Iterate over paths
     for iPath=1:numel(pathList)
         if ~isempty(regexpi(pathList{iPath}, '(spm|cat12|lst|fsl)'))
-            % this part is useful for debugging
-%                 pathList{iPath,2} = 1;
-%             else
-%                 pathList{iPath,2} = 0;
-            if isempty(regexp(pathList{iPath}, fullfile('ExploreASL', 'External')))
-                % if this path is not an ExploreASL-contained toolbox
+            condition_xasl = regexp(pathList{iPath}, fullfile('ExploreASL', 'External'), 'once');
+            condition_matlab = regexp(pathList{iPath}, fullfile('matlab', 'toolbox'), 'once');
+            if isempty(condition_xasl) || isempty(condition_matlab)
+                % If this path is not an ExploreASL-contained toolbox
                 rmpath(pathList{iPath});
-                warning(['Removed Matlab path to avoid conflicts: ' pathList{iPath}]);
+                if verboseMode
+                    warning(['Removed Matlab path to avoid conflicts: ' pathList{iPath}]);
+                end
             end
         end
     end
-    
-    fprintf('\n');
+    if verboseMode
+        fprintf('\n');
+    end
 
 
     % Define paths (should be equal when loading data or only initializing)
--- a/ExploreASL_Initialize.m
+++ b/ExploreASL_Initialize.m
@@ -224,8 +224,8 @@ function addExploreASLDirectory(MyPath)
     for iPath=1:numel(pathList)
         if ~isempty(regexpi(pathList{iPath}, '(spm|cat12|lst|fsl)'))
             condition_xasl = regexp(pathList{iPath}, fullfile('ExploreASL', 'External'), 'once');
-            condition_matlab = regexp(pathList{iPath}, fullfile('matlab', 'toolbox'), 'once');
-            if isempty(condition_xasl) || isempty(condition_matlab)
+            condition_matlab = regexp(pathList{iPath}, fullfile(version('-release'), 'toolbox'), 'once');
+            if isempty(condition_xasl) && isempty(condition_matlab)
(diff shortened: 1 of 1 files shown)


jan-petr (approved):
OK

MichaelStritt:
@jan-petr & @HenkMutsaerts: did some changes to reduce the crazy verbosity, I bet users would greatly complain about that. I also added a condition for matlab related toolboxes, because that leads to super annoying bugs otherwise.

MichaelStritt (approved):

HenkMutsaerts:
> @jan-petr & @HenkMutsaerts: did some changes to reduce the crazy verbosity, I bet users would greatly complain about that. I also added a condition for matlab related toolboxes, because that leads to super annoying bugs otherwise.
> 
> Edit: please add the "how to test" section next time, we kind of agreed on this.

* Thanks, but we have to remove the Matlab toolbox condition. There are users having e.g., SPM in their matlab toolbox, this is what we want to remove from paths. This is why I had a filesep in front of the regular expressions, so it wouldn't detect `/alst` but it would detect `/lst`

* I disagree with the verbose mode, or we have to default it to true. Users should not have initialized other spm versions, cat12 versions, or lst versions, this should always give a warning. We could reduce the number of warnings to max 1 per toolbox.

MichaelStritt:
> > @jan-petr & @HenkMutsaerts: did some changes to reduce the crazy verbosity, I bet users would greatly complain about that. I also added a condition for matlab related toolboxes, because that leads to super annoying bugs otherwise.
> > Edit: please add the "how to test" section next time, we kind of agreed on this.
> 
>     * Thanks, but we have to remove the Matlab toolbox condition. There are users having e.g., SPM in their matlab toolbox, this is what we want to remove from paths. This is why I had a filesep in front of the regular expressions, so it wouldn't detect `/alst` but it would detect `/lst`
> 
>     * We also have to remove the `version` related regex, why is this there? Again, this won't cause a warning if you have multiple ExploreASL or toolbox versions per se. This will only cause a warning if you have initialized (i.e., added to the Matlab path) multiple versions; which is actually what we want to rule out.
> 
>     * I disagree with the verbose mode, or we have to default it to true. Users should not have initialized other spm versions, cat12 versions, or lst versions, this should always give a warning. We could reduce the number of warnings to max 1 per toolbox.
> 
>     * Note that this isn't a feature, but a bug. Users that have ExploreASL-specific toolboxes installed also in `Matlab/Toolbox` — such as Wibeke had — will get strange errors when running SPM functions inside ExploreASL, related to the version differences between our SPM version and the version installed outside of ExploreASL

@HenkMutsaerts & @jan-petr: For me all of this is asking for trouble. I got a ton of warnings and some errors afterwards just during coding, because Matlab was complaining about missing paths. I definitely wouldn't merge it like this. If a user decides to put spm inside the internal matlab directory, something is wrong on their side on not in ExploreASL.

HenkMutsaerts:
> > > @jan-petr & @HenkMutsaerts: did some changes to reduce the crazy verbosity, I bet users would greatly complain about that. I also added a condition for matlab related toolboxes, because that leads to super annoying bugs otherwise.
> > > Edit: please add the "how to test" section next time, we kind of agreed on this.
> > 
> > 
> > ```
> > * Thanks, but we have to remove the Matlab toolbox condition. There are users having e.g., SPM in their matlab toolbox, this is what we want to remove from paths. This is why I had a filesep in front of the regular expressions, so it wouldn't detect `/alst` but it would detect `/lst`
> > 
> > * We also have to remove the `version` related regex, why is this there? Again, this won't cause a warning if you have multiple ExploreASL or toolbox versions per se. This will only cause a warning if you have initialized (i.e., added to the Matlab path) multiple versions; which is actually what we want to rule out.
> > 
> > * I disagree with the verbose mode, or we have to default it to true. Users should not have initialized other spm versions, cat12 versions, or lst versions, this should always give a warning. We could reduce the number of warnings to max 1 per toolbox.
> > 
> > * Note that this isn't a feature, but a bug. Users that have ExploreASL-specific toolboxes installed also in `Matlab/Toolbox` — such as Wibeke had — will get strange errors when running SPM functions inside ExploreASL, related to the version differences between our SPM version and the version installed outside of ExploreASL
> > ```
> 
> @HenkMutsaerts & @jan-petr: For me all of this is asking for trouble. I got a ton of warnings and some errors afterwards just during coding, because Matlab was complaining about missing paths. I definitely wouldn't merge it like this. If a user decides to put spm inside the internal matlab directory, something is wrong on their side on not in ExploreASL.
> 
> Especially if multiple users share a license they get like 100 warnings every time they initialize. That's crazy. Maybe you don't see it on your side, but that's not tested properly.

OK then it needs to be made more stable, such that it will only throw away important paths. Or we can for now include only spm, that should be sufficient as CAT12 & LST are usually in spm paths, and fsl is dealt with separately. Let me add a change and can you test it then again Michael?

MichaelStritt:
> 
> 
> > > > @jan-petr & @HenkMutsaerts: did some changes to reduce the crazy verbosity, I bet users would greatly complain about that. I also added a condition for matlab related toolboxes, because that leads to super annoying bugs otherwise.
> > > > Edit: please add the "how to test" section next time, we kind of agreed on this.
> > > 
> > > 
> > > ```
> > > * Thanks, but we have to remove the Matlab toolbox condition. There are users having e.g., SPM in their matlab toolbox, this is what we want to remove from paths. This is why I had a filesep in front of the regular expressions, so it wouldn't detect `/alst` but it would detect `/lst`
> > > 
> > > * We also have to remove the `version` related regex, why is this there? Again, this won't cause a warning if you have multiple ExploreASL or toolbox versions per se. This will only cause a warning if you have initialized (i.e., added to the Matlab path) multiple versions; which is actually what we want to rule out.
> > > 
> > > * I disagree with the verbose mode, or we have to default it to true. Users should not have initialized other spm versions, cat12 versions, or lst versions, this should always give a warning. We could reduce the number of warnings to max 1 per toolbox.
> > > 
> > > * Note that this isn't a feature, but a bug. Users that have ExploreASL-specific toolboxes installed also in `Matlab/Toolbox` — such as Wibeke had — will get strange errors when running SPM functions inside ExploreASL, related to the version differences between our SPM version and the version installed outside of ExploreASL
> > > ```
> > 
> > 
> > @HenkMutsaerts & @jan-petr: For me all of this is asking for trouble. I got a ton of warnings and some errors afterwards just during coding, because Matlab was complaining about missing paths. I definitely wouldn't merge it like this. If a user decides to put spm inside the internal matlab directory, something is wrong on their side on not in ExploreASL.
> > Especially if multiple users share a license they get like 100 warnings every time they initialize. That's crazy. Maybe you don't see it on your side, but that's not tested properly.
> 
> OK then it needs to be made more stable, such that it will only throw away important paths. Or we can for now include only spm, that should be sufficient as CAT12 & LST are usually in spm paths, and fsl is dealt with separately. Let me add a change and can you test it then again Michael?

Sure. I guess we should just change our general method. Instead of removing everything besides ExploreASL, we should remove only the things that could be troublesome, so all paths besides the ExploreASL ones that include our internal toolboxes. Meaning we only throw out other xASL, spm, bids-matlab, cat, and lst related paths, right?

HenkMutsaerts:
> > > > > @jan-petr & @HenkMutsaerts: did some changes to reduce the crazy verbosity, I bet users would greatly complain about that. I also added a condition for matlab related toolboxes, because that leads to super annoying bugs otherwise.
> > > > > Edit: please add the "how to test" section next time, we kind of agreed on this.
> > > > 
> > > > 
> > > > ```
> > > > * Thanks, but we have to remove the Matlab toolbox condition. There are users having e.g., SPM in their matlab toolbox, this is what we want to remove from paths. This is why I had a filesep in front of the regular expressions, so it wouldn't detect `/alst` but it would detect `/lst`
> > > > 
> > > > * We also have to remove the `version` related regex, why is this there? Again, this won't cause a warning if you have multiple ExploreASL or toolbox versions per se. This will only cause a warning if you have initialized (i.e., added to the Matlab path) multiple versions; which is actually what we want to rule out.
> > > > 
> > > > * I disagree with the verbose mode, or we have to default it to true. Users should not have initialized other spm versions, cat12 versions, or lst versions, this should always give a warning. We could reduce the number of warnings to max 1 per toolbox.
> > > > 
> > > > * Note that this isn't a feature, but a bug. Users that have ExploreASL-specific toolboxes installed also in `Matlab/Toolbox` — such as Wibeke had — will get strange errors when running SPM functions inside ExploreASL, related to the version differences between our SPM version and the version installed outside of ExploreASL
> > > > ```
> > > 
> > > 
> > > @HenkMutsaerts & @jan-petr: For me all of this is asking for trouble. I got a ton of warnings and some errors afterwards just during coding, because Matlab was complaining about missing paths. I definitely wouldn't merge it like this. If a user decides to put spm inside the internal matlab directory, something is wrong on their side on not in ExploreASL.
> > > Especially if multiple users share a license they get like 100 warnings every time they initialize. That's crazy. Maybe you don't see it on your side, but that's not tested properly.
> > 
> > 
> > OK then it needs to be made more stable, such that it will only throw away important paths. Or we can for now include only spm, that should be sufficient as CAT12 & LST are usually in spm paths, and fsl is dealt with separately. Let me add a change and can you test it then again Michael?
> 
> Sure. I guess we should just change our general method. Instead of removing everything besides ExploreASL, we should remove only the things that could be troublesome, so all paths besides the ExploreASL ones that include our internal toolboxes. Meaning we only throw out other xASL, spm, bids-matlab, cat, and lst related paths, right?

But this was exactly what the script did. It searched for folders that contained these toolboxes and only removed those paths.

MichaelStritt (approved):

PR #922, opened 2021-11-15
Closes #913 Refactor xASL_im_ResampleLinearFair

### LINKED ISSUE #913: xASL_im_ResampleLinearFair: simplify code
### Description

We should do a small follow-up on 35feb2f & 28fea18 from #799, where we simplify the code. We can easily merge `xASL_im_ResampleLinearFair_1D_1/2/3` into a single sub-function, if we introduce the current dimension as an input parameter.

Idea: 

```matlab
%% 1-dimensional change n-th dimension
function output_res = xASL_im_ResampleLinearFair_1D(old_res,new_res,dim)

for k=1:size(new_res,1)
    
    startDim              =(k-1)* (size(old_res,dim) / size(new_res,dim));
    endDim                =k    * (size(old_res,dim) / size(new_res,dim));
    
    ...

end

end
```

### Requirements

### How to test

`xASL_im_ResampleLinearFair` is used in a few places. One of them is even in `
xASL_init_VisualizationSettings.m`. We should test at least one 1D example.

### Release notes

Refactored `xASL_im_ResampleLinearFair` code.

MichaelStritt:
### Linked issue

Check out the issue description in #913.

### How to test

If you run `ExploreASL` the function will already be used in the visualization settings e.g., but there are of course a lot of more refined test options. Loading a NIFTI and resampling it would be a good idea.


MichaelStritt:
### Test

image

jan-petr (changes requested):
First - this will work for 1D and 3D changes, but resampling in 2D will be skipped. I fixed that. But this function simply has to be replaced by interp3+ngrid instead of having this enormous thing...

jan-petr:
% Define test data
newsize = [4 3 3];
imagein = [1 1;2 2];
imagein(:,:,2) = [3 1;5 2]

% Define the new grid
[Xq,Yq,Zq] = ndgrid(1:newsize(1),1:newsize(2),1:newsize(3));

% Adapt the new grid to perfectly match the original grip - i.e. stretch or squeeze to the original grid coordinates
Xq = (Xq-1)/(newsize(1)-1)*(size(imagein,1)-1) + 1;
Yq = (Yq-1)/(newsize(2)-1)*(size(imagein,2)-1) + 1;
Zq = (Zq-1)/(newsize(3)-1)*(size(imagein,3)-1) + 1;
outputim = interpn(imagein,Xq,Yq,Zq)

MichaelStritt:
@jan-petr 

> First - this will work for 1D and 3D changes, but resampling in 2D will be skipped. I fixed that. But this function simply has to be replaced by interp3+ngrid instead of having this enormous thing...

Correct, but that didn't work before either, right?

ExploreASL_Initialize.m
```diff
@@ -190,7 +190,7 @@ function xASL_init_PrintLogo()
     '                    ## |                                                                         \n'...
```
HenkMutsaerts, line 190: The changes here in `ExploreASL_Initialize` have nothing to do with `xASL_im_ResampleLinearFair` right, I would move them to another issue

Functions/xASL_im_ResampleLinearFair.m
```diff
@@ -1,502 +1,89 @@
...
-%               {{NB:}} all dimensions of new_res should have equal size
+% DESCRIPTION:  Downsample or upsample an image from its old to a new resolution.
 %
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
 % EXAMPLE:      image = xASL_io_Nifti2Im('.\M0.nii.gz');
-%               output = xASL_im_ResampleLinearFair(image,[2,2,2]);
+%               output = xASL_im_ResampleLinearFair(image, [2,2,2]);
 %
 % __________________________________
 % Copyright 2015-2021 ExploreASL
```
HenkMutsaerts, line 19: Can you please test with interpolating several image matrices that you get **exactly identical** output, with upscaling and downscaling? I created this function as interpolation methods as SPM etc did degrade the quality of scans by interpolation artifacts, or changing the actual resolution. This was in the context of doing the motion correction in the higher resolution space, where we don't want interpolation artifacts (so no interpolation! just regridding). So please only change my work here if the shortening really gives **identical** results in all cases.

Testing/UnitTests/xASL_ut_function_xASL_im_ResampleLinearFair.m
```diff
@@ -0,0 +1,177 @@
+function UnitTest = xASL_ut_function_xASL_im_ResampleLinearFair(TestRepository)
```
HenkMutsaerts, line 1: Here we should test that when we make the voxel-size 3 times as small, the sum, mean, median, min, max, etc remain identical to the original voxel-size. Then we make the voxel-size 3 times as large again, and do the same test. Even negligible or small deviations should not be allowed, as that was the whole purpose of the fair linear interpolation.

HenkMutsaerts (changes requested):

HenkMutsaerts:
> This will do exactly the same job and probably way faster. You will only need to test if the image is really 3D etc.
> 
> ```
> % Define test data
> newsize = [4 3 3];
> imagein = [1 1;2 2];
> imagein(:,:,2) = [3 1;5 2]
> ```
> 
>  ```
> % Define the new grid
> [Xq,Yq,Zq] = ndgrid(1:newsize(1),1:newsize(2),1:newsize(3));
> ```
> 
>  ```
> % Adapt the new grid to perfectly match the original grip - i.e. stretch or squeeze to the original grid coordinates
> Xq = (Xq-1)/(newsize(1)-1)*(size(imagein,1)-1) + 1;
> Yq = (Yq-1)/(newsize(2)-1)*(size(imagein,2)-1) + 1;
> Zq = (Zq-1)/(newsize(3)-1)*(size(imagein,3)-1) + 1;
> outputim = interpn(imagein,Xq,Yq,Zq)
> ```

No, it needs to be tested that the output matrices are identical to what my function did!

Testing/UnitTests/xASL_ut_function_xASL_im_ResampleLinearFair.m
HenkMutsaerts: Let's keep my function as subfunction in the unit testing, because this is what we should test, that it keeps the regridding "fair" by giving the identical result as my "more manual" function did.

ExploreASL_Initialize.m
MichaelStritt: Can't you just review them here? It was really only cosmetics and not worth a separate branch.

Functions/xASL_im_ResampleLinearFair.m
MichaelStritt: Improved unit test: 912f792

Testing/UnitTests/xASL_ut_function_xASL_im_ResampleLinearFair.m
MichaelStritt: Option to use the previous method: 7f3d4c5

jan-petr:
> @jan-petr
> 
> > First - this will work for 1D and 3D changes, but resampling in 2D will be skipped. I fixed that. But this function simply has to be replaced by interp3+ngrid instead of having this enormous thing...
> 
> Correct, but that didn't work before either, right?

Yes - I tried to write that. That was an error that was there indeed previously :) But so important that I had to fix it :)

jan-petr (changes requested):
We should be careful about not dividing by zero when one of the dimension is 1. So mainly input size or new size [X Y 1] should flip it to a 2D mode instead of a 3D mode. Inner dimensions being 1 is a problem as well, but no nice solution - perhaps there we simply calculate the Xq like a mean coordinate. Do you know what I mean?!

jan-petr:
@MichaelStritt Have you tested if there's a major difference between the old and new? If yes, we can try to find out what it was - some boundary conditions or special cases etc might have a different output despite that the thing in general should be exactly the same.

MichaelStritt:
> We should be careful about not dividing by zero when one of the dimension is 1. So mainly input size or new size [X Y 1] should flip it to a 2D mode instead of a 3D mode. Inner dimensions being 1 is a problem as well, but no nice solution - perhaps there we simply calculate the Xq like a mean coordinate. Do you know what I mean?!

image

Testing/UnitTests/xASL_ut_function_xASL_im_ResampleLinearFair.m
HenkMutsaerts: Thanks, though I wouldn't call it "old method". The methods are not newer or older, the difference is the matrix operation or the iteration over lines/voxels. Probably better to call them fast vs slow.

Functions/xASL_im_ResampleLinearFair.m
HenkMutsaerts: Does the updated unit test now resample up and down with the old and new technique, and compare both?
I actually thought that the old/slow technique would be the ideal unit test of the newer function Jan wrote.

Testing/UnitTests/xASL_ut_function_xASL_im_ResampleLinearFair.m
HenkMutsaerts: Yes, though if Jan's method actually works in all cases, then we could keep my function in the unit test only.
Though I wonder if Jan's method is Matlab-version dependent. The unit test with the old function would find this.

Testing/UnitTests/xASL_ut_function_xASL_im_ResampleLinearFair.m
```diff
@@ -171,7 +171,7 @@
 %% Test run 5
 
 % Give your individual subtest a name
-UnitTest.tests(5).testname = 'Resample 3D (up, down, etc.)';
+UnitTest.tests(5).testname = 'Resample 3D (up, down, old, etc.)';
```
HenkMutsaerts, line 174: No this is wrong. All unit tests should be using the old/slow version as reference I would say. So you run up, and then again down (not separately but in serial, taking the output from up and use it again for down), both compare them with the old/slow version, and then we're fine.

HenkMutsaerts (changes requested):

HenkMutsaerts:
Apologies for the perhaps confusing review, but the GUI jumps from `files changed` to `commit` to `conversion` so I just put everything in where I was, hope this works.

Testing/UnitTests/xASL_ut_function_xASL_im_ResampleLinearFair.m
MichaelStritt: It does some tri-linear resampling and because of that there are some super minimal differences. I wouldn't make the unit test even more complex. I personally like it the way it is right now.

Functions/xASL_im_ResampleLinearFair.m
MichaelStritt: It's not really common to test one script functionality based on another one... This is getting so theoretical. I don't see any benefits of improving this right now. I just wanted to clean up some code and this whole issue got out of hand and I don't want to spend even more time on this ATM.

Testing/UnitTests/xASL_ut_function_xASL_im_ResampleLinearFair.m
MichaelStritt: As said above, I don't think we should do that and I don't think we should spend even more time on this. Originally I only wanted to clean up some code, then we suddenly started to implement new features because a lot of newer users stumbled upon this function and were annoyed by how slow it is. Now everything's kind of convoluted.

jan-petr: Let's shortly discuss this on Monday. But this function is currently only used for VTK export and for resampling 2 colormaps. We should test if old and new are equal or similar and then throw away the old function. There is absolutely no reason to keep it since it is not used in any critical imaging code. Kill your darlings.

Also, we could rename `ResampleLinearFair` to `ResampleLinear`.

Testing/UnitTests/xASL_ut_function_xASL_im_ResampleLinearFair.m
jan-petr: Super simple function - no reason this will differ across Matlab versions. I would keep the new version only. Especially since we only use it in one case of resampling a colormap.

Testing/UnitTests/xASL_ut_function_xASL_im_ResampleLinearFair.m
MichaelStritt: > Also, we could rename `ResampleLinearFair` to `ResampleLinear`

Agree 👍

Testing/UnitTests/xASL_ut_function_xASL_im_ResampleLinearFair.m
MichaelStritt: > Super simple function - no reason this will differ across Matlab versions. I would keep the new version only. Especially since we only use it in one case of resampling a colormap.

Agree 👍

HenkMutsaerts:
I discussed this with Jan. The old function `ResampleLinearFair` should remain untouched, and be moved to the Development folder, as we may use this later to improve motion correction in high resolution.

Good idea then to create this new function called `ResampleLinear` and use this instead of `ResampleLinearFair`.

Is that OK for you?

HenkMutsaerts (changes requested):
I discussed this with Jan. 

We revert the commits on `xASL_im_ResampleLinearFair`, and `git mv` this old function to the development folder, as we may use this later to improve motion correction in high resolution. (it might be faster to do this in a new branch to keep history clean)

Good idea then to create this new function called ResampleLinear and use this instead of ResampleLinearFair.

Again, nice work!

MichaelStritt:
> I discussed this with Jan.
> 
> We revert the commits on `xASL_im_ResampleLinearFair`, and `git mv` this old function to the development folder, as we may use this later to improve motion correction in high resolution. (it might be faster to do this in a new branch to keep history clean)
> 
> Good idea then to create this new function called ResampleLinear and use this instead of ResampleLinearFair.
> 
> Again, nice work!

Sounds good, I cleaned it up. Should have nice individual commits for each file now and it's less convoluted.

jan-petr (approved):
OK.

PR #933, opened 2021-11-20
Closes #929 Improved warning

### LINKED ISSUE #929: Import continues with incorrect input
### Description

If `sourceStructure`'s `folderHierarchy` is wrongly defined, this gives several warnings and errors but unclear to the user that the sourceStructure was wrong or no files were found. The problem is that it doesn't crash when no files are found, but continues, and crashes with confusing other warnings.

image

### Requirements

Solution is probably to let the import module crash if no files were found, i.e. replacing the warning on `xASL_imp_ReadSourceData` (line 47) by an error.

### How to test

Run `ExploreASL('path2datasetRoot',[1 0 0 0], 0);` with incorrect `sourceStructure` or no image data in `/datasetRoot/sourcedata/'.

### Release notes

MichaelStritt:
### Linked issue

#929

### DIFF
--- a/Modules/SubModule_Import/xASL_imp_DetermineSubjectStructure.m
+++ b/Modules/SubModule_Import/xASL_imp_DetermineSubjectStructure.m
@@ -62,7 +62,6 @@ function x = xASL_imp_DetermineSubjectStructure(x)
     end
     
     % SESSIONS DUMMY
-    % fprintf(2,'Currently x.SESSIONS is not supported for the import...\n');
     x.SESSIONS = {''};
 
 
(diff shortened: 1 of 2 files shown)


HenkMutsaerts (approved):

HenkMutsaerts (changes requested):
Sorry I approved too fast :)
I would add some more feedback to the user. "No matching files" could mean anything to the user, so perhaps 1 or 2 lines explaining to the user what could be wrong (like you say in your comment, which the user that doesn't dig into the code, won't see).

Thinking about it, it would be nice if Matlab had some sort of comment/fprintf linking in case of errors :)

MichaelStritt:
@HenkMutsaerts: Please check if you're happy with the new version. It includes a link to the documentation, because otherwise we would have to explain how the `sourceStructure.json` works there and that would be overkill.

image

Issue #930, opened 2021-11-20
Move log files to log directory


HenkMutsaerts:
### Description

`derivatives/ExploreASL/log` contains only import log files, this is potentially confusing

### Requirements

Can we move the log files to the log directory?

### How to test

Optional: insert description about how to test the code changes here

### Release notes

This is a follow-up for import revamp #799, so in the release, this should be added to this issue.

MichaelStritt:
The main reason was that we have an import log for each subject in derivatives, but there are no dedicated subject directories in derivatives (only sub_ses).

I'd suggest to start working on the derivatives structure, so that we are closer to bids. It should become sub_<sub-id>/ses_<ses-id> instead of sub_<sub-id>_<ses-number>, right?

We can easily move the other logs there too, but then we have to give them more specific names (sub_<sub-id>_ses_<ses-id>_xASL_module_<module>.log e.g.)

For me this would be moving forward instead of just renaming temporarily.

PR #935, opened 2021-11-20
#930 xASL_Iteration: move logs

MichaelStritt:
### Linked issue

#930

### DIFF
--- a/Functions/xASL_Iteration.m
+++ b/Functions/xASL_Iteration.m
@@ -110,7 +110,7 @@ function [bAborted, xOut] = xASL_Iteration(x, moduleName, dryRun, stopAfterError
     if ~isempty(regexp(ModName, '(DARTEL|Population|Analyze)', 'once'))
         dbSettings.diaryFile = ['<ROOT>/log/' moduleName '.log'];
     elseif ~isempty(regexp(ModName,'(ASL|func|dwi)', 'once'))
-        dbSettings.diaryFile = ['<SESSIONDIR>/' moduleName '.log'];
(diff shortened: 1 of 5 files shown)


MichaelStritt:
# Test

## derivatives

image

### lock

image

### log

image

MichaelStritt:
## derivatives

image

### lock

image

### log

image

HenkMutsaerts (changes requested):
Only now I see the figure, very nice!

Can we at the same time make sure that the naming is the same throughout. I wouldn't add the 'sub-' as this should have been done in the conversion. I would just use <SUBJECT>, without changing them or interpreting them. 

Note that `xASL_module_ASL_012S4093_2` should be `xASL_module_ASL_012S4093_2_ASL_1` or `xASL_module_ASL_012S4093_2_ASL_2`. So for session modules it should be <SUBJECT>_<SESSION>.

In the future, we could move the file reports (Missing*.csv etc) there as well, but this will confuse the users. We shouldn't have too many small changes from version to version, or the users will get crazy. Let's add this to our wishlist for version 2.0 :)

MichaelStritt:
@HenkMutsaerts: Should work with c7051c9. If a user defines `sub-[...]` as the subject id instead of `[...]` we would have `sub-sub-[...]` though. There's no way around it though, since you can't use regular expressions for the symbolic fields within xASL_Iteration. You could implement that later on, but that wouldn't be too clean either. I'd leave it as it is right now.

jan-petr (approved):
OK.

PR #937, opened 2021-11-20
Closes #936 Check directories for reruns/crashed imports

### LINKED ISSUE #936: Import module incorrectly says "has been run before"
### Description

If the import module has been run but crashed, it will still not allow rerunning — unless the user removes the `derivatives` folder. I thought the import module would now modularly check each subject? Does it only check if the subject folder exists? Or does it also check if the folder contains valid NIfTI files etc?

It should use the lock files appropriately, if import was run before, but crashed, these lock files won't exist, so should rerun. It should not be dependent on existence of folders.

### Requirements

Optional: add bug requirements here

### How to test

Put an error in different places of the import modules, the modularity should work in each case, and be independent of existence of folders etc. Also try putting an error at the beginning of the import module, this should also work.

### Release notes

MichaelStritt:
### Linked issue

Closes #936

### DIFF
--- a/Functions/xASL_init_DetermineRequiredPaths.m
+++ b/Functions/xASL_init_DetermineRequiredPaths.m
@@ -34,12 +34,21 @@ function [x] = xASL_init_DetermineRequiredPaths(x)
     % BIDS DatasetRoot directory
     x.dir.DatasetRoot = x.opts.DatasetRoot;
     x.opts.dataParType = 'directory';
+    
+    % Set other basic BIDS directories
+    x.dir.SourceData = fullfile(x.opts.DatasetRoot,'sourcedata');
+    x.dir.RawData = fullfile(x.opts.DatasetRoot,'rawdata');
+    x.dir.Derivatives = fullfile(x.opts.DatasetRoot,'derivatives');
+    x.dir.xASLDerivatives = fullfile(x.dir.Derivatives,'ExploreASL');
+    
     % Search for descriptive JSON files
     fileListSourceStructure = xASL_adm_GetFileList(x.dir.DatasetRoot, '(?i)^sourcestructure.*\.json$');
     fileListStudyPar = xASL_adm_GetFileList(x.dir.DatasetRoot, '(?i)^studypar.*\.json$');
     fileListDataDescription = xASL_adm_GetFileList(fullfile(x.dir.DatasetRoot, 'rawdata'), '(?i)^dataset_description\.json$');
+    
     % First try the derivatives folder
     fileListDataPar = xASL_adm_GetFileList(fullfile(x.dir.DatasetRoot, 'derivatives', 'ExploreASL'), '(?i)^datapar.*\.json$');
+    
     % Check for x.D.ROOT
     if exist(fullfile(x.dir.DatasetRoot, 'derivatives', 'ExploreASL'),'dir')
         x.D.ROOT = fullfile(x.dir.DatasetRoot, 'derivatives', 'ExploreASL');
@@ -48,6 +57,7 @@ function [x] = xASL_init_DetermineRequiredPaths(x)
         % Derivatives maybe does not exist already, we'll try study root
         fileListDataPar = xASL_adm_GetFileList(x.dir.DatasetRoot, '(?i)^datapar.*\.json$');
     end
+    
     % Assign fields
     if ~isempty(fileListSourceStructure)
 		if length(fileListSourceStructure) > 1
--- a/Modules/SubModule_Import/xASL_bids_BIDS2Legacy.m
+++ b/Modules/SubModule_Import/xASL_bids_BIDS2Legacy.m
@@ -1,10 +1,9 @@
-function [x] = xASL_bids_BIDS2Legacy(pathStudy, x, bOverwrite)
+function [x] = xASL_bids_BIDS2Legacy(x, bOverwrite)
 %xASL_bids_BIDS2Legacy Convert BIDS rawdata to ExploreASL legacy format
 %
-% FORMAT: [x] = xASL_bids_BIDS2Legacy(pathStudy, x[, bOverwrite])
+% FORMAT: [x] = xASL_bids_BIDS2Legacy(x[, bOverwrite])
 % 
 % INPUT:
-%   pathStudy  - path to the study folder containing the BIDS data in rawdata subfolder (REQUIRED)
 %   x          - ExploreASL x structure (REQUIRED, STRUCT)
 %   bOverwrite - boolean, true for overwriting files (OPTIONAL, DEFAULT = true)
 %   
@@ -33,7 +32,7 @@ function [x] = xASL_bids_BIDS2Legacy(pathStudy, x, bOverwrite)
 % 6. Clean up
 % 
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
-% EXAMPLE: [x] = xASL_bids_BIDS2Legacy(pathStudy, x, bOverwrite);
+% EXAMPLE: [x] = xASL_bids_BIDS2Legacy(x, bOverwrite);
 % __________________________________
 % Copyright 2015-2021 ExploreASL
 
@@ -41,28 +40,28 @@ function [x] = xASL_bids_BIDS2Legacy(pathStudy, x, bOverwrite)
 %% 0. Admin
 
 % Verify the input parameters
-if nargin<1 || isempty(pathStudy)
-	error('pathStudy is a required parameter.');
+if isempty(x.dir.DatasetRoot)
+	error('x.dir.DatasetRoot is a required parameter.');
 end
 
-if nargin<3 || isempty(bOverwrite)
+if nargin<2 || isempty(bOverwrite)
     bOverwrite = 1;
 end
 
 % Verify that the rawdata subfolder exists
-if ~exist(fullfile(pathStudy,'rawdata'),'dir')
+if ~exist(fullfile(x.dir.DatasetRoot,'rawdata'),'dir')
     warning('Invalid folder selected, not containing rawdata folder');
     return;
 end
 
 % Creates the derivatives directory
-pathLegacy = fullfile(pathStudy, 'derivatives', 'ExploreASL');
-if exist(pathLegacy, 'dir') && bOverwrite
+x.dir.pathLegacy = fullfile(x.dir.DatasetRoot, 'derivatives', 'ExploreASL');
+if exist(x.dir.pathLegacy, 'dir') && bOverwrite
     fprintf('The derivatives directory already exists, overwriting...\n');
-elseif exist(pathLegacy, 'dir')
-    fprintf('%s\n', [pathLegacy ' exists, merging']);
+elseif exist(x.dir.pathLegacy, 'dir')
+    fprintf('%s\n', [x.dir.pathLegacy ' exists, merging']);
 else
-    xASL_adm_CreateDir(pathLegacy);
+    xASL_adm_CreateDir(x.dir.pathLegacy);
 end
 
 % Loads the configuration for file renaming from the BIDS configuration file
@@ -70,11 +69,11 @@ bidsPar = xASL_bids_Config();
 
 
 %% 1. Parse a folder using bids-matlab
-BIDS = bids.layout(fullfile(pathStudy,'rawdata'));
+BIDS = bids.layout(fullfile(x.dir.DatasetRoot,'rawdata'));
 nSubjects = numel(BIDS.subjectName);
 nVisits = numel(BIDS.sessionName); % this is called sessions in BIDS
 % we use this below to see if the legacy subjectname gets _1 as visit suffix or not
-[~, studyName] = fileparts(pathStudy);
+[~, studyName] = fileparts(x.dir.DatasetRoot);
 fprintf('Converting from BIDS to Legacy: %s   \n', studyName);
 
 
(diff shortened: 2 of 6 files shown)


MichaelStritt:
@jan-petr & @HenkMutsaerts: I don't really know how this is supposed to work in some cases. For NII2BIDS it is somewhat easy, because we can just check if the rawdata ASL or T1 folder already exists and delete it for the reruns or when the pipeline crashed. For BIDS2Legacy this is problematic, because we do the import on subject level and as I said before, as long as we don't have subject directories in legacy, this is kind of impossible. For DCM2NII I know there was a previous option with bOverwrite and so on. Maybe we can use that. Other than that I don't know how I should check for existing directories if one directory can be the top of the other one (T1 on top of ASL e.g.) or if we don't exactly know all of the output scans of DCM2NII beforehand. Nevertheless, I think the implemented code is an improvement for NII2BIDS and there are some simplifications for BIDS2Legacy.

Modules/SubModule_Import/xASL_bids_BIDS2Legacy.m
```diff
@@ -97,21 +96,23 @@
 
         %% 3. Define Session
         iVisit = find(strcmp(BIDS.sessionName, SessionID));
-        % remove iteration for iVisit=1 % iterate visit/session in this "BIDS.subjects" (always 1 session per BIDS.subjects)
+        % Remove iteration for iVisit = 1 -> iterate visit/session in this "BIDS.subjects" (always 1 session per BIDS.subjects)
         % ExploreASL uses visit as a number (e.g. _1 _2 _3 etc)
         if nVisits==1
-            pathLegacy_SubjectVisit = fullfile(pathLegacy, SubjectID);
+            pathLegacy_SubjectVisit = fullfile(x.dir.pathLegacy, SubjectID);
```
HenkMutsaerts, line 102: If we have `x.dir.DatasetRoot`, would make it more consistent to also have `x.dir.raw` and `x.dir.derivatives`? `x.dir.pathLegacy` is a new name, not clear to the new reader, but pathDerivatives or something similar is clear (and this is what we mean here right?

Modules/SubModule_Import/xASL_imp_BIDS2Legacy.m
```diff
@@ -50,39 +50,37 @@
...
-		if length(fListDataPar) > 1
-			warning('Multiple dataPar*.jsons exist. Using the first: %s\n',fListDataPar{1});
-		end
-		% Fill the dataPars with the provided parameters
-		x.dataPar = spm_jsonread(fListDataPar{1});
-		x = xASL_bids_BIDS2Legacy(localDatasetRoot, x, 1);
+	fListDataPar = xASL_adm_GetFileList(x.dir.DatasetRoot,'(?i)(^dataPar.*\.json$)', 'FPList', [], 0);
+    if length(fListDataPar) < 1
+        % Fill the dataPars with default parameters
+        fprintf('There is no dataPar.json file in the study root directory. Default settings will be used...\n');
```
HenkMutsaerts, line 73: What happens if, at the start of the whole import, only `/DatasetRoot/derivatives/ExploreASL/dataPar.json` exists? That would be the first thing a user would try right? At least this shouldn't be overwritten. Perhaps we can detect this, copy to the datasetRoot, and issue a warning that this was found and will be used.

Modules/SubModule_Import/xASL_imp_DCM2NII_ConvertScan.m
```diff
@@ -88,7 +88,8 @@
             bPutInSessionFolder = false;
     end
 
-    if ~isempty(strfind(char(imPar.folderHierarchy(end)),'PAR'))
+    % Check for old PARREC data
```
HenkMutsaerts, line 91: I just removed this in #938

HenkMutsaerts (changes requested):
For the rerun, I would:

* base rerunning and overwriting or skipping per subject only on the existence of the lock files, identical to how this is managed in the processing modules. So if a lock/status file doesn't exist, this means that the current subject wasn't correctly run, and the previous rawdata should be deleted
* ensure that the lock/status files are only created after successful import of a single subject
* always overwrite existing rawdata or derivatives for the current subject (since overwriting/skipping is managed by the lock/status files)

So instead of checking folders, files, what if they exist partly, etc, etc, you only rely on the lock files (unless you create a `bCompletelyRerunImport` parameter of course, in that case you delete all lock files). Makes the code/program also much more transparent.

MichaelStritt:
> So instead of checking folders, files, what if they exist partly, etc, etc, you only rely on the lock files

And that doesn't work on your side? If one of the import steps crashes for me, then the lock file is not created and on a rerun it reruns the sub-module. I think we should make it cleaner and remove some files/folders for the import reruns, because I'm not sure if every xASL_Move command or the used tools work with overwrite, but other than that this part should be fine. Please actually describe the defect/failure behavior.

Modules/SubModule_Import/xASL_imp_DCM2NII_ConvertScan.m
MichaelStritt: Okay? Then it will be a part of the conflict resolving, I guess?

Modules/SubModule_Import/xASL_bids_BIDS2Legacy.m
jan-petr: Very good comment. `DatasetRoot` is good for the root to all directories. `RawData`, `SourceData`, `Derivatives` might be used as well. So while we are at it, we can probably change `pathLegacy` to `Derivatives`, but it can't be just derivatives - that would preclude further extensions and `xASLDerivatives` might be better - just to show that we don't mean root derivatives folder. In general, regarding use of `Legacy` - apart from this case, we don't need to rename it that much, because all Legacy will be later renamed to fit better the BIDS derivatives format.

Modules/SubModule_Import/xASL_imp_DCM2NII_ConvertScan.m
jan-petr: Yes, that's going to be part of resolving indeed. BUT, in #938, Henk did an equivalent fix as you did here and moved the code elsewhere. It is thus easier to restore this (with a new commit) to the previous state - it will save us some hassle with the conflict solving.

jan-petr (changes requested):
Minor comments only.

Modules/SubModule_Import/xASL_imp_DCM2NII_ConvertScan.m
MichaelStritt: 3e79423 ?

MichaelStritt: Reverted 👍

Modules/SubModule_Import/xASL_bids_BIDS2Legacy.m
MichaelStritt: Done

Modules/SubModule_Import/xASL_imp_BIDS2Legacy.m
MichaelStritt: The workflow already checks for dataPar.json files in derivatives. I added a print statement for you 2fc744d

MichaelStritt:
Still think we should implement a sub/ses/run structure in the xASL derivatives and optimize the temp structure, so that reruns can be done in an easier way by removing pre-existing data from pipeline crashes etc.

HenkMutsaerts:
> @jan-petr & @HenkMutsaerts: I don't really know how this is supposed to work in some cases. For NII2BIDS it is somewhat easy, because we can just check if the rawdata ASL or T1 folder already exists and delete it for the reruns or when the pipeline crashed. For BIDS2Legacy this is problematic, because we do the import on subject level and as I said before, as long as we don't have subject directories in legacy, this is kind of impossible. For DCM2NII I know there was a previous option with bOverwrite and so on. Maybe we can use that. Other than that I don't know how I should check for existing directories if one directory can be the top of the other one (T1 on top of ASL e.g.) or if we don't exactly know all of the output scans of DCM2NII beforehand. Nevertheless, I think the implemented code is an improvement for NII2BIDS and there are some simplifications for BIDS2Legacy.

I would use the existence of lock files, so we don't have to check folders, because that is arbitrary. `BIDS2Legacy` is much faster than the other import submodules, so I wouldn't worry too much if that needs rerunning; plus this needs to be moved to the processing part anyway (and we will add subject directories in legacy as you say in the future).

The import should always overwrite (so `bOverwrite` is always true), and if a lock file exists, importing of a subject is skipped. That should work right?

HenkMutsaerts:
> > So instead of checking folders, files, what if they exist partly, etc, etc, you only rely on the lock files
> 
> And that doesn't work on your side? If one of the import steps crashes for me, then the lock file is not created and on a rerun it reruns the sub-module. I think we should make it cleaner and remove some files/folders for the import reruns, because I'm not sure if every xASL_Move command or the used tools work with overwrite, but other than that this part should be fine. Please actually describe the defect/failure behavior.

OK if this works for you, then this is fine. Your reasoning is correct here. I think my point here was that the import incorrectly said "has been run before". So this was an error on the beginning, not in the iteration.

HenkMutsaerts:
> Still think we should implement a sub/ses/run structure in the xASL derivatives and optimize the temp structure, so that reruns can be done in an easier way by removing pre-existing data from pipeline crashes etc.

Yep good point for version 2.0

HenkMutsaerts:
PS: why are we discussing here, shouldn't we discuss in the issue?

Functions/xASL_init_DetermineRequiredPaths.m
```diff
@@ -34,12 +34,21 @@
     % BIDS DatasetRoot directory
     x.dir.DatasetRoot = x.opts.DatasetRoot;
     x.opts.dataParType = 'directory';
+    
+    % Set other basic BIDS directories
+    x.dir.SourceData = fullfile(x.opts.DatasetRoot,'sourcedata');
```
HenkMutsaerts, line 39: Nice to have all in `x.dir`. I don't understand the difference between `x.opts` and `x.dir` here. Is x.opt necessary as a separate input options field? Don't all input options fall under some other field?

Modules/SubModule_Import/xASL_imp_NII2BIDS_Subject.m
```diff
@@ -34,6 +34,11 @@
     subjectDirectory = fullfile(imPar.BidsRoot,['sub-' bidsLabel.subject]);
     if ~xASL_exist(subjectDirectory,'dir')
```
HenkMutsaerts, line 35: You can simplify this by moving `xASL_adm_CreateDir` outside, and below, the if statement.
`xASL_adm_CreateDir` always checks if a folder exists before creating it.

Modules/SubModule_Import/xASL_imp_DCM2NII_ConvertScan.m
```diff
@@ -149,6 +149,15 @@
     end
 
     %% 4.6 Start the conversion if this scan should not be skipped
+
```
HenkMutsaerts, line 152: Indeed, if everything is done on the subject level, then everything should be managed at the subject level. So we either delete everything for a single subject and rerun everything for this subject, or we don't at all.

Modules/SubModule_Import/xASL_imp_BIDS2Legacy.m
HenkMutsaerts: I would solve this differently. `dataPar.json` contains settings for processing, and you recommend the user to put this in `/datasetRoot`, and then at the end of the `BIDS2legacy` submodule, this `dataPar.json` will be merged with the new fields that are generated automatically, if these fields didn't exist (e.g., `x.Quality`, `x.SUBJECTS` etc). At least, this should happen!

My point here was that, if the user created a `/datasetRoot/derivatives/ExploreASL/dataPar.json` instead of `/datasetRoot/dataPar.json`, you move the `/datasetRoot/derivatives/ExploreASL/dataPar.json` to `/datasetRoot/dataPar.json`, so that in the end of the `BIDS2legacy` submodule, they will be merged appropriately.

If the user reruns ExploreASL and forgets to delete the previous `/datasetRoot/derivatives/ExploreASL/dataPar.json`, this is the user's fault (or we could warn the user at the beginning of all modules everything, if no lock files exist yet, if you want).

Modules/SubModule_Import/xASL_bids_BIDS2Legacy.m
```diff
@@ -33,48 +32,52 @@
...
 
 % Verify that the rawdata subfolder exists
-if ~exist(fullfile(pathStudy,'rawdata'),'dir')
+if ~exist(fullfile(x.dir.DatasetRoot,'rawdata'),'dir')
     warning('Invalid folder selected, not containing rawdata folder');
     return;
 end
 
+% Check derivatives of ExploreASL
+if ~isfield(x.dir,'xASLDerivatives')
```
HenkMutsaerts, line 58: Note that nobody knows what "xASL" means, that is just an abbreviation that we use internally. Best to replace this by "ExploreASL derivatives folder" or something that doesn't lead to the user asking you what you meant ;)

HenkMutsaerts (changes requested):
Some suggestions, hope this helps you also understanding this from a user perspective, which can be very useful in software development!

Functions/xASL_init_DetermineRequiredPaths.m
MichaelStritt: Don't think this is the right issue/PR or the right time to restructure this again.

Modules/SubModule_Import/xASL_imp_NII2BIDS_Subject.m
MichaelStritt: 10deab0

Modules/SubModule_Import/xASL_bids_BIDS2Legacy.m
MichaelStritt: This should only occur if it really crashes. In that case it doesn't make sense for the user to know that one x field which isn't called "ExploreASL derivatives folder" is missing. 

We should differentiate between warnings for users and warnings for developers.

PR #939, opened 2021-11-21
Closes Bug #938 redundant dcm2nii behavior

### LINKED ISSUE #938: Redundant dcm2nii_version behavior
### Description

1. `dcm2nii_version` should be managed at one location, for correct switching of versions based on e.g., PAR/REC.
2. `dcm2nii_version` should not try to detect PAR/REC from sourcedata.json information, but only if explicitly called.
3. dcm2nii2011 - needs to be fixed as it doesn't read the correct parameters config parameters from .ini file.

### Requirements

So I removed it from `xASL_imp_ConvertScan`, `xASL_imp_TokenBackwardsCompatibility`
and kept this in `xASL_imp_CheckImportSettings` only, and I correct the defaulting behavior.

### Release notes

dcm2nii handling done in one place and corrected for PAR/REC.

HenkMutsaerts:
### Linked issue

Closes #938

### How to test

### Comments

### DIFF
--- a/Functions/xASL_io_dcm2nii.m
+++ b/Functions/xASL_io_dcm2nii.m
@@ -25,8 +25,8 @@ function [niifiles, ScanNameOut, usedinput, msg] = xASL_io_dcm2nii(inpath, destd
 % 2. Parse parameters
 % 3. Locate dcm2nii executable
 % 4. Set default arguments dcm2nii
-% 5. Set dcm2niiX initialization loading
-% 6. Check if we are reading a DICOM folder
+% 5. Check if we are reading a DICOM folder
+% 6. Set dcm2niiX initialization loading
 % 7. Check for existing targets
 % 8. Create temporary subfolder for converting
 % 9. Run dcm2nii and move files to final destination using specified series name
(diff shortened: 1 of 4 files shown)


jan-petr (approved):
Looks good to me. Michael should approve as well and check if there won't be any indirect issues with moving the lines (I don't foresee anything).

jan-petr:
Many things need to be changed to make this work for PAR-REC again.

1. Setting to version 2011 instead of 2019 cannot be done in settings because we don't know that we will have PAR/RECs - this doesn't need to be said in the sourcestructure.json. Maybe you do say it sometimes, but that's not a rule. So this decision has to be made before calling dcm2nii for each file.
2. Inside xASL_io_dcm2nii, we first loaded option file based on the fact if we are treating PAR/REC and only then we run the code to recognize if we are handling PAR/REC. So this code had to be reversed.

jan-petr:
3. The PARREC flavor is still not read correctly - it crashes on copying the two generated NIFTI files.

MichaelStritt:
Hm... cool that we don't check in three places of the workflow now, but if it doesn't work at all right now, we should definitely work on the actual bugfix. It also seems a bit risky just to check if file or folder is called PAR. We should probably restrict it to bMatchDirectories=false, so that it only works for DICOM files starting with PAR?

MichaelStritt:
The nicest solution would probably be if there was a specific DICOM tag for those datasets.

jan-petr (approved):
All good.

HenkMutsaerts (commented):
Looks good

PR #949, opened 2021-11-26
Closes #948 Update x.SUBJECT/S and x.overview

### LINKED ISSUE #948: Update x.SUBJECT/S and x.overview for subjects with illegal characters
### Description

- Found this after some more extensive testing of the `EMI_001` dataset 
- We need to update `x.SUBJECT/S` and `x.overview` for subjects with illegal characters after DCM2NII
- We didn't see this bug before, because we only ran each import sub-module individually, the failure only occurs if you run steps one and two in one go though

### Release notes

Update x.SUBJECT/S and x.overview for subjects with illegal characters.

MichaelStritt:
### Linked issue

Check out the description in #948

### Comments

Seems like we overlooked this minor bug in the previous merge.

### DIFF
--- a/Modules/SubModule_Import/xASL_imp_DCM2NII.m
+++ b/Modules/SubModule_Import/xASL_imp_DCM2NII.m
@@ -1,4 +1,4 @@
-function xASL_imp_DCM2NII(x, imPar)
+function x = xASL_imp_DCM2NII(x, imPar)
 %xASL_imp_DCM2NII Run the dcm2nii part of the import.
 %
 % FORMAT: xASL_imp_DCM2NII(x, imPar)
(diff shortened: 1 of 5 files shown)


Modules/SubModule_Import/xASL_imp_DCM2NII.m
```diff
@@ -1,4 +1,4 @@
-function xASL_imp_DCM2NII(x, imPar)
+function x = xASL_imp_DCM2NII(x, imPar)
 %xASL_imp_DCM2NII Run the dcm2nii part of the import.
 %
 % FORMAT: xASL_imp_DCM2NII(x, imPar)
```
jan-petr, line 4: "x = " missing

jan-petr (approved):
OK.

jan-petr (changes requested):
One minor header thing

Modules/SubModule_Import/xASL_imp_DCM2NII.m
MichaelStritt: Fixed

jan-petr (approved):
OK

PR #955, opened 2021-11-29
Closes #940 Check folder hierarchy

### LINKED ISSUE #940: Managing `bMatchDirectories` is false
### Description

* If `bMatchDirectories` is false, `folderHierarchy` should end with an extension, like dcm or par. This allows switching the `dcm2nii_version` using `imPar.folderHierarchy{end}`. 

### Requirements

If bMatchDirectories is false, a warning should be given if no extension is provided in `folderHierarchy`, and we automatically add `.dcm` as extension (assuming that this concerns enhanced DICOMs, which is the case mostly).

If a folder only contains one enhanced DICOM, this probably won't be an issue, but `bMatchDirectories` is actually useful when a folder contains multiple enhanced DICOMs, in which case each file is a separate "scan token".

This also requires `folderHierarchy` `dcmExtFilter` to be case insensitive.

### How to test

### Release notes

MichaelStritt:
### Linked issue

Check out #940


jan-petr (changes requested):
Besides XML, I have one issue:
1. What if the Extension is given without the fullstop - just "PAR", but then just "PAR" can be mistaken with something else. But we can't look at it as the last string "PAR$" cause there can be some bracket etc.

I have no real idea how to solve this. Better to send to Henk for review as well - even though it will take some time...

Modules/SubModule_Import/xASL_imp_ReadSourceData.m
```diff
@@ -61,3 +65,33 @@
...
+           ~isempty(regexpi(lastElement,'.par')) || ...
+           ~isempty(regexpi(lastElement,'.rec'))
+           warning('The sourceStructure folderHierarchy includes a file extension...');
+        end
+    else
+        % Check for extension in last folder hierachy element
+        if isempty(regexpi(lastElement,'.dcm')) && ...
+           isempty(regexpi(lastElement,'.ima')) && ...
+           isempty(regexpi(lastElement,'.par')) && ...
+           isempty(regexpi(lastElement,'.rec'))
```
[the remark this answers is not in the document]
MichaelStritt: Fixed

jan-petr (approved):
I'm fine with it and let's see if Henk has any idea how to address my comment in PR...

HenkMutsaerts (changes requested):
We should add an elseif for another (unknown) extension, then it can just issue a warning that we detected an unknown extension.

MINOR:
I would replace the regular expressions by a single one, then you can do the `regexp` once outside the if statement. 

`hasKnownExtension = ~isempty(regexpi(lastElement,'.(dcm|ima|xml|par|rec)'))`;

and you can combine `else` and `if` on line 84 into `elseif`.

Otherwise nice!

HenkMutsaerts:
> Besides XML, I have one issue:
> 
> 1. What if the Extension is given without the fullstop - just "PAR", but then just "PAR" can be mistaken with something else. But we can't look at it as the last string "PAR$" cause there can be some bracket etc.
> 
> I have no real idea how to solve this. Better to send to Henk for review as well - even though it will take some time...

Try using Matlab's `fileparts` for this. This should separate the content of the last folder level for extension. Check if it gives an empty file or empty extension properly.

Modules/SubModule_Import/xASL_imp_ReadSourceData.m
```diff
@@ -74,6 +74,9 @@ function xASL_imp_ReadSourceData_CheckFolderHierarchy(x)
     % Condition for file extension
     conditionFile = '.(dcm|ima|xml|par|rec)';
 
+    % Other extension
+    conditionExtension = '.';
```
jan-petr, line 78: '.' is **any** character in regexpi notation. You want '\\.'

Modules/SubModule_Import/xASL_imp_ReadSourceData.m
```diff
@@ -74,6 +74,9 @@ function xASL_imp_ReadSourceData_CheckFolderHierarchy(x)
     % Condition for file extension
     conditionFile = '.(dcm|ima|xml|par|rec)';
```
jan-petr, line 75: '.' is any character in regexpi notation. You want '\\.'

jan-petr (changes requested):
Small edit needed

Modules/SubModule_Import/xASL_imp_ReadSourceData.m
MichaelStritt: c503604

Modules/SubModule_Import/xASL_imp_ReadSourceData.m
MichaelStritt: c503604

jan-petr (approved):
OK

PR #961, opened 2021-12-01
Closes #947 Update finish import script

### LINKED ISSUE #947: xASL_imp_FinishImport violates modularity
### Description

We strive towards modularity, where submodules are iterated and take care over their own "householding". It is important that the user can always combine these modules to its own liking, which works with the lock files. if a lock file doesn't exist for a current iteration of a current module, the submodule will start with householding (deleting previous derivatives for this iteration, etc), then running all submodules/wrappers, and then doing the final householding (deleting temporary files etc).

So in other words, all general householding functions (either initializing or "finish" should do general stuff only.

### Requirements

So the contents of `xASL_imp_FinishImport` should be distributed across their actual (sub-)modules. 

### How to test

If a user runs dcm2nii and nii2BIDS, the derivatives folder should be untouched, unchecked, no warning or printing given on the screen. All the touching, checking, printing, warning of derivatives should be done in BIDS2Legacy.

Likewise, if a user runs BIDS2Legacy, the sourcedata and temp folders should be untouched, unchecked, and no warning or printing on this.

### Release notes

No need for release notes, this is just improving good coding practice.

MichaelStritt:
### Linked issue

Check out #947

### DIFF
--- a/ExploreASL_ImportMaster.m
+++ b/ExploreASL_ImportMaster.m
@@ -44,7 +44,10 @@ function [x] = ExploreASL_ImportMaster(x)
         % Here we run the subject-wise ExploreASL xASL_module_Import. In future releases xASL_Iteration should 
         % help us to run a parallelized import and also to enable reruns if parts of the import crashed.
         [~, x] = xASL_Iteration(x,'xASL_module_Import');
-        x = xASL_imp_FinishImport(x);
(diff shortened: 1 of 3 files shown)


HenkMutsaerts (approved):
Nice thanks!

Issue #963, opened 2021-12-01
Philips Hadamard decoding


jan-petr:
### Description

Implement the decoding for Philips Hadamard matrix.
Philips data should be a standard Hadamard working with the normal code. So I have only cleaned and optimized the code.

### Requirements

-> inside xASL_quant_HadamardDecoding function:
- [x] clean and optimize the previous code
- [x] add default Philips decoding matrix
- [x] multiply the decoding matrix per -1 if x.Q.Vendor is Philips 

-> inside xASL_wrp_ResampleASL function: 
- [x] after decoding, average across repetiitons (as we already do for multiPLD data) - save as PWI4D.nii (=keep as is).
- [x] before averaging across PLDs, we take only the first TE of each PLD and we average across PLDs - save that additionally as PWI.nii (to be added similar like for multiPLD).

### Release notes

Clear the Hadamard decoding code. Add decoding matrix for Philips

PR #964, opened 2021-12-02
Feature #963 hadamard decoding philips

jan-petr:
### Linked issue

#963

### DIFF
--- a/Functions/xASL_quant_HadamardDecoding.m
+++ b/Functions/xASL_quant_HadamardDecoding.m
@@ -101,7 +101,12 @@ else
                                  1  1 -1 -1  1  1 -1 -1;
                                  1 -1  1 -1  1 -1  1 -1];
 		end    
-	end
+    end
+    
+    if strcmp(xQ.Vendor,'Philips')
+        xQ.TimeEncodedMatrix = xQ.TimeEncodedMatrix * -1;
+    end
+    
 end
 
 % Load time-series nifti
--- a/Functions/xASL_quant_HadamardDecoding.m
+++ b/Functions/xASL_quant_HadamardDecoding.m
@@ -1,7 +1,7 @@
-function [imASLReordered] = xASL_quant_HadamardDecoding(imPath, xQ)
+function [imDecoded] = xASL_quant_HadamardDecoding(imPath, xQ)
 %xASL_quant_HadamardDecoding Hadamard-4 & Hadamard-8 Decoding
 %
-% FORMAT:       [imASLReordered] = xASL_quant_HadamardDecoding(imPath, xQ)
+% FORMAT:       [imDecoded] = xASL_quant_HadamardDecoding(imPath, xQ)
 %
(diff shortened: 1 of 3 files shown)


MichaelStritt:
For me personally dcm2niix still crashes (even though the up-to-date flavor library was used):

```
Importing subject = Sub1_1:   
Subject = Sub1, visit = 1, session = ASL_1, scan = ASL4D
executing: ["M:\SoftwareDevelopment\MATLAB\m.stritt\Server_xASL\ExploreASL\External\MRIcron\20190902\dcm2nii.exe" 
-f "%f_%p_%t_%s_%r" -o "M:\SoftwareDevelopment\MATLAB\m.stritt\Server_xASL\FlavorDatabase\Siemens_PCASL_3DGRASE_VD13A_Hadamard4FME_1\derivatives\ExploreASL\temp\Sub1\ASL_1\dcm2nii_temp_ASL4D"
"M:\SoftwareDevelopment\MATLAB\m.stritt\Server_xASL\FlavorDatabase\Siemens_PCASL_3DGRASE_VD13A_Hadamard4FME_1\sourcedata\Sub1\ASL\IM0007.0001.4914180.IMA"]

status: -1073740791
Warning: No NIfTIs found! Either illegal characters in filenames or other code defect... 
> In xASL_io_dcm2nii (line 251)
```

Functions/xASL_quant_HadamardDecoding.m
```diff
@@ -101,7 +101,12 @@
                                  1  1 -1 -1  1  1 -1 -1;
                                  1 -1  1 -1  1 -1  1 -1];
 		end    
-	end
+    end
+    
+    if strcmp(xQ.Vendor,'Philips')
```
MichaelStritt, line 106: x.Q.Vendor

jan-petr: That was correct. We only pass x.Q as input, not the whole x, so inside this function. It is called xQ ;)

BeatrizPadrela:
Okay guys, I had to put an if loop here

image

because with Amnah's data (FME), we only have 2 repetitions, so this line makes sense and works for the next for loop of the averaging across repetitions.
However, with Philips data, since we have 12 repetitions, the size of ASL_im, divided by the 12 repetitions, gives the PLDs per repetition (7).  If I used the same NumberofPLDs line that I used for FME data, in the foor loop, in line 251, in the last iteration, we would have 12 (number of rep) * 12 (84 / Had8-1 = 84/7 = 12), which would overfit the matrix size

Tell me what you think @MichaelStritt  @jan-petr

MichaelStritt:
> 

@jan-petr & @BeatrizPadrela: Possibly a stupid question from my side, but wouldn't it be easier to determine that based on the matrix size? Then your for loop can't crash.

By the way, there are no if _loops_, we call them if _statements_.

BeatrizPadrela:
@MichaelStritt  that's a good idea!
And yes indeed, if _statements_ and for _loops_, yesterday was already late :)

BeatrizPadrela (approved):

PR #966, opened 2021-12-02
Closes #873 Avoid double ASL4D registration

### LINKED ISSUE #873: xASL_wrp_LinearReg_T1w2MNI detecting ASL4D twice
### Description

In the xASL_wrp_LinearReg_T1w2MNI, which is needed for the registration, there are predefined paths of where the ASL4D.nii should be, and they it searches for other scans and identifies the ASL4D.nii again (line71) even if there's no dwi and func folders:

image
image
image

So what happens is that the ASL4D is registered twice to the T1, so the second time it becomes an image that is clearly not registered to the T1. Henk and I discovered this when I was trying to convert the CEST (as ASL4D) nifti to MNI space, because right after the registration (where other files that are in the subject's folder are also registered to the T1), the CEST image and the T1 weren't matching (when overlapped with fsl eyes).

 

### How to test

I tested it this way:
Creating the SubjectsFolders, where it looks for folders inside the Subject's folder and only keeps the ones that do exist.
image

### Release notes

Avoid spurious double registration of ASL to T1w.

MichaelStritt:
### Linked issue

Check out #873 

### How to test

### Comments

### DIFF
--- a/Modules/SubModule_Structural/xASL_wrp_LinearReg_T1w2MNI.m
+++ b/Modules/SubModule_Structural/xASL_wrp_LinearReg_T1w2MNI.m
@@ -57,16 +57,16 @@ for iS=1:length(Lesion_ROI_list)
     OtherList{end+1,1} = Lesion_ROI_list{iS};
 end
 % Add ASL images to the registration list
-for iSess = 1:x.dataset.nSessions
-    OtherList{end+1,1} = x.P.Path_ASL4D;
-    OtherList{end+1,1} = x.P.Path_M0;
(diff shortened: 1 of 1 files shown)


HenkMutsaerts:
I think I wrote this, if Beatriz tested this, this should work. We should check carefully that it doesn't select ancillary files that shouldn't be coregistered.

Note that `xASL_adm_GetFileList` also checks for `.nii.gz` if `.nii` is requested, so that will work.

HenkMutsaerts (approved):

jan-petr (approved):
OK

Issue #902, opened 2021-11-12
Do not use parfor in xASL_Iteration or elsewhere


MichaelStritt:
Remove use of parfor

Also `parfor` needs to be removed from:
* `/Functions/xASL_Iteration` line 243
* `/External/SPMmodified/toolbox/cat12/cat_surf_epivolsurf.m`
* `/External/SPMmodified/spm_dcm_fit.m`
* `/Development/StatsAnalysis/StatTestCompare/StatTrial.m`

### Release note
Stop using Distributed computing and Parallel Toolboxes and parfor.

jan-petr:
@MichaelStritt I have no idea how to read this - can you please add a bit more details to the issue description.

MichaelStritt:
> 
> 
> @MichaelStritt I have no idea how to read this - can you please add a bit more details to the issue description.

I used the GitHub feature where you create a new issue based on a comment 😆
You can nicely click on the comment though. 
I'm actually not so sure what Henk wants either 😉 

![image](https://user-images.githubusercontent.com/46593074/141679749-43ea4eef-5607-42c4-997a-ec7df9038480.png)

MichaelStritt:
@HenkMutsaerts suggested to fade out the parfor support.

Issue #888, opened 2021-11-04
Distribution toolbox initialization


jan-petr:
### Description

Try to use the distributed computing toolbox only when necessary - i.e. only when trying to run commands that use it. Otherwise don't try to block it by default.

### Requirements

Optional: add feature requirements here

### How to test

Optional: insert description about how to test the code changes here

### Release notes

Required: summarize the changes for the release notes here

MichaelStritt:
@jan-petr: isn't this related to #902?

MichaelStritt:
@jan-petr: so what's the plan here?

In `xASL_Iteration` we do this check ...

```matlab
    % Check if parallel computing toolbox is available
    poolsize = 0;
    if license('test','Distrib_Computing_Toolbox') && license('checkout','Distrib_Computing_Toolbox')
        if verLessThan('matlab','8.2') 
            % < R2013b
            if exist('matlabpool','file')
                poolsize = matlabpool('size');
            end
        else
            % >= R2013b
            if  exist('gcp','file') % bugfix, some even have the Distrib_Computing_Toolbox, but don't have the Parallel_Computing_Toolbox
                poolobj = gcp('nocreate'); % If no pool, do not create new one.
                if ~isempty(poolobj)
                    poolsize = poolobj.NumWorkers;
                end
            end
        end
    end
```

... and I know that's the problem which you're facing at your institute. I just personally only see two options ...

1. Remove the parallel toolbox completely
2. Keep it like it is

... because we already query if the toolbox is there or not and only is it in that case.

PR #968, opened 2021-12-02
Optimize #902 remove parfor

jan-petr:
### Linked issue

#902

### DIFF
--- a/Development/StatsAnalysis/StatTestCompare/StatTrial.m
+++ b/Development/StatsAnalysis/StatTestCompare/StatTrial.m
@@ -151,7 +151,7 @@ StepSize        = (pMax - pMin) ./ ROCRes;
 
 StatCalc(:,1)   = [pMin:StepSize:pMax]';
 
-parfor iStat=1:size(StatCalc,1)
+for iStat=1:size(StatCalc,1)
     P           = zeros( size(Group1,1), size(Group1,2));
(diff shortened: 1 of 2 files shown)


jan-petr:
Shall we remove code on line 182 as well?

MichaelStritt (approved):
👍

MichaelStritt:
> Shall we remove code on line 182 as well?

Hm... not sure. Let's not change too much to keep everything stable before the release.

HenkMutsaerts (changes requested):
Nice, what about the other two occurrences? Do we leave them in?

`cat_surf_epivolsurf`
`zspm_dcm_fit`

Nobody uses these, so safe to replace the `parfor` with `for` here to be sure Matlab never checks for this toolbox.

Then we can close the other issue as "duplicate".

jan-petr:
> zspm_dcm_fit



> Nice, what about the other two occurrences? Do we leave them in?
> 
> `cat_surf_epivolsurf` `zspm_dcm_fit`
> 
> Nobody uses these, so safe to replace the `parfor` with `for` here to be sure Matlab never checks for this toolbox.

There it is already commented out. No need to remove.

> Then we can close the other issue as "duplicate".

Which other issue? #888 this is already closed...

HenkMutsaerts (approved):

PR #974, opened 2021-12-06
Closes #841 Import zipped files

### LINKED ISSUE #841: Sourcedata zipping on scantype level
### Description

Just heard from Paul Groot that the DICOMs make life very difficult on slow server disks that are specialized for backup. So the request is to us to zip these on the folder level, preferably on the scan type level.

•	Paul shares his zip-script, to zip all DICOMs per scantype
•	Then just before the dcm2niiX conversion, we need to unzip per scantype, and delete this after the conversion

### Requirements

### How to test

### Release notes

MichaelStritt:
### Linked issue

Check out #841

### How to test

I took `GE_PCASL_3Dspiral_14.0LX_1` and just zipped all the ASL and T1w data. Then I changed the `sourceStructure.json` to:

```json
{"folderHierarchy":["^(.)+$","^(ASL|T1w|M0|T2|FLAIR)$","\\.(zip|dcm)"],
"tokenOrdering":[1,0,2],
"tokenSessionAliases":["",""],
"tokenScanAliases":["^ASL$","ASL4D","^T1w$","T1w","^M0$","M0","^T2$","T2w","^FLAIR$","FLAIR"],
"bMatchDirectories":false}
```

Most of it seems to work. @jan-petr: for me it used to crash later on. Am I doing something wrong in `sourceStructure.json`?

```
ERROR: Import module terminated for subject 1: Sub103

ans =

    'Index exceeds the number of array elements (1).
     
     Error in xASL_bids_Dicom2JSON (line 419)
     	if  isfield(t_parms{parmsIndex},'AcquisitionTime')
```

I'm using this temporary fix right here https://github.com/ExploreASL/ExploreASL/pull/974#discussion_r769957679

### Comments

### DIFF
--- a/Functions/xASL_bids_Dicom2JSON.m
+++ b/Functions/xASL_bids_Dicom2JSON.m
@@ -51,7 +51,10 @@ function [parms, pathDcmDictOut] = xASL_bids_Dicom2JSON(imPar, pathIn, pathJSON,
 	end
 	if nargin<5
 		pathDcmDictIn = [];
-	end
+    end
+    if ~exist('dcmfields','var')
+        dcmfields = [];
+    end
 	
 	pathDcmDictOut = pathDcmDictIn;
     
@@ -322,40 +325,7 @@ function [parms, pathDcmDictOut] = xASL_bids_Dicom2JSON(imPar, pathIn, pathJSON,
 			
 			
 			% Do this only once, and do not reset the manufacturer for other files (assume the same is for all files within the directory)
-			if ~exist('dcmfields','var')
-				%% -----------------------------------------------------------------------------
-				% Identify Manufacturer and select the important header parameters to read
-				% -----------------------------------------------------------------------------
-				if  isfield(temp, 'Manufacturer')
-					manufacturer    = lower(temp.Manufacturer);
-					if ~isempty(strfind(manufacturer,'ge'))
-						bManufacturer = 'GE';
-					elseif ~isempty(strfind(manufacturer,'philips'))
-						bManufacturer = 'Philips';
-					elseif ~isempty(strfind(manufacturer,'siemens'))
-						bManufacturer = 'Siemens';
-					else
-						warning('xASL_adm_Dicom2JSON: Manufacturer unknown for %s', filepath);
-					end
-				else
-					warning('xASL_adm_Dicom2JSON: Manufacturer unknown for %s', filepath);
-				end
-				
-				dcmfields = DcmFieldList;
-				
-				switch bManufacturer
-					case 'GE'
-						dcmfields(end+1:end+4) = {'AssetRFactor', 'EffectiveEchoSpacing'...
-							'GELabelingDuration' 'InversionTime' }; % (0043,1083) (0043,102c)
-					case 'Philips'
-						dcmfields(end+1:end+4) = {'MRSeriesWaterFatShift', 'MRSeriesEPIFactor'...
-							'TemporalPositionIdentifier'  'PhilipsNumberTemporalScans'}; % (2001,1022) (2001,1013)
-					case 'Siemens'
-						dcmfields(end+1) = {'BandwidthPerPixelPhaseEncode'}; % (0019,1028)
-					otherwise
-						% do nothing
-				end
-			end
+            [bManufacturer,dcmfields] = xASL_bids_Dicom2JSON_IdentifyManufacturer(temp,dcmfields,DcmFieldList,bManufacturer);
 			
             %% -----------------------------------------------------------------------------
             % Obtain the selected DICOM parameters from the header
@@ -654,13 +624,54 @@ end
 
 %% Obtain a value from a structure 
 function val = GetDicomValue(I, fieldname, default)
-if nargin>2
-	val = default;
-else
-	val = [];
+
+    if nargin>2
+        val = default;
+    else
+        val = [];
+    end
+
+    if isfield(I, fieldname)
+        val = I.(fieldname);
+    end
+    
 end
 
-if isfield(I, fieldname)
-	val = I.(fieldname);
-end
+%% Identify Manufacturer
+function [bManufacturer,dcmfields] = xASL_bids_Dicom2JSON_IdentifyManufacturer(temp,dcmfields,DcmFieldList,bManufacturer)
+
+    % Identify Manufacturer and select the important header parameters to read
+    if isempty(dcmfields)
+        
+        if  isfield(temp, 'Manufacturer')
+            manufacturer    = lower(temp.Manufacturer);
+            if ~isempty(strfind(manufacturer,'ge'))
+                bManufacturer = 'GE';
+            elseif ~isempty(strfind(manufacturer,'philips'))
+                bManufacturer = 'Philips';
+            elseif ~isempty(strfind(manufacturer,'siemens'))
+                bManufacturer = 'Siemens';
+            else
+                warning('xASL_adm_Dicom2JSON: Manufacturer unknown for %s', filepath);
+            end
+        else
+            warning('xASL_adm_Dicom2JSON: Manufacturer unknown for %s', filepath);
+        end
+
+        dcmfields = DcmFieldList;
+
+        switch bManufacturer
+            case 'GE'
+                dcmfields(end+1:end+4) = {'AssetRFactor', 'EffectiveEchoSpacing'...
+                    'GELabelingDuration' 'InversionTime' }; % (0043,1083) (0043,102c)
+            case 'Philips'
+                dcmfields(end+1:end+4) = {'MRSeriesWaterFatShift', 'MRSeriesEPIFactor'...
+                    'TemporalPositionIdentifier'  'PhilipsNumberTemporalScans'}; % (2001,1022) (2001,1013)
+            case 'Siemens'
+                dcmfields(end+1) = {'BandwidthPerPixelPhaseEncode'}; % (0019,1028)
+            otherwise
+                % do nothing
+        end
+    end
+
 end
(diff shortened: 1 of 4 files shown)


HenkMutsaerts:
Nice work, let's ask Paul Groot for an example dataset and we can test this after January 13th, sorry ;)

MichaelStritt:
> Nice work, let's ask Paul Groot for an example dataset and we can test this after January 13th, sorry ;)

It's super easy to build one, but as shown above, there's still some minor work to do.

MichaelStritt:
### To Do

- [ ] Use xASL (un)zipping instead of Matlab (un)zipping

Modules/SubModule_Import/xASL_imp_ReadSourceData.m
```diff
@@ -99,3 +118,19 @@ function xASL_imp_ReadSourceData_CheckFolderHierarchy(x)
...
 
 
+%% Unzip all '.zip' files in subdirectories and delete original files
+function xASL_imp_UnzipAll(dirToUnzip)
+
+    % Iterate over zip file
+    zipFiles = xASL_adm_GetFileList(dirToUnzip,'^.+\.zip$','FPListRec');
+    for iFile = 1:numel(zipFiles)
+        currentDirectory = xASL_fileparts(zipFiles{iFile});
+        unzip(zipFiles{iFile},currentDirectory);
```
HenkMutsaerts, line 128: Preferably, see if we can use our own function that is faster and/or doesn't require jvm

HenkMutsaerts (changes requested):

Modules/SubModule_Import/xASL_imp_ReadSourceData.m
MichaelStritt: For me `xASL_adm_UnzipOrCopy` crashes and says there are too many output files (line 64). In addition it internally uses the exact same command...

```matlab
if strcmpi(ext,'.gz')
	destpath = fullfile(destDir, name );
	if bOverwrite || ~exist(destpath,'file')
		X = gunzip(srcpath, destDir);
	end
elseif strcmpi(ext,'.zip')
	destpath = fullfile(destDir, name );
	if bOverwrite || ~exist(destpath,'file')
		X = unzip(srcpath, destDir);
	end
else
	destpath = fullfile(destDir, entries(ii).name);
	if bOverwrite || ~exist(destpath,'file')
		copyfile(srcpath, destDir);
		X = destpath;
	end
end
```

... so no real gained benefit for me.

Modules/SubModule_Import/xASL_imp_ReadSourceData.m
```diff
@@ -66,31 +72,44 @@
...
 %% Check folder hierarchy
-function xASL_imp_ReadSourceData_CheckFolderHierarchy(x)
+function x = xASL_imp_ReadSourceData_CheckFolderHierarchy(x)
 
     % Get last element
     lastElement = lower(x.modules.import.imPar.folderHierarchy{end});
 
     % Condition for file extension
-    conditionFile = '\.(dcm|ima|xml|par|rec)';
+    conditionFile = '(dcm|ima|xml|par|rec|zip)';
```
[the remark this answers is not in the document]
MichaelStritt: I'm starting with zip for now. We can then easily add more functionality. One step after the other.

Functions/xASL_bids_Dicom2JSON.m
```diff
@@ -416,20 +416,31 @@
...
-						tempAcquisitionTime = t_parms{parmsIndex}(1).AcquisitionTime;
-						t_parms{parmsIndex} = rmfield(t_parms{parmsIndex},'AcquisitionTime');
-						t_parms{parmsIndex}(1).AcquisitionTime = tempAcquisitionTime;
-					end
+                if parmsIndex>numel(t_parms)
+                    % This statement catches the cases where we iterate out of valid ranges of the t_parms struct. Right now I
+                    % mostly saw that bug during the import using bMatchDirectory = false! We probabaly determine the index wrong there!
+                    warning('Dicom to JSON: Invalid index for t_parms...');
+                    return
+                else
```
MichaelStritt, line 424: @jan-petr: Could you help me here? I think there is a bug related to `bMatchDirectories=false`, probably no one really uses it or I am doing something wrong.

Modules/SubModule_Import/xASL_imp_ReadSourceData.m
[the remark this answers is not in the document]
MichaelStritt: Needs to be tested 575c826

Functions/xASL_bids_Dicom2JSON.m
```diff
@@ -4,652 +4,799 @@
...
-%        imPar               - struct with import parameters (REQUIRED)
-%        pathIn (PATH)       - path to the RAW files (REQUIRED)
-%        pathJSON (PATH)     - cell with paths to the JSON file for saving parameters (OPTIONAL, DEFAULT = don't save)
-%        dcmExtFilter (STR)  - wildcards specifying the allowed extensions for the RAW files
-%        bUseDCMTK (BOOL)    - if yes, then use DCMTK instead of dicominfo
-%        pathDcmDictIn (STR) - path to the dicom dictionary in case DCMTK fails and DICOMINFO is used
+%        imPar            - struct with import parameters (STRUCT, REQUIRED)
+%        pathIn           - path to the RAW files (PATH, REQUIRED)
+%        pathJSON         - cell with paths to the JSON file for saving parameters (PATH, OPTIONAL, DEFAULT = don't save)
+%        dcmExtFilter     - wildcards specifying the allowed extensions for the RAW files (STRING)
```
jan-petr, line 10: specify that these are OPTIONAL

jan-petr: bUseDCMTK = true

jan-petr: pathDcmDictIn = []

jan-petr (changes requested):
Small things only.

HenkMutsaerts:
We discussed that this would be transparant, so the `sourceStructure.json` would stay the same, and if instead a folder we find a zip file with the same name, this is used instead. Not sure if this is very tricky to program. The problem is, that a datasets can be partly zipped.

HenkMutsaerts (changes requested):
Very nice!

We discussed that this would be transparant, so the `sourceStructure.json` would stay the same, and if instead a folder we find a zip file with the same name, this is used instead. Not sure if this is very tricky to program. The problem is, that a datasets can be partly zipped.

The refactoring is nice, but is a different issue right (unrelated to zipping)? Let's keep this clean.

MichaelStritt:
> Very nice!
> 
> We discussed that this would be transparant, so the `sourceStructure.json` would stay the same, and if instead a folder we find a zip file with the same name, this is used instead. Not sure if this is very tricky to program. The problem is, that a datasets can be partly zipped.
> 
> The refactoring is nice, but is a different issue right (unrelated to zipping)? Let's keep this clean.

If you start making things even more complex now, this will only prolong the development. Let's start with zipped or not zipped, not partly zipped. I would check the partly thing later on. The sourceStructure can't stay the same, because the solution right now is to check if we match files and if the file extension is zip. I wouldn't know how to do this otherwise, since the matching function that you originally wrote has to look for files or directories, not some weird combination thing.

MichaelStritt:
@jan-petr: Let'se use this flavor for testing https://github.com/ExploreASL/FlavorDatabase/tree/main/Siemens_PASL_2DEPI_VB15A_ADNI2_941_S_4764_1

jan-petr (changes requested):
The ADNI data in general works. BUT! 
1. The ZIP runs in a MatchDirectory=false mode - because that's a file. Which is fine - you made the sourceStructure.json correctly. But once this is unzipped, it suddenly doesn't work anymore with the same sourceStructure.json - because the directory structure is preserved, but there are a lot of files now and it tries to import sort of each one separately. It would have probably been possible to fix this, but that's not the way to go IMHO.
2. I think that it was wrong in the first place to specify files-matching and adding the .ZIP as another layer - because you shouldn't modify the sourceStructure.json based on what you are expecting.
3. Instead, I would keep the sourcestructure.json exactly the same as it was with DCMs in - bmatchdirectory = true and not adding the .zip/.dcm. And once you start scanning the directory, then before running dcm2niix, you'll have to identify that you have zip files and then unzip them. This approach has two advantages - everything stays in directories, and you will be able to use the same sourcestructure.json for the temp-data as well without modification. And the biggest advantage - you'll be able to accept both unzipped and zipped files without any effort associated.

But this all means that you'll have to modify certain parts of the code. I can help with that...

MichaelStritt:
> 3\. Instead, I would keep the sourcestructure.json exactly the same as it was with DCMs in - bmatchdirectory = true and not adding the .zip/.dcm. And once you start scanning the directory, then before running dcm2niix, you'll have to identify that you have zip files and then unzip them. This approach has two advantages - everything stays in directories, and you will be able to use the same sourcestructure.json for the temp-data as well without modification. And the biggest advantage - you'll be able to accept both unzipped and zipped files without any effort associated.

Sounds like a good plan to me. Then we need to change the matching function, which I was a bit afraid of. Since it's a different road to go down, I would probably start from the beginning and copy paste what I need from this branch.

HenkMutsaerts:
We also should test this branch for the following dataset:
`L:\basic\divi\Projects\ExploreASL\Temp\1242_epod-3_20200414_8_channel` 
as real-life example from Paul's Flux server

no need anymore, there is an example in the flavors

https://github.com/ExploreASL/FlavorDatabase/tree/main/Siemens_PASL_2DEPI_VB15A_ADNI2_941_S_4764_1/sourcedata/Sub1/ASL:

/Siemens_PASL_2DEPI_VB15A_ADNI2_941_S_4764_1/sourcedata/Sub1/ASL/ADNI_941_S_4764_MR_ASL_PERFUSION.zip

/Siemens_PASL_2DEPI_VB15A_ADNI2_941_S_4764_1/sourcedata/Sub1/ASL/ADNI_941_S_4764_MR_ASL_PERFUSION/0001.dcm

Issue #971, opened 2021-12-06
xASL_im_GetMeanControl


jan-petr:
### Description

- [x] xASL_im_GetMeanControl general function to create mean control from Hadamard, single PLD etc to a single function.
- [x] Fix header of GetControlLabelOrder function (note that this works also for multiPLD without changes needed.

-> inside the new xASL_quant_GetMeanControl function: 
- [x] for Hadamard - don't take only the 1st repetition, but all repetitions of control (which are later averaged)
- [x] for Hadamard - take only the first TE.

### Release notes

Mean control image creation in a separate function.

BeatrizPadrela:
@jan-petr the code I've added, works for the repetitions of Controls and also for the first TE because the first volume of the control is also the first TE. What do you think?

PR #977, opened 2021-12-08
Optimize #971 get mean control

BeatrizPadrela:
### Linked issue

Should work alreayd for Hadamard and for the other data formats

### DIFF
--- a/Functions/xASL_quant_GetControlLabelOrder.m
+++ b/Functions/xASL_quant_GetControlLabelOrder.m
@@ -1,11 +1,10 @@
 function [ControlIm, LabelIm, OrderContLabl] = xASL_quant_GetControlLabelOrder(ASLTimeSeries)
 %xASL_quant_GetControlLabelOrder Get control-label order of ASL time-series
 %
-% FORMAT: [ControlIM, LabelIM, OrderContLabl] = xASL_quant_GetControlLabelOrder(FramesIn, x)
+% FORMAT: [ControlIm, LabelIm, OrderContLabl] = xASL_quant_GetControlLabelOrder(ASLTimeSeries)
 %
 % INPUT:
 %   ASLTimeSeries - 4D matrix with multiple 3D ASL image volumes over time (REQUIRED)
-%   x             - structure containing fields with all information required to run this submodule (REQUIRED)
 %
 % OUTPUT:
 %   ControlIm     - 4D matrix with 3D ASL control image volumes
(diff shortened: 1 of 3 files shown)


Functions/xASL_quant_GetMeanControl.m
```diff
@@ -0,0 +1,75 @@
+function imMeanControl = xASL_quant_GetMeanControl(x, imASLTimeSeries)
```
MichaelStritt, line 1: Nice header 👍

MichaelStritt:
On the first look I can't see any hard errors. If you say you already tested it on a few flavors then I'm fine, because we'll re-test everything for the release really soon anyway. Let's wait for Jan's feedback though 👍

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
```diff
@@ -130,17 +130,10 @@ function xASL_wrp_ResampleASL(x)
...
-        ControlIm = Im(:,:,:,1);    
-        xASL_io_SaveNifti(x.P.Path_rdespiked_ASL4D, x.P.Path_mean_control, ControlIm, [], 0);
-    else
-        % Create mean control in native space
-        ControlIm = xASL_quant_GetControlLabelOrder(xASL_io_Nifti2Im(x.P.Path_rdespiked_ASL4D));
-        IM_mean = xASL_stat_MeanNan(ControlIm, 4);
-        xASL_io_SaveNifti(x.P.Path_rdespiked_ASL4D, x.P.Path_mean_control, IM_mean, [], 0);
-    end
+	% Obtain the mean control image
+	imMeanControl = xASL_quant_GetMeanControl(x, xASL_io_Nifti2Im(x.P.Path_rdespiked_ASL4D));
```
MichaelStritt, line 134: Nice use of one function as the argument of another 👍

Functions/xASL_quant_GetControlLabelOrder.m
```diff
@@ -16,11 +15,12 @@
...
+%              with alternative control/label or label/control order
 %
 % EXAMPLE:
-% [~, ~, OrderContLabl] = xASL_quant_GetControlLabelOrder(ASLTimeSeries, x);
-% [ControlIm, LabelIm] = xASL_quant_GetControlLabelOrder(ASLTimeSeries, x);
+% [~, ~, OrderContLabl] = xASL_quant_GetControlLabelOrder(ASLTimeSeries);
+% [ControlIm, LabelIm] = xASL_quant_GetControlLabelOrder(ASLTimeSeries);
 % 
 % __________________________________
 % Copyright (C) 2015-2019 ExploreASL
```
MichaelStritt, line 26: Maybe update the date while we're at it, otherwise perfect 👍

MichaelStritt (approved):
Let's wait for Jan's approval 👍

jan-petr:
Some comments: NumberOfAverages might not be reliable for Hadamard, so we should really calculate the size of each Hadamard blockSize = MatrixSize*numberTEs

Then we also don't need to precalculate all the other sizes as we can then index the imASLTimeSeries directly with 1:blockSize:end

BeatrizPadrela:
> Some comments: NumberOfAverages might not be reliable for Hadamard, so we should really calculate the size of each Hadamard blockSize = MatrixSize*numberTEs
> 
> Then we also don't need to precalculate all the other sizes as we can then index the imASLTimeSeries directly with 1:blockSize:end

Indeed! Looks great now

Functions/xASL_quant_GetControlLabelOrder.m
jan-petr: Done!

jan-petr (approved):
OK

PR #985, opened 2021-12-09
Closes #984 General population based report

### LINKED ISSUE #984: xASL Report
### Description

I'm working on a more general population based report. I had some ideas and did an exemplary implementation. An Example is shown here: [xASL_report.pdf](https://github.com/ExploreASL/ExploreASL/files/7687469/xASL_report.pdf)

I know the images still have to be rotated and maybe we can get them to align a bit better, but in general I think a report over multiple pages could be super useful.

### How to test

- [ ] We should test subjects with one to five sessions, since it is supposed to cut off after four images at the moment.
- [ ] It is also supposed to skip missing images.
- [ ] It is also supposed to overwrite existing population based reports.

### Release notes

Duplicate. Now done in #1225

MichaelStritt:
### Linked issue

Check out #984

### How to test

There are instructions what should be tested in #984. Would be good if someone could test all of this on a more "messy" dataset. Please also check out the exemplary PDF in the issue.

### Comments

This is an initial version and can probably be improved with some love from all developers 👍


HenkMutsaerts:
Nice idea! Let's discuss. We already have the single-page report, which is intertwined with the QC output, I would continue this work for the multi-page report. And we should try to see if we can use functions that don't require a GUI (very important) and don't require java virtual machine(nice to have but probably very tricky)

MichaelStritt:
> Nice idea! Let's discuss. We already have the single-page report, which is intertwined with the QC output, I would continue this work for the multi-page report. And we should try to see if we can use functions that don't require a GUI (very important) and don't require java virtual machine(nice to have but probably very tricky)

Don't know why you're talking about GUIs and java, I only used Matlabs `publish` method.

Functions/xASL_report_Main.m
```diff
@@ -0,0 +1,313 @@
...
+    % Get template and output path
+    reportTemplate = fullfile(x.opts.MyPath, 'Functions', 'xASL_report_Template.m');
+    populationDir = fullfile(x.dir.xASLDerivatives, 'Population');
+    htmlDir = fullfile(populationDir, 'html');
+    reportPDFhtml = fullfile(htmlDir, 'xASL_report.pdf');
+    reportOutput = fullfile(populationDir, 'xASL_report.m');
+    reportPDFpop = fullfile(populationDir, 'xASL_report.pdf');
+    
+    % Change to population directory
+    cd(populationDir);
```
HenkMutsaerts, line 28: `pathOriginal = pwd;`
`cd(x.P.PopDir);`

Functions/xASL_report_Main.m
```diff
@@ -0,0 +1,313 @@
...
+%
+% -----------------------------------------------------------------------------------------------------------------------------------------------------
+% EXAMPLE:      reportFile = xASL_report_Main(x);
+%
+% __________________________________
+% Copyright 2015-2021 ExploreASL
+
+    % Get template and output path
+    reportTemplate = fullfile(x.opts.MyPath, 'Functions', 'xASL_report_Template.m');
+    populationDir = fullfile(x.dir.xASLDerivatives, 'Population');
```
HenkMutsaerts, line 21: this already existed also in `x.P.PopDir` or something

Functions/xASL_report_Main.m
```diff
@@ -0,0 +1,313 @@
...
+            if x.settings.DELETETEMP
+                textArray = vertcat(textArray,['% Delete temporary files: ' 'yes']);
+                textArray = vertcat(textArray,'% ');
+            else
+                textArray = vertcat(textArray,['% Delete temporary files: ' 'no']);
+                textArray = vertcat(textArray,'% ');
+            end
+        end
+    end
+
```
HenkMutsaerts, line 309: `cd(pathOriginal);`

Functions/xASL_report_Main.m
```diff
@@ -0,0 +1,313 @@
...
+            textArray = vertcat(textArray,'% ');
+        end
+    end
+    
+    % Add settings
+    textArray = vertcat(textArray,'%% Settings');
+    textArray = vertcat(textArray,'% ');
+    
+    % Print settings
+    if isfield(x,'Version')
```
HenkMutsaerts, line 272: Replace by loop,
PrintFields = {'Version', 'Settings;Quality', '...};

HenkMutsaerts (changes requested):
Consider using `x.Output` and `x.Output_im`

HenkMutsaerts:
> > Nice idea! Let's discuss. We already have the single-page report, which is intertwined with the QC output, I would continue this work for the multi-page report. And we should try to see if we can use functions that don't require a GUI (very important) and don't require java virtual machine(nice to have but probably very tricky)
> 
> Don't know why you're talking about GUIs and java, I only used Matlabs `publish` method.

Matlab has Figure/graphic-creating functions that requires that you run the GUI (they won't run when running Matlab -nodesktop). Then Matlab also has several functions running through the JVM, which you ideally want to skip (it can be buggy sometimes).

Issue #867, opened 2021-10-04
Code optimisation: Rearranging ASL quantification single- and multi-PLD layers


MDijsselhof:
### Description

Currently, we divide single- and multi-PLD quantification in step 8 of the ASL module (`Perform Quantification`) and perform the preprocessing steps twice. For future code changes we need to merge this together and perform the specific single or multi PLD quantification steps a layer down (follow-up issue to #759). Furthermore, all scaling can be in separate function, so we might as well do this in a separate function.

### Release notes
Duplicate #1002

PR #986, opened 2021-12-09
Closes #978 Multi PLD quantification

### LINKED ISSUE #978: Fix multiPLD quantification
### Description

Fix several errors in multi-PLD quantification to make it work with general multi-PLD and Hadamard as well.

### Requirements

- [x] ASL module line 310 crashes for uneven number of volumes - we need to skip in case we have deltams
- [x] Make sure that multiple repetitions are treated correctly for multiPLD. ResampleASL line 265 - make averaging based on PLD vector and don't assume any specific PLD order.
- [x] Clean the code to make the multi-PLD uniquely for BASIL and remove the unused non-BASIL code.
- [x] Clear separation between multi-PLD and single-PLD calls of BASIL.
- [x] Make sure the correct PLDs are used for BASIL.
- [x] Correct all BASIL input and model parameters.
- [x] Configure BASIL for Look-Locker, multi-PLD, and Hadamard.
- [x] Output ATT maps from BASIL.
- [x] Full processing of ATT maps.

### How to test

### Release notes

BASIL for multi-PLD, Look-Locker, and Hadamard. Correct BASIL configuration. Output ATT maps.

jan-petr:
### Linked issue

#978

### DIFF
--- a/Functions/xASL_bids_parms2BIDS.m
+++ b/Functions/xASL_bids_parms2BIDS.m
@@ -54,7 +54,7 @@ convertTimeFieldsRange = [0.5        5                10            10          
                           500        20000            10000         5000               5000                 5000            400               5000      500  5000       400                10000];% Maximum in ms   
 					  
 % Fields that are entered under the subfield 'Q' for xASL on the output
-xASLqFields = {'LabelingType' 'Initial_PLD' 'BackGrSupprPulses' 'LabelingDuration' 'SliceReadoutTime' 'NumberOfAverages' 'BloodT1'...
+xASLqFields = {'LabelingType' 'Initial_PLD' 'BackGrSupprPulses' 'LookLocker' 'LabelingDuration' 'SliceReadoutTime' 'NumberOfAverages' 'BloodT1'...
 	           'BackgroundSuppressionPulseTime' 'BackgroundSuppressionNumberPulses' 'TimeEncodedMatrixSize' 'NumberEchoTimes' 'TimeEncodedMatrixType' 'TimeEncodedMatrix'};
 
 % Some JSON fields need to be updated to fit the BIDS definition
--- a/Functions/xASL_init_FileSystem.m
+++ b/Functions/xASL_init_FileSystem.m
@@ -160,7 +160,8 @@ if isfield(x.P,'SubjectID')
     x.P.Pop_Path_MaskVascular = fullfile(x.D.PopDir, ['MaskVascular_' x.P.SubjectID '_' x.P.SessionID '.nii']);
     x.P.Path_Pop_MaskSusceptibility = fullfile(x.D.PopDir, ['rMaskSusceptibility_' x.P.SubjectID '_' x.P.SessionID '.nii']);    
 	
-	x.P.Path_ASL4D_aslcontext = fullfile(Path{2},'ASL4Dcontext.tsv');
+	x.P.Path_ASL4Dcontext = fullfile(Path{2},'ASL4Dcontext.tsv');
+	x.P.Path_ASL4Dcontext_Source = fullfile(Path{2},'ASL4Dcontext_Source.tsv');
 end
 
 
--- a/Functions/xASL_init_FileSystem.m
+++ b/Functions/xASL_init_FileSystem.m
@@ -159,6 +159,8 @@ if isfield(x.P,'SubjectID')
     end
     x.P.Pop_Path_MaskVascular = fullfile(x.D.PopDir, ['MaskVascular_' x.P.SubjectID '_' x.P.SessionID '.nii']);
     x.P.Path_Pop_MaskSusceptibility = fullfile(x.D.PopDir, ['rMaskSusceptibility_' x.P.SubjectID '_' x.P.SessionID '.nii']);    
+	
+	x.P.Path_ASL4D_aslcontext = fullfile(Path{2},'ASL4D_aslcontext.tsv');
 end
 
 
--- a/Functions/xASL_init_FileSystem.m
+++ b/Functions/xASL_init_FileSystem.m
@@ -160,7 +160,7 @@ if isfield(x.P,'SubjectID')
     x.P.Pop_Path_MaskVascular = fullfile(x.D.PopDir, ['MaskVascular_' x.P.SubjectID '_' x.P.SessionID '.nii']);
     x.P.Path_Pop_MaskSusceptibility = fullfile(x.D.PopDir, ['rMaskSusceptibility_' x.P.SubjectID '_' x.P.SessionID '.nii']);    
 	
-	x.P.Path_ASL4D_aslcontext = fullfile(Path{2},'ASL4D_aslcontext.tsv');
+	x.P.Path_ASL4D_aslcontext = fullfile(Path{2},'ASL4Dcontext.tsv');
 end
 
 
--- a/Functions/xASL_quant_Basil.m
+++ b/Functions/xASL_quant_Basil.m
@@ -1,7 +1,7 @@
-function [CBF_nocalib, resultFSL] = xASL_quant_Basil(PWI, x)
+function [CBF_nocalib, ATT_map, resultFSL] = xASL_quant_Basil(PWI, x)
 %xASL_quant_Basil Perform quantification using FSL BASIL
 
-% FORMAT: [CBF_nocalib] = xASL_quant_Basil(PWI, x)
+% FORMAT: [CBF_nocalib, ATT_map, resultFSL] = xASL_quant_Basil(PWI, x)
 % 
 % INPUT:
 %   PWI             - image matrix of perfusion-weighted image (REQUIRED)
@@ -10,6 +10,7 @@ function [CBF_nocalib, resultFSL] = xASL_quant_Basil(PWI, x)
 % OUTPUT:
 % CBF_nocalib       - Quantified CBF image
 %                     (if there is no FSL/BASIL installed, we return the original PWI)
+% ATT_Map           - ATT map (if possible to calculate with multi-PLD)
 % resultFSL         - describes if the execution was successful
 %                     (0 = successful, NaN = no FSL/BASIL found, 1 or other = something failed)
 %
@@ -85,11 +86,19 @@ function [CBF_nocalib, resultFSL] = xASL_quant_Basil(PWI, x)
     fprintf('%s\n', 'The following warning (if mentioned above) can be ignored:');
     fprintf('%s\n', '/.../fsl/bin/basil: line 124: imcp: command not found');
     
-    pathBasilMean = xASL_adm_GetFileList(dirBasilOutput, '^mean_ftiss\.nii$', 'FPListRec');
-    pathBasilMean = pathBasilMean{end}; % we assume the latest iteration (alphabetically) is optimal. also converting cell to char array
+    pathBasilCBF = xASL_adm_GetFileList(dirBasilOutput, '^mean_ftiss\.nii$', 'FPListRec');
+    pathBasilCBF = pathBasilCBF{end}; % we assume the latest iteration (alphabetically) is optimal. also converting cell to char array
        
-    CBF_nocalib = xASL_io_Nifti2Im(pathBasilMean);
+    CBF_nocalib = xASL_io_Nifti2Im(pathBasilCBF);
         
+	pathBasilATT = xASL_adm_GetFileList(dirBasilOutput, '^mean_delttiss\.nii$', 'FPListRec');
+	if ~isempty(pathBasilATT)
+		pathBasilATT = pathBasilATT{end}; % we assume the latest iteration (alphabetically) is optimal. also converting cell to char array
+		ATT_map = xASL_io_Nifti2Im(pathBasilATT);
+	else
+		ATT_map = [];
+	end
+	
     %% 6. Scaling to physiological units
     % Note different to xASL_quant_SinglePLD since Fabber has T1 in seconds
     % and does not take into account labeling efficiency
--- a/Functions/xASL_quant_Basil.m
+++ b/Functions/xASL_quant_Basil.m
@@ -157,6 +157,7 @@ fprintf(FIDoptionFile, '--t1=%f\n', x.Q.TissueT1/1000);
 % Labelling type - PASL or pCASL
 switch lower(x.Q.LabelingType)
 	case 'pasl'
+		% PASL model is assumed by default and does not need to be specified in the config file
 		fprintf('BASIL: PASL model\n');
 		TIs = (unique(x.Q.Initial_PLD))'/1000;
 
@@ -176,7 +177,10 @@ switch lower(x.Q.LabelingType)
 		
 		% Either print bolus duration or unspecify it
 		if isfield(x.Q,'LabelingDuration') && x.Q.LabelingDuration
-			fprintf(FIDoptionFile, '--tau=%.2f\n', x.Q.LabelingDuration/1000);
+			if length(unique(x.Q.LabelingDuration))>1
+				warning('PASL multi-PLD currently supports only a single Labeling Duration');
+			end
+			fprintf(FIDoptionFile, '--tau=%.2f\n', x.Q.LabelingDuration(1)/1000);
 		else
 			% Bolus duration not know. If multi-TI, then try to infer it
 			if length(TIs) > 1
@@ -186,6 +190,7 @@ switch lower(x.Q.LabelingType)
 		end
 		
 	case {'casl','pcasl'}
+		% Specify that we run the PCASL/CASL model
 		fprintf(FIDoptionFile, '--casl\n');
 		fprintf('BASIL: (P)CASL model\n');
 		% Print all the PLDs and LabDurs
@@ -225,8 +230,11 @@ fprintf(FIDoptionFile, '--slicedt=%f\n', x.Q.BasilSliceReadoutTime/1000);
 
 if isfield(x.Q,'LookLocker') && x.Q.LookLocker 
 	if isfield(x.Q,'FlipAngle')
-		fprintf(option_file, '--FA=%f\n', x.Q.FlipAngle);
-		fprintf('BASIL: Flip angle for Look-Locker readout: %f\n', x.Q.FlipAngle);
+		if length(unique(x.Q.FlipAngle))>1
+			warning('Look-Locker quantification with multiple flip angles, e.g. QUASAR, is not implemented yet');
+		end
+		fprintf(option_file, '--FA=%f\n', x.Q.FlipAngle(1));
+		fprintf('BASIL: Flip angle for Look-Locker readout: %f\n', x.Q.FlipAngle(1));
 	else
 		warning('BASIL: Unknown flip angle for Look-Locker\n');
 	end
--- a/Functions/xASL_quant_Basil.m
+++ b/Functions/xASL_quant_Basil.m
@@ -59,7 +59,7 @@ function [CBF_nocalib, ATT_map, resultFSL] = xASL_quant_Basil(PWI, x)
     % FIXME would be good to have a brain mask at this point -> PM: if this would be a brainmask as well, we can skip creating a dummy input image here
     PWI(isnan(PWI)) = 0;
     
-    if (x.modules.asl.bTimeEncoded == 0) && (x.modules.asl.bMultiPLD == 0)
+    if ~x.modules.asl.bMultiPLD
         % SinglePLD
         xASL_io_SaveNifti(x.P.Path_PWI, pathBasilInput, PWI, [], 0); % use PWI path
     else
@@ -161,7 +161,7 @@ switch lower(x.Q.LabelingType)
 		TIs = (unique(x.Q.Initial_PLD))'/1000;
 
 		% Print all the TIs
-		if x.modules.asl.bTimeEncoded || x.modules.asl.bMultiPLD
+		if x.modules.asl.bMultiPLD
 			% For Time-encoded, we skip the first volume
 			if x.modules.asl.bTimeEncoded
 				TIs = TIs(2:end);
@@ -203,7 +203,7 @@ switch lower(x.Q.LabelingType)
 			LDs = LDs(2:end);
 		end
 		
-		if x.modules.asl.bTimeEncoded || x.modules.asl.bMultiPLD
+		if x.modules.asl.bMultiPLD
 			% For Time-encoded, we skip the first volume
 			for iPLD = 1:length(PLDs)
 				fprintf(FIDoptionFile, '--pld%d=%.2f\n', iPLD, PLDs(iPLD));
@@ -248,7 +248,7 @@ switch lower(x.Q.LabelingType)
 		fprintf(FIDoptionFile, '--bat=1.3\n');
 end
 
-if x.modules.asl.bTimeEncoded || x.modules.asl.bMultiPLD
+if x.modules.asl.bMultiPLD
 	% Multi-PLD or Time Encoded data allows to fit arrival times
 	
 	% Set the variance of ATT estimation
--- a/Functions/xASL_quant_Basil.m
+++ b/Functions/xASL_quant_Basil.m
@@ -87,6 +87,9 @@ function [CBF_nocalib, ATT_map, resultFSL] = xASL_quant_Basil(PWI, x)
     fprintf('%s\n', '/.../fsl/bin/basil: line 124: imcp: command not found');
     
     pathBasilCBF = xASL_adm_GetFileList(dirBasilOutput, '^mean_ftiss\.nii$', 'FPListRec');
+	if isempty(pathBasilCBF)
+		error('FSL BASIL failed');
+	end
     pathBasilCBF = pathBasilCBF{end}; % we assume the latest iteration (alphabetically) is optimal. also converting cell to char array
        
     CBF_nocalib = xASL_io_Nifti2Im(pathBasilCBF);
@@ -183,7 +186,11 @@ switch lower(x.Q.LabelingType)
 		
 		[PLDs, ind] = unique(x.Q.Initial_PLD);
 		PLDs = PLDs'/1000;
-		LDs  = (x.Q.LabelingDuration(ind))'/1000;
+		if length(x.Q.LabelingDuration)>1
+			LDs = (x.Q.LabelingDuration(ind))'/1000;
+		else
+			LDs = ones(size(PLDs))*x.Q.LabelingDuration/1000;
+		end
 		% For Time-encoded, we skip the first volume
 		if x.modules.asl.bTimeEncoded
 			PLDs = PLDs(2:end);
--- a/Functions/xASL_quant_Basil.m
+++ b/Functions/xASL_quant_Basil.m
@@ -121,34 +121,100 @@ end
 function [BasilOptions] = xASL_quant_Basil_Options(pathBasilOptions, x, PWI)
 % Save a Basil options file and store CLI options for Basil
 
-    %% Create option_file that contains options which are passed to Fabber
-    % basil_options is a character array containing CLI args for the Basil command
-    
-    FIDoptionFile = fopen(pathBasilOptions, 'w+');
-    BasilOptions = '';
+%% Create option_file that contains options which are passed to Fabber
+% basil_options is a character array containing CLI args for the Basil command
 
-    %% Basic acquisition and tissue parameters
-     % create CSV style text for TIs input
+FIDoptionFile = fopen(pathBasilOptions, 'w+');
+BasilOptions = '';
 
-     % print parameters
-     fprintf(FIDoptionFile, '# Basil options written by ExploreASL\n');
-     fprintf(FIDoptionFile, '--iaf=diff\n'); % as input is PWI
-     
-     TIs = (x.Q.LabelingDuration + x.Q.Initial_PLD)'/1000;
-     PLDAmount = length(TIs);
-     for TIsingle = 1:PLDAmount
-         fprintf(FIDoptionFile, ['--ti%d=%.2f\n'], TIsingle, TIs(TIsingle));
-     end
-     fprintf(FIDoptionFile, '--repeats=%i\n', size(PWI, 4)/PLDAmount);
-     fprintf(FIDoptionFile, '--t1b=%f\n', x.Q.BloodT1/1000);
-     fprintf(FIDoptionFile, '--tau=%f\n', x.Q.LabelingDuration/1000); % FIXME tau could be list
-     fprintf(FIDoptionFile, '--slicedt=%f\n', x.Q.BasilSliceReadoutTime/1000);
-     fprintf(FIDoptionFile, '--save-model-fit\n');
+fprintf(FIDoptionFile, '# Basil options written by ExploreASL\n');
+fprintf(FIDoptionFile, '--iaf=diff\n'); % as input is PWI
+
+%% Basic tissue parameters
+fprintf(FIDoptionFile, '--t1b=%f\n', x.Q.BloodT1/1000);
+
+%% Basic acquisition parameters
+
+% Labelling type - PASL or pCASL
+switch lower(x.Q.LabelingType)
+	case 'pasl'
+		fprintf('Basil: PASL model\n');
+		TIs = (unique(x.Q.Initial_PLD))'/1000;
+
+		% Print all the TIs
+		if x.modules.asl.bTimeEncoded || x.modules.asl.bMultiPLD
+			% For Time-encoded, we skip the first volume
+			if x.modules.asl.bTimeEncoded
+				TIs = TIs(2:end);
+			end
+			for iTI = 1:length(TIs)
+				fprintf(FIDoptionFile, '--ti%d=%.2f\n', iTI, TIs(iTI));
+			end
+			
+		else
+			fprintf(FIDoptionFile, '--ti=%.2f\n', TIs);
+		end
+		
+		% Either print bolus duration or unspecify it
+		if isfield(x.Q,'LabelingDuration') && x.Q.LabelingDuration
+			fprintf(FIDoptionFile, '--tau=%.2f\n', x.Q.LabelingDuration/1000);
+		else
+			% Bolus duration not know. If multi-TI, then try to infer it
+			if length(TIs) > 1
+				fprintf(FIDoptionFile, '--infertau\n');
+				fprintf('Basil: Infer bolus duration component\n')
+			end
+		end
+		
+	case {'casl','pcasl'}
+		fprintf(FIDoptionFile, '--casl\n');
+		fprintf('Basil: CASL/PCASL model\n');
+		% Print all the PLDs and LabDurs
+		
+		[PLDs, ind] = unique(x.Q.Initial_PLD);
+		PLDs = PLDs'/1000;
+		LDs  = (x.Q.LabelingDuration(ind))'/1000;
+		% For Time-encoded, we skip the first volume
+		if x.modules.asl.bTimeEncoded
+			PLDs = PLDs(2:end);
+			LDs = LDs(2:end);
+		end
+		
+		if x.modules.asl.bTimeEncoded || x.modules.asl.bMultiPLD
+			% For Time-encoded, we skip the first volume
+			for iPLD = 1:length(PLDs)
+				fprintf(FIDoptionFile, '--pld%d=%.2f\n', iPLD, PLDs(iPLD));
+			end
+			for iLD = 1:length(LDs)
+				fprintf(FIDoptionFile, '--tau%d=%.2f\n', iLD, LDs(iLD));
+			end
+		else
+			fprintf(FIDoptionFile, '--tau=%.2f\n', LDs);
+		end
+end
+
+% Right now, we assume that we have averaged over PLDs
+%fprintf(FIDoptionFile, '--repeats=%i\n', size(PWI, 4)/PLDAmount);
+fprintf(FIDoptionFile, '--repeats=1\n');
+	
+% Slice-timing
+fprintf(FIDoptionFile, '--slicedt=%f\n', x.Q.BasilSliceReadoutTime/1000);
+
+if isfield(x.Q,'LookLocker') && x.Q.LookLocker && isfield(x.Q,'FlipAngle')
+	fprintf(option_file, '--FA=%f\n', x.Q.FlipAngle);
+	fprintf('Basil: Flip angle for Look-Locker readout: %f\n', x.Q.FlipAngle);
+end
+	 
 
 
+
+
+
+
+
+
+	 
     %% FIXME Aquisition options we might be able to use in the future
-    %fprintf(option_file, '--FA=%f\n', fa);
-    %fprintf('Basil: Flip angle for look-locker readout: %f\n', fa);
     %fprintf(option_file, '--sliceband=%i\n', sliceband);
     %fprintf('Basil: Multi-band setup with number of slices per band: %i\n', slicedband);
 
@@ -159,14 +225,7 @@ function [BasilOptions] = xASL_quant_Basil_Options(pathBasilOptions, x, PWI)
     %fprintf(option_file, '--t1im=%s\n', t1im)
     %fprintf('Basil: Using supplied T1 (tissue) image in BASIL: %s\n', $t1im)
 
-    % Labelling type - PASL or pCASL
-    switch lower(x.Q.LabelingType)
-        case 'pasl'
-            fprintf('Basil: PASL model\n');
-        case 'casl'
-            fprintf(FIDoptionFile, '--casl\n');
-            fprintf('Basil: CASL/PcASL model\n');
-    end
+    fprintf(FIDoptionFile, '--save-model-fit\n');
 
 
     %% Model option - 1 or 2 compartment
@@ -185,16 +244,6 @@ function [BasilOptions] = xASL_quant_Basil_Options(pathBasilOptions, x, PWI)
             fprintf(FIDoptionFile, '--t1=%f\n', x.Q.TissueT1/1000);
             fprintf('Basil: 2-compartment - ATT=%fs\n', x.Q.ATT/1000);
     end
-
-    
-    %%  Bolus duration\labeling duration typically fixed but can be inferred
-    if isfield(x.Q,'BasilInferTau') && x.Q.BasilInferTau
-         fprintf(FIDoptionFile, '--infertau\n');
-         fprintf('Basil: Infer bolus duration component\n')
-    else
-        fprintf('Basil: Fixed bolus duration component\n')
-    end
-
     
     %% ATT and arterial component inference only possible with multi-PLD
     if (x.modules.asl.bTimeEncoded == 0) && (x.modules.asl.bMultiPLD == 0)
--- a/Functions/xASL_quant_Basil.m
+++ b/Functions/xASL_quant_Basil.m
@@ -35,7 +35,7 @@ function [CBF_nocalib, ATT_map, resultFSL] = xASL_quant_Basil(PWI, x)
     
 
     %% Admin
-    fprintf('%s\n','Quantification CBF using FSL Basil:');    
+    fprintf('%s\n','Quantification CBF using FSL BASIL:');    
     
     %% 1. Define paths
     pathBasilInput = fullfile(x.dir.SESSIONDIR, 'PWI4D_BasilInput.nii');
@@ -132,8 +132,14 @@ end
 
 function [BasilOptions] = xASL_quant_Basil_Options(pathBasilOptions, x, PWI)
 % Save a Basil options file and store CLI options for Basil
+% 1. Create an option file
+% 2. Basic tissue parameters
+% 3. Basic acquisition parameters
+% 4. Model fiting parameters
+% 5. Extra features on demand
+% 6. Save BASIL options file
 
-%% Create option_file that contains options which are passed to Fabber
+%% 1. Create option_file that contains options which are passed to Fabber
 % basil_options is a character array containing CLI args for the Basil command
 
 FIDoptionFile = fopen(pathBasilOptions, 'w+');
@@ -142,16 +148,16 @@ BasilOptions = '';
 fprintf(FIDoptionFile, '# Basil options written by ExploreASL\n');
 fprintf(FIDoptionFile, '--iaf=diff\n'); % as input is PWI
 
-%% Basic tissue parameters
+%% 2. Basic tissue parameters
 fprintf(FIDoptionFile, '--t1b=%f\n', x.Q.BloodT1/1000);
 fprintf(FIDoptionFile, '--t1=%f\n', x.Q.TissueT1/1000);
 
-%% Basic acquisition parameters
+%% 3. Basic acquisition parameters
 
 % Labelling type - PASL or pCASL
 switch lower(x.Q.LabelingType)
 	case 'pasl'
-		fprintf('Basil: PASL model\n');
+		fprintf('BASIL: PASL model\n');
 		TIs = (unique(x.Q.Initial_PLD))'/1000;
 
 		% Print all the TIs
@@ -175,13 +181,13 @@ switch lower(x.Q.LabelingType)
 			% Bolus duration not know. If multi-TI, then try to infer it
 			if length(TIs) > 1
 				fprintf(FIDoptionFile, '--infertau\n');
-				fprintf('Basil: Infer bolus duration component\n')
+				fprintf('BASIL: Infer bolus duration component\n')
 			end
 		end
 		
 	case {'casl','pcasl'}
 		fprintf(FIDoptionFile, '--casl\n');
-		fprintf('Basil: CASL/PCASL model\n');
+		fprintf('BASIL: (P)CASL model\n');
 		% Print all the PLDs and LabDurs
 		
 		[PLDs, ind] = unique(x.Q.Initial_PLD);
@@ -217,12 +223,16 @@ fprintf(FIDoptionFile, '--repeats=1\n');
 % Slice-timing
 fprintf(FIDoptionFile, '--slicedt=%f\n', x.Q.BasilSliceReadoutTime/1000);
 
-if isfield(x.Q,'LookLocker') && x.Q.LookLocker && isfield(x.Q,'FlipAngle')
-	fprintf(option_file, '--FA=%f\n', x.Q.FlipAngle);
-	fprintf('Basil: Flip angle for Look-Locker readout: %f\n', x.Q.FlipAngle);
+if isfield(x.Q,'LookLocker') && x.Q.LookLocker 
+	if isfield(x.Q,'FlipAngle')
+		fprintf(option_file, '--FA=%f\n', x.Q.FlipAngle);
+		fprintf('BASIL: Flip angle for Look-Locker readout: %f\n', x.Q.FlipAngle);
+	else
+		warning('BASIL: Unknown flip angle for Look-Locker\n');
+	end
 end
 	 
-%% Model fiting parameters
+%% 4. Model fiting parameters
 
 % This helps avoid failure on the structural-space image
 fprintf(FIDoptionFile, '--allow-bad-voxels\n');
@@ -251,10 +261,10 @@ end
 
 % 	%% FIXME Aquisition options we might be able to use in the future
 %     %fprintf(option_file, '--sliceband=%i\n', sliceband);
-%     %fprintf('Basil: Multi-band setup with number of slices per band: %i\n', slicedband);
+%     %fprintf('BASIL: Multi-band setup with number of slices per band: %i\n', slicedband);
 % 	
 % 	%fprintf(option_file, '--t1im=%s\n', t1im)
-%     %fprintf('Basil: Using supplied T1 (tissue) image in BASIL: %s\n', $t1im)
+%     %fprintf('BASIL: Using supplied T1 (tissue) image in BASIL: %s\n', $t1im)
 
 %     %% Noise specification
 %     % For small numbers of time points we need an informative noise prior. 
@@ -265,48 +275,48 @@ end
 % 
 %     if size(PWI, 4) < 5
 %         x.Q.BasilNoisePrior = 1;
-%         fprintf('Basil: Small number of volumes (%i < 5): informative noise prior will be used\n', size(PWI, 4));
+%         fprintf('BASIL: Small number of volumes (%i < 5): informative noise prior will be used\n', size(PWI, 4));
 %     end
 %     if isfield(x.Q,'BasilNoisePrior') && x.Q.BasilNoisePrior
 %         % Use an informative noise prior
 %         if ~isfield(x.Q,'BasilNoiseSD') || ~x.Q.BasilNoiseSD
-%             fprintf('Basil: Using SNR of %f to set noise std dev\n', x.Q.BasilSNR);
+%             fprintf('BASIL: Using SNR of %f to set noise std dev\n', x.Q.BasilSNR);
 %             % Estimate signal magntiude FIXME brain mask assume half of voxels
 %             mag_max = max(PWI, [], 4);
 %             brain_mag = 2*xASL_stat_MeanNan(mag_max(:));
-%             fprintf('Basil: Mean maximum signal across brain: %f\n', brain_mag);
+%             fprintf('BASIL: Mean maximum signal across brain: %f\n', brain_mag);
 %             % This will correspond to whole brain CBF (roughly) - about 0.5 of GM
 %             x.Q.BasilNoiseSD = sqrt(brain_mag * 2 / x.Q.BasilSNR);
 %         end
-%         fprintf('Basil: Using a prior noise std.dev. of: %f\n', x.Q.BasilNoiseSD);
+%         fprintf('BASIL: Using a prior noise std.dev. of: %f\n', x.Q.BasilNoiseSD);
 %         fprintf(FIDoptionFile, '--prior-noise-stddev=%f\n', x.Q.BasilNoiseSD);
 %     end
 
 
-%% Extra features on demand
+%% 5. Extra features on demand
 if isfield(x,'BasilSpatial') && x.Q.BasilSpatial
-	fprintf('Basil: Instructing BASIL to use automated spatial smoothing\n');
+	fprintf('BASIL: Instructing BASIL to use automated spatial smoothing\n');
 	BasilOptions = [BasilOptions ' --spatial'];
 end
 
 if isfield(x.Q,'BasilInferT1') && x.Q.BasilInferT1
 	fprintf(FIDoptionFile, '--infert1\n');
-	fprintf('Basil: Instructing BASIL to infer variable T1 values\n');
+	fprintf('BASIL: Instructing BASIL to infer variable T1 values\n');
 end
 
 if isfield(x.Q,'BasilInferATT') && x.Q.BasilInferATT
 	fprintf(FIDoptionFile, '--inferart\n');
-	fprintf('Basil: Infer arterial component');
-	fprintf('Basil: Variable arterial component arrival time');
+	fprintf('BASIL: Infer arterial component');
+	fprintf('BASIL: Variable arterial component arrival time');
 end
 
 if isfield(x.Q,'BasilExch')
-	fprintf('Basil: Using exchange model: %s\n', x.Q.BasilExch);
+	fprintf('BASIL: Using exchange model: %s\n', x.Q.BasilExch);
 	fprintf(FIDoptionFile, '--exch=%s\n', x.Q.BasilExch);
 end
 
 if isfield(x.Q,'BasilDisp')
-	fprintf('Basil: Using dispersion model: %s\n', x.Q.BasilDisp);
+	fprintf('BASIL: Using dispersion model: %s\n', x.Q.BasilDisp);
 	fprintf(FIDoptionFile, '--disp=%s\n', x.Q.BasilDisp);
 end
 
@@ -315,7 +325,7 @@ if isfield(x.Q,'BasilDebug') && x.Q.BasilDebug
 end
 
 
-%% Save Basil options file
+%% 6. Save Basil options file
 fclose(FIDoptionFile);
     
 end
--- a/Functions/xASL_quant_Basil.m
+++ b/Functions/xASL_quant_Basil.m
@@ -40,8 +40,7 @@ function [CBF_nocalib, resultFSL] = xASL_quant_Basil(PWI, x)
     pathBasilInput = fullfile(x.dir.SESSIONDIR, 'PWI4D_BasilInput.nii');
     pathBasilOptions = fullfile(x.dir.SESSIONDIR, 'Basil_ModelOptions.txt');
     dirBasilOutput = fullfile(x.dir.SESSIONDIR, 'BasilOutput');
-    
-    
+        
     %% 2. Delete previous BASIL output
     xASL_adm_DeleteFileList(x.dir.SESSIONDIR, '(?i)^basilOutput.*$', 1, [0 Inf]);
     FolderList = xASL_adm_GetFileList(x.dir.SESSIONDIR, '(?i)^basilOutput.*$', 'FPList', [0 Inf], 1);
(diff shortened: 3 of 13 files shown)


Functions/xASL_quant_Basil.m
```diff
@@ -59,21 +59,18 @@
     % FIXME would be good to have a brain mask at this point -> PM: if this would be a brainmask as well, we can skip creating a dummy input image here
     PWI(isnan(PWI)) = 0;
     
-    if size(PWI,4) == 1 
+    if (x.modules.asl.bTimeEncoded == 0) && (x.modules.asl.bMultiPLD == 0)
```
MichaelStritt, line 62: Probably a stupid question, but we are sure that those sub-fields are always defined, right?

Modules/xASL_module_ASL.m
```diff
@@ -187,6 +187,22 @@
 	x.Q.NumberEchoTimes = 1;
 end
 
+%% D6. DeltaM parsing - check if all/some volumes are deltams
```
MichaelStritt, line 190: We could add the D6 in the header description where the other sections are listed, too 😃

MichaelStritt (approved):
You nicely cleaned up the code. Especially the switch statements and the io of the functions seems a lot better now. All headers look fine. I added two minor comments. Would be good if someone with an FSL/BASIL installation (@BeatrizPadrela or @MDijsselhof) could test this though (either now or during the release testing).

Functions/xASL_quant_Basil.m
jan-petr: No - a very important question! Yes, they are always defined at the start of ASL_module, so that we don't have to check all the time if the field exist. These fields are not defined in dataPar, but rather derived at the start of the module from the data. Good that you check these things...

jan-petr:
We have several flavors to test this with. I've tried them, but a re-test by others would be welcome:
These work nicely and give CBF and TT maps
- GE_PCASL_3Dspiral_25.0LX_eASLPLD3_1
- GE_PCASL_3Dspiral_25.0LX_eASLPLD7_1
- Philips_PCASL_2DEPI_5.6.1.2_Hadamard_1

These two crashes in FABBER as multi-TE is not yet supported - we will do that later
- Siemens_PCASL_3DGRASE_VD13A_Hadamard8FME_1
- Siemens_PCASL_3DGRASE_VD13A_Hadamard4FME_1

This one crashes badly, but the reason is that it doesn't have M0 and has BSup. And our BSup correction for M0 doesn't yet work for multi-PLD. I already created a future issue for that in 2.1.0
- Siemens_PCASL_3DGRASE_VB17A_multiPLD_1

This one used to crash because there are way too many volumes for motion correction. However, I have disabled motion correction for Look-Locker, so it should now run. And we will adapt the motion correction for Look-Locker in the future - an issue for that has been already created:
- Philips_PASL_2DEPI_5.4.0.2-LookLocker_1

@MichaelStritt Can you install FSL in WSL?
@BeatrizPadrela Can you test these yourself as well?
@MDijsselhof Can you test your testing set if it works?
@everybody: Any other ideas or testing data?

Modules/xASL_module_ASL.m
jan-petr: Good point - added. I though about adding that already previously, and I knew that being lazy doesn't pay off and somebody will find out :D

Functions/xASL_quant_GetMeanControl.m
```diff
@@ -36,29 +36,36 @@
...
-		% Find all dynamics with that PLD
-		imMeanControl = imMeanControl(:,:,:,Initial_PLD==idealPLD);
 	end
+	
+	% Sort the unique PLDs and pick the ideal PLD
+	idealPLD = sort(unique(Initial_PLD),'ascend');
+	
+	if (isfield(x.Q,'LookLocker') && x.Q.LookLocker) || x.modules.asl.bDeltaM
+		% For Look-Locker, get the middle one
+		idealPLD = idealPLD(round(numel(idealPLD)/2));
```
HenkMutsaerts, line 56: We discussed with @MDijsselhof that it would be most comparable with single PLD to use the PLD closest to the white paper PLD (let's say closest to a PLD of 2000 ms).

Plus, this would decrease the variance between multiPLD types

Functions/xASL_quant_MultiPLD.m
```diff
@@ -263,21 +221,19 @@
...
 
 %% 6    Print parameters used
 fprintf('%s\n',' model with parameters:');
 
 if x.Q.ApplyQuantification(3)
     switch lower(x.Q.LabelingType)
         case 'pasl'
-            fprintf('%s',['TI1 = ' num2str(x.Q.LabelingDuration) ' ms, ']);
-            fprintf('%s',['TI (ms) = ' num2str(x.Q.Initial_PLD')]);  % transpose for multiPLD
+            fprintf('%s\n',['TI1 = ' num2str(unique(x.Q.LabelingDuration(:)')) ' ms,']);
```
HenkMutsaerts, line 232: Perhaps nice to add to `xASL_num2str` that it will always do the `(:)'`.

Modules/SubModule_ASL/xASL_wrp_Quantify.m
```diff
@@ -390,13 +390,14 @@ function xASL_wrp_Quantify(x, PWI_Path, OutputPath, M0Path, SliceGradientPath)
 
 %% ------------------------------------------------------------------------------------------------
 %% 8.   Perform Quantification
-if size(PWI,4) == 1 % single PLD quantification
+if (x.modules.asl.bTimeEncoded == 0) && (x.modules.asl.bMultiPLD == 0)% single PLD quantification
```
HenkMutsaerts, line 393: same as above

Modules/SubModule_ASL/xASL_wrp_Quantify.m
```diff
@@ -390,13 +390,14 @@ function xASL_wrp_Quantify(x, PWI_Path, OutputPath, M0Path, SliceGradientPath)
...
     fprintf('%s\n',['Performing single PLD quantification']);
     [~, CBF] = xASL_quant_SinglePLD(PWI, M0_im, SliceGradient, x, x.Q.bUseBasilQuantification); % also runs BASIL, but only in native space!
+	ATT = [];
```
HenkMutsaerts, line 396: or `NaN`?

Modules/SubModule_ASL/xASL_wrp_Quantify.m
```diff
@@ -421,6 +422,15 @@ function xASL_wrp_Quantify(x, PWI_Path, OutputPath, M0Path, SliceGradientPath)
 
 xASL_io_SaveNifti(PWI_Path, OutputPath, CBF, 32, 0);
 
+if ~isempty(ATT)
```
HenkMutsaerts, line 425: Perhaps nicer to have `pathOutput_CBF` and `pathOutput_ATT` from the start of this wrapper, this is a bit quick and dirty

Modules/SubModule_ASL/xASL_wrp_Quantify.m
```diff
@@ -456,5 +466,8 @@ function xASL_wrp_Quantify(x, PWI_Path, OutputPath, M0Path, SliceGradientPath)
     end
     
     xASL_spm_deformations(x, {x.P.Path_CBF}, {x.P.Pop_Path_CBF}, [], [], AffineTransfPath, x.P.Path_y_ASL);
+	if xASL_exist(x.P.Path_TT,'file')
+		xASL_spm_deformations(x, {x.P.Path_TT}, {x.P.Pop_Path_TT}, [], [], AffineTransfPath, x.P.Path_y_ASL);
```
HenkMutsaerts, line 470: Should it be `Path_ATT` instead? `Path_TT` I used in the past for the FEAST output, we could clean this up in the future

Modules/SubModule_ASL/xASL_wrp_RealignASL.m
```diff
@@ -61,21 +61,18 @@ function xASL_wrp_RealignASL(x,bSubtraction)
 PercExcl = NaN;
 MinimumtValue = NaN;
 
-if x.modules.asl.bMultiPLD || x.modules.asl.bMultiTE
+if x.modules.asl.bMultiPLD || x.modules.asl.bMultiTE || x.modules.asl.bDeltaM
```
HenkMutsaerts, line 64: Not sure what `bDeltaM` is, but throughout ExploreASL it assumes that if `ASL4D` has only a single volume, this is a `deltaM` volume. Motion correction is disabled by default for a single volume.

BTW, why should ENABLE be disabled for DeltaM images? It should rather skip the subtraction.

Ah I think I now get it, you needed `bDeltaM` for GE eASL. But I would have created a separate issue for this then, because now parts of the pipeline use `bDeltaM` and other parts don't.

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
```diff
@@ -220,6 +220,23 @@ function xASL_wrp_ResampleASL(x)
         % Apparently, subtraction was already done on the scanner/reconstruction
         fprintf('%s\n', PathPWI{iSpace});
         xASL_io_SaveNifti(PathASL{iSpace}, PathPWI{iSpace}, ASL_im, 32, false);
+	elseif x.modules.asl.bDeltaM
+		% We have subtraction images
```
HenkMutsaerts, line 224: Please be clear in the comments (like in a paper, otherwise the comments are not helpful!). You meant here that we don't have control-label images in the raw data, but only subtraction images, right?

(because otherwise, we always have subtraction images somewhere

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
```diff
@@ -220,6 +220,23 @@ function xASL_wrp_ResampleASL(x)
         % Apparently, subtraction was already done on the scanner/reconstruction
         fprintf('%s\n', PathPWI{iSpace});
         xASL_io_SaveNifti(PathASL{iSpace}, PathPWI{iSpace}, ASL_im, 32, false);
+	elseif x.modules.asl.bDeltaM
+		% We have subtraction images 
+		
+		% For multi-PLD, save both PWI4D and PWI
```
HenkMutsaerts, line 226: Also for single PLD, we should save both! This should have been there already, why did you change this?

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
```diff
@@ -249,20 +266,22 @@ function xASL_wrp_ResampleASL(x)
...
         % Average PWI - multiPLD
         if isfield(x.modules.asl,'bMultiPLD') && x.modules.asl.bMultiPLD 
-            % multiPLD PWI
-            nRepeatsPLD = size(ASL_im,4)/length(x.Q.Initial_PLD); % we still need to create a single Repeats in X struct
-			PWI = zeros(size(ASL_im,1), size(ASL_im,2), size(ASL_im,3), length(x.Q.Initial_PLD)); % preallocate PWI
-            for nPLD = 1:length(x.Q.Initial_PLD)
-                iStartRepeatsPLD = 1 + (nPLD-1) * nRepeatsPLD; % Location of the first repeat for a given single PLD
-                iEndRepeatsPLD = nPLD * nRepeatsPLD; % Location of the last repeat for a given single PLD
-                PWI(:,:,:,nPLD) = xASL_stat_MeanNan(ASL_im(:,:,:,iStartRepeatsPLD:iEndRepeatsPLD), 4); % Averaged PWI4D 
+			% Skip every second PLD due to the subtraction
```
HenkMutsaerts, line 275: what do you mean by this? this is confusing to me. You probably meant:

% Skip every other value because x.Q.Initial_PLD was stored for both control and label images and we need it now for the pairwise subtractions only

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
```diff
@@ -249,20 +266,22 @@ function xASL_wrp_ResampleASL(x)
...
+			subtractedPLD = x.Q.Initial_PLD(1:2:end);
```
HenkMutsaerts, line 276: Confusing as well, you have not "subtracted PLDs". Plus, these are still initial PLDs right?

So how about `Initial_PLD_PWI` ?

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
```diff
@@ -249,20 +266,22 @@ function xASL_wrp_ResampleASL(x)
...
+        
+			% After averaging across PLDs, we'll obtain these unique PLDs
```
HenkMutsaerts, line 278: Again here, please make the comment and the variable name clearer

Modules/xASL_module_ASL.m
```diff
@@ -41,6 +41,7 @@
 % - D3. Multi-PLD parsing
 % - D4. TimeEncoded parsing
 % - D5. Multi-TE parsing
+% - D6. DeltaM parsing - check if all/some volumes are deltams
```
HenkMutsaerts, line 44: Nice, but then this should include that if we have a single volume only, we assume it is a `deltaM`

Modules/xASL_module_ASL.m
```diff
@@ -187,6 +188,22 @@
 	x.Q.NumberEchoTimes = 1;
 end
 
+%% D6. DeltaM parsing - check if all/some volumes are deltams
+% If TSV file exist
+[pathDir, pathASL4D] = fileparts(x.P.Path_ASL4D);
+pathASL4Dcontext = fullfile(pathDir, [pathASL4D '_aslcontext.tsv']);
```
HenkMutsaerts, line 194: perhaps good to use the same file naming as the rest, so this would be `x.P.Path_ASL4Dcontext`. Even better if you can add this in the general file used for creating all paths (`xASL_ini_FileSystem`), so you don't have to repeat this in different places

Modules/xASL_module_ASL.m
```diff
@@ -187,6 +188,22 @@
...
+% We don't have a subtraction image by default
```
HenkMutsaerts, line 195: Unless there is only a single volume

Modules/xASL_module_ASL.m
```diff
@@ -187,6 +188,22 @@
...
+x.modules.asl.bDeltaM = 0;
+if xASL_exist(pathASL4Dcontext,'file')
+	% Load TSV file
+	aslContext = xASL_tsvRead(pathASL4Dcontext);
+	bidsPar = xASL_bids_Config;
+	% Check for presence of deltaM subtraction volumes
+	if numel(regexpi(strjoin(aslContext(2:end)),bidsPar.stringDeltaM)) > 0
```
HenkMutsaerts, line 202: Can we not find a better name for `stringDeltaM`? Is this supposed to contain a string with all `deltaM` volumes? Why not logic?

And if you have at least 1 deltaM (according to this `if` statement), does this mean that we assume that all `ASL4D` volumes are `deltaM`? In that case I would add a warning (unless vendor is GE perhaps)

HenkMutsaerts (changes requested):
In general, this issue contains multiple issues, making it harder to review this.

`xASL_wrp_Resample` already saved both `PWI` and `PWI4D`, and it should still do this, irregardless of the number of PLDs. At the end of the quantification wrapper we delete PWI4D as it can take up quite some space. And for BASIL we always provide PWI4D, regardless the number of PLDs. (at least, this is how it should be and how I programmed this)

Functions/xASL_bids_parms2BIDS.m
```diff
@@ -54,7 +54,7 @@
                           500        20000            10000         5000               5000                 5000            400               5000      500  5000       400                10000];% Maximum in ms   
 					  
 % Fields that are entered under the subfield 'Q' for xASL on the output
-xASLqFields = {'LabelingType' 'Initial_PLD' 'BackGrSupprPulses' 'LabelingDuration' 'SliceReadoutTime' 'NumberOfAverages' 'BloodT1'...
+xASLqFields = {'LabelingType' 'Initial_PLD' 'BackGrSupprPulses' 'LookLocker' 'LabelingDuration' 'SliceReadoutTime' 'NumberOfAverages' 'BloodT1'...
```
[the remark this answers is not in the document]
jan-petr: Reference added. LookLocker is a BIDS field and is directly a data property. Makes no sense to change it to `bLookLocker`.

Functions/xASL_quant_Basil.m
```diff
@@ -1,7 +1,7 @@
-function [CBF_nocalib, resultFSL] = xASL_quant_Basil(PWI, x)
+function [CBF_nocalib, ATT_map, resultFSL] = xASL_quant_Basil(PWI, x)
```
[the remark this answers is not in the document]
jan-petr: Not sure what you talk about. `isvar` is not used in the function. We only check existence of ATT_map output from BASIL. That cannot be initialized, but has to be checked.

Functions/xASL_quant_Basil.m
jan-petr: Currently, we have multiPLD and TimeEncoded as two different things. I'm fine to redefine this and change this as TimeEncoded sub-case of multiPLD. But lets discuss this - note that these are internal parameters.

The second point. Makes sense to always save as PWI4D, but I didn't write this if-statement. Shall we completely remove it then?

Functions/xASL_quant_GetMeanControl.m
```diff
@@ -36,29 +36,36 @@
...
     imMeanControl = imASLTimeSeries(:,:,:,1:blockSize:end);
-else
-	% Create mean control in native space (works also for multi-PLD)
-	imMeanControl = xASL_quant_GetControlLabelOrder(imASLTimeSeries);
 	
-	% For multi-PLD, we need to select one PLD
-	if isfield(x.modules.asl,'bMultiPLD') && x.modules.asl.bMultiPLD
+elseif isfield(x.modules.asl,'bMultiPLD') && x.modules.asl.bMultiPLD
+	% Create mean control in native space
+	if x.modules.asl.bDeltaM
```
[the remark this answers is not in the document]
jan-petr: We should probably rename to `bContainsDeltaM`.
Yes, of course, it is automatically detected option. But we have the automatic detection centralized and then use this variable, instead of re-detecting at each point. 
Previously, this was wrong as it detected this as having a single volume - which is not with multi-PLD not an effective thing to do.

Functions/xASL_quant_MultiPLD.m
```diff
@@ -56,7 +57,11 @@
 end
 
 if nargin<5 || isempty(bUseBasilQuantification)
-    bUseBasilQuantification = false;
+    bUseBasilQuantification = true;
+end
+
+if bUseBasilQuantification == false
```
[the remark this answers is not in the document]
jan-petr: Fixed

Functions/xASL_quant_MultiPLD.m
```diff
@@ -105,14 +106,10 @@
...
-				if max(SliceReadoutTime)>0 && length(SliceReadoutTime) > 1
-					x.Q.BasilSliceReadoutTime = SliceReadoutTime(2)-SliceReadoutTime(1);
-				else
-					x.Q.BasilSliceReadoutTime = 0;
-				end
+			if max(SliceReadoutTime)>0 && length(SliceReadoutTime) > 1
+				x.Q.BasilSliceReadoutTime = SliceReadoutTime(2)-SliceReadoutTime(1);
 			else
-				ScaleImage = ScaleImage.*(x.Q.Initial_PLD + SliceReadoutTime(imSliceNumber)); % effective/net PLD            
+				x.Q.BasilSliceReadoutTime = 0;
```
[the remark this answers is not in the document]
jan-petr: No it shouldn't. Cause BASIL wouldn't read it then correctly - most likely.

Functions/xASL_quant_MultiPLD.m
```diff
@@ -121,48 +118,9 @@
...
-                error('Unknown x.Q.nCompartments');
-        end
-
-
-        %% 3    Scaling to physiological units
-        ScaleImage = ScaleImage.*60000.*100.*x.Q.Lambda;
-        % (For some reason, GE sometimes doesn't need the 1 gr->100 gr conversion)
-        % & old Siemens sequence also didn't need the 1 gr->100 gr conversion
-    end
+	[PWI, ATT] = xASL_quant_Basil(PWI, x);
```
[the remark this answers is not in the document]
jan-petr: Yes. Known issue. We indeed plan to merge those. #867

Functions/xASL_quant_Basil.m
```diff
@@ -89,19 +86,29 @@
     fprintf('%s\n', 'The following warning (if mentioned above) can be ignored:');
     fprintf('%s\n', '/.../fsl/bin/basil: line 124: imcp: command not found');
     
-    pathBasilMean = xASL_adm_GetFileList(dirBasilOutput, '^mean_ftiss\.nii$', 'FPListRec');
-    pathBasilMean = pathBasilMean{end}; % we assume the latest iteration (alphabetically) is optimal. also converting cell to char array
+    pathBasilCBF = xASL_adm_GetFileList(dirBasilOutput, '^mean_ftiss\.nii$', 'FPListRec');
+	if isempty(pathBasilCBF)
+		error('FSL BASIL failed');
```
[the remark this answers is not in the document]
jan-petr: Yes. In a future issue.

Functions/xASL_quant_Basil.m
```diff
@@ -89,19 +86,29 @@
...
        
-    CBF_nocalib = xASL_io_Nifti2Im(pathBasilMean);
+    CBF_nocalib = xASL_io_Nifti2Im(pathBasilCBF);
         
+	pathBasilATT = xASL_adm_GetFileList(dirBasilOutput, '^mean_delttiss\.nii$', 'FPListRec');
+	if ~isempty(pathBasilATT)
+		pathBasilATT = pathBasilATT{end}; % we assume the latest iteration (alphabetically) is optimal. also converting cell to char array
+		ATT_map = xASL_io_Nifti2Im(pathBasilATT);
+	else
+		ATT_map = [];
```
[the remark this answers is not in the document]
jan-petr: Yes. And if ATT map is not calculated (single-PLD case), then we don't have missing values, but missing entire variable. Hence []. Or not?

Functions/xASL_quant_Basil.m
```diff
@@ -123,170 +130,193 @@
 
 
 
-function [BasilOptions] = xASL_quant_Basil_Options(pathBasilOptions, x, PWI, bMultiPLD)
+function [BasilOptions] = xASL_quant_Basil_Options(pathBasilOptions, x, PWI)
```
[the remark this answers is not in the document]
jan-petr: Yes. I already divided and restructured it. Now adding section numbers.

Functions/xASL_quant_Basil.m
```diff
@@ -123,170 +130,193 @@
...
+%% Basic acquisition parameters
+
+% Labelling type - PASL or pCASL
+switch lower(x.Q.LabelingType)
+	case 'pasl'
+		fprintf('Basil: PASL model\n');
+		TIs = (unique(x.Q.Initial_PLD))'/1000;
+
+		% Print all the TIs
+		if x.modules.asl.bTimeEncoded || x.modules.asl.bMultiPLD
```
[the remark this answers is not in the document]
jan-petr: As above.

Functions/xASL_quant_Basil.m
```diff
@@ -123,170 +130,193 @@
...
+			% Bolus duration not know. If multi-TI, then try to infer it
+			if length(TIs) > 1
+				fprintf(FIDoptionFile, '--infertau\n');
+				fprintf('Basil: Infer bolus duration component\n')
+			end
+		end
+		
+	case {'casl','pcasl'}
+		fprintf(FIDoptionFile, '--casl\n');
+		fprintf('Basil: CASL/PCASL model\n');
```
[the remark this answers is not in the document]
jan-petr: Done.

Functions/xASL_quant_Basil.m
```diff
@@ -123,170 +130,193 @@
...
+		[PLDs, ind] = unique(x.Q.Initial_PLD);
+		PLDs = PLDs'/1000;
+		if length(x.Q.LabelingDuration)>1
+			LDs = (x.Q.LabelingDuration(ind))'/1000;
+		else
+			LDs = ones(size(PLDs))*x.Q.LabelingDuration/1000;
+		end
+		% For Time-encoded, we skip the first volume
+		if x.modules.asl.bTimeEncoded
+			PLDs = PLDs(2:end);
```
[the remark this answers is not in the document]
jan-petr: We have this twice - for PASL and PCASL. And for PCASL, we need to do this for LD as well. So doing it at one location for PASL and PCASL with PLD would mess it up for LD. So I would keep this. Unless you see it more than twice.

Functions/xASL_quant_Basil.m
```diff
@@ -123,170 +130,193 @@
...
 
-    
-    %% ATT and arterial component inference only possible with multi-PLD
-    if ~bMultiPLD
-        fprintf('Basil: Single-delay data - cannot infer ATT or arterial component\n');
-        x.Q.BasilInferATT = 0;
-        x.Q.BasilInferArt = 0;
-    end
+if isfield(x.Q,'LookLocker') && x.Q.LookLocker && isfield(x.Q,'FlipAngle')
+	fprintf(option_file, '--FA=%f\n', x.Q.FlipAngle);
```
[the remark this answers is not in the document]
jan-petr: `LookLocker` - see above. Warning added for missing FlipAngle.

Functions/xASL_quant_Basil.m
```diff
@@ -123,170 +130,193 @@
...
-        fprintf('Basil: Variable arterial component arrival time');
-    end
+fprintf(FIDoptionFile, '--save-model-fit\n');
 
+switch lower(x.Q.LabelingType)
+	case 'pasl'
+		% Default initial ATT for PASL is 0.7
+		fprintf(FIDoptionFile, '--bat=0.7\n');
+	case {'pcasl','casl'}
+		% Default initial ATT for PCASL is 1.3
```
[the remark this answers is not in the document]
jan-petr: No. This is standard arrival time. `--bat` option defines it's initial value for fitting. But not after crushing thing, but normal arrival time. At least I understand it like this from the manual.

Functions/xASL_quant_GetMeanControl.m
```diff
@@ -36,29 +36,36 @@
```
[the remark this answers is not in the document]
jan-petr: As above. We shall discuss this on Monday meeting.

Functions/xASL_quant_GetMeanControl.m
jan-petr: Agree. Done.

Functions/xASL_quant_MultiPLD.m
jan-petr: I have tested `xASL_num2str` and it doesn't do the (:)' automatically. 

`xASL_num2str([1;2;3])

ans =

  3×1 char array

    '1'
    '2'
    '3'`

Modules/SubModule_ASL/xASL_wrp_Quantify.m
jan-petr: Agree done.

Modules/SubModule_ASL/xASL_wrp_Quantify.m
jan-petr: Up to you, I have used the previous notation. But I'm fine to rename this. Do you want to use ATT for both FEAST and non-FEAST?

Modules/SubModule_ASL/xASL_wrp_RealignASL.m
jan-petr: bDeltaM == bContainsDeltaM . It tries to distinguish sequences containing deltaM from those containing control/labels.

Yes, the assumption that single-volume ASL4D has a deltaM is wrong - that's why I had to change this. GE multi-PLD has multiple deltaMs.

ENABLE needs to be disabled for deltaM as either it is a single volume, or it is multi-PLD and you can't start removing certain PLDs. DeltaM == true when the original sequence already contains the subtractions.

Yes - you got it right. We perhaps need an extra issue to use bDeltaM everywhere. But I needed to do this here - cause I couldn't split the multi-PLD eASL implementation to two issues - that wouldn't be possible to test. I can use bDeltaM everywhere - in this issue, or in a different issue?

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
jan-petr: Good point. Thanks ;)

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
jan-petr: OK. Will do that. But it wasn't really me who changed that.

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
jan-petr: Yes! Corrected.

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
jan-petr: Good point. Modified.

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
jan-petr: Good point. Modified.

Modules/xASL_module_ASL.m
jan-petr: Good point!!! Will do that - see comments above.

Modules/xASL_module_ASL.m
jan-petr: Good point. Will do.

Modules/xASL_module_ASL.m
jan-petr: Agree. Will change.

Modules/xASL_module_ASL.m
jan-petr: this you got wrong. `bidsPar` is a function and variable where we set all the default variable and tag names for BIDS and also set all the dependencies so that they are configured in a single file and easily resolved.

So `bidsPar.stringDeltaM` contains the string `deltam` so that in case this is changed in BIDS, we only have to change it in the `xASL_bids_Config` ? That OK? This has been there for half a year ;)

If we have a mixture of control/label/deltaM, then we are doomed - we don't really support these more complex stuff. Lets add a warning and implement late?

Modules/xASL_module_ASL.m
jan-petr: On the other hand. A single volume will still have deltam in the context, so we don't need to program this option as this already works. ASLcontext always have to be present and is checked in import.

Functions/xASL_bids_parms2BIDS.m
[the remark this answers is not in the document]
MichaelStritt: I agree with Jan here.

Functions/xASL_quant_Basil.m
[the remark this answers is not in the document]
MichaelStritt: Couldn't find `isvar` either.

Functions/xASL_quant_Basil.m
```diff
@@ -123,170 +130,193 @@
...
+		else
+			LDs = ones(size(PLDs))*x.Q.LabelingDuration/1000;
+		end
+		% For Time-encoded, we skip the first volume
+		if x.modules.asl.bTimeEncoded
+			PLDs = PLDs(2:end);
+			LDs = LDs(2:end);
+		end
+		
+		if x.modules.asl.bTimeEncoded || x.modules.asl.bMultiPLD
```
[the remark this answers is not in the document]
MichaelStritt: I don't get it, I thought we can have multiple TIs and TEs @BeatrizPadrela ?

Functions/xASL_quant_Basil.m
HenkMutsaerts: We discussed this multiple times, and I have always emphasized that bMultiPLD should be true when bTimeEncoded is true. So TimeEncoded is a special case of multiPLD, this is to make life simpler. Throughout the pipeline, both multiPLD and TimeEncoded do exactly the same, except for a few locations only — e.g., when you decode Hadamard volumes into multiPLD volumes.

The change is probably small: we should:
1) verify at the start of the ASL module that this is the case
2) any time that you request bMultiPLD OR bTimeEncoded, you remove bTimeEncoded

Yes, I think I wrote this if-statement because it didn't work with PWI4D or something similar. If this is easy changed it would be good, otherwise we can do this later.

Most important probably is to do the comparison with Martin Craig.

Functions/xASL_bids_parms2BIDS.m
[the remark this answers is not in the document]
HenkMutsaerts: Ah OK, just note that we use it internally as a boolean, whereas BIDS fields usually contain (more) information. E.g., we also use `bMultiPLD`, whereas this also has a BIDS field (though bit more complicated). I'm fine either way, we should just be consistent, and it should be clear what datatype a variable is (that's why I like `bLookLocker`, the b tells you that this is a boolean/logical.

Functions/xASL_quant_Basil.m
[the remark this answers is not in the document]
HenkMutsaerts: What I mean, is that if there is no output from BASIL (e.g., with singlePLD), you will use `NaN` for ATT_map. To avoid the need of `if ~exist(ATT_map,'var');ATT_map = NaN; end` at the end, you can initialize `ATT_map` with `NaN` at the start. This is useful as singlePLD should still be the default (unless multiPLD is specifically designated).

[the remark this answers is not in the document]
HenkMutsaerts: Currently we would need something like `isvar` (probably `exist(,'var')` instead) at the end. I want to avoid this. But it wasn't there, I was just giving this as example. If by whatever if statements that are skipped, the function end is reached, we need to catch `ATT_map` not existing as variable. Better (cleaner, easier to read and debug) to initialize this at the beginning.

Functions/xASL_quant_Basil.m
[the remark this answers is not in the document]
HenkMutsaerts: Hadamard/TimEncoded is a subtype of multiPLD, these are all labeling strategies. Multiple TEs (multiple echo times) is a type of readout. Has nothing to do with TimeEncoding (though I understand your confusion ;)

Functions/xASL_quant_GetMeanControl.m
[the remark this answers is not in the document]
HenkMutsaerts: Yes if you also add `bContainsLookLocker` ;)

Functions/xASL_quant_Basil.m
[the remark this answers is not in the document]
HenkMutsaerts: We could discuss how we want to standardize this, but usually it is `NaN`. Think of the variable always having to be numeric, so you can calculate it and the pipeline will never crash (which you can with NaN). When you have no variable, you need a lot of `if statements`. So I would go for `NaN` like we use in many other situations.

Functions/xASL_quant_Basil.m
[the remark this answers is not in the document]
HenkMutsaerts: OK perhaps just nice to avoid 1 extra PLD or LD variable and do this at a single location. Is easier to read for others, you can then do this at the top of the function with the admin, so you don't have to through the function.

Functions/xASL_quant_Basil.m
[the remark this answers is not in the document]
HenkMutsaerts: OK but then it is really short. Note that it should be more around 2 seconds for clinical populations. I agree that we should do the same as the FSL group does.

Functions/xASL_quant_MultiPLD.m
HenkMutsaerts: I meant to add this feature to xASL_num2str as we probably want this in all cases. Read my comment, I said "add to `xASL_num2str`, I didn't say "add `xASL_num2str`" :)

Modules/SubModule_ASL/xASL_wrp_Quantify.m
HenkMutsaerts: Perhaps use something that works for now, and soon update to BIDS terminology?

Modules/SubModule_ASL/xASL_wrp_RealignASL.m
HenkMutsaerts: OK I agree.

I would have liked splitting this issue in multiple issues, e.g., `bDeltaM` should be a separate issue at all (not only halfly implementing it in this issue and fixing it in another ;)

But I'm fine either way. If you want to do this properly, search for `nVolumes>1` or something similarly, especially in the realign function.

Modules/xASL_module_ASL.m
HenkMutsaerts: The mixture of `control-label-deltaM` is realistic. I'm fine with all this, as long as we add proper warnings in place!

Modules/xASL_module_ASL.m
HenkMutsaerts: Yes, but then this needs to be automated in the import, always putting 1 volume to `deltaM` with a warning, if `ASLcontext` wasn't provided yet.

Functions/xASL_bids_parms2BIDS.m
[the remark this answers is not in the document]
jan-petr: Yes. The BIDS field `LookLocker` is a boolean. So the use is consistent. OK to resolve the issue?!

Functions/xASL_quant_Basil.m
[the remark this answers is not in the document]
jan-petr: As previously discussed - we keep this as it is. Can you resolve?!

Functions/xASL_quant_Basil.m
jan-petr: Fixed bTimeEncoded being always bMultiPLD a10bfd4

Functions/xASL_quant_Basil.m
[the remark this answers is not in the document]
jan-petr: Done in a10bfd4

Functions/xASL_quant_Basil.m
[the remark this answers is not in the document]
jan-petr: Agreed that we keep it as it is, right? Can you resolve?

Functions/xASL_quant_Basil.m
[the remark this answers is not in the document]
jan-petr: Agreed that it is more transparent to keep as it is. Can you resolve this?

Functions/xASL_quant_Basil.m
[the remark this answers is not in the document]
jan-petr: Yes, short, but proposed by FSL. we might need to fine tune later. Also, that's the initial value and will be still optimized to match the data.

Functions/xASL_quant_GetMeanControl.m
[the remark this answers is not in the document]
jan-petr: Fixed in a10bfd4

Modules/SubModule_ASL/xASL_wrp_Quantify.m
jan-petr: Done. a10bfd4

Modules/SubModule_ASL/xASL_wrp_Quantify.m
jan-petr: We agreed on this?! Resolve?

Modules/SubModule_ASL/xASL_wrp_Quantify.m
jan-petr: Yes. Can you resolve?

Functions/xASL_quant_GetMeanControl.m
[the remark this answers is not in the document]
jan-petr: `bDeltaM` renamed to `bContainsDeltaM` 312b844 .Can you resolve?

Functions/xASL_quant_MultiPLD.m
jan-petr: New issue made for this. OK to resolve?
https://github.com/ExploreASL/ExploreASL/issues/995

Modules/SubModule_ASL/xASL_wrp_RealignASL.m
jan-petr: I have reviewed all use of `nVolumes` and either `bContainsDeltaM` is there or it doesn't need to be there. So we are OK to resolve this one.

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
jan-petr: I've checked again the code and currently, it always save PWI and PWI4D with the exception of single-volume or single-PLD files, where it saves PWI.nii only. 

But BASIL, later on, tries to read PWI4D.nii and quantifies with that in preference and only goes to PWI when PWI4D is not there. So all works as it should.
OK to resolve?

MichaelStritt:
### Open tasks

- [ ] Add `LookLocker` to dataPar markdown template https://github.com/ExploreASL/ExploreASL/pull/986#discussion_r767737279
- [ ] ASL context https://github.com/ExploreASL/ExploreASL/pull/986#discussion_r767751276
- [ ] Renaming https://github.com/ExploreASL/ExploreASL/pull/986#discussion_r767262642
- [ ] ASL context (single volume) https://github.com/ExploreASL/ExploreASL/pull/986#discussion_r767514654
- [ ] Add warning https://github.com/ExploreASL/ExploreASL/pull/986#discussion_r767512917


### Can be resolved

- [ ] PLD variable (discussed on monday) https://github.com/ExploreASL/ExploreASL/pull/986#discussion_r767250625
    ▶️ No fix necessary
- [ ] NaN default (discussed on monday) https://github.com/ExploreASL/ExploreASL/pull/986#discussion_r767260136
    ▶️ No fix necessary
- [ ] nVolumes and deltaM https://github.com/ExploreASL/ExploreASL/pull/986#discussion_r767261309
    ▶️ Fix https://github.com/ExploreASL/ExploreASL/pull/986/commits/312b84473c9bfb371dd24b40bda1a4d9df16bb34
- [ ] PWI and PWI4D https://github.com/ExploreASL/ExploreASL/pull/986#discussion_r767261525 
    ▶️  Fix https://github.com/ExploreASL/ExploreASL/pull/986/commits/9995c40c6459fd439ff1ce46bda19c1f33212a69

Modules/xASL_module_ASL.m
MichaelStritt: The mix wouldn't be correct according to BIDS AFAIK, so the validation should already alert you. Right now we still recommend to run the BIDS validation. We don't say anywhere that we do perfect validation ourselves. If we start validating that stuff from start to finish, there are a hundred new issues you can theoretically open. That's not really part of this PR IMHO.

Functions/xASL_bids_parms2BIDS.m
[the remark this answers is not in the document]
jan-petr: It is x.Q field directly obtained from BIDS, not a dataPar field. Same as e.g. x.Q.BackgroundSuppression. I know that we have all those fields in dataParTemplate, but that's an obsolete thing. We have `Initial_PLD`, `readoutDim` and all those field in the Legacy format, which will only confuse the user - these things are now not supposed to be added through the dataPar.json, but they should be added through studyPar.json in the BIDS format. And we should remove all these from here in the future. 

So I can add LookLocker there for now, but that goes against our current philosophy.

Modules/xASL_module_ASL.m
jan-petr: Agree. Added this warning.

Modules/xASL_module_ASL.m
jan-petr: Vendors do all crazy things of putting `control` and `label` and `deltaM` and `cbf` in the same sequence. Besides, it is not strictly against BIDS definition to have all these there. I don't foresee why would anybody have some controls and some deltams in the same sequence that wouldn't be simply duplicite. But that's nowhere in our Flavors. I would thus recommend to make this BIDS check in the import and not here. I've made a new issue for that. One of the reasons why not to do it now is that we don't have any such flavors, we therefore don't know what to expect and how to properly react to that.

Modules/xASL_module_ASL.m
jan-petr: I've checked this function and it needs a revamp. It creates a HUGE structure of 30.000 cells. 
I've made a new issue for that #1006 and I will add the ASL4Dcontext path there. It makes no sense to add this new path there quickly without revamping the function as it would only make it grow even more.

Functions/xASL_bids_parms2BIDS.m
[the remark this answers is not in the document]
HenkMutsaerts: Agree!

Modules/xASL_module_ASL.m
HenkMutsaerts: Sounds good

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
HenkMutsaerts: OK I updated this in #997

Functions/xASL_init_FileSystem.m
```diff
@@ -159,6 +159,8 @@
     end
     x.P.Pop_Path_MaskVascular = fullfile(x.D.PopDir, ['MaskVascular_' x.P.SubjectID '_' x.P.SessionID '.nii']);
     x.P.Path_Pop_MaskSusceptibility = fullfile(x.D.PopDir, ['rMaskSusceptibility_' x.P.SubjectID '_' x.P.SessionID '.nii']);    
+	
+	x.P.Path_ASL4D_aslcontext = fullfile(Path{2},'ASL4D_aslcontext.tsv');
```
HenkMutsaerts, line 163: This filename is incorrect, in ExploreASL we currently exchange `asl` for `ASL4D`. So to be consistent:

BIDS -> legacy
`_asl.json` -> `ASL4D.json`
`_aslcontext.tsv` -> `ASL4Dcontext.tsv`

imho

Modules/SubModule_ASL/xASL_wrp_RealignASL.m
```diff
@@ -61,21 +61,18 @@ function xASL_wrp_RealignASL(x,bSubtraction)
 PercExcl = NaN;
 MinimumtValue = NaN;
 
-if x.modules.asl.bMultiPLD || x.modules.asl.bMultiTE
+if x.modules.asl.bMultiPLD || x.modules.asl.bMultiTE || x.modules.asl.bContainsDeltaM
     % ENABLE is disabled if multiPLD/TE
```
HenkMutsaerts, line 65: Change into:
`% ENABLE is temporary disabled if multiPLD/TE.`

HenkMutsaerts, line 65: I would move `x.modules.asl.bContainsDeltaM` to a separate if-statement at the beginning of `xASL_wrp_RealignASL`, because in this case we want to skip motion correction completely right?

Modules/SubModule_ASL/xASL_wrp_RealignASL.m
```diff
@@ -139,7 +136,7 @@ function xASL_wrp_RealignASL(x,bSubtraction)
 %     end
 
 % Run motion correction for corresponding case
-if nFrames>2 && bSubtraction && (x.modules.asl.bMultiPLD  ||  x.modules.asl.bMultiTE  || x.modules.asl.bTimeEncoded)
+if nFrames>2 && bSubtraction && (x.modules.asl.bMultiPLD  ||  x.modules.asl.bMultiTE)
```
HenkMutsaerts, line 139: Incorrect, this is a check to state if we can run motion correction at all right, this is irrespective of multiPLD and multiTE...

Also, I don't know exactly what `bSubtraction` does here, this boolean is here to differentiate between `spm_realign` (`bSubtraction==false`) and `spm_realign` with zig-zag (`bSubtraction==true`).
We might replace `bSubtraction` by `bZigZag` for clarification

Modules/xASL_module_ASL.m
```diff
@@ -150,15 +151,16 @@
 %% D4. TimeEncoded parsing
 % Check if TimeEncoded is defined
 if isfield(x.Q,'TimeEncodedMatrixType') || isfield(x.Q,'TimeEncodedMatrixSize') 
-    x.modules.asl.bTimeEncoded = 1;
+    x.modules.asl.bTimeEncoded = true;
```
HenkMutsaerts, line 154: We should add a warning here:

>> warning('We detected TimeEncoded data (e.g., Hadamard or Walsh), we will process this but this is still an experimental feature');

And similarly for multiPLD in this admin section:

>> warning('We detected multiPLD data, we will process this but this is still an experimental feature');

Modules/xASL_module_ASL.m
```diff
@@ -323,7 +348,7 @@
             xASL_im_CenterOfMass(x.P.Path_ASL4D, OtherList, 10); % set CenterOfMass to lower accepted distance for when rerunning wrong registration
         end
 
-        if  nVolumes>1
+        if  nVolumes>1 && ~x.modules.asl.bContainsDeltaM && (~isfield(x.Q,'LookLocker') || (x.Q.LookLocker == 0))
             % Run motion Correction
             xASL_wrp_RealignASL(x);
```
HenkMutsaerts, line 353: Here we do check for `bContainsDeltaM`, but to be sure I would also do this inside `xASL_wrp_RealignASL`, only once, at the top of the wrapper where the admin is done. We can remove it elsewhere in `xASL_wrp_RealignASL`.

HenkMutsaerts (changes requested):
Few minor fixes left

Modules/SubModule_ASL/xASL_wrp_RealignASL.m
MichaelStritt: 2ac98fc

Modules/xASL_module_ASL.m
MichaelStritt: ccec3a5

MichaelStritt: a67a7bd

Modules/xASL_module_ASL.m
MichaelStritt: I'll quickly try to implement what you want, but this isn't how reviewing should be done. We shouldn't stack feature request over feature request again and again.

MichaelStritt: 9256d4e

Modules/SubModule_ASL/xASL_wrp_Quantify.m
```diff
@@ -71,9 +71,16 @@ function xASL_wrp_Quantify(x, PWI_Path, OutputPath, M0Path, SliceGradientPath)
...
-    OutputPath = x.P.Pop_Path_qCBF;
+
+if nargin<3 || isempty(pathOutputCBF)
+    pathOutputCBF = x.P.Pop_Path_qCBF;
 end
+% Define output path for ATT map based on the CBF output path
+% Replace CBF with ATT in the output path
+iStringCBF = regexpi(pathOutputCBF, 'CBF');
+iStringCBF = iStringCBF(end);
+pathOutputTT = [pathOutputCBF(1:(iStringCBF-1)) 'TT' pathOutputCBF((iStringCBF+3):end)];
```
MDijsselhof, line 82: 'TT' Is this a typo?

Modules/SubModule_ASL/xASL_wrp_Quantify.m
MDijsselhof: I apologise for opening this again, but what is the difference? Both produce ATT maps?

Modules/SubModule_ASL/xASL_wrp_RealignASL.m
MDijsselhof: I agree this is unclear, maybe we should add comments for these `if` statements to show what is happening

Functions/xASL_quant_Basil.m
```diff
@@ -123,170 +130,203 @@
...
+		else
+			% Bolus duration not know. If multi-TI, then try to infer it
+			if length(TIs) > 1
+				fprintf(FIDoptionFile, '--infertau\n');
+				fprintf('BASIL: Infer bolus duration component\n')
+			end
+		end
+		
+	case {'casl','pcasl'}
+		fprintf(FIDoptionFile, '--casl\n');
```
MDijsselhof, line 189: Although this is fine, we should add `--pasl` to the PASL case to make it clear

Functions/xASL_quant_Basil.m
```diff
@@ -123,170 +130,203 @@
...
+			end
+			for iTI = 1:length(TIs)
+				fprintf(FIDoptionFile, '--ti%d=%.2f\n', iTI, TIs(iTI));
+			end
+			
+		else
+			fprintf(FIDoptionFile, '--ti=%.2f\n', TIs);
+		end
+		
+		% Either print bolus duration or unspecify it
```
MDijsselhof, line 177: Is this going correctly for multiple LDs? Or are we improving this later?

Functions/xASL_quant_Basil.m
```diff
@@ -123,170 +130,203 @@
...
-    
-    %% ATT and arterial component inference only possible with multi-PLD
-    if ~bMultiPLD
-        fprintf('Basil: Single-delay data - cannot infer ATT or arterial component\n');
-        x.Q.BasilInferATT = 0;
-        x.Q.BasilInferArt = 0;
-    end
+% Right now, we assume that we have averaged over PLDs
+%fprintf(FIDoptionFile, '--repeats=%i\n', size(PWI, 4)/PLDAmount);
+fprintf(FIDoptionFile, '--repeats=1\n');
```
MDijsselhof, line 221: Why, I thought we were going to supply BASIL with multiple repeats?

Functions/xASL_quant_Basil.m
```diff
@@ -123,170 +130,203 @@
...
+	
+% Slice-timing
+fprintf(FIDoptionFile, '--slicedt=%f\n', x.Q.BasilSliceReadoutTime/1000);
+
+if isfield(x.Q,'LookLocker') && x.Q.LookLocker 
+	if isfield(x.Q,'FlipAngle')
+		fprintf(option_file, '--FA=%f\n', x.Q.FlipAngle);
+		fprintf('BASIL: Flip angle for Look-Locker readout: %f\n', x.Q.FlipAngle);
```
MDijsselhof, line 229: I know of only fixed FA Look-Locker, but is it possible to have multiple FAs? Would complicate the reconstruction I would say, just curious

MDijsselhof (changes requested):
Couple of questions!

Functions/xASL_quant_Basil.m
MDijsselhof: This is about the code below !

Modules/SubModule_ASL/xASL_wrp_Quantify.m
jan-petr: No worries. There's no difference. It was just called TT previously, so while ATT is better, we will keep the initial name. And we also don't want to have both TT and ATT as this would be confusing.

There's of course a difference in the algorithms - one works with multi-PLD, the other with single-PLD and crusher ON/OFF. 

If that explains it, you can resolve this one.

Functions/xASL_init_FileSystem.m
jan-petr: Very good comment! But we need to fix this everywhere ;) See this 26fa3e2d
Note that to properly test this, we will need to change flavors as well, and we can only do the full test upon merging with `develop` that has some new edits in import.

Modules/SubModule_ASL/xASL_wrp_RealignASL.m
jan-petr: Yes. Unclear. Plus it should have reused bZigZag assigned previously here :D This is now fixed.

`bSubtraction` was there previously, to activate ZigZag. This is now renamed to `bASL` and for non-ASL, `bZigZag` is turned off. And the whole `bZigZag` assignment is done above.

With all the above - makes no sense to rename `bSubtraction` to `bZigZag` as there are cases with `bSubtraction==1` that still have `bZigZag` off. Like if it contains deltaMs - is a subtraction-based technique, but ZigZag has to be off - this is all explained in the comment above and fixed in this commit: 3d240036

Modules/SubModule_ASL/xASL_wrp_RealignASL.m
jan-petr: The input parameter `bSubtraction` was confusing as it wasn't clear if this refers to the fact that input is 'already subtracted' or 'subtraction-based images'. I have changed that to `bASL` is it is literally only ASL that has this option true. And it is only an internal parameter. This way, we won't confuse it with the `bContainsDeltaM` parameter.

To your question: It's a bit more complex.
- bContainsDeltaM turns off ENABLE and ZigZag as both need control/labels currently
- multi-PLD turns of both as well as with ENABLE it will be really tricky to exclude something, and ZigZag needs to be specially adapted for multi-PLD - but that's temporarily.
- However, single-PLD and bContainsDeltaM with multiple averages (though not sure if we have such a flavor) can run motion correction with ENABLE=0 and ZigZag = 0.

Finally, bZigZag was ignored currently in the code and was assigned 'manually'. I have fixed that as well.

All in 3d240036

Modules/xASL_module_ASL.m
jan-petr: See all my comments above. `bContainsDeltaM` by itself is not a reason not to run motion correction. It's a combination with other factors that disable or modify motion correction. So warning removed. And removed `bContainsDeltaM` from this if-statement - see explanations above.

Modules/SubModule_ASL/xASL_wrp_Quantify.m
jan-petr: No it isn't - historically, we're having it this way and decided to keep it like this and only change it when BIDS-derivatives names are introduced.

jan-petr: See also my reply in the re-opened comment.

Functions/xASL_quant_Basil.m
jan-petr: Yes, in the future, we will always provide all repeats to the PWI4D for BASIL. But currently, we average and we would have to change it on several places and properly test. So currently, we keep it averaged with a single repeat and we have a new issue for this already.

Functions/xASL_quant_Basil.m
jan-petr: Good point. Look-Locker practically have a single-FA only per volume (or a single volume batch after a single labeling), but you can indeed have different FAs for different repetitions - like in QUASAR. Currently, BASIL seems to accept only a single FA value here, and QUASAR has it's own quantification function. So since almost nobody uses Look-Locker, and nobody uses QUASAR, no point implementing QUASAR now and we can leave it here for single FA. I added a warning though: 4dffd3ca

Functions/xASL_quant_Basil.m
jan-petr: Good point. For PASL, the labeling duration is not defined by its actual duration (cause that is almost instantaneous) but by the timing of the bolus saturation pulse. And I guess that this doesn't really differ for multi-PLD sequences - or at least we don't really have the example for this. Still a valid point and I'll add a warning for future possible cases! Fixed here: 4dffd3ca

Functions/xASL_quant_Basil.m
jan-petr: Good point. But PASL is the default option and it doesn't need to be added to the config file. Besides, that option doesn't exist ;)
I added some explanatory comments to the code here: 4dffd3ca

MichaelStritt:
@jan-petr, @BeatrizPadrela, @MDijsselhof & @HenkMutsaerts: Let's merge as is and test in release branch. Easier and better IMHO 👍

MDijsselhof (approved):

BeatrizPadrela (approved):

Functions/xASL_init_FileSystem.m
HenkMutsaerts: Sure, we will revamp this anyway when going fully to BIDS. But nice if we can keep ourselves adherent to BIDS as much as possible, that will make the future changes smaller.

Modules/SubModule_ASL/xASL_wrp_RealignASL.m
HenkMutsaerts: Well you can say the same for `bASL`, there are also cases where `bASL` is true and `bZigZag` is off.
I would keep it simplier.

We need as options inside `xASL_wrp_Realign`:
`bMotionCorrection`
`bSubtraction` (used previously only to switch between ZigZag off-on) -> `bZigZag`
`bENABLE`

In the header of `xASL_wrp_Realign`, we set these three options based on the type of ASL (`bMultiPLD`, `bContainsDeltaM`, `bMultiTE`, etc)

Modules/SubModule_ASL/xASL_wrp_RealignASL.m
HenkMutsaerts: See below.

Modules/xASL_module_ASL.m
HenkMutsaerts: See above. Let's replace this by `bMotionCorrection` and `bZigzag` et, which we all parse nicely in the header. Then it's easier to discuss what needs to happen and you don't have to read the whole code to understand this.

Modules/SubModule_ASL/xASL_wrp_Quantify.m
HenkMutsaerts: > @jan-petr, @BeatrizPadrela, @MDijsselhof & @HenkMutsaerts: Let's merge as is and test in release branch. Easier and better IMHO 👍

We should first agree on the structure, then test this branch, and then test the release branch. But we can fuse the last two to speed things up, for now. And we could move things to a new issue.

But the clarity in `xASL_wrp_Realign` what is done when, is something I would like to implement now, saves time testing and discussing.

Functions/xASL_init_FileSystem.m
jan-petr: Sure. But can you resolve this?! ;)

Modules/SubModule_ASL/xASL_wrp_RealignASL.m
jan-petr: That's not what is described in the header. Header say bSubtraction == bASL == ASL yes/no

Yes, we have now both `bZigZag` and `bASL`. We don't need `bMotionCorrection`, all is motion corrected. We only turn of ZigZag, but not depending on function parameters, but on data parameters. We need to keep this transparent and adjust based on data. To me, it makes no sense to start gluing up extra functionality - that's what you always protest against ;)

Modules/xASL_module_ASL.m
jan-petr: Makes no sense to me. We moved all the details inside wrappers to keep ASL-module clean. It doesn't make sense to start moving things back from wrappers again.

`bContainsDeltaM` is now removed from here. And we can remove `nVolumes` and `LookLocker` and put it inside the wrapper if you want. But any larger changes do not make sense here.

HenkMutsaerts: Not outside the wrapper, just to the header of the wrapper

jan-petr: I see. Sorry, missed that. Yes, we can move that in.

Modules/SubModule_ASL/xASL_wrp_RealignASL.m
MichaelStritt: If you want to change functionality without messing up things with too much backwards compatibility, we can just tackle this in the next major release. For me the current header and descriptions seem pretty clear.

Modules/xASL_module_ASL.m
jan-petr: Fixed 791df878

HenkMutsaerts (approved):
I added 2 commits, for clarity and to avoid running through `xASL_wrp_Realign` when `bMotionCorrection`

PR #1043, opened 2022-01-04
Closes #995 num2str changes

### LINKED ISSUE #995: xASL_num2str output always row vector
### Description

Add (:)' inside num2str to make it always output a row vector. Because right now it gives several rows

`xASL_num2str([1;2;3])`
`ans =`
`3×1 char array`
`'1'`
`'2'`
`'3'`

Review the use of xASL_num2str everywhere to make sure always a single number or a row vector of output is expected.

### Requirements

### How to test

### Release notes

MichaelStritt:
### Linked issue

Check out #995. Did some minor changes to the original idea. Someone still has to go through the pipeline and check for potential problems.

### How to test

### Comments

### DIFF
--- a/Functions/xASL_num2str.m
+++ b/Functions/xASL_num2str.m
@@ -1,13 +1,13 @@
-function [DataOut] = xASL_num2str(DataIn, f, bConcatenate, strDelimiter)
+function [DataOut] = xASL_num2str(DataIn, stringFormat, bConcatenate, stringDelimiter)
 %xASL_num2str Wrapper around Matlab builtin 'num2str', bypassing strings/characters & BIDS-compatible
 %
-% FORMAT: [DataOut] = xASL_num2str(DataIn[, f, bConcatenate, strDelimiter])
+% FORMAT: [DataOut] = xASL_num2str(DataIn[, stringFormat, bConcatenate, stringDelimiter])
 %
 % INPUT:
-%   DataIn 	     - input data (can be any format) (REQUIRED)
-%   f            - second argument of "num2str". Can be a format, but mostly used as number of characters to print (effectively rounding) (OPTIONAL)
-%   bConcatenate - concatenate multiple-lines to a single line with ',' as a delimiter (OPTIONAL, DEFAULT = 1)
-%   strDelimiter - string to be used as a delimiter for concatenating when bConcatenate is TRUE, empty delimiters are allowed (OPTIONAL, DEFAULT = ',')
+%   DataIn 	        - input data (can be any format) (REQUIRED)
+%   stringFormat    - second argument of "num2str". Can be a format, but mostly used as number of characters to print (effectively rounding) (OPTIONAL)
+%   bConcatenate    - concatenate multiple-lines to a single line with ',' as a delimiter (OPTIONAL, DEFAULT = 1)
+%   stringDelimiter - string to be used as a delimiter for concatenating when bConcatenate is TRUE, empty delimiters are allowed (OPTIONAL, DEFAULT = ',')
 %
 % OUTPUT:
 %   DataOut - numbers converted to string, or bypassed data
@@ -46,8 +46,8 @@ if isnumeric(DataIn)
 end
 
 % Default: no format
-if nargin < 2 || isempty(f)
-	f = '';
+if nargin < 2 || isempty(stringFormat)
+	stringFormat = '';
 end
 
 % Default: set concatenate to 1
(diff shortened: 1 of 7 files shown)


Functions/xASL_num2str.m
```diff
@@ -93,24 +99,28 @@
...
-		for iSpace = 1:length(indStart)
-			DataOut = [DataOut(1:indStart(iSpace)) strDelimiter DataOut(indEnd(iSpace)+1:end)];
-			indShift = length(strDelimiter) - (indEnd(iSpace)-indStart(iSpace));
-			indStart = indStart + indShift;
-			indEnd   = indEnd + indShift;
-		end
-	end
+    % If the bConcatenate option is ON, then check if concatenation is needed
+    if bConcatenate
+        if size(DataOut,1) > 1
```
HenkMutsaerts, line 104: This seems unnecessary complicated, line 120 below should suffice?

Not sure what you fixed here, it seems the concatenation option was there already?

HenkMutsaerts (changes requested):
Looks fine but unnecessary complicated?

Functions/xASL_num2str.m
HenkMutsaerts: Or did you only insert a tab here? :)

MichaelStritt: Matlab showed the wrong indentation thing so I did a tab 😆

MichaelStritt: As mentioned in the comments I solved it a bit diffently now. Everything becomes a column vector now and later on the rows are merged in the concat stage. If you don't do this, the default concat for a row vector wouldn't add the default separator (minor bug from before).

MichaelStritt: It still kind of always adds a space after arrays, we could get rid of that, too.

And it would also be nice if someone would check the effect on the pipeline, because we use it quite often, right?

Functions/xASL_num2str.m
```diff
@@ -47,7 +52,7 @@
 
 % Default: set concatenate to 0
```
jan-petr, line 53: Default goes to 1 also in the comment here

Functions/xASL_num2str.m
```diff
@@ -1,13 +1,12 @@
 function [DataOut] = xASL_num2str(DataIn, f, bConcatenate, strDelimiter)
```
jan-petr, line 1: I know that this was there before, but can we change it to something meaningful - like 'format' or 'stringFormat'?

Functions/xASL_num2str.m
```diff
@@ -1,13 +1,12 @@
...
-%xASL_num2str Wrapper around Matlab builtin 'num2str', bypassing strings/characters
-% & BIDS-compatible
+%xASL_num2str Wrapper around Matlab builtin 'num2str', bypassing strings/characters & BIDS-compatible
 %
 % FORMAT: [DataOut] = xASL_num2str(DataIn[, f, bConcatenate, strDelimiter])
 %
 % INPUT:
 %   DataIn 	     - input data (can be any format) (REQUIRED)
 %   f            - second argument of "num2str". Can be a format, but mostly used as number of characters to print (effectively rounding) (OPTIONAL)
```
jan-petr, line 8: please add default to f - DEFAULT = ''

Functions/xASL_num2str.m
jan-petr: Yes, we should remove the extra space 'manually'.

Theoretically - I've checked the code - all should be fine. But we will only see upon re-runing 10TestDataSet and import ;)

jan-petr (changes requested):
A few minor things. And I've added xASL_num2str to a few more places just in case.

Functions/xASL_num2str.m
HenkMutsaerts: The DEFAULT for f should be empty, so '[]' (see how this is handled in the rest of the function)

Functions/xASL_num2str.m
HenkMutsaerts: I think I copied this from `num2str`, so if `stringFormat` is correct then yes, good idea. I would then also use `stringDelimiter` instead of `strDelimiter`

Functions/xASL_num2str.m
HenkMutsaerts: Ah OK cool, I didn't think of the separator

HenkMutsaerts (approved):

Functions/xASL_num2str.m
MichaelStritt: 9ebbd27

Functions/xASL_num2str.m
MichaelStritt: ea1ec1d

Functions/xASL_num2str.m
MichaelStritt: ea1ec1d

Functions/xASL_num2str.m
MichaelStritt: Fixed in 8143da2

jan-petr (approved):
Very good.

PR #1047, opened 2022-01-07
Closes #1046 Fix M0 with multi-TE

### LINKED ISSUE #1046: Use first TE value for M0 calculation
### Description

When using M0 calculation from a single value, then in TE-T2 effect application only use the first TE.

### Tasks

- [x] Fix `xASL_wrp_Quantify`

### Release notes

M0 value scaling for T2-effect use first TE only.

jan-petr:
### Linked issue

#1046 

### Comments

The quantification of multi-TE Hadamard still doesn't work as we need the new BASIL version, but at least it calculates PWI and doesn't crash before.


BeatrizPadrela (approved):
Ok!

PR #1069, opened 2022-01-25
Closes #1068 Fix create analysis mask

### LINKED ISSUE #1068: CreateSusceptibilityMask major errors
### Description

As caught by one of the users. There were major errors in the  `xASL_im_CreateAnalysisMask` function.

### Release notes

Fix errors in susceptibility mask in ` xASL_im_CreateAnalysisMask`.

jan-petr:
### Linked issue

#1068 
An error we had there for a long time, incurred during a revamp. And since there were two errors, they canceled themselves out, completely putting this function out of work. Thanks to the Japanese group who have noticed that!

It should all be fine - but please see if it makes sense to you.


HenkMutsaerts (approved):
Looks good. I wonder if we need the legacy masking subfunction at all, we can check this later.

MichaelStritt (approved):
👍

Issue #1073, opened 2022-01-31
New category of development functions


jan-petr:
### Description

These functions are supposed to be used by developers to create documentation, atlases, and run other maintenance functions for xASL. But not directly used by the users.
We will use the notation `xASL_dev_FunctionName`

### Tasks

- [x] Create a folder `Functions/Development`
- [x] Add folder to init
- [x] Add folder to Documentation + crawler + pages
- [x] Add the following function: `xASL_io_ConvertNiiGz2Mat.m`
- [x] Add the following function: `/Functions/Administration/xASL_adm_DocCrawler.m`, `xASL_dev_DocInitialize.m`, and other docu functions
- [x] Check all other development/Functions etc for stuff to be moved here.

### How to test

Optional: insert description about how to test the code changes here

### Release notes

Duplicate. Done in #1075.

PR #1074, opened 2022-01-31
Closes #823 development cleaning

### LINKED ISSUE #823: Clean development folder
### Description

Clean the development folder

### Requirements
- [x] `Siemens_ConcatenateNIfTI.m` we probably don't need this one anymore? @jan-petr 
-> HM says: this was a function that I created in the past, I think you also wrote something like this
-> Jan: Yes, we have concatenation as part of the import, and more robust than this version.
- [x] Can be moved to `CustomScript/ExploreASLLegacy`: `BIDS_createTemplatesImportConfig.m`, `xASL_im_TestDummyNIfTIOrientation.m`
-> HM says: good idea
- [x] Old import can be moved to `CustomScript/ExploreASLLegacy`: `xASL_ImportFull.m`
-> HM says: yes
- [x] Config files for old studies are moved to `CustomScript/ExploreASLLegacy`
-> HM says: why not a separate folder `CustomScript/StudyConfigs`
- [x] Can be moved to `CustomScript/ExploreASLLegacy`: `xASL_io_ConvertNiiGz2Mat.m`.? Do you agree? @MichaelStritt
-> HM says: no this function is needed for the creation of new atlases.
-> Agree. #1073 
- [x] Can be moved to `CustomScript/ExploreASLLegacy`: `QualityControl/xASL_qc_TestExploreASL_HM.m` 
-> HM says: Yes good idea
- [x] Can be moved to `CustomScript/ExploreASLLegacy`: `xASL_spm_jobman.m`? JAN: I have checked it and it is not used anywhere, so lets move it. Do you agree?  @HenkMutsaerts 
-> HM says: I think I wrote this for running spm jobman deployed, but if we don't use it then it can go to `CustomScript/ExploreASLLegacy` indeed
- [x] `UnitTest_xASL_io_DcmtkRead.m` I wrote this for Jan to test `xASL_io_DcmtkRead`, but he didn't check it out. Can we delete it? JAN: IMHO, lets delete this as we use DcmtkRead actively in import and that's a super robust test on 50+ flavors that it works.
-> HM says: this function iterates across multiple DICOMs and their fields and compares them between Matlab's `dicominfo` and `xASL_io_DcmtkRead`
- [x] `/Functions/Administration/xASL_adm_DocCrawler.m` -> `/Functions/Development/xASL_dev_DocCrawler.m`
`/Functions/Administration/xASL_adm_DocInitialize.m` -> `/Functions/Development/ xASL_dev_DocInitialize.m`
(and other docu functions)
-> Agree. #1073

-> HM says: As discussed, we create a new folder `/Functions/Development`, where functions live with the prefix `xASL_dev_*`, e.g. `xASL_dev_ConvertNiiGz2Mat.m`. These are functions that aren't used in the pipeline and that developers use to improve the pipeline.
-> Agree. #1073 

### Release notes

Cleaned unused functions in the `Development` folder.

jan-petr:
### Linked issue

#823. Agreed with all comments. New `Development` folder under functions will be created in an issue #1073 - it needs a new issues, because we also need to adapt the documentation and folder initialization etc.


HenkMutsaerts (changes requested):
We should give the good example to the kids, "clean develop folder" is not really a specific commit title. "Clean develop: delete study config files" would be better.

I don't understand why we make a list in a single issue and then only do 1 item from this list ;)

The docu functionality needs to be improved anyway, as we discussed this morning, so why not moving the other development files to the new folder in this issue, and updating the docu files in this issue as well? Splitting issues for real functional differences is good but spreading too thin in this case imho.

BTW: we should adapt the docu crawling, that it doesn't crawl across a fixed number of folders but across all folders inside `/Functions`, then we don't have to change the docu crawler function for each time we add a new folder to `/Functions`.

MichaelStritt:
> We should give the good example to the kids, "clean develop folder" is not really a specific commit title. "Clean develop: delete study config files" would be better:)

The kids are fine with it 👍

jan-petr:
The list was to start the discussion - so a part was done (several things), a part will be done in a different issue, part I found out doesn't need to be done.

HenkMutsaerts (approved):

PR #1076, opened 2022-02-01
Closes #1059 hadamard import corrections

### LINKED ISSUE #1059: Philips HAD4 and HAD8 import
### Description

Import of TE and PLD for Philips HAD8 is not entirely correct. With hopefully no impact on the later quantification though.

The aslcontext.json of FME HAD4 is not with the correct order of C and L.
It should repeat C based on the nr of TEs.
Ex: HAD4 = 4 PLDs, each PLD with 8 TEs.

### Tasks

- [x] 1. HAD8 - TE is repeated for each volume despite only a single unique TE should be used. 1427e09
- [x] 2. HAD8 - Repeated TE for M0 as well. 1427e09
- [x] 3. HAD8 - PLDs are repeated correctly number of times, but they are grouped by the same PLDs instead of grouping them by Hadamard cycles. [1 1 1 2 2 2 3 3 3...] instead of [1 2 3 1 2 3 1 2 3...] ->this works for Amnah's data because it's only one repetition. So I tested it with Saima's data as well (see comment below)
- [x] 4. HAD4 - aslcontext should be: 8 x C and then Ls ( C C C C C C C C L L L L L L L L L L L L L L L L L L....), as the TEs of the first PLD should be controls, as the first PLD is the control image.
- [x] 5. Change the TotalAcquiredPairs of Hadamard to count the number of controls from the asl_context (ex: C,L,L,L,C,L,L,L to be as 2 acquired pairs) (duplicate #1042)
- [x] 6. Remove this warning from Population module 'Parms.EchoTime had multiple values' (duplicate #1054) - "'Parms.EchoTime had multiple values' (xASL_stat_GetDICOMStatistics, line 98)" 1427e09

### Release notes

Fix TE and ASLcontext in Hadamard import.

BeatrizPadrela:
### Linked issue

#1059 

### How to test

I ran the Hadamard datasets...

- [x] FME HAD4 
- [x] FME HAD8
- [x] Philips HAD
- [x] Siemens VE11E_HAD4 (Saima's)
- [x] Siemens VE11E_HAD8 (Saima's)

...and all of the issues are solved. Import done correctly  👍

### DIFF
--- a/Functions/BIDS/xASL_bids_BIDSifyASLJSON.m
+++ b/Functions/BIDS/xASL_bids_BIDSifyASLJSON.m
@@ -350,12 +350,21 @@ if dimASL(4) ~= lengthASLContext
 	% Check if we can simply repeat it
 	if mod(dimASL(4),lengthASLContext)
 		error('Cannot find a match between the ASLContext and the 4th dimension of the NIFTI');
-	else
-		numRepeat = dimASL(4)/lengthASLContext;
-		tmpStr = jsonOut.ASLContext;
-		for iRepeat = 2:numRepeat
-			jsonOut.ASLContext = sprintf('%s\n%s',jsonOut.ASLContext,tmpStr);
-		end
+    else
+        % Check number of Echo times
+        if length(unique(jsonOut.EchoTime)) > 1
+            nVolumes = dimASL(4);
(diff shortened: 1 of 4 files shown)


jan-petr (approved):
One minor edit at the end. It looks all good now.

PR #1139, opened 2022-08-03
Bug #1138 x asl wrp bids2 legacy

### LINKED ISSUE #1138: xASL_wrp_BIDS2Legacy: Fix subject parsing for M0
### Description

`xASL_wrp_BIDS2Legacy` didn't correctly parse `"M0Type":"Included". Debugging showed me that the subject searching had an incorrect regular expression that always required the subject string to have symbols before and after the identified subject name. Instead, this should not require anything before or after the subject name, but allow the "sub-" prefix and visit suffices.

### Tasks

- [x] xASL_wrp_BIDS2Legacy: Fix subject parsing for M0
- [x] Question to Jan Petr: `xASL_wrp_BIDS2Legacy` first loops over all subjects/visits/sessions for all modalities (sections 2-4) but then does this separately for M0 (section 5). Shouldn't M0 be included in sections 2-4 or was this done on purpose?

### Release notes

Fix subject directory name parsing for M0.

HenkMutsaerts:
### Linked issue

Closes #1138
@jan-petr please also check my question in #1138.

That's on purpose - we need to know that there are all the ASL files, because for M0, we have to fill in the BIDS tag "IntendedFor", therefore we need to know what ASL files are there before processing the M0 files.


jan-petr (approved):
OK.

PR #1141, opened 2022-08-03
Closes #1140 spm load

### LINKED ISSUE #1140: spm_load
### Description

When creating a TSV such as `participants.tsv`, it is easy for users to leave a few empty cells at the end of a row. This creates problems for `xASL_tsvRead`->`spm_load`, which wants the same number of elements for each row. Now I programmed a script to fix this, assuming that the columns in the header are correct.

### Tasks

- [x] `spm_load`: Automatic fixing empty CSV/TSV cells
- [x] `README_SPM.txt`: Add `spm_load` edit
- [x] - I can report a list of lines with too few/too much delimiters
- [x] I will shorten/fill with empty accordingly
- [x] if no end of lines are in the data, then we revert to a simple total check - anything else is just guessing
- [x] I will save repaired.
- [x] I will only cut/add fields, but not lines. 

- [x] @jan-petr This still goes wrong: when loading a line that has sufficient number of delimiters to get to the correct number of cells per line; `textscan` on line 194 still only reads 8 (so it somehow ignores the last "empty cell"). We could ignore this, but then it is annoying for the user that the `participants.tsv`, even when "fixed" by `xASL_tsvRead`->`spm_load`, will keep repeating a warning.

-> In the first screenshot, the selection contains two tabs at the end of line two; leading to 9 "cells" and a correct line length. However, in the second screenshot, one sees that `textscan` has loaded only 8 "cells"; somehow ignoring the last `\t`.

Screenshot 2022-08-13 at 07 50 12
Screenshot 2022-08-13 at 07 52 41

### Release notes

Handles an incorrect number of delimiters per line in TSV/CSV/DSV files loaded by spm_load.

HenkMutsaerts:
### Linked issue

Closes #1140

### How to test

### Comments


External/SPMmodified/spm_load.m
```diff
@@ -191,3 +225,50 @@
...
+%   N = number of columns
+    
+    emptyCells = find(cellfun(@(y) isempty(y), d{1}));
+    % N = number of columns
+    % so if any index-1 of the empty cells equals to the number of columns,
+    % this means that the empty cell is at the end of a line and potentially
+    % crashing this function trying to open a delimiter-separated value file
+    % (dsv)
+    
+    illegalEmptyCells = (emptyCells-1)/N == round((emptyCells-1)/N);
```
jan-petr, line 271: That will not work correctly if N is an integer. Correct function is "mod(emptyCells-1,N) == 0"

External/SPMmodified/spm_load.m
```diff
@@ -191,3 +225,50 @@
...
+
+%% ========================================================================
+
+
+function [illegalEmptyCells] = DetectIllegalEmptyCells(d, N)
+%DetectIllegalEmptyCells % EXPLOREASL HACK: detect & fix potentially erroneous empty cells
+%   d = all cells from input delimiter-separate value file (without header)
+%   N = number of columns
+    
+    emptyCells = find(cellfun(@(y) isempty(y), d{1}));
```
jan-petr, line 264: is d{1} intended?

External/SPMmodified/spm_load.m
jan-petr: Besides - what happens if there's an entire empty row, then this will be fine. What happens if half of a row is empty and half of another row is empty?

External/SPMmodified/spm_load.m
```diff
@@ -191,3 +225,50 @@
     x = struct2cell(x);
     x = [x{:}];
 end
+
+
+end
+
+%% ========================================================================
+
+function [d] = FixIllegalEmptyCells(d, N)
```
jan-petr, line 234: Sorry, but that's a very messy function. Best would be to detect entire rows with empty cells and remove a list of rows and not run this iteratively, as this is quite inefficient.

External/SPMmodified/spm_load.m
```diff
@@ -162,11 +164,43 @@
 end
 d = textscan(S,'%s','Delimiter',delim);
 
-if rem(numel(d{1}),N)
+% EXPLOREASL HACK: detect potentially erroneous empty cells
+illegalEmptyCells = DetectIllegalEmptyCells(d, N);
+
+if ~isempty(illegalEmptyCells)
+    warning(['Found illegal empty cells in ' f ', trying to repair']);
+    d = FixIllegalEmptyCells(d, N);
```
jan-petr, line 172: I don't like that you first detect, then run fix that does detection as the first thing.

External/SPMmodified/spm_load.m
```diff
@@ -162,11 +164,43 @@
...
+
+    % Check if it was repaired
+    if rem(numel(d{1}),N) % can we divide cells/columns?
```
jan-petr, line 175: that's not a robust check

jan-petr (changes requested):
Sorry, but this needs a complete overhaul. Can you send the wrong file?! And I'll code it nicely.

HenkMutsaerts:
Great refactoring, but you now report and fix all empty cells; whereas we only should report and fix the empty cells at the end of lines (so too many delimiters at end of lines). Details:


171 "If there's a single line-end only in the data, then consider everything a single line a divide (if possible) by the number of columns" -> what do you mean? difficult to read

174 rem(numel(d{1}, N) -> this used to be the check if we can reshape by number of columns, so this check is now redundant?

208 Am I correct that you are here filling empty cells with empty strings? That's not what we should do. Empty cells are not the problem (if they were, this should be filled with 'n/a' per BIDS (or NaN)). By the way, difficult to read for me, I need more comments ;)

217 -> this warning is incorrect, it should only warn for empty cells at row ends, not for any empty cells

222 -> this warning is correct! We should add +1 though to compensate for the header.
And the repaired file is the original file, so the line endings still have too many delimiters (or empty cells)

`FixIllegalEmptyCells` and `DetectIllegalEmptyCells` are not used anymore?

jan-petr:
> Great refactoring, but you now report and fix all empty cells; whereas we only should report and fix the empty cells at the end of lines (so too many delimiters at end of lines). Details:

It fixes too much or too little cells per line. It reports all empty for convenience of user that he knows he has empty cells, but I can disable that.

> 171 "If there's a single line-end only in the data, then consider everything a single line a divide (if possible) by the number of columns" -> what do you mean? difficult to read

Fixed description.
 
> 174 rem(numel(d{1}, N) -> this used to be the check if we can reshape by number of columns, so this check is now redundant?

Not redundant, if there are no line ends (unlikely), then we cannot do anything else than check the total number of cells and divide by the number of columns. Any other strategy is wrong by design and error-prone.

> 208 Am I correct that you are here filling empty cells with empty strings? That's not what we should do. Empty cells are not the problem (if they were, this should be filled with 'n/a' per BIDS (or NaN)). By the way, difficult to read for me, I need more comments ;)

If cells are empty, then they are initialised as empty. I am only changing type of the empty cell from double to char. Because all the read cells are chars.

> 217 -> this warning is incorrect, it should only warn for empty cells at row ends, not for any empty cells

Sure. That was just for user information. Removed.

> 222 -> this warning is correct! We should add +1 though to compensate for the header. And the repaired file is the original file, so the line endings still have too many delimiters (or empty cells)

I added +1 to compensate for header. 
Repaired file is correct. I has the correct number of delimiters to have the same number of delimiters for each line. That is - having delimiters also for empty cells. It rectifies the number of delimiters as this should correctly be. Instead of removing the delimiters for empty cells which would have been wrong.

> `FixIllegalEmptyCells` and `DetectIllegalEmptyCells` are not used anymore?

jan-petr:
> > 208 Am I correct that you are here filling empty cells with empty strings? That's not what we should do. Empty cells are not the problem (if they were, this should be filled with 'n/a' per BIDS (or NaN)). By the way, difficult to read for me, I need more comments ;)
> 
> If cells are empty, then they are initialised as empty. I am only changing type of the empty cell from double to char. Because all the read cells are chars.
>
> HENK: Are you sure it won't read any cells as numeric? If all cells are read as char then this is fine of course; otherwise this is just redundant code and we can skip this whole part. We don't care about empty cells that are not at the eol.

Yes - numbers are read also as chars.
So this code makes it consistent. Maybe not really necessary, but it helps. Maybe not now, but for future functions, it is good to have clean outputs.
 

> > `FixIllegalEmptyCells` and `DetectIllegalEmptyCells` are not used anymore?
> HENK: ???

Yes - these are deleted - not used anymore.


> NEW TEST:
> 
> image
> 
> It still mentions all empty cells. It should only mention the empty cells at the end of the line (making a line having more cells than the header has). Did you remove the wrong warning? The warning on line 222 was correct, the warning on line 217 was incorrect.

Sorry. Yes I deactivated the wrong warning. Function was intact though. Just the warnings were switched.

As before. It finds lines with too few or too much delimiters (or empty cells). It doesn't really care if the cell is empty or not. Just checks really the delimiters. So reporting is fixed, but function unchanged.

> image
> 
> The code still doesn't remove empty cells at the end of the line... It should remove the empty cells at the end of the line, such that the number of delimiters is the same for each line.

I don't understand. This sentence seems contradictory. The number of delimiters is the same for each line - same as in the header. But that means that there can be empty cells at the end of the line. 

So the header has 9 cells. And each line should have 8 delimiters = 9 normal cells, or 5 normal cells and 4 empty, or 4 normal and 5 empty etc. And that's what you see in the figure you've sent - 6 filled cells and then 3 empty ones. 

I can of course try to write a function that has 6 normal cells on the line and no empty cells. But if I remove empty cells, then there won't be the delimiters and there will only be 5 delimiters per line = that's not what we want. And probably tsv_write doesn't support this anyway.

Please check and say exactly what to do with empty cells/delimiters at the end of the line. Previous comments were unclear.

Still - all that was in the code before. Only the warnings and comments in code changed....

jan-petr:
The function textscan treats a last empty cell (so a delimiter follow by line-end) as a non-existing cell (not empty, but completely skipping it). So "text,text,text EOL" is three cells (comma being a delimiter). "text,,,text EOL" is four cells. But "text, EOL" one cell. Though "text,,EOL" two cells.

So I've inserted an extra fix that adds an extra cell at the end of the line if it ends with a delimiter.

Now it all works.

jan-petr (approved):
OK.

HenkMutsaerts:
Funnily, I think that none of the changes were necessary for the `participants.tsv` that I now worked with, except for your very last commit ...

PR #1143, opened 2022-08-06
Closes #1142 xASL_init_checkDataSetRoot

### LINKED ISSUE #1142: Try correcting incorrect path specification of the user
### Description

When the user specifies the path including a subfolder such as `sourceStructure` or `derivatives/ExploreASL`, this should give a warning that points the user to this problem, and this can easily be attempted to correct to help the user and avoid them asking us for help.

BTW: Our ExploreASL_Initialization is very convoluted. E.g., it should only run `xASL_init_checkDataSetRoot` if a folder was specified as first argument, but instead it now always runs this function, as this function also deals with a missing dataSetRoot. So the modularity is missing of our initialization scripts.

### Tasks

- [x] `xASL_init_checkDataSetRoot`: Try fixing incorrectly specified folder or filenames

### How to test

### Release notes

Ignores JSON files or subdirectories in the root-directory input path.

HenkMutsaerts:
### Linked issue

Closes #1142

### How to test

### Comments

### DIFF
--- a/Functions/Initialization/xASL_init_checkDatasetRoot.m
+++ b/Functions/Initialization/xASL_init_checkDatasetRoot.m
@@ -30,55 +30,57 @@ function [x] = xASL_init_checkDatasetRoot(x)
     end
 
     % Track if a valid path was provided
-    validPath = true;
+    bValidPath = true;
 
     % Check if user correctly inserted a dataset root directory
     % Trying to fix incorrectly specified files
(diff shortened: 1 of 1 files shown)


HenkMutsaerts:
Nice, note that it the "the path" (when the path is specified) and "a path" (when it is not specified yet). So ExploreASL requires "a path" (not specified yet), and "the provided path" should be a file (already specified).

I avoided printing the path as this leads to very long warnings. But fine as it is now, you can merge with my final changes if you agree.

PR #1156, opened 2022-08-15
#1155 Fix path update before BIDS2Legacy

### LINKED ISSUE #1155: Path update before NII2BIDS
### Description

Import from BIDS to Legacy crashes when running import+processing at once.

Dataset_description.json path is not known until the conversion from DICOM2BIDS is finished. So this path ``x.dir.dataset_description`` can be only initialized once we have data in BIDS. So this path is first checked on initialization, and then updated after DICOM2BIDS - this was previously done. But previously, NII2BIDS was a part of import. And we now throw the initialization after import and start again with processing. So this path update was lost. 

So we simply need to move the path update function from IMPORT to PROCESSING (part NII2BIDS).

### Release notes

Fix bug in path update before NII2BIDS conversion.

jan-petr:
### Linked issue

Close #1155.

### DIFF
--- a/Modules/xASL_module_BIDS2Legacy.m
+++ b/Modules/xASL_module_BIDS2Legacy.m
@@ -44,6 +44,9 @@ function [result, x] = xASL_module_BIDS2Legacy(x)
     
     % Print feedback
     xASL_adm_BreakString('BIDS to ExploreASL LEGACY CONVERSION');
+	
+	% Update paths
+	x = xASL_imp_UpdateDatasetRoot(x);
     
     if ~isfield(x,'dir')
         error('Missing directories field...');


Modules/xASL_module_BIDS2Legacy.m
```diff
@@ -44,6 +44,9 @@
     
     % Print feedback
     xASL_adm_BreakString('BIDS to ExploreASL LEGACY CONVERSION');
+	
+	% Update paths
+	x = xASL_imp_UpdateDatasetRoot(x);
```
HenkMutsaerts, line 49: OK, this function already existed, and is now removed from the import module?

HenkMutsaerts (changes requested):

Modules/xASL_module_BIDS2Legacy.m
jan-petr: I've kept it in the import module not to mess up defacing.

HenkMutsaerts (approved):

Issue #1085, opened 2022-02-24
Add Time of exchange maps stats (ROI statistics)


BeatrizPadrela:
### Description

As ExploreASL does with the perfusion maps, would be nice to have stats for the time of exchange maps

### Release notes

Nothing. Done in #927

PR #1176, opened 2022-08-26
Feature #927 fabber quantification new

### LINKED ISSUE #927: FSL/FABBER model for multi-TE quantification
### Description

- @BeatrizPadrela got the FSL/FABBER scripts/model from @amahroo. Now we need to compile all of this and add it to ExploreASL.
- @jan-petr: Am I right that we have to add it in `xASL_quant_Basil` and specifically its sub-function `xASL_quant_Basil_Options`?
- How do we handle the compilation? Do we get something like a mex or do we get sh and exe files?

### Requirements

### How to test

### Release notes

jan-petr:
### Linked issue

Closes #927 

### How to test

I've tested it with BBB_VUmc data (Siemens PCASL 3D GRASE) both with 20 and 64ch coils

### Comments
 I merged quant_Fabber inside quant_Basil, with if statements for b.modules.asl.bMultiTE, but for the creation of the txt with the options to run FSL, since they were very different, I kept xASL_quant_Basil_Options and xASL_quant_Fabber_Options

the xASL_fsl_RunFSL stayed the same as now I'm using my personal directory with the copy of FSL plus Fabber. This path I add to the dataPar.json as:
image


Modules/SubModule_Population/xASL_wrp_GetROIstatistics.m
```diff
@@ -124,6 +124,8 @@ function xASL_wrp_GetROIstatistics(x)
     x.S.unit = 'mL/100g/min';
 elseif ~isempty(strfind(x.S.InputDataStr,'T1'))
     x.S.unit = 'mL';
+elseif ~isempty(strfind(x.S.InputDataStr,'TExch'))
```
jan-petr, line 127: Should we also add units for ATT? To be fixed.

Modules/xASL_module_Population.m
```diff
@@ -227,8 +227,8 @@
     % x.S.InputAtlasPath            = fullfile(x.D.MapsSPMmodifiedDir,'WholeBrain.nii');
     % xASL_wrp_GetROIstatistics( x);
 
-    x.S.InputDataStr = 'qCBF'; % 'SD' 'TT' 'M0' 'R1' 'ASL_HctCohort' 'ASL_HctCorrInd'
-	x.S.InputDataStrNative = 'CBF'; % 'SD' 'TT' 'M0' 'R1' 'ASL_HctCohort' 'ASL_HctCorrInd'
+    x.S.InputDataStr = 'qCBF'; % 'TExch' 'SD' 'TT' 'M0' 'R1' 'ASL_HctCohort' 'ASL_HctCorrInd'
```
jan-petr, line 230: Should we add here ATT? To be fixed.

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
```diff
@@ -244,15 +244,19 @@ function xASL_wrp_ResampleASL(x)
     elseif x.modules.asl.bTimeEncoded
         % Decoding of TimeEncoded data (Nifti is saved inside the function)
         ASL_im = xASL_quant_HadamardDecoding(PathASL{iSpace}, x.Q);
+        HadamardBlock = size(ASL_im,4) / (x.Q.TimeEncodedMatrixSize-1);
```
jan-petr, line 247: I am quite lost here what this means and how this is later used. Can you please improve comments in lines 247-257 and always very clearly say what should each variable contain?

HadamardBlock contains the number of repetitions of the block or possibly also repetitions*nTE (or rather only nTE, since repetitions are averaged in Decoding), so this will be always >=1. Then on line 249 - this condition will never be satisfied. For once, number of volume is always <= number of unique PLDs, and always < if you add HadamardBlock, which should be always >=1. 

I would need more explanation to understand this. Possibly fix the error, and more specific comments at each line to explain.

jan-petr (commented):
Very good!! I found several smaller bugs (that might not work for other sequences) and edited some comments and headers. I have partly fixed them, so I added a comment just so that you understand why I did that and can verify this. And there are also a few comments that still need fixing.

Before merging, we should run UnitTest and TestDataSet to be sure that we haven't messed up the quantification for non-DEBBIE sequences.

jan-petr:
One more question - can we close #1085 as this is already done here?

jan-petr:
> One more question - can we close #1085 as this is already done here?

Closing. But please check if Texch maps are correctly exported and if we can generate stats for it.

Functions/Quantification/xASL_quant_Basil.m
```diff
@@ -67,29 +70,45 @@
...
+        error('Something went wrong running FSL BASIL...'); 
     end
     
     fprintf('%s\n', 'The following warning (if mentioned above) can be ignored:');
     fprintf('%s\n', '/.../fsl/bin/basil: line 124: imcp: command not found');
     
     pathBasilCBF = xASL_adm_GetFileList(dirBasilOutput, '^mean_ftiss\.nii$', 'FPListRec');
-	if isempty(pathBasilCBF)
+    
+	if isempty(pathBasilCBF) && x.modules.asl.bMultiTE
```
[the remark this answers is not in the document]
BeatrizPadrela: to discuss

Functions/Quantification/xASL_quant_Basil.m
```diff
@@ -196,17 +223,18 @@
 		% Print all the PLDs and LabDurs
 		
 		[PLDs, ind] = unique(x.Q.Initial_PLD);
-		PLDs = PLDs'/1000;
+        PLDs = PLDs'/1000;
+        % For Time-encoded, we skip the first volume
+        if x.modules.asl.bTimeEncoded
+            N = length(PLDs)/x.Q.TimeEncodedMatrixSize;
+            PLDs = PLDs(1+N:end); % ex: if HAD8 with 2 different repetitions (16 PLDs), we will want 7+7 PLDs for the quantification. So PLDs(1+1:end) = 14 PLDs
```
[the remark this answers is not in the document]
BeatrizPadrela: to discuss

Functions/Quantification/xASL_quant_Basil.m
```diff
@@ -338,4 +366,87 @@
     
 end
 
+function [FIDoptionFile] = xASL_quant_Fabber_Options(pathFabberOptions, x, PWI, pathFabberInput) %have two functions, basil options and fabber options
```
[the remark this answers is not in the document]
BeatrizPadrela: to discuss

Functions/Quantification/xASL_quant_Basil.m
```diff
@@ -338,4 +366,87 @@
...
+% basil_options is a character array containing CLI args for the Basil command
+
+FIDoptionFile = fopen(pathFabberOptions, 'w+');
+
+[~,PWIfileName, ext] = fileparts(pathFabberInput);
+PWIfile = [PWIfileName ext];
+
+fprintf(FIDoptionFile, '# Fabber options written by ExploreASL\n');
+%fprintf(FIDoptionFile, 'fabber_asl\n');
+fprintf(FIDoptionFile, '--output=Basil_Output\n');
```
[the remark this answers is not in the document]
BeatrizPadrela: to discuss

Modules/SubModule_ASL/xASL_wrp_Quantify.m
```diff
@@ -435,10 +436,11 @@ function xASL_wrp_Quantify(x, PWI_Path, pathOutputCBF, M0Path, SliceGradientPath
     fprintf('%s\n', 'Performing single PLD quantification');
     [~, CBF] = xASL_quant_SinglePLD(PWI, M0_im, SliceGradient, x, x.Q.bUseBasilQuantification); % also runs BASIL, but only in native space!
 	ATT = [];
+	TExch = [];
```
HenkMutsaerts, line 439: We should come up with a default for 1) missing data and 2) processing going wrong, across the pipeline. BIDS has NaN or n/a for missing data/not available. So we could use 1) missing data == NaN or n/a 2) processing going wrong == []

Modules/SubModule_ASL/xASL_wrp_Quantify.m
```diff
@@ -509,6 +516,10 @@ function xASL_wrp_Quantify(x, PWI_Path, pathOutputCBF, M0Path, SliceGradientPath
     xASL_spm_deformations(x, {x.P.Path_CBF}, {x.P.Pop_Path_CBF}, [], [], AffineTransfPath, x.P.Path_y_ASL);
 	if xASL_exist(x.P.Path_TT,'file')
 		xASL_spm_deformations(x, {x.P.Path_TT}, {x.P.Pop_Path_TT}, [], [], AffineTransfPath, x.P.Path_y_ASL);
+    end
+    
+    if xASL_exist(x.P.Path_TExch,'file')
```
HenkMutsaerts, line 521: Do we have this for other outputs, e.g. BASIL ATT?

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
```diff
@@ -244,15 +244,23 @@ function xASL_wrp_ResampleASL(x)
     elseif x.modules.asl.bTimeEncoded
         % Decoding of TimeEncoded data (Nifti is saved inside the function)
         ASL_im = xASL_quant_HadamardDecoding(PathASL{iSpace}, x.Q);
+        PLD_vector_size = length(unique(x.Q.Initial_PLD));
```
HenkMutsaerts, line 247: PLD_vectorSize please (camelCase :)

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
```diff
@@ -244,15 +244,23 @@ function xASL_wrp_ResampleASL(x)
...
+			if mod(size(ASL_im,4), blockSize)
+				% Number of total volumes cannot be divided by blockSize
+				error(['Total number of volumes ' xASL_num2str(size(ASL_im,4)) ' cannot be composed of blocks of size ' xASL_num2str(blockSize)]);
+			else
+				PWI = zeros(size(ASL_im,1), size(ASL_im,2), size(ASL_im,3), blockSize); % preallocate PWI
+				for iBlock = 1:blockSize
+					PWI(:,:,:,iBlock) = xASL_stat_MeanNan(ASL_im(:,:,:,iBlock:blockSize:end), 4); % Averaged PWI4D across repetitions
+				end
+			end
+        else % if unique PLDs == decoded image size + HadBlock -> repetitions have different PLDs = we don't want to average
```
HenkMutsaerts, line 260: HadBlock -> Hadamard block?

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
```diff
@@ -244,15 +244,23 @@ function xASL_wrp_ResampleASL(x)
     elseif x.modules.asl.bTimeEncoded
         % Decoding of TimeEncoded data (Nifti is saved inside the function)
```
HenkMutsaerts, line 245: Please explain what you mean here by "Nifti is saved in the function"

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
```diff
@@ -244,15 +244,23 @@ function xASL_wrp_ResampleASL(x)
     elseif x.modules.asl.bTimeEncoded
         % Decoding of TimeEncoded data (Nifti is saved inside the function)
         ASL_im = xASL_quant_HadamardDecoding(PathASL{iSpace}, x.Q);
+        PLD_vector_size = length(unique(x.Q.Initial_PLD));
+        blockSize = x.Q.NumberEchoTimes * (PLD_vector_size - (PLD_vector_size/x.Q.TimeEncodedMatrixSize)); % Number of volumes per Hadamard block = EchoTimes*DecodedMatrixSize
```
HenkMutsaerts, line 248: Please add (or correct first):
"A Hadamard block is a individual labeling duration unit in the Hadamard labeling matrix design."

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
```diff
@@ -244,15 +244,23 @@ function xASL_wrp_ResampleASL(x)
...
         
-        % Hadamard Block size is calculated as number of TEs and the HadamardMatrixSize-1
-		blockSize = x.Q.NumberEchoTimes * (x.Q.TimeEncodedMatrixSize-1);
-		        
-        PWI = zeros(size(ASL_im,1), size(ASL_im,2), size(ASL_im,3), blockSize); % preallocate PWI
-        
-        for iBlock = 1:blockSize
-            PWI(:,:,:,iBlock) = xASL_stat_MeanNan(ASL_im(:,:,:,iBlock:blockSize:end), 4); % Averaged PWI4D across repetitions
+        if blockSize ~= size(ASL_im,4) % In case there is more than 1 block, we will average across repetitions
```
HenkMutsaerts, line 250: Is "Size" here duration? A block is a single LabelingDuration right? Should this then be `blockDuration` to clarify this?

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
```diff
@@ -244,15 +244,23 @@ function xASL_wrp_ResampleASL(x)
     elseif x.modules.asl.bTimeEncoded
```
HenkMutsaerts, line 244: Sorry that I only see this now, but `x.Q.NumberEchoTimes` is now only specified and used with `bTimeEncoded`, but this should also be used without Time encoded (i.e., for multiPLD or singlePLD). Does this always go correctly? Is this fourth dimension always echo time, for all type of Hadamard/multiPLD/singlePLD scans? Please generalize this.

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
```diff
@@ -244,15 +244,23 @@ function xASL_wrp_ResampleASL(x)
...
+					PWI(:,:,:,iBlock) = xASL_stat_MeanNan(ASL_im(:,:,:,iBlock:blockSize:end), 4); % Averaged PWI4D across repetitions
+				end
+			end
+        else % if unique PLDs == decoded image size + HadBlock -> repetitions have different PLDs = we don't want to average
+            PWI = ASL_im;
         end
+
         
         % Save PWI4D
         fprintf('%s\n', PathPWI4D{iSpace});
```
HenkMutsaerts, line 266: Did you test `xASL_wrp_ResampleASL` sections 9-11?
Do the SD images outputted make sense? And the wsCV reported?
How about the reported spatial CoV?
Any other temporary files that need deleting?

Modules/SubModule_ASL/xASL_wrp_VisualQC_ASL.m
```diff
@@ -211,29 +211,32 @@ function xASL_wrp_VisualQC_ASL(x)
...
 
 T.ColorMapIs{10}= x.S.jet256;
 T.ColorMapIs{11}= {x.S.jet256};
 
-T.NameExt       = {[] [] TextpWM []};
-T.NameExt( 5:8) = {[] [] TextpGM []};
-T.NameExt(9:11) = {[] TextpWM 'AnalysisMask'};
+T.NameExt        = {[] [] TextpWM []};
+T.NameExt( 5: 8) = {[] [] TextpGM []};
+T.NameExt( 9:10) = {[] TextpWM };
```
HenkMutsaerts, line 238: Why are 9:10 and 11:12 specified separately if they are the same?
And why is `AnalysisMask` removed?

Modules/xASL_module_ASL.m
```diff
@@ -208,6 +208,10 @@
 
 if ~isfield(x.Q, 'bUseBasilQuantification') || isempty(x.Q.bUseBasilQuantification)
 	x.Q.bUseBasilQuantification = false;
+
```
HenkMutsaerts, line 211: Nice, but this goes wrong if the user accidentally specified `bUseBasilQuantification` as `false`, this would skip multiTE and multiPLD. In this case, there should be a warning to the user that we only have multi-TE or multi-PLD processing with BASIL.

HenkMutsaerts (changes requested):
Can you please update/clean the issue, so it's easier to read? Now it still has text from Michael Stritt, and it says that the issue is halfway done.

Modules/SubModule_ASL/xASL_wrp_Quantify.m
```diff
@@ -80,6 +80,7 @@ function xASL_wrp_Quantify(x, PWI_Path, pathOutputCBF, M0Path, SliceGradientPath
 iStringCBF = regexpi(pathOutputCBF, 'CBF');
 iStringCBF = iStringCBF(end);
 pathOutputTT = [pathOutputCBF(1:(iStringCBF-1)) 'TT' pathOutputCBF((iStringCBF+3):end)];
```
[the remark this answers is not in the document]
jan-petr: I'm fine with that... Up to you. Currently, we just used the current parameter `TT`.

Functions/Quantification/xASL_quant_Basil.m
```diff
@@ -31,20 +34,28 @@
...
     pathBasilOptions = fullfile(x.dir.SESSIONDIR, 'Basil_ModelOptions.txt');
-    dirBasilOutput = fullfile(x.dir.SESSIONDIR, 'BasilOutput');
+    dirBasilOutput = fullfile(x.dir.SESSIONDIR, 'Basil_Output');
+	
+	% Define if BASIL or Fabber is used - so that this is 
+	if isfield(x.modules.asl, 'bMultiTE') && x.modules.asl.bMultiTE == 1
+		bUseFabber = 1;
+	else
+		bUseFabber = 0;
+	end
```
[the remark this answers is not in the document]
jan-petr: This was of course the case already - producing an empty image. If you want, we'll switch to `NaN` instead. Done in 031559f8

Functions/Quantification/xASL_quant_MultiPLD.m
```diff
@@ -120,9 +121,11 @@
 
     if xASL_stat_SumNan(ScaleImage(:))==0
         error('Wrong PLD definition!');
-	end
+    end
+    
+    % BASIL/FABBER quantification
+    [PWI, ATT, TExch] = xASL_quant_Basil(PWI, x);
```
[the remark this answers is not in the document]
jan-petr: They were defaulted to `[]` and it was all fine. But I'm fine switching to `NaN` - though it's not really that elegant as we need to check first for size and then for being `NaN`. Done in 031559f8

Modules/SubModule_ASL/xASL_wrp_Quantify.m
jan-petr: I've changed that to `NaN` already in 031559f8

Functions/Quantification/xASL_quant_HadamardDecoding.m
```diff
@@ -140,6 +147,19 @@
 	imEncoded = imEncoded(:,:,:,vectorOldOrder);
```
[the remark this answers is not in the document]
jan-petr: This is not about Walsh/Hadamard. But about switching order of TEs and PLDs.
So we have original (old order) and reordered (new order).

So we have old/new, or we can have reordered/original etc. I don't think that OLD is particularly bad, but can rename to something else.

jan-petr (commented):
See answers to your commits (I wrote some answers while checking the code, so to make them appear, I have to submit a review :)).

Functions/Quantification/xASL_quant_HadamardDecoding.m
```diff
@@ -149,7 +169,7 @@
             
             indexPositive = find(xQ.TimeEncodedMatrix(iTI,:)==1);
             indexNegative = find(xQ.TimeEncodedMatrix(iTI,:)==-1);
-            imDecoded(:,:,:,((iTE-1)*nDecodedTI+iTI)+(iRepetition-1)*nDecodedVolumes) = mean(imEncoded(:,:,:,(indexPositive+idx)),4) - mean(imEncoded(:,:,:,(indexNegative+idx)),4);
+            imDecoded(:,:,:,((iTE-1)*nDecodedTI+iTI)+(iRepetition-1)*nDecodedVolumes) = sum(imEncoded(:,:,:,(indexPositive+idx)),4) - sum(imEncoded(:,:,:,(indexNegative+idx)),4);
             
         end
         idx = idx+xQ.TimeEncodedMatrixSize;
```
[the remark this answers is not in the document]
jan-petr: Nono. It shouldn't. This is a local copy of x.Q. We were making the function more lightweight, so were passing only x.Q and not the entire x, so inside, we call it xQ.

Modules/SubModule_ASL/xASL_wrp_Quantify.m
jan-petr: I don't understand. Yes - we have this for TExch and ATT (see the line above).

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
jan-petr: Done in 9f3ee421

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
jan-petr: Fixed in 9f3ee421

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
jan-petr: Fixed in  9f3ee421

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
jan-petr: Fixed in 9f3ee421
Explained in a comment below.

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
jan-petr: Yes. But where there's no demand, there's no offer. We don't have such a sequence and it would take a lot of time to implement and test something that is then never used...

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
jan-petr: Good point - this was probably not addressed and tested. Do you want to do this here, or in a new issue?

Modules/SubModule_ASL/xASL_wrp_VisualQC_ASL.m
jan-petr: Of course not the same! That's the name of the file (which is also similar for CBF), but 9:10 is TT, 11:12 is TExch. So they of course have a similar naming convention, but each displays something else and goes to a different folder.

Modules/xASL_module_ASL.m
jan-petr: Yes - that warning of course is in the quantification module...

Modules/SubModule_ASL/xASL_wrp_Quantify.m
jan-petr: Using ATT instead of TT for multi-PLD: 6eeff88f

Modules/SubModule_ASL/xASL_wrp_Quantify.m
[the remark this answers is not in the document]
jan-petr: Added ATT for all multi-PLD: 6eeff88f

Functions/Quantification/xASL_quant_Basil.m
```diff
@@ -31,20 +34,28 @@
...
     
     %% 1. Define paths
+    % For Basil and Fabber
     pathBasilInput = fullfile(x.dir.SESSIONDIR, 'PWI4D_BasilInput.nii');
     pathBasilOptions = fullfile(x.dir.SESSIONDIR, 'Basil_ModelOptions.txt');
-    dirBasilOutput = fullfile(x.dir.SESSIONDIR, 'BasilOutput');
+    dirBasilOutput = fullfile(x.dir.SESSIONDIR, 'Basil_Output');
+	
+	% Define if BASIL or Fabber is used - so that this is 
+	if isfield(x.modules.asl, 'bMultiTE') && x.modules.asl.bMultiTE == 1
```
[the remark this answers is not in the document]
HenkMutsaerts: Then add it there please. I take the time to come up with these improvements, then you cannot say that we will do something in another issue and don't write the list there ;)

Functions/Quantification/xASL_quant_Basil.m
```diff
@@ -11,18 +11,21 @@
 % CBF_nocalib       - Quantified CBF image
 %                     (if there is no FSL/BASIL installed, we return the original PWI)
 % ATT_Map           - ATT map (if possible to calculate with multi-PLD)
+% Texch_map         - Time of exchange map of transport across BBB (if possible to calculate with multi-TE)
 % resultFSL         - describes if the execution was successful
 %                     (0 = successful, NaN = no FSL/BASIL found, 1 or other = something failed)
 %
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
-% DESCRIPTION: This script performs quantification of the PWI using the FSL Basil pipeline. Final calibration to
+% DESCRIPTION: This script performs quantification of the PWI using the FSL Basil/Fabber pipeline. Final calibration to
```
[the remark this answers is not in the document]
HenkMutsaerts: Mention it there then please

Functions/Quantification/xASL_quant_Basil.m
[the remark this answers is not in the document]
HenkMutsaerts: OK let's discuss how to standardize this in the future. You can resolve this.

Functions/Quantification/xASL_quant_Basil.m
```diff
@@ -31,20 +34,28 @@
...
+	if isfield(x.modules.asl, 'bMultiTE') && x.modules.asl.bMultiTE == 1
+		bUseFabber = 1;
+	else
+		bUseFabber = 0;
+	end
         
-    %% 2. Delete previous BASIL output
-    xASL_adm_DeleteFileList(x.dir.SESSIONDIR, '(?i)^basilOutput.*$', 1, [0 Inf]);
-    FolderList = xASL_adm_GetFileList(x.dir.SESSIONDIR, '(?i)^basilOutput.*$', 'FPList', [0 Inf], 1);
+    %% 2. Delete previous BASIL/Fabber output
```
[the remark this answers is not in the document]
HenkMutsaerts: Then please add it there ;)

This concerns a bit more. Let's say you first ran BASIL for getting a multi-PLD ATT and CBF map. Then you run it again to get the Texch map. In this case, you don't want to overwrite the first results with the other results. So at least the output should be saved somewhere. If we always ignore the output of BASIL/FABBER/whatever, then we might as well delete it when DELETETEMP is true, then we don't have this ambiguity.

Functions/Quantification/xASL_quant_Basil.m
```diff
@@ -67,29 +78,53 @@
...
+    else
+        [~, resultFSL] = xASL_fsl_RunFSL(['basil -i ' xASL_adm_UnixPath(pathBasilInput, ispc) ' -@ ' xASL_adm_UnixPath(pathBasilOptions, ispc) ' -o ' xASL_adm_UnixPath(dirBasilOutput, ispc) ' ' BasilOptions], x);
+    end
     
     % Check if FSL failed
-    if isnan(resultFSL)
-        error('FSL BASIL was not found, exiting...');
+    if isnan(resultFSL) 
+		if bUseFabber
+			error('FSL FABBER was not found, exiting...');
```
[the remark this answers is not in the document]
HenkMutsaerts: OK then please move this idea there.

Functions/Quantification/xASL_quant_Basil.m
```diff
@@ -100,7 +135,15 @@
 		ATT_map = xASL_io_Nifti2Im(pathBasilATT);
 	else
 		ATT_map = [];
-	end
+    end
+    
+    pathFabberTexch = xASL_adm_GetFileList(dirBasilOutput, '^mean_T_exch\.nii$', 'FPListRec');
+	if ~isempty(pathFabberTexch)
+		pathFabberTexch = pathFabberTexch{end}; % we assume the latest iteration (alphabetically) is optimal. also converting cell to char array
```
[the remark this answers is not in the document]
HenkMutsaerts: BASIL runs on anything UNIX (so at least Linux, macOS, and WSL). You discovered yesterday that alphabetic ordering might be slightly different, putting the M0 and deltaM in a different order for different OSes (#1201). I was just thinking if we need to take this into account also for other potential cases where we use sorting (such as (end) here. But here it is probably fine.

You can resolve this if you read this.

Functions/Quantification/xASL_quant_Basil.m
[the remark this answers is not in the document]
HenkMutsaerts: Sure but then PLEASE add this to the #1181 list.

Functions/Quantification/xASL_quant_Basil.m
```diff
@@ -338,4 +366,87 @@
...
+
+% For Time-encoded, we skip the first volume
+if x.modules.asl.bTimeEncoded
+    PLDs = PLDs(2:end);
+end
+
+if length(x.Q.LabelingDuration)>1
+    LDs = (x.Q.LabelingDuration(ind))'/1000;
+else
+    LDs = ones(size(PLDs))*x.Q.LabelingDuration/1000;
```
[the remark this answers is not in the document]
HenkMutsaerts: At least write it out, `LabelingDurations` instead of `LD`. Nobody uses LD

Functions/Quantification/xASL_quant_MultiPLD.m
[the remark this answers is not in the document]
HenkMutsaerts: OK let's discuss this and find a general solution across the pipeline. YOu can resolve this.

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
HenkMutsaerts: The rule is to generalize. If there is a special case then try to generalize. So we HAVE to come up with something that would work in all cases, that shouldn't be too difficult. E.g., (and please move this to #1181 ;)

Reason is: demand will come easily. By only creating new if-statements without generalization, code easily becomes convoluted. We need to simplify:

* `x.Q.numberEchoTimes` is always used and defaulted to 1
* Dimension 4 is always EchoTimes, Dimension 5 is always PLDs, Dimension 6 is always labeling duration (or some other order). These are all defaulted to 1

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
HenkMutsaerts: OK then this should be blockCount or number right? Size suggests a multimensional count, for a 1D count (e.g., sum, count, number of) you don't usually use Size.

In fact, `blockSize` suggests a continuous measure, like the duration of a single labeling block. Not a count of integers, like the number of blocks.

Unless this is used by everyone and chosen in OSIPI TF4.1, we should avoid confusing language.

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
HenkMutsaerts: I added this question to #1205, you can resolve this.

Modules/xASL_module_ASL.m
HenkMutsaerts: Let's move all those warnings concerning WHAT the ASL module will run, with what booleans, to the start of the ASL module. This makes it clear to the user right at the start of a module, rather than in the middle of running something, don't you agree? You're right that this is quantification-specific, but in a way this concerns the whole ASL module. To the user, it doesn't matter which submodule or wrapper lacks functionality with a wrong setting.

Shall we create a new issue or move this to an existing issue:
* Ensure that important warnings for boolean/settings/sequence/flavors combinations are issued at the start of the ASL module rather than somewhere in the middle of the code

HenkMutsaerts (changes requested):
Nearly done. Main thing is that either:

A) you agree with a comment but want to do it later: move it explicitly to a new issue and refer to that
B) you disagree with a comment and explain why
C) you agree with a comment and do it right away
D) you agree with a comment and ask the commenter to move it to a new issue or solve it

So you cannot say "this will be fixed in the future (issue bla bla) without explicitly listing it there, just thinking we will read it again. Everything we need to re-read costs time.

Functions/Quantification/xASL_quant_Basil.m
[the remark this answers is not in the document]
jan-petr: You can't run BASIL again to get Texch map - because it will always also recalculate the CBF and ATT map. Pointless not to overwrite. We should make it part of DELETETEMP. But it is good to overwrite here as well to avoid crashing.

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
jan-petr: I don't agree. We are not counting blocks, we are not labeling the blocks, we are saying how big a block is - how many fields it has. 

Sorry, but your comment does not really make sense. BlockSize is perfectly fitting for this. Same as vector size or vector length refer to the number of fields. Size doesn't suggest a continuous measure at all.

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
jan-petr: ok. make a new issue for that.

Modules/SubModule_ASL/xASL_wrp_ResampleASL.m
HenkMutsaerts: You say vector length (1D) and matrix size (>1D) right?

HenkMutsaerts (approved):
Approved

BeatrizPadrela (approved):
Niceee

PR #1192, opened 2022-09-21
Feature #435 update dcm2nii

### LINKED ISSUE #435: Update dcm2nii version
### Description
Update dcm2nii to a newer version

### Tasks
- [x] download and rename Windows and Linux files
- [x] download and rename Mac-OS installation
- [x] Run Flavors and identify mismatches
- [x] dcm2nii have identified previous errors with Philips scale slopes - we might want to update the dcm2nii version once it is out (15.3.21 was announced to fix this).
- [x] We also want to check if it helps with some MOSAIC errors.
- [x] See if this fixes some issues in #342 
- [x] See if this fixes some issues in #1167 
- [x] See if this fixes some issues in #1170 
- [x] Write HK guys that their GE RX28.0 is ready and working.
- [x] See if this fixes some issues in #1133  
- [x] This will help Yeva with her FLAIR datasets, also GE DV22.0 is struggling with this, and also GE_Premier_RX28. See #1170.
- [x] We need to test if all the imports and scalings are applied in the same way as before.
- [x] We should have an option for use of older dcm2nii in studyPar.json or elsewhere.
- [x] In Philips scaling function - we should check for dcm2nii version so that this is used correctly.
- [x] Let Maurice know about this as he wants to test his GE dataset.
- [x] Fix all GE flavors
- [x] Fix all Siemens flavors
- [x] Fix all Philips flavors
- [x] Philips_PCASL_2DEPI_3.2.2.1_2 - RVW, different echo time, different scaling. It applies the scalings in an incorrect direction - multiplies instead of divides.
- [x] Philips_PCASL_2DEPI_5.3.1.1_1 - RVW, different scaling!!

Need another merge:
- [x] Philips_PCASL_2DEPI_PARREC_1 - ASL.json. OK, but need to merge M0 fix to make it work properly.
- [x] Siemens_PASL_2DEPI_VB15A_ADNI2_941_S_4764_1 - All missing. It won't work as it has zipped DICOMs
- [x] Siemens_PASL_3DGRASE_VE11C_1 - M0.json. OK, but need to merge M0 fix to make it work properly.
- [x] Siemens_PASL_3DGRASE_VE11C_2 - M0.json. OK, but need to merge M0 fix to make it work properly.
- [x] Siemens_PCASL_3DGRASE_VB17A_multiPLD_1 - M0.json. OK, but need to merge M0 fix to make it work properly.

### Release notes
Updated dcm2nii to a 20220720 version

jan-petr:
### Linked issue

Closes #435


Functions/Administration/xASL_adm_GetPhilipsScaling.m
```diff
@@ -29,45 +30,59 @@
...
+			end
+			
+			if isfield(parms,'MRScaleSlope') && length(parms.MRScaleSlope)>1
+				parms.MRScaleSlope = parms.MRScaleSlope(idx);
+			end
 		end
 		
-		if isfield(parms,'RescaleSlope') && length(parms.RescaleSlope)>1
-			parms.RescaleSlope = parms.RescaleSlope(idx);
+		if ~isempty(parms.RWVSlope)
```
HenkMutsaerts, line 64: Isn't it called `RealWorldValue` fully in DICOM? I would write this out fully (or add multiple comments and explain in the header that RWV = RealWorldValue and Philips specific).

You said in another comment that an internal variable only doesn't need to be in the header, but I believe it's nice  that a header helps reading the code as well as using it.

Functions/Administration/xASL_adm_GetPhilipsScaling.m
```diff
@@ -29,45 +30,59 @@
...
+			if ~isnear(parms.RWVSlope,rescaleSlopeNifti,parms.RWVSlope/100) && (rescaleSlopeNifti ~= 1)
```
HenkMutsaerts, line 65: Here you only print stuff, you don't actually execute what you print here that you will do.

I would put the execution in the same if-statement as the printing, so we are sure that the messages on the screen/log file represent what was actually done.

Functions/BIDS/xASL_bids_BIDSifyASLJSON.m
```diff
@@ -128,6 +130,14 @@
 	end
 end
 
+% In Philips and dcm2niix20220720, often RepetitionTimePreparation is assigned the time right after the last readout and 
+% RepetitionTimeExcitation the length of the whole cycle - that needs to be fixed.
+if strcmpi(jsonOut.Manufacturer,'Philips')
+	if isfield(jsonOut,'RepetitionTimeExcitation') && jsonOut.RepetitionTimeExcitation >= jsonOut.RepetitionTimePreparation
```
HenkMutsaerts, line 136: This is weird, we should choose one out of the two options:

Option 1. jsonOut == jsonIn and we run `rmfield` for each exception
Option 2. jsonOut == empty and we fill it per structure/field.

Now we use option 2 above for `RepetitionTimePreparation`and option 1 here?

Functions/BIDS/xASL_bids_BIDSifyASLJSON.m
```diff
@@ -151,8 +161,16 @@
 		if dimASL(4)>=numel(jsonOut.GELabelingDuration)
 			warning('Labeling duration mismatch with GE private field.');
```
HenkMutsaerts, line 162: which private field?

Functions/BIDS/xASL_bids_BIDSifyASLJSON.m
```diff
@@ -151,8 +161,16 @@
...
 			jsonOut.LabelingDuration = jsonOut.GELabelingDuration;
+		elseif dimASL(4)>=numel(unique(jsonOut.GELabelingDuration))
+			warning('Labeling duration mismatch with GE private field.');
```
HenkMutsaerts, line 165: Which private field? Add the name/number to the warning. Helps us debugging.

Functions/BIDS/xASL_bids_BIDSifyASLJSON.m
```diff
@@ -151,8 +161,16 @@
...
+			labDurTemp = unique(jsonOut.GELabelingDuration);
+			if labDurTemp(1) == 0
+				labDurTemp(1:end-1) = labDurTemp(2:end);
+				labDurTemp(end) = 0;
+			end
+			jsonOut.LabelingDuration = labDurTemp;
```
HenkMutsaerts, line 171: `labelingDurationTemp` instead of the abbreviation

Functions/BIDS/xASL_bids_BIDSifyASLJSON.m
```diff
@@ -151,8 +161,16 @@
 		if dimASL(4)>=numel(jsonOut.GELabelingDuration)
 			warning('Labeling duration mismatch with GE private field.');
 			jsonOut.LabelingDuration = jsonOut.GELabelingDuration;
+		elseif dimASL(4)>=numel(unique(jsonOut.GELabelingDuration))
+			warning('Labeling duration mismatch with GE private field.');
+			labDurTemp = unique(jsonOut.GELabelingDuration);
```
HenkMutsaerts, line 166: For consistency labelingDurationX throughout? `labelingDuration`, `labelingDurationTemp`, `labelingDurationGE`

Functions/BIDS/xASL_bids_BIDSifyASLJSON.m
```diff
@@ -169,7 +187,7 @@
 				warning('PostLabelingDelay mismatch with the GE-DICOM value in Inversion time.');
```
HenkMutsaerts, line 187: `GE-DICOM value in Inversion time` -> `GE DICOM Inversion time value` right?
Now it could suggest that Inversion time contains multiple values, or multiple vendors, or whatever.

Functions/BIDS/xASL_bids_BIDSifyASLJSON.m
```diff
@@ -178,6 +196,13 @@
 	end
 end
 
+% For GE and multi-PLD or single-PLD not defined in the GELabelingDurationField, we prefer LabelingDuration from study par due to issues with eASL
+if strcmp(jsonIn.Manufacturer, 'GE') && isfield(studyPar,'LabelingDuration') && ~isequal(studyPar.LabelingDuration,jsonOut.LabelingDuration) &&...
+	( length(jsonOut.LabelingDuration)>1 || ~isfield(jsonIn,'GELabelingDuration'))
+	warning('Labeling duration differs in DICOM and studyPar.');
```
HenkMutsaerts, line 202: differs between. Otherwise, it could strictly speaking differ within the DICOM (having two values) and within the studyPar (having two values)

Functions/BIDS/xASL_bids_Config.m
```diff
@@ -28,22 +28,22 @@
...
-	'SaturationStopTime','BaseResolution','DerivedVendorReportedEchoSpacing','RawImage','PhaseOversampling','BolusDuration','scaleFactor','EchoNumber'}; % Fields to exclude as not defined in BIDS
+	'EchoTrainLength','PhaseEncodingSteps','BodyPartExamined','ShimSetting','TxRefAmp','PhaseResolution','InstanceNumber', 'WaterFatShift',...
+	'RefLinesPE','BandwidthPerPixelPhaseEncode','ImageComments','ConsistencyInfo','WipMemBlock','Interpolation2D','TotalAcquiredVolumes', 'NumberOfExcitations', 'PrescanReuseString', ...
+	'SaturationStopTime','BaseResolution','DerivedVendorReportedEchoSpacing','RawImage', 'T1', 'PercentSampling', 'ImageOrientationText', 'PhaseOversampling','BolusDuration','scaleFactor','EchoNumber'}; % Fields to exclude as not defined in BIDS
 
 % A list of fields to remove the ASL-BIDS JSON files only
 bidsPar.listFieldsRemoveASL = {'SliceReadoutTime','RepetitionTime','InversionTime','LabelOffset','PostLabelDelay','NumRFBlocks',...
 	'GELabelingDuration','RFGap','MeanGzx10','PhiAdjust','M0','LabelingType','ScanType','SequenceType'}; % Fields to exclude from ASL only
 
 % A list of fields to remove the ASL-M0 JSON files only
```
HenkMutsaerts, line 45: to remove from? Otherwise unclear.

Why do we remove fields here, because we copied the `asl.json` to `m0.json`? Or because Chris Rorden outputs too many fields? (Please clarify this here in a comment)

Functions/BIDS/xASL_bids_determineImageTypeGE.m
```diff
@@ -40,14 +40,26 @@
     if length(jsonPar.ImageType) == 5 && strcmpi(jsonPar.ImageType{1},'DERIVED') && strcmpi(jsonPar.ImageType{2},'PRIMARY') &&...
```
HenkMutsaerts, line 40: Some of these typical names might be vendor-independent, if it helps to know this in the future.

Functions/Import/xASL_imp_NII2BIDS_RunPerf.m
```diff
@@ -60,11 +60,30 @@ function xASL_imp_NII2BIDS_RunPerf(imPar, bidsPar, studyPar, subjectSessionLabel
     end
     if xASL_exist(fullfile(inSessionPath, [aslLabel '.nii']),'file')
         headerASL = xASL_io_ReadNifti(fullfile(inSessionPath, [aslLabel '.nii']));
```
HenkMutsaerts, line 62: What is "Label" here? Can we use the BIDS nomenclature? E.g. `subjectSessionValue` or `subjectSessionKey`? Can we replace "aslLabel" by what it is (e.g., `ASL4D_legacyLabel`, or `asl_FileName`). Now I have to open the actual code to understand what this means (though it probably means the FileName?

Functions/InputOutput/xASL_io_dcm2nii.m
```diff
@@ -79,34 +79,34 @@
...
 		case {'MACI','MACI64'}
-			ExePath = fullfile(mricron_version_path,'dcm2nii-osx');
+			ExePath = fullfile(mricron_version_path,'dcm2niix-osx');
 		otherwise
 			error('Unknown computer architecture: %s',computer());
 	end
     [stat, attr] = fileattrib(ExePath);
     if ~stat
-        error('dcm2nii application not found: %s',ExePath);
+        error('dcm2niix application not found: %s',ExePath);
```
HenkMutsaerts, line 101: Same as above. not required to replace all `dcm2nii` to `dcm2niiX`, but fine here in the code.

Functions/InputOutput/xASL_io_dcm2nii.m
```diff
@@ -150,6 +150,36 @@
     %   -x : crop (y/n, default n)
     %   -z : gz compress images (y/i/n/3, default n) [y=pigz, i=internal:miniz, n=no, 3=no,3D]
 
+	% -1..-9 : gz compression level (1=fastest..9=smallest, default 6)
```
HenkMutsaerts, line 153: Why do we add this list, also: do these parameters change between dcmnii versions?

HenkMutsaerts (changes requested):
* Perhaps we can delete the dcm2nii versions that we don't use anymore.
* Note that `dcm2nii` became `dcmniiX` a few years ago (approx when he started outputting JSON), so 2013 and 2010 may be wrongly renamed
*

Functions/Administration/xASL_adm_GetPhilipsScaling.m
```diff
@@ -29,45 +30,59 @@
...
-    fprintf('=========================================== Scaling ==========================================\n');
-	if length(parms.RWVSlope)>1
-		[~,idx] = min(abs(parms.RWVSlope - rescaleSlopeNifti));
-		parms.RWVSlope = parms.RWVSlope(idx);
-		if isfield(parms,'RescaleSlopeOriginal') && length(parms.RescaleSlopeOriginal)>1
-			parms.RescaleSlopeOriginal = parms.RescaleSlopeOriginal(idx);
+	% Obtains dcm2niix version used for conversion and ExploreASL version
+	% If ExploreASL version >= 1.10.0 or dcm2niix version >= 20210317
+	if isfield(parms,'ConversionSoftware') && (strcmp(parms.ConversionSoftware,'dcm2nii') || strcmp(parms.ConversionSoftware,'dcm2niix')) && isfield(parms,'ConversionSoftwareVersion')
+		dcm2niixVersion = parms.ConversionSoftwareVersion;
```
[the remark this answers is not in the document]
jan-petr: No guessing. There is the software name and version. So that we're not limited to dcm2nii.

jan-petr:
> * Perhaps we can delete the dcm2nii versions that we don't use anymore.
> * Note that `dcm2nii` became `dcmniiX` a few years ago (approx when he started outputting JSON), so 2013 and 2010 may be wrongly renamed

Up to you. I'm fine removing old version. I would just keep 2019 in case we have issues with 2022. All our flavors work with 2022 or 2019 - older are not necessary.

Functions/Administration/xASL_adm_GetPhilipsScaling.m
```diff
@@ -29,45 +30,59 @@
...
+		dcm2niixVersion = str2num(dcm2niixVersion(end-7:end));
+	else
+		dcm2niixVersion = 0;
+	end
+	
+	% It has to use the correct dcm2niix version that is able to handle this parameter and it also needs to report that it did do that
+	if dcm2niixVersion >= 20210317 && isfield(parms,'UsePhilipsFloatNotDisplayScaling') && (parms.UsePhilipsFloatNotDisplayScaling == 1)
```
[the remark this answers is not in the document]
jan-petr: `>=` is correct. Because starting with this version, the bug was fixed. So all newer versions handle it correctly, that's why we need `>=` and not `==`.

Functions/Administration/xASL_adm_GetPhilipsScaling.m
```diff
@@ -29,45 +30,59 @@
...
+		scaleFactor = 0;
```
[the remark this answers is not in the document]
jan-petr: scaleFactor == 0 is correct. It won't give errors - it is an internal parameter that is supposed to be 0 when scalings are to be kept as they were. It is not missing (`NaN`), it means that it shouldn't be applied. There won't be any confusion, this value is only used internally in this function and it means that you should keep scalings as they are.

Functions/Administration/xASL_adm_GetPhilipsScaling.m
```diff
@@ -29,45 +30,59 @@
...
+	else
+		% If the RealWorldValue is present, then dcm2nii scales to them and ignores everything else
+		fprintf('=========================================== Scaling ==========================================\n');
```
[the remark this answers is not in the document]
jan-petr: The whole function is called GetPhilipsScaling - we probably don't need to mention Philips also inside or have vendor sections :D Siemens and GE doesn't have anything similar now. Their specific needs will be addressed in the future. And then we can redesign this - but not now.

Functions/Administration/xASL_adm_GetPhilipsScaling.m
jan-petr: We are sure - the execution is 5 lines below as `RWVSlope` is always used in this paragraph. So no point writing it five times in each branch, when it is executed in all branches.

Functions/BIDS/xASL_bids_BIDSifyASLJSON.m
jan-petr: No point adding DICOM tag here - cause that is read elsewhere or in dcm2nii, so that would be prone to future error.

Functions/BIDS/xASL_bids_determineImageTypeGE.m
jan-petr: In the future yes. Currently, they are pretty vendor specific.

Functions/InputOutput/xASL_io_dcm2nii.m
jan-petr: Yes - there was no need to have any of those helps. I removed them all. See 43d03f7b

Functions/BIDS/xASL_bids_Config.m
jan-petr: Yes - there are far too many fields that don't exist in BIDS, so we remove them. Explained in a comment - see 43d03f7b

Functions/BIDS/xASL_bids_BIDSifyASLJSON.m
jan-petr: Fixed in 43d03f7b

Functions/BIDS/xASL_bids_BIDSifyASLJSON.m
jan-petr: Fixed in 43d03f7b

Functions/BIDS/xASL_bids_BIDSifyASLJSON.m
jan-petr: Fixed in 43d03f7b

Functions/BIDS/xASL_bids_BIDSifyASLJSON.m
jan-petr: Yes, can fix. but that's a hell of a work - so will take me some time.

Functions/BIDS/xASL_bids_BIDSifyASLJSON.m
jan-petr: clarified that it's the studyPar versus DICOM. See 43d03f7b

Functions/Administration/xASL_adm_GetPhilipsScaling.m
jan-petr: Sure. Explaining now in the header and giving the DICOM tag numbers. See 43d03f7b

Functions/Administration/xASL_adm_GetPhilipsScaling.m
```diff
@@ -29,45 +30,59 @@
...
+		if length(parms.RWVSlope)>1
```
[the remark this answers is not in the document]
jan-petr: Agree. Fixed in 43d03f7b

Functions/Administration/xASL_adm_GetPhilipsScaling.m
```diff
@@ -29,45 +30,59 @@
...
+	% Obtains dcm2niix version used for conversion and ExploreASL version
+	% If ExploreASL version >= 1.10.0 or dcm2niix version >= 20210317
+	if isfield(parms,'ConversionSoftware') && (strcmp(parms.ConversionSoftware,'dcm2nii') || strcmp(parms.ConversionSoftware,'dcm2niix')) && isfield(parms,'ConversionSoftwareVersion')
+		dcm2niixVersion = parms.ConversionSoftwareVersion;
+		dcm2niixVersion = str2num(dcm2niixVersion(end-7:end));
+	else
+		dcm2niixVersion = 0;
+	end
+	
+	% It has to use the correct dcm2niix version that is able to handle this parameter and it also needs to report that it did do that
```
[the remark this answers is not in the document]
jan-petr: Good catch. Clarified in 43d03f7b

Functions/BIDS/xASL_bids_BIDSifyASLJSON.m
jan-petr: GELabelingDuration is a DICOM tag. All our DICOM tags have vendor name as first. So it is better to keep the DICOM tag names consistent (also with dcm2nii style). So I changed to tempLabelingDuration - now it is all consistent.

Functions/Import/xASL_imp_NII2BIDS_RunPerf.m
jan-petr: `label` IS a BIDS nomenclature :)
But `aslLabel` was indeed a reference to the legacy name. So better to use legacy and not label.

Functions/BIDS/xASL_bids_BIDSifyASLJSON.m
jan-petr: Not weird - DICOM, vendors, and dcm2nii behaves very weirdly. So we have to correct for their inconsistencies ad-hoc. Unfortunately. What you propose would only work if these people would stick to standards.

If a messy standard is on input and you want to provide a solid standard on output, you have to go for ad-hoc exceptions and fixes. The specific reason is given in the comment.

Functions/Administration/xASL_adm_GetPhilipsScaling.m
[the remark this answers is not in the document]
HenkMutsaerts: But why not using the pre-existing field?

Also, there's also a comment above this one

Functions/Administration/xASL_adm_GetPhilipsScaling.m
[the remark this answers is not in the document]
HenkMutsaerts: OK then I propose having another boolean `bChangeScaleSlopes` or something; this is confusing. You could potentially have both at the same time right: you want to keep the scale slope but it also has a value.

Functions/BIDS/xASL_bids_determineImageTypeGE.m
HenkMutsaerts: OK can we then make a Philips section (and for other vendors)? That shouldn't take too much time, and does help code readability a lot

Functions/Administration/xASL_adm_GetPhilipsScaling.m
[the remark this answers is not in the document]
jan-petr: Because there are no pre-existing field for conversionsoftware name and conversion software version. I had to create new fields.

Functions/BIDS/xASL_bids_determineImageTypeGE.m
jan-petr: Yes. but we don't have yet sections for other, because we can't read that for others. We have issues for that. So best we make sections for Philips and Siemens once we have them.

HenkMutsaerts:
Had some issues with testing, see Whatsapp and #1194. Otherwise, it ran nicely.

* there is illegal character issue for [ ]. Perhaps we can fix this by putting the paths in strings for Linux/macOS: " ", or by removing more illegal characters from the path with our replacement functions. But the latter would involve overwriting the sourcedata which should be read-only, so it has to be the " ". Or first testing if the folder names don't have illegal characters (BTW: Windows should be much worse for illegal characters...) 

>> executing: [/Users/henk/ExploreASL/ExploreASL/External/MRIcron/20220720/dcm2niix-osx -o /Users/henk/ExploreASL/FlavorDatabase/GE_PCASL_3Dspiral_DV25.0_product_1/derivatives/ExploreASL/temp/001/ASL_1/dcm2nii_temp_ASL4D /Users/henk/ExploreASL/FlavorDatabase/GE_PCASL_3Dspiral_DV25.0_product_1/sourcedata/sub-001/20/S 007 [3D ASL]/Im181.0.dcm]
zsh:1: bad pattern: [3D

Functions/Administration/xASL_adm_GetPhilipsScaling.m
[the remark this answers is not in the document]
jan-petr: I found a way to fix this that is more clean.
I have now both a boolean.
And missing/non-applying/default scaleSlope is 1 and this is fixed in all connected functions.
See bc5022ed

Functions/Administration/xASL_adm_GetPhilipsScaling.m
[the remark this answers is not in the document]
HenkMutsaerts: I thought we had `impar.dcm2nii_version`. Are you sure we got rid of that? It would be nice to use the same names throughout the import.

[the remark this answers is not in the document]
jan-petr: Yes. we have that. but that's a specific field for dcm2nii. I wanted to have general fields that would describe: (i) what software it was (if other than dcm2nii) and (ii) what was the version of that software (not necessarily dcm2nii). 
So the current single field dcm2nii wasn't really good for that.

So I revamped stuff going from very specific, non-modular, hard to read stuff. To a nice new revamped stuff.

Functions/InputOutput/xASL_io_dcm2nii.m
```diff
@@ -73,9 +73,6 @@
     msg = [];
 
     %% 3. Locate dcm2nii executable
-	%if regexpi(inpath,'PAR$') % Automatic detection of PAR-REC, now disabled
-	%	imPar.dcm2nii_version = '20101105';
-	%end
 	
 	if ismac && str2num(imPar.dcm2nii_version(1:4))<2014
```
HenkMutsaerts, line 77: I guess this can move out, we don't have any dcm2nii_version before 2014 anymore :)
And macOS is fine with the versions that we have now

Functions/Administration/xASL_adm_GetPhilipsScaling.m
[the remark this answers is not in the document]
HenkMutsaerts: OK I just don't understand what other versions we have, we already have a field for ExploreASL software, and we don't use anything else than dcm2niiX? And just so I understand: we still have the same field but you renamed it? Let's move this discussion to the new issue, then you can resolve this conversation :)

HenkMutsaerts (changes requested):
...

Functions/Administration/xASL_adm_GetPhilipsScaling.m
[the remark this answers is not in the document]
jan-petr: Yes - but that's talking about the whole import.
And dcm2nii_version is a parameter on input, and it is still there and can be used.

We need a field that will record the software actually used for DICOM2NII.

Functions/InputOutput/xASL_io_dcm2nii.m
jan-petr: removed 06d28576

Functions/Administration/xASL_adm_GetPhilipsScaling.m
HenkMutsaerts: OK then only improve the comment on line 92 "set the scaling to remove"

Functions/BIDS/xASL_bids_BIDSifyASLJSON.m
HenkMutsaerts: OK, so `xASL_bids_BIDSifyASLJSON` merges `studyPar` and `jsonIn` to `jsonOut` right?

Can we then change it such that:
1) you always take `jsonOut` = `jsonIn` OR `studyPar` (NOT `jsonIn` = `jsonOut` or `jsonOut` = `jsonOut`)
2) merge `xASL_bids_BIDSifyASLJSON` and `xASL_bids_MergeStudyPar`

(Again I'm fine not doing that here but moving this to a new issue)

Functions/Administration/xASL_adm_GetPhilipsScaling.m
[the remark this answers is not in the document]
HenkMutsaerts: OK then we should create a general single Provenance field, rather than having to search for the specific Provenance names you came up with ;)

You always use the input parameter field as version right?
Otherwise, throughout the processing, we would need to record from each parameter what the user wanted and what the pipeline did in the end... I would only record what the pipeline did.

Functions/BIDS/xASL_bids_BIDSifyASLJSON.m
HenkMutsaerts: No sorry that's not what I meant. I meant that you actually print WHICH private field mismatched. E.g.,:

>> warning('studyPar parameter LabelingDuration=1.6 mismatch with DICOM field LabDur=1.8, using 1.8.'); (in the case that the DICOM field is called LabDur)

Again, this makes it easier for us to help users and check for bugs, as you have 1000 potential mismatch warnings, you better specifiy what they are.

jan-petr: Sure. Clarifying warnings makes sense. Done as suggested in 337e00c4

Functions/Administration/xASL_adm_GetPhilipsScaling.m
jan-petr: Done 337e00c4

Functions/BIDS/xASL_bids_BIDSifyASLJSON.m
jan-petr: Agree that the mixing of variables was confusing. I'm now merging studyPar+jsonIn to jsonInMerged. jsonIn and studyPar are used, but never modified. jsonInMerged is then cleaned, but we always write in the direction jsonInMerged->jsonOut.

HenkMutsaerts (approved):

PR #1207, opened 2022-10-13
Closes #1197 xASL_module_Population: add x.S.DataTypes

### LINKED ISSUE #1197: Iterate Statistics over multiple DataTypes (E.g., TExch, SD, etc)
### Description

If we had another string to line 230 (see printscreen below), we need to iterate the ROIstatistics through the measures as well (so not only qCBF but also TExch, SD, etc)
image

### Tasks

- [x] Check if x.S.InputDataStr can be a cell array
- [x] Add this option to dataPar.json, so that we don't need to hardcode everytime we need the stats for TExch e.g.

### Release notes
Allow configuring the image contrast for statistics in dataPar.json.

HenkMutsaerts:
### Linked issue

Closes #1197

### How to test

- [x] Add "x.S.DataTypes": ["qCBF", "Texch"] to `dataPar.json` and run the population module @BeatrizPadrela to test if this new feature works
- [ ] Run the population module without this new setting to see if the default still works (it should default to `qCBF`as it originally did)

### Comments

@jan-petr Should we add this new parameter to the `dataParTemplate.md` (this is now removed and only online right?)

### DIFF
--- a/Modules/xASL_module_Population.m
+++ b/Modules/xASL_module_Population.m
@@ -209,6 +209,8 @@ if ~x.mutex.HasState(StateName{8})
     % Default datatypes
     if ~isfield(x.S,'DataTypes')
         x.S.DataTypes = {'qCBF'}; % Default
+        % Alternatives: 'Texch' 'ATT' 'SD' 'TT' 'M0' 'R1' 'ASL_HctCohort' 'ASL_HctCorrInd'
+        % These can be added in the dataPar manually
     end
     
     
@@ -223,14 +225,13 @@ if ~x.mutex.HasState(StateName{8})
     % ROI statistics
     % x.S.SubjectWiseVisualization =1; % set this on to visualize the subject-wise masks
     % over CBF maps (takes lot of extra time though)
-    
-    x.S.InputDataStr = 'qCBF'; % 'Texch' 'ATT' 'SD' 'TT' 'M0' 'R1' 'ASL_HctCohort' 'ASL_HctCorrInd'
-	x.S.InputDataStrNative = 'CBF'; % 'Texch' 'ATT' 'SD' 'TT' 'M0' 'R1' 'ASL_HctCohort' 'ASL_HctCorrInd'
+
     
     % Iterate over DataTypes
     for iDataType=1:length(x.S.DataTypes)
         x.S.InputDataStr = x.S.DataTypes{iDataType};
-        if isempty(regexp(x.S.InputDataStr(1),'(q|r)'))
+        
+        if isempty(regexp(x.S.InputDataStr(1), '(q|r)', 'once'))
             % Some DataTypes have a prefix in standard space but not in
             % native space: e.g.,
             % qCBF in standard space is CBF in native space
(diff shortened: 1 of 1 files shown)


Modules/xASL_module_Population.m
```diff
@@ -201,75 +201,84 @@
...
-    % x.S.InputAtlasPath            = fullfile(x.D.MapsSPMmodifiedDir,'WholeBrain.nii');
-    % % x.S.InputAtlasPath            = fullfile(x.D.AtlasDir,'HOcort_CONN.nii');
-    % xASL_wrp_GetROIstatistics( x);
-    % x.S.InputDataStr              = 'SNR'; % 'SD' 'TT' 'M0' 'R1' 'ASL_HctCohort' 'ASL_HctCorrInd'
-    % x.S.InputAtlasPath            = fullfile(x.D.MapsSPMmodifiedDir,'WholeBrain.nii');
-    % xASL_wrp_GetROIstatistics( x);
-
-    x.S.InputDataStr = 'qCBF'; % 'SD' 'TT' 'M0' 'R1' 'ASL_HctCohort' 'ASL_HctCorrInd'
-	x.S.InputDataStrNative = 'CBF'; % 'SD' 'TT' 'M0' 'R1' 'ASL_HctCohort' 'ASL_HctCorrInd'
+    x.S.InputDataStr = 'qCBF'; % 'Texch' 'ATT' 'SD' 'TT' 'M0' 'R1' 'ASL_HctCohort' 'ASL_HctCorrInd'
```
jan-petr, line 227: These lines can be probably commented out. This value will be always overwritten.

Modules/xASL_module_Population.m
```diff
@@ -201,75 +201,84 @@
...
     
-    % Default & Fallback for Atlases
+    % Default atlases/ROIs
     if ~isfield(x.S,'Atlases')
         x.S.Atlases = {'TotalGM','DeepWM'}; % Default
     end
-
+    
+    % Default datatypes
+    if ~isfield(x.S,'DataTypes')
```
jan-petr, line 210: Probably add
|| isempty(x.S.DataTypes)
just to be sure.

jan-petr (changes requested):
Two minor things.
Also, we should add this new dataPar parameter to the Documentation.
Can you please draft something and I'll put it here?
https://exploreasl.github.io/Documentation/1.10.0beta/ProcessingParameters/

Modules/xASL_module_Population.m
HenkMutsaerts: Done

Modules/xASL_module_Population.m
HenkMutsaerts: Done

HenkMutsaerts:
Can you re-review?

For the dataPar template online:

x.S.[...]

DataTypes = Vector of cells specifying which images to take data from, in the ROI analysis. E.g., {'Tex’, 'ATT' 'SD' 'M0' ‘rc1T1’}

OPTIONAL, DEFAULT = {‘qCBF’}

jan-petr (approved):
OK

PR #1219, opened 2022-10-21
#1218 xASL_bids_MergeNifti Fix TE ordering

### LINKED ISSUE #1218: Ordering HAD4 & HAD8 volumes - import module
### Description

Before quantification, the volumes should be: all 8TEs for PLD1 then all 8TEs for PLD2, etc.
However, in DEBBIE LCBC scans, the acquisition was: all PLDs for TE1, all PLDs for TE2 etc.

If we don't reorder the volumes, the quantification doesn't work (creates a negative TExch map).

The PLDs vector written in rawdata's ASL.json is correct with the order of acquisition. What we need to ensure is that we reorganize them as 8TEs for PLD1 and then all 8TEs for PLD2, so [3.5x8 0.5x8 1.5x8 2.5x8] - for HAD4

Issue #1216 is labelled as duplicate but it still has code that we can use for this issue!

### Tasks for Bea to check

- [x] The problem was in DCM2NII in merging several NII+JSON - it hasn't recognized that it's a FME sequence, so it hasn't merged the NIIs. So it didn't write the TEs in the merged JSON.
- [x] Once we have the correct TE vector, then in DCM2NII, it correctly reorders it.
- [x] Then NII2BIDS step doesn't even need to reorder anything.
- [x] We only need to check if ASL module correctly identifies that the sequence is Time-encoded multi-TE. But I don't expect issues there.
- [x] And we want to add this to flavors as well (Jan will do).

### Release notes

Fix issue with Hadamard NII merging and ordering.

### BEA & HENK
* Quick & Dirty solution: Bea manually sorts them; editing the basil options text file, as this is the same for all scans per site.
* Future solution: register images to vascular template. Henk creates proximal to distal template based on MRA template & ATT template; and this proximal-distance signal gradient is then used to sort the control-label volumes. The sum(signal) of a whole control-label pair can help sorting the later phases for T1w-decay.

jan-petr:
### Linked issue

Closes #1218

### DIFF
--- a/Functions/BIDS/xASL_bids_MergeNifti.m
+++ b/Functions/BIDS/xASL_bids_MergeNifti.m
@@ -660,17 +660,22 @@ function pathOut = xASL_bids_MergeNifti_Merge(NiftiPaths, indexSortedFile, nameM
             if exist(fullfile(jsonPathX, [jsonNameX '.json']),'file')
                 tmpCheckJSON = spm_jsonread(fullfile(jsonPathX, [jsonNameX '.json']));
                 % Check EchoTimes
-                if isfield(tmpCheckJSON,'SeriesDescription')
-                    isHadamardFME = ~isempty(regexp(char(tmpCheckJSON.SeriesDescription),...
-                                    '(Encoded_Images_Had)\d\d(_)\d\d(_TIs_)\d\d(_TEs)', 'once')) ... % ASL format
-                                 || ~isempty(regexp(char(tmpCheckJSON.SeriesDescription),...
-                                    '(ss_TE)\d\d(_TI)\d\d\d\d', 'once'));                            % M0 format
-                    if isHadamardFME
-                        if isfield(tmpCheckJSON,'EchoTime')
-                            EchoTimes{iFileCheck,1} = tmpCheckJSON.EchoTime;
-                        end
-                    end
-                end
+				if isfield(tmpCheckJSON, 'SeriesDescription')
+					isHadamardFME = ~isempty(regexp(char(tmpCheckJSON.SeriesDescription),...
+						'(Encoded_Images_Had)\d\d(_)\d\d(_TIs_)\d\d(_TEs)', 'once')) ... % ASL format
+						|| ~isempty(regexp(char(tmpCheckJSON.SeriesDescription),...
+						'(ss_TE)\d\d(_TI)\d\d\d\d', 'once'));                            % M0 format
+				elseif isfield(tmpCheckJSON, 'SequenceName')
+					isHadamardFME = ~isempty(regexp(char(tmpCheckJSON.SequenceName), 'fme_asl', 'once'));
+				else
+					isHadamardFME = false;
+				end
+				
+				if isHadamardFME
+					if isfield(tmpCheckJSON,'EchoTime')
+						EchoTimes{iFileCheck,1} = tmpCheckJSON.EchoTime;
+					end
+				end
             end
         end
         % Add echo number array if it exists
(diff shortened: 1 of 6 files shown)


BeatrizPadrela (approved):
Works nicely!!

Functions/BIDS/xASL_bids_MergeNifti.m
```diff
@@ -659,18 +659,15 @@
...
-                                 || ~isempty(regexp(char(tmpCheckJSON.SeriesDescription),...
-                                    '(ss_TE)\d\d(_TI)\d\d\d\d', 'once'));                            % M0 format
-                    if isHadamardFME
-                        if isfield(tmpCheckJSON,'EchoTime')
-                            EchoTimes{iFileCheck,1} = tmpCheckJSON.EchoTime;
-                        end
-                    end
-                end
+                % Check if FME Hadamard
+				isHadamardFME = xASL_imp_CheckIfFME(tmpCheckJSON);
```
HenkMutsaerts, line 663: Should this be an `imp` function (ExploreASL-specific) or a `bids` function (BIDS-specific)?

HenkMutsaerts: Also, you replaced "check EchoTimes" by "FME Hadamard": that's not generalizable right? What if we have a non-Hadamard non-FME sequence with multiple echotimes?

jan-petr: `imp` - it's really needed for general Import.

This is supposed to be non-generalizable. Cause it is dealing with the peculiarities of FME-BBB-ASL. For further multi-TE sequences, we will have to first check if they work right away (as they should) or if they need a special tweak.

HenkMutsaerts: OK then please call this section (line 649) "FME Hadamard sequence-specific fixes/checks" rather than "If at the end and all went well"

Just for clarity, a section is either 1) general (checking different type of non-standard sequences) 2) Hadamard (checking different Hadamard sequences) or 3) FME-specific

Functions/Import/xASL_imp_DCM2NII_ReorderTimeEncoded.m
```diff
@@ -22,28 +25,62 @@ function xASL_imp_DCM2NII_ReorderTimeEncoded(nii_files, bTimeEncoded, resultJSON
 
     if numel(nii_files)>=1
         [resultPath, resultFile] = xASL_fileparts(nii_files{1});
-        % Check if we the current sequence is a Hadamard or not
-        if bTimeEncoded
+        % Check if the current sequence is a Hadamard from FME or not
+		if bTimeEncoded && bTimeEncodedFME
```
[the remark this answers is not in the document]
HenkMutsaerts: OK then we just wait until we will have `bTimeEncodedGE` `bTimeEncodedPhilips` and `bTimeEncodedDanny` as well ;) Just saying that you want to separate generalizable and specific functions/code sections, that's all.

This function has a general time encoded name, but works only for FME TimeEncoded, (otherwise it is skipped). So I would then rename this function to *ReorderTimeEncodedFME or something similar.

HenkMutsaerts (changes requested):
Final 2 minor remarks, just for keeping generalizability and specific cases separate.

Functions/Import/xASL_imp_DCM2NII_ReorderTimeEncoded.m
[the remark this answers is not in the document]
jan-petr: Done

Functions/BIDS/xASL_bids_MergeNifti.m
jan-petr: Done

HenkMutsaerts (approved):

PR #1234, opened 2022-11-07
Feature #1079 miscellaneous

### LINKED ISSUE #1079: Improve clarity
### Description

Here, I list some suggestions I got when going through the newest version of ExploreASL.

### Tasks

- [ ] `/ExploreASL/Templates` is not a useful name, as the word "templates" is also used for atlas templates, so let's reserve this word for that and rename this folder (or move the json-templates elsewhere)
- [ ] `/ExploreASL/Maps/` (==`x.D.MapsDir`) only contains the folder `/ExploreASL/Maps/Templates` (==`x.D.TemplateDir`), we can simplify this by removing one folder layer. 
- [ ] `xASL_module_ASL: *_CheckASL`: `skip` should be `bSkip`
- [ ] `xASL_module_ASL:` *_CheckASL`: `result` is unused and can be deleted
- [ ] `xASL_module_ASL`: `xASL_adm_DefineASLSequence` was performed twice
- [ ] When `bRegistrationCBF` is off, no need to create a `VascularTemplate` file
- [ ] What happens if `x.modules.asl.bRegistrationContrast==0` but no `mean_control` exists? Should issue warning & skip registration

### How to test

### Release notes

jan-petr:
### Linked issue

Closes #1079

### DIFF
--- a/Modules/xASL_module_ASL.m
+++ b/Modules/xASL_module_ASL.m
@@ -62,8 +62,8 @@ result = false;
 
 
 %% A. Check if ASL exists, otherwise skip this module
-[x,result,skip] = xASL_module_ASL_CheckASL(x,result);
-if skip, return; end
+[x, bSkip] = xASL_module_ASL_CheckASL(x);
+if bSkip, return; end
 x = xASL_init_FileSystem(x); % Do this only here, to save time when skipping this module
 oldFolder = cd(x.dir.SESSIONDIR); % Change working directory to make sure that unspecified output will go there...
 
@@ -580,9 +580,9 @@ end
 
 
 %% Check if ASL exists, otherwise skip this module
-function [x,result,skip] = xASL_module_ASL_CheckASL(x,result)
+function [x, bSkip] = xASL_module_ASL_CheckASL(x)
 
-    skip = false;
+    bSkip = false;
     x.P.Path_ASL4D = fullfile(x.dir.SESSIONDIR, 'ASL4D.nii');
     x.P.Path_ASL4D_json = fullfile(x.dir.SESSIONDIR, 'ASL4D.json');
     x.P.Path_ASL4D_parms_mat = fullfile(x.dir.SESSIONDIR, 'ASL4D_parms.mat');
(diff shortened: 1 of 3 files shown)


Modules/xASL_module_ASL.m
```diff
@@ -62,8 +62,8 @@
 
 
 %% A. Check if ASL exists, otherwise skip this module
-[x,result,skip] = xASL_module_ASL_CheckASL(x,result);
-if skip, return; end
+[x, bSkip] = xASL_module_ASL_CheckASL(x);
```
jan-petr, line 65: Not true that `result` is unused and can be deleted. `result` and also `x.result` are output parameters of this function - so they should both be assigned before doing `return`. Please revert and add `x.result`.

jan-petr (commented):
One small comment.

Modules/xASL_module_ASL.m
HenkMutsaerts: Fixed this now in [c24f23ad](https://github.com/ExploreASL/ExploreASL/pull/1234/commits/c24f23ad5a6af40ea9d8dc823da17aa900e50fc5). But it seems that `x.results` is not really used, it is initialized in `xASL_module_Import`, `xASL_module_ASL`, `xASL_init_InitializeMutex` on several locations but not interpreted anywhere, so shall I remove it?

HenkMutsaerts (changes requested):
Funny to notice after the fact that I'm reviewing my own work ;)

Modules/xASL_module_ASL.m
jan-petr: Also note that you need to add not only `result` here but also `x.result` that was previously missing - just that it's consistent before we decide to remove it (just in case we don't).

jan-petr:
> Funny to notice after the fact that I'm reviewing my own work ;)

I know. That's the drawback when somebody else creates an issue :D

HenkMutsaerts (changes requested):
Can you check my last few additions? I just added it here, I know I was the one wanting to keep issues separate in the past, so sorry about my inconsistency ;) This was a small change, should work fine.

jan-petr:
> Can you check my last few additions? I just added it here, I know I was the one wanting to keep issues separate in the past, so sorry about my inconsistency ;) This was a small change, should work fine.

I have no issue with that. So if you don't complain about it yourself, I'm fine :D

HenkMutsaerts (approved):

PR #1244, opened 2022-11-23
#1222 Update links in README

### LINKED ISSUE #1222: Change ReadMe to include exploreasl.lab@gmail.com in 1.10
### Release 1.10
We recently added the exploreasl.lab@gmail.com mailbox as a way to reach out to us, we should add it to the read me instead of/alongside with the email adresses of Jan and Henk-Jan. 
Henk-Jan suggested to also include a link to the discussion section of ExploreAsl as alternative place for questions. 
The Explore ASL website already links to the newly created mailbox.

jan-petr:
### Linked issue

Closes #1222

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Additionally, this (semi-)automatic pipeline allows us to minimize manual interv
 
 ## Documentation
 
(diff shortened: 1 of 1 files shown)


HenkMutsaerts (approved):
I would then also remove our e-mail addresses at the end. If they really want them they can always find them through google, but let's have them default to the discussions and the new e-mail address.

jan-petr:
> I would then also remove our e-mail addresses at the end. If they really want them they can always find them through google, but let's have them default to the discussions and the new e-mail address.

Agree. will do.

PR #1246, opened 2022-11-24
Bug #1124 b sup m0 correction multi pld

### LINKED ISSUE #1124: BSup fixing fails with non-integer timing
### Description

BSup timings not being integers in BSup fixing crashes the calculation on line xASL_quant_BSupCalculation (line 139). The same happens to multi-PLD readout. Now PLD 2000 is selected and then applied also for this correction.

### Release notes

Allow basic background-suppression correction for M0 in multi-PLD.

jan-petr:
### Linked issue

Closes #1124

### DIFF
--- a/Functions/Quantification/xASL_quant_M0.m
+++ b/Functions/Quantification/xASL_quant_M0.m
@@ -292,14 +292,14 @@ function [M0IM, x] = xASL_quant_RevertBsupFxControl(M0IM, x)
         end
 	end
 	   
-	% First obtain initial_PLD parameter
+	% First obtain initial_PLD parameter taking into account backwards compatibility of the naming
 	if isfield(x.Q,'Initial_PLD')
 		Initial_PLD = x.Q.Initial_PLD;
 	else
 		Initial_PLD = x.Initial_PLD;
 	end
 	
-	% And adjust initial_PLD for multi-PLD (a value closest to 2000 ms is currently selected)
+	% And adjust initial_PLD for multi-PLD - a value closest to 2000 ms is currently selected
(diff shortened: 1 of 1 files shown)


Functions/Quantification/xASL_quant_M0.m
```diff
@@ -292,18 +292,34 @@
...
 	   
+	% First obtain initial_PLD parameter taking into account backwards compatibility of the naming
+	if isfield(x.Q,'Initial_PLD')
+		Initial_PLD = x.Q.Initial_PLD;
+	else
+		Initial_PLD = x.Initial_PLD;
+	end
+	
+	% And adjust initial_PLD for multi-PLD - a value closest to 2000 ms is currently selected
+	if x.modules.asl.bMultiPLD
```
maartenhammer, line 303: Can't this also be replaced with an _if (length(Initial_PLD) >  1)_ to incorporate all cases where there is more than one PLD?  Even unexpected and/or undesired cases where more than 1 pld exists would then be handled as well.

jan-petr: Good comment. In theory, if we have more than 1 different PLD, then the bMultiPLD option should be set to 1. This is something done already at the start of ASL module, so that we don't have to check again at every time if we have 1 or more PLDs. So from that sense, we could trust this and don't need to check PLD length.

But it is of course possible that there are multiple PLDs and all the same for a single-PLD acquisition - that is permitted according to BIDS.
So I have changed that to avoid possible errors.

maartenhammer (approved):

PR #1272, opened 2022-12-24
Closes #684

### LINKED ISSUE #684: TemplateNumberName: Incorrect subject number population templates
### Description

In a dataset with subjects that not all include ASL4D.nii's, the amount of scans/subjects is correctly mentioned in the name of the templates present in the Population folder. However, when looking for these templates, xASL still searches for the original nSubjects instead of the correct nSubjects. The data showing the issue is located at: `lood_storage/divi/Projects/ExploreASL/ABBA/NORMENT/StrokeMRI.`

See images below: 

Searching for incorrect name:
image

Correct name:
image

The origin of the problem is the regular expression (in function `xASL_im_CreateAnalysisMask.m` line 144): 

` PathFoV = xASL_adm_GetFileList(x.D.TemplatesStudyDir, ['^FoV_n' xASL_num2str(x.dataset.nSubjectsSessions) '_bs-mean\.nii$'], 'FPList', [1 1]);` 

The name of the file is defined in `xASL_im_CreateAnalysisMask.m`, where in the subfunction `xASL_stat_RobustMean` empty, corrupt and outlier images are removed in the cell-structure. The length of the remaining cell structure is then used to determine the remaining files from which a template is made and the file is assigned that name in: ` NameIM = [TemplateNameList{iScanType} '_n' num2str(length(NotOutliers))];` 

Proposed solution:
A) in this issue, I fix that the name created in the templates is stored n the x-struct and reused throughout the pipeline. 
B) we create a general provenance stamp in the x-struct/json: the ExploreASL version and date-time stamp. When ExploreASL is rerun with a different version than the x-struct/json has, it will issue a very clear warning. 

Future issues:
We would need to check if the x-struct is always saved and reloaded for provenance, not to rely only on memory (for pipeline crashing). Note that the population module doesn't have to run in parallel.
B) 

### Requirements

- [x] replace `x.datasets.nSubjectsSessions` by `x.S.TemplateNumberName` that can be used throughout the population module
Note that the number of sessions is still not nicely processed, so we keep the number of subjects only throughout
- [x] replace `x.nSubjects` by `x.dataset.nSubjects` (to sync with where also other n parameter-fields live, like `x.dataset.nSessions` etc)
- [x] Make sure that already the first lines of xASL_module_Population have x.S.Atlases field. `xASL_module_Population` like 48 initializes Atlases field. But at this point, it should be already read from the dataPar.json.

### How to test

@MDijsselhof can you please test this? I couldn't test this with your folder, which only contains 1 subject (templates are only created with more than 6 subjects, as you need a minimal sample size to create a robust template)

### Release notes

Fix issue with subject number in population filenames.

HenkMutsaerts:
### Linked issue

Closes #684


jan-petr (approved):
OK

MDijsselhof (approved):

Issue #1052, opened 2022-01-17
xASL_adm_GzipAllFiles.m improve linux behaviour


MDijsselhof:
### Description

Requested for by @HenkMutsaerts:

Current `xASL_adm_GipAllFiles.m` behaviour is not optimised in Linux. 

JAN: Might have been fixed in #1164 
HENK: May be addressed in #1285

MichaelStritt:
We should also make sure that Maurice's code for the gzipping on windows works in compiled mode or we default to the slower version right away. Right now it crashes and defaults to the slower version.

HenkMutsaerts:
Is adressed in #1164 and hopefully in #1285

PR #1286, opened 2023-01-15
Closes #1285_supergzip_rust_implementation_update

### LINKED ISSUE #1285: SuperGzip Rust implementation update
### Description

Noticed that gzipping runs after every BIDS2Legacy run. Knew it could be faster since the original was written in Python. Indeed, it is superior by all metrics if re-written in Rust:

- 2.5x faster on a single thread
- Orders faster on multithread, depending on the system
- Slower than native find + grep + gzip on Unix on single thread b/c it validates that a filepath still exists. I'd argue this potentially  makes it safer than find + grep + gzip if multiple ExploreASL workers are simultaneously trying to access similar Nifti files that may or may not exist by the time one of the workers is done gzipping.
- Faster than find + grep + gzip on Unix when --num-threads = 3 or greater. 
- Executable size is ~9x smaller (~1.1MB vs Python implementation being ~10MB)

All major desktop releases available here: https://github.com/MauricePasternak/SuperGZip/releases/tag/v0.2.0

### Tasks

- [ ] Make adjustments to `xASL_adm_GzipAllFiles.m`
- [x] Test on macOS:
        HENK: with 8 cores on macbook Pro intel, zipping `/ExploreASL/External/TestDataSet` takes 2.6 s, unzipping 0.5 s.
        Using `find | grep` zipping takes 2.7 s.
        When I clone `/ExploreASL/External/TestDataSet/rawdata` 8 times, unzipping takes 2.7 s, zipping 7.4 s.
        Using `find | grep` zipping takes 2.9 s (but it skipped foldernames with spaces).
        Removing the spaces in foldernames:
        Using `find | grep` zipping takes 21.6 s (but it skipped foldernames with spaces).
        
        So for macOS: 3-fold increase in performance, is more flexible with foldernames, only downside is that we have yet 
        another mac application that needs approval when ran first time (but that's fine I would say)
- [x] Test on Windows server (28 cores; 32 files)
        zipping 16.2 s
        Windows Matlab method: 81.5 s
        so 5-fold increase (because of parallelization probably)

        Advantage: we don't need to use WSL for fast zipping, eases the code.

- [ ] Test on old Linux server (28 cores): gives an error, can you include these libraries in the compilation?

>> /scratch/hjmutsaerts/ExploreASL/External/SuperGZip/super-gzip_linux: /lib64/libc.so.6: version `GLIBC_2.25' not found (required by /scratch/hjmutsaerts/ExploreASL/External/SuperGZip/super-gzip_linux)
/scratch/hjmutsaerts/ExploreASL/External/SuperGZip/super-gzip_linux: /lib64/libc.so.6: version `GLIBC_2.28' not found (required by /scratch/hjmutsaerts/ExploreASL/External/SuperGZip/super-gzip_linux)
/scratch/hjmutsaerts/ExploreASL/External/SuperGZip/super-gzip_linux: /lib64/libc.so.6: version `GLIBC_2.33' not found (required by /scratch/hjmutsaerts/ExploreASL/External/SuperGZip/super-gzip_linux)
/scratch/hjmutsaerts/ExploreASL/External/SuperGZip/super-gzip_linux: /lib64/libc.so.6: version `GLIBC_2.18' not found (required by /scratch/hjmutsaerts/ExploreASL/External/SuperGZip/super-gzip_linux)
/scratch/hjmutsaerts/ExploreASL/External/SuperGZip/super-gzip_linux: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by /scratch/hjmutsaerts/ExploreASL/External/SuperGZip/super-gzip_linux)
/scratch/hjmutsaerts/ExploreASL/External/SuperGZip/super-gzip_linux: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by /scratch/hjmutsaerts/ExploreASL/External/SuperGZip/super-gzip_linux)

Only disadvantage of SuperGzip (but `find | grep` has the same) is that we don't have a progress tracker, @MauricePasternak would this be possible to print on the screen? Or only when we select the non-verbose mode?

Minor spelling mistakes:
Deompressing -> Decompressing

Also: our imaging IT specialist Paul Groot asks if SuperGzip is faster than pigz?

### How to test

```matlab
glob_pat = join(['"' ROOT '/**/*.nii"']);
command = [PathToSuperGzip ' gzip ' glob_pat ' -n ' num2str(numCores) ' --verbose' ];
[exit_code, system_result] = system(command);
```

### Release notes

Required: Updated SuperGzip to use the newer Rust implementation.

MauricePasternak:
### Linked issue

Closes #1285 

### Comments

Tested on 1000+ files in a Population folder and then re-running the study. Noticeable improvement over the old Python version.

As in the original issue, the new version of the program can beat out native gzip in Unix under the condition that at least 3 threads are used. No logic was added for Unix, as this was meant to be a simple language & CLI change, but it wouldn't be difficult to make a check for whether is it reasonable to do so. I leave it to the discretion of reviewers on preferred logic.


HenkMutsaerts (commented):
What I don't understand: the original SuperGZip was implemented by Michael for Windows specifically. But now you state that you created something that performs faster (un)zipping in Unix? That would be great of course, but we need to keep `xASL_adm_GzipAllFiles` to work for all platforms.

Minor name comments:
* a) CamelCase would give `SuperGZip.exe`
* b) Why do you call it `unzip`, the `gzip` that you replace does both zip and unzip right?

Should we improve `xASL_adm_GzipNifti` and `xASL_adm_UnzipNifti` accordingly?

MauricePasternak:
Couple of points:

- the original SuperGzip was written by me in Python. It was only packaged for Windows. Michael presumably wrote the logic behind invoking the packaged executable from matlab, since my knowledge of version control and general programming outside of Python still wasn't that great. Despite being Python, it still offered a boost over matlab's gzip by virtue of multiprocessing.

- this version written by me in Rust offers the benefits of using a [compiled language over an interpreted one.](https://i.redd.it/40t7alhvlac91.png) It's a CLI tool I decided to write as a quick mini project after seeing that gzipping a couple of thousand Nifti files still takes a bit of time. 

- Yes it performs both gzip and unzip. That is part of the reason for the lowercase name difference. The other reason being that CLI-based tools that can be used on Unix should not be uppercase by convention, if I recall correctly. 

- I use Github Workflows to package this newer Rust version [for all major desktop platforms.](https://github.com/MauricePasternak/SuperGZip/releases). That shouldn't be an issue.

- It performs best when asked to batch process many files in one go. Even if Rust > Matlab  in terms of speed, the initial overhead of a `system` call in matlab may not make the venture worthwhile just to gunzip one or two files. So with regards to adding calls to this executable in other ExploreASL functions, I'd first ask whether it'd only be used to gunzip one or two files at a time. If so, not worth it. If used to perform cleanup/startup operations on a whole subject/module/study, then the performance benefits start to shine.

HenkMutsaerts:
@MauricePasternak Thanks for clarifying this! If we make this change, it should work with both multi-core processors and single-threaded processors (if ExploreASL is ran in a server, it could be ran single-threaded but with multiple ExploreASL instances).

Can we also address #1052 in here (implementing this also for Linux and macOS)?

Modules/SubModule_Population/xASL_adm_GzipAllFiles.m
```diff
@@ -58,14 +58,14 @@ function xASL_adm_GzipAllFiles(ROOT, bFolder, bUseLinux, pathExternal)
             if ~isempty(PathList)
                 fprintf('\n%s\n',['G-zZzZipping ' num2str(length(PathList)) ' files']);
                 numCores = feature('numcores');
+                glob_pat = join(['"' ROOT '/**/*.nii"']);
```
HenkMutsaerts, line 67: We use camelCase with readable variable names, which enhances readability for others. 
Plus, note that you use a Unix slash / that won't work with a Windows system \\

So this should probably be:
if ispc
    globalPattern = ['"' ROOT '\**\*.nii"']);
else
    globalPattern = ['"' ROOT '/**/*.nii"']);
end

Modules/SubModule_Population/xASL_adm_GzipAllFiles.m
```diff
@@ -58,14 +58,14 @@ function xASL_adm_GzipAllFiles(ROOT, bFolder, bUseLinux, pathExternal)
...
                 % Get SuperGzip path
-                PathToSuperGzip = fullfile(pathExternal, 'SuperGZip', 'SuperGZip_Windows.exe');
+                PathToSuperGzip = fullfile(pathExternal, 'SuperGZip', 'super-gunzip.exe');
```
HenkMutsaerts: I would still not call it `super-gunzip.exe` if we also (or even more frequently) use it for zipping; fine to rename to `super-zip.exe`?

HenkMutsaerts (changes requested):
Nice work! See also my other comment above

Modules/SubModule_Population/xASL_adm_GzipAllFiles.m
MauricePasternak: It actually shouldn't care about forward slashs or backslashes. The glob library in Rust that this program calls is very robust in that regard. Example of mixing slashes in Windows:

image

Modules/SubModule_Population/xASL_adm_GzipAllFiles.m
MauricePasternak: Sounds good to me. I'll keep calling it super-gunzip in the original repo & its release page. But when copying over to ExploreASL, will change name to just super-zip.

HenkMutsaerts: How does it perform compared to the default Linux parallel zipping algorithm, called `p7zip`?

HenkMutsaerts: I would still say that that is weird. No zip package calls itself unzip as primary name right? 7zip not 7unzip. Winrar not WinUnrar. Etc. But perhaps you have another reason to call it this way.

PR #1289, opened 2023-01-17
Closes #1288 hotfix

### LINKED ISSUE #1288: Incorrectly import and BIDS2Legacy of multiple anatomical images.
### Description

Dataset that imports anatomical images as duplicates first creates two anatomical runs in rawdata, and then when doing BIDS2Legacy it messes up the order of the multiple-runs. We fix both errors, while fixing one of them would have been sufficient for this dataset.

`xASL_bids_BIDS2Legacy_ParseModality` line 43 sorts the table `reference`.
Indices (e.g., `TypeRunIndex`) are then created according to the sorted `reference` table.
But `ModalityFields` (and perhaps others?) does not get sorted accordingly, so `xASL_bids_BIDS2Legacy_CompilePathsForCopying` line 70 had unmatched `TypeRunIndex` and `ModalityFields` in the case of multiple ASL or anatomical runs.

### Tasks

- [x] Fix the sorting from `xASL_bids_BIDS2Legacy_ParseModality`
- [x] Add a final filename check before copying in `xASL_bids_BIDS2Legacy_ParseScanType`
- [x] Some minor warning improvements in `xASL_bids_parseM0` & `xASL_imp_UpdateDatasetRoot`
- [x] Default output when `SessionID` is empty in `xASL_wrp_BIDS2Legacy` (tricky to default to 1 as this may be incorrect, so just added "unknown" as default)
- [x] If multiple anatomical files are converted to NII, but they are not assigned a specific run-label, then covert only the non-labeled one from the first run.
- [x] Inform Mervin about merging.

### Release notes

Correct import and BIDS2Legacy with duplicate anatomical scans.

HenkMutsaerts:
### Linked issue

Closes #1288 hotfix, see issue


jan-petr (approved):
All working and Flavors test running.

HenkMutsaerts (commented):
This is an import-change, that strictly speaking has nothing to do with this issue (which is about `BIDS2Legacy`)

HenkMutsaerts (commented):
1) Are you sure this sorting is required? I don't think the sorting is needed.
2) Are you sure that `ModalityFields` was the only other field that was not sorted yet? (there may be more dependencies)

jan-petr:
You are right. This fixes the same problem Mervin had, but in import and not in BIDS2Legacy. I will rename issue and release note to reflect this.

jan-petr:
> 1. Are you sure this sorting is required? I don't think the sorting is needed.
> 2. Are you sure that `ModalityFields` was the only other field that was not sorted yet? (there may be more dependencies)

I have checked that the sorting is now complete. I sort of assumed that when the sorting was there, it is needed. But maybe this was some historical reason for this and it is not needed anymore. Shall we create a new issue for checking this, or we keep it as benign?

jan-petr:
By mistake, we have created a PR with 'main' as the base branch. And after merging to 'develop', it signals here that nothing was merged, and unfortunately, it is not possible to recreate a new PR or change the base to develop. So everything is fine, the commits are correctly merged, just this PR signals a problem.

PR #1296, opened 2023-01-20
#1295 Write WMH count to the subject dir

### LINKED ISSUE #1295: WMH count issue with parallelization
### Description

In some cases, structural volume counts WMH lesions and saves it into a CSV file, but immediately after that fails to find this file. This is most probably an issue in parallelization when files are saved to a common folder and likely cleaned by another worked. We need to move this temporary file saving to the subject folder instead of the common population folder.

### Release notes

Save temporary CSV with WMH counts to subject folder and not population folder.

jan-petr:
### Linked issue

Closes #1295

### DIFF
--- a/External/SPMmodified/xASL/xASL_bids_csv2tsvReadWrite.m
+++ b/External/SPMmodified/xASL/xASL_bids_csv2tsvReadWrite.m
@@ -1,7 +1,7 @@
-function [PathTSV, CellContents] = xASL_bids_csv2tsvReadWrite(PathIn, bDeleteCSV, bWriteTSV)
+function [PathTSV, CellContents] = xASL_bids_csv2tsvReadWrite(PathIn, bDeleteCSV, bOverwrite)
 %xASL_bids_csv2tsvReadWrite Multi-purpose read & convert csv/tsv
 %
-% FORMAT: [PathTSV, CellContents] = xASL_bids_csv2tsvReadWrite(PathIn[, bDeleteCSV, bWriteTSV])
+% FORMAT: [PathTSV, CellContents] = xASL_bids_csv2tsvReadWrite(PathIn[, bDeleteCSV, bOverwrite])
 %
 % INPUT:
 %   PathIn          - path to CSV or TSV input file (REQUIRED)
@@ -24,14 +24,14 @@ function [PathTSV, CellContents] = xASL_bids_csv2tsvReadWrite(PathIn, bDeleteCSV
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
 % EXAMPLE: xASL_adm_tsvWrite(ParticipantsMetadata, '/MyStudy/participants.tsv');
 % __________________________________
-% Copyright 2015-2020 ExploreASL
+% Copyright 2015-2023 ExploreASL
 
 
 
 %% -------------------------------------------------------
 %% Admin
-if nargin<3 || isempty(bWiteTSV)
-    bWriteTSV = true;
+if nargin<3 || isempty(bOverwrite)
+    bOverwrite = true;
 end
 
 if nargin<2 || isempty(bDeleteCSV)
@@ -45,7 +45,7 @@ end
 
 [Fpath, Ffile, Fext] = xASL_fileparts(PathIn);
 
-if isempty(regexp(Fext, '^\.(tsv|csv)$'))
+if isempty(regexp(Fext, '^\.(tsv|csv)$', 'once'))
     warning('Wrong extension, resetting to .csv');
     PathIn = fullfile(Fpath, [Ffile '.csv']);
 end
@@ -75,8 +75,10 @@ end
 
 %% -------------------------------------------------------
 %% 2) Write the TSV file (if requested)
-if ~exist(PathTSV, 'file') && bWriteTSV
+if ~exist(PathTSV, 'file') || bOverwrite
     xASL_tsvWrite(CellContents, PathTSV, 1); % overwrite
+else
+	warning(['File ' PathTSV ' already exists and bOverwrite is set to false'])
 end
 
 
--- a/External/SPMmodified/xASL/xASL_bids_csv2tsvReadWrite.m
+++ b/External/SPMmodified/xASL/xASL_bids_csv2tsvReadWrite.m
@@ -64,6 +64,15 @@ if strcmp(Fext, '.csv') && ~exist(PathIn, 'file')
     end
 end
 
+if strcmp(Fext, '.tsv') && ~exist(PathIn, 'file')
+    % if we want to read a TSV file, but it doesnt exist, try reading a CSV file instead
+    Fext = '.csv';
+	PathIn = fullfile(Fpath,[Ffile '.csv']);
+    if ~exist(PathIn, 'file')
+        warning([PathIn ' did not exist, skipping']);
+        return;
+    end
+end
 
 %% -------------------------------------------------------
 %% 1) Read the CSV or TSV file
--- a/External/SPMmodified/xASL/xASL_bids_csv2tsvReadWrite.m
+++ b/External/SPMmodified/xASL/xASL_bids_csv2tsvReadWrite.m
@@ -1,12 +1,13 @@
-function [PathTSV, CellContents] = xASL_bids_csv2tsvReadWrite(PathIn, bDeleteCSV, bOverwrite)
+function [PathTSV, CellContents] = xASL_bids_csv2tsvReadWrite(PathIn, bDeleteCSV, bWriteTSV)
 %xASL_bids_csv2tsvReadWrite Multi-purpose read & convert csv/tsv
 %
-% FORMAT: [PathTSV, CellContents] = xASL_bids_csv2tsvReadWrite(PathIn[, bDeleteCSV, bOverwrite])
+% FORMAT: [PathTSV, CellContents] = xASL_bids_csv2tsvReadWrite(PathIn[, bDeleteCSV, bWriteTSV])
 %
 % INPUT:
 %   PathIn          - path to CSV or TSV input file (REQUIRED)
 %   bDeleteCSV      - boolean specifying if an original CSV should be deleted (OPTIONAL, DEFAULT=true)
-%   bOverwrite      - boolean specifying if the CSV contents should be written to a new TSV file (OPTIONAL, DEFAULT=true)
+%   bWriteTSV       - boolean specifying if the CSV contents should be
+%                     written to a new TSV file (OPTIONAL, DEFAULT=true) (note: also overwrites any pre-existing TSV-file)
 %
 % OUTPUT:
 %   PathTSV         - path to TSV file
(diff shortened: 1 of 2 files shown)


External/SPMmodified/xASL/xASL_bids_csv2tsvReadWrite.m
```diff
@@ -1,7 +1,7 @@
-function [PathTSV, CellContents] = xASL_bids_csv2tsvReadWrite(PathIn, bDeleteCSV, bWriteTSV)
+function [PathTSV, CellContents] = xASL_bids_csv2tsvReadWrite(PathIn, bDeleteCSV, bOverwrite)
```
HenkMutsaerts: All changes of `bWriteTSV` to `bOverwrite` are illegal here. You can add `bOverwrite`, but you cannot remove `bWriteTSV`.

If `bWriteTSV` is false, and you read a CSV, it reads the content and puts it in memory. If `bWriteTSV` is true, it also reads to memory but additionally converts to TSV.

External/SPMmodified/xASL/xASL_bids_csv2tsvReadWrite.m
```diff
@@ -45,7 +45,7 @@
 
 [Fpath, Ffile, Fext] = xASL_fileparts(PathIn);
 
-if isempty(regexp(Fext, '^\.(tsv|csv)$'))
+if isempty(regexp(Fext, '^\.(tsv|csv)$', 'once'))
```
HenkMutsaerts, line 49: You keep adding 'once' because Matlab complaints, I would argue this just makes stuff less readable, we won't notice the speed increase ;)

Modules/SubModule_Structural/xASL_wrp_GetVolumetrics.m
```diff
@@ -94,21 +94,29 @@ function xASL_wrp_GetVolumetrics(x)
     xASL_adm_DeleteFileList(x.D.TissueVolumeDir, '^LST_tlv_.*\.(csv|tsv)$', [], [0 Inf]);
     xASL_adm_DeleteFileList(x.D.TissueVolumeDir, ['WMH_LST_(LGA|LPA)_' x.P.SubjectID '.(csv|tsv)'], [], [0 Inf]);
     
-    % run the LST lesion volume calculator
+    % Run the LST lesion volume calculator
+	% Change directory to the SUBJECT DIRECTORY as the count function outputs to the current directory
```
HenkMutsaerts: "count function" is not clear, it took me some time to understand that you meant `ps_LST_tlv`, (am I right?) Please be explicit ;)

Modules/SubModule_Structural/xASL_wrp_GetVolumetrics.m
```diff
@@ -94,21 +94,29 @@ function xASL_wrp_GetVolumetrics(x)
...
+    cd(x.dir.SUBJECTDIR);
     [~, FileName] = ps_LST_tlv(x.P.Path_WMH_SEGM, ~bVerbose, BinThresh, MinimalLesionVolume);
     cd(CurrDir);
     % get current filename:
-    FileName = fullfile(x.D.TissueVolumeDir, FileName);
+    FileName = fullfile(x.dir.SUBJECTDIR, FileName);
     % define new filename:
-    OutputPath = fullfile(x.D.TissueVolumeDir, ['WMH_LST_' x.WMHsegmAlg '_' x.P.SubjectID '.csv']);
-    % Then move our file
+    OutputPathCSV = fullfile(x.dir.SUBJECTDIR, ['WMH_LST_' x.WMHsegmAlg '_' x.P.SubjectID '.csv']);
```
HenkMutsaerts: From here below you need to revert, it is too long and wrong; this doesn't concern temporary files, and the output should remain in the population folder.

We can eventually move all structural output to subject folders, but that would be a large future issue.

HenkMutsaerts (changes requested):
How much time did you spend on this, would it be worth checking with me first?

External/SPMmodified/xASL/xASL_bids_csv2tsvReadWrite.m
jan-petr: The speed is not important - but it removes Matlab warnings - which is nice to have. Not to optimize code, but to have code without warnings. But we can for sure skip this the next time...

External/SPMmodified/xASL/xASL_bids_csv2tsvReadWrite.m
jan-petr: Not entirely true - the code previously had bOverwrite in the code description, so I was only cleaning code. Not adding anything. So up to you to decide what names you want. I was just syncing it...

Modules/SubModule_Structural/xASL_wrp_GetVolumetrics.m
jan-petr: sure - you can update the comment...

Modules/SubModule_Structural/xASL_wrp_GetVolumetrics.m
jan-petr: Actually. It has to stay like this. The temporary files in population folder were completely messed up during parallelization. As this is structural module outputting to population folder. If you still want to output temp files there, then you need to change the entire structure and debug. I don't see issue outputting temp files to subject folder.

jan-petr:
> How much time did you spend on this, would it be worth checking with me first?

Not much. But read again - this has to be fixed. The previous implementation was causing bugs in parallel mode...

External/SPMmodified/xASL/xASL_bids_csv2tsvReadWrite.m
HenkMutsaerts: The functionality is very different. `bOverwrite` means just overwriting yes or no. `bWriteTSV` means if you should convert a CSV-file to a TSV file (or only output the read CSV-file to memory). So this needs reverting.

Modules/SubModule_Structural/xASL_wrp_GetVolumetrics.m
HenkMutsaerts: Done

Modules/SubModule_Structural/xASL_wrp_GetVolumetrics.m
HenkMutsaerts: No, I agreed with the first part where you change the temporary file to the subject folder (assuming LST outputs the same name for each subject, which gets messed up during parallelization).

It was just the last part where you added unnecessary code.

HenkMutsaerts (approved):
See my comments, this should fix my issues

External/SPMmodified/xASL/xASL_bids_csv2tsvReadWrite.m
jan-petr: Agree. I was confused and tried to clean it. But I agree with your change.

Modules/SubModule_Structural/xASL_wrp_GetVolumetrics.m
jan-petr: OK. Now I understand! But there's still an issue.
1. The former version saved the temporary to POPULATION dir and then renamed it - this was causing issues in parallelization.
2. So I made a new version that saved the temporary to SUBJECTDIR and then moved the final to POPULATION. But I agree that it was way to complicated.
3. But you edit saves everything to SUBJECTDIR and never moves it to the population.
4. So I made a last fix that saves temporary to SUBJECTDIR and then moves it to POPULATION and there converts to TSV. Please check if this is clean enough.

External/SPMmodified/xASL/xASL_bids_csv2tsvReadWrite.m
```diff
@@ -64,6 +64,15 @@
     end
 end
 
+if strcmp(Fext, '.tsv') && ~exist(PathIn, 'file')
```
HenkMutsaerts, line 67: This was not what this function was intended to. This function was to read a CSV-file and at the same time convert to a TSV-file, which happens a lot when taking output of LST, CAT12, etc, and converting to BIDS (BIDS likes TSV).

Otherwise, we would simply use `xASL_tsvRead` or `xASL_csvRead`.

Was this not clear from the header?

I'm not sure your change here changes the concept, because if we already have a TSV-file, LST outputs a CSV-file with the same name, it should always overwrite. Now, it would take the old TSV-context and save it as if this was the new TSV.

So we should probably revert your edit here, but the header should be more clear. And rename this function, so it is more clear, e.g. `xASL_bids_csv2tsv.m` instead of ReadWrite

Modules/SubModule_Structural/xASL_wrp_GetVolumetrics.m
```diff
@@ -95,22 +95,26 @@ function xASL_wrp_GetVolumetrics(x)
...
     FileName = fullfile(x.dir.SUBJECTDIR, FileName);
-    % define new filename:
-    OutputPath = fullfile(x.dir.SUBJECTDIR, ['WMH_LST_' x.WMHsegmAlg '_' x.P.SubjectID '.csv']);
+    % Define new filenames:
+    OutputPath = fullfile(x.D.TissueVolumeDir, ['WMH_LST_' x.WMHsegmAlg '_' x.P.SubjectID '.csv']);
     % Then move our file
     if ~exist(FileName, 'file')
          warning(['Missing:' FileName]);
 	else
+		% First rename the output file to the correct filename format in the population dir
```
HenkMutsaerts, line 112: If you remove the word "format" here, I'm fine :)
Because the format is changed below.

And I like that you make the comments this explicit! ;)

HenkMutsaerts (changes requested):
Minor question but mostly fine

External/SPMmodified/xASL/xASL_bids_csv2tsvReadWrite.m
jan-petr: I don't want to do "a Henk" - but please read my comments before you reply :)
I know what was this issue supposed to do - previously, it saved the CSV in a way that it was conflicting and crashing during parallelization, so we had to save it elsewhere, convert CSV to TSV and then move TSV to the population directory. See all my comments above. I'm fine with the csv2tsv function but it has to be run in the subject dir...

Modules/SubModule_Structural/xASL_wrp_GetVolumetrics.m
jan-petr: that was to convince you :)
Sure "format" was a typo ;)

External/SPMmodified/xASL/xASL_bids_csv2tsvReadWrite.m
jan-petr: We can remove this commit, but I just wanted to catch all possible misuses of this function. So this is a safety net, not a function ;)But we can remove it if you want...

Modules/SubModule_Structural/xASL_wrp_GetVolumetrics.m
jan-petr: I'll do this once I'm done with running other tests. So please only check the function above.

HenkMutsaerts (approved):

PR #1304, opened 2023-01-31
Closes #1300 NIfTI import

### LINKED ISSUE #1300: NIfTIs import
### Description

Return the original ability to import NIfTIs, which was removed when we added NIfTI2BIDS to the import module. ExploreASL’s import should work with pre-existing NIfTIs. If these don’t have a JSON sidecar, this is not BIDS compliant and thus not desired, but we should support it. Anatomical scans that are NIfTI-only should be supported.

### Tasks

- [x] `NIfTIs` in import should be dealt with, by using `sourceStructure.json`>`bMatchDirectories:false` and `\.nii` as last token.
- [x] `xASL_imp_DCM2NII_ConvertScan` line 82 and further, it detects the scantype.
T1w was not recognised here, which is why it came into a session folder (T1w_1) like ASL,
which NII2BIDS expects in `xASL_wrp_NII2BIDS_Subject` > line 42
 (so I didn’t fix this) Otherwise, this is hacked such that it can survive NIfTIs without JSONs.
Unsure how this relates to #1288.
- [x] `xASL_wrp_NII2BIDS_Subject` > listRuns should also detect T1 (not only T1w)
- [x] `xASL_imp_ReadSourceData` >  Shoulld recognise .nii file extension 
- [x] `xASL_imp_DCM2NII_Subject_StartConversion` > destination should be zipped
- [x] `xASL_imp_DCM2NII_Subject_StartConversion` line 31 has a bug. If `bSkipThisOne` is true (which it probably isn't anymore after Michael's import revamp), it will try to access undefined variables `iSubject` `iVisit` etc. Easiest solution is to just add these variables to the input variables, but perhaps we can remove `bSkipThisOne` completely, if this simplifies the code. -> fixed by #1294
- [x] Only create a rawdata subfolder (e.g., `perf`) if we move NIfTIs there
-> `xASL_adm_CreateDir` should be removed at `xASL_imp_NII2BIDS_Run` > line 44 & line 56, because this will create the folder whether this folder will be populated or not
Rather, this should happen in the ASLBIDSify functions (where the JSON & NIfTIs are moved. I did some changes for this, but @jan-petr you should check where the ASL folder needs to be created and change this. Alternatively, we move this to a future issue if this is too complicated.
- [x] Allow importing both `.nii` and `.nii.gz`; when either `.nii` or `.nii.gz` is provided, it should import both `.nii` and `.nii.gz`
    -> when detecting `nii` at the end, it should become (nii|nii\.gz)
- [x] @jan-petr  Add new NIfTI flavor to flavordatabase (see branch #1300 in `/FlavorDatabase/DDI_Siemens_Skyra_E11_NIfTI_Only`)
- [x] In the flavor - convert manually TXT sidecars to JSONs
- [x] Test the flavor to see if JSONs are also copied
- [x] Unverified commits have to be fixed before merging

This also closes duplicate #1128.

### Minor
- [X] `xASL_bids_CompleteBIDS2Legacy` gave confusing messages; stemming from Michael Stritt taking running import before ExploreASL processing as default behavior (whereas default behavior should be running the pipeline separate, with BIDS as input)

>> There is a dataPar.ison in derivatives already.
Overwriting x.dir.dataPar...

* heads up when `dataPar.json` already existed, which should be default behavior, so I removed this
* it said it overwrote `x.dir.dataPar`, which can confuse the user as if a file is overwritten, so I removed this

### How to test

Import the new dataset `/FlavorDatabase/DDI_Siemens_Skyra_E11_NIfTI_Only` which has only NIfTIs. There will be warnings because of the JSON lacking, which I believe we should keep (it is not BIDS compliant so not really supported behavior).

Check to sync this with #1128.

### Release notes

Allow direct import of NIfTIs.

HenkMutsaerts:
### Linked issue

Closes #1300

### How to test

See #1300 and please also check and merge flavor #1300 in the FlavorDatabase.

### Comments


Functions/Import/xASL_imp_DCM2NII_ConvertScan.m
```diff
@@ -77,16 +79,21 @@
...
+    bOneScanIsEnough = false; % PM: THIS USED TO BE TRUE FOR ANATOMICAL SCANS
+    bPutInSessionFolder = true; % THIS USED TO BE FALSE FOR ANATOMICAL SCANS
     switch scan_name
         case {'ASL4D', 'M0', 'ASL4D_RevPE', 'func_bold'}
             bPutInSessionFolder = true;
-        case {'T1', 'WMH_SEGM', 'FLAIR', 'T2', 'T1c'}
-            bPutInSessionFolder = false;
+            bAnatomical = false;        
+        case {'T1w' 'T2w' 'T1', 'WMH_SEGM', 'FLAIR', 'T2', 'T1c'}
+%             bPutInSessionFolder = false;
```
jan-petr, line 90: I don't get this change. bPutInSessionFolder is initialized as true, so if you comment it here, it will put all anat to session folders....

Functions/Import/xASL_imp_DCM2NII_ConvertScan.m
```diff
@@ -77,16 +79,21 @@
...
+%     % Defaults
+    bOneScanIsEnough = false; % PM: THIS USED TO BE TRUE FOR ANATOMICAL SCANS
+    bPutInSessionFolder = true; % THIS USED TO BE FALSE FOR ANATOMICAL SCANS
     switch scan_name
         case {'ASL4D', 'M0', 'ASL4D_RevPE', 'func_bold'}
             bPutInSessionFolder = true;
-        case {'T1', 'WMH_SEGM', 'FLAIR', 'T2', 'T1c'}
-            bPutInSessionFolder = false;
+            bAnatomical = false;        
+        case {'T1w' 'T2w' 'T1', 'WMH_SEGM', 'FLAIR', 'T2', 'T1c'}
```
jan-petr, line 89: isn't scan_name a keyword? So if you change it here, you might also need to make sure that this is correctly recognized in NII2BIDS. Is there a support for both T1w and T1?

Functions/Import/xASL_imp_DCM2NII_ConvertScan.m
```diff
@@ -77,16 +79,21 @@
...
-    bPutInSessionFolder = true; 
+%     % Defaults
+    bOneScanIsEnough = false; % PM: THIS USED TO BE TRUE FOR ANATOMICAL SCANS
+    bPutInSessionFolder = true; % THIS USED TO BE FALSE FOR ANATOMICAL SCANS
     switch scan_name
         case {'ASL4D', 'M0', 'ASL4D_RevPE', 'func_bold'}
             bPutInSessionFolder = true;
-        case {'T1', 'WMH_SEGM', 'FLAIR', 'T2', 'T1c'}
-            bPutInSessionFolder = false;
+            bAnatomical = false;
```
jan-petr, line 88: it would be good to initialize this also upfront. not only inside branches

Functions/Import/xASL_imp_DCM2NII_ConvertScan.m
```diff
@@ -162,35 +168,42 @@
...
-    [x,nii_files, summary_line, thisSubject.globalCounts] = ...
-        xASL_imp_DCM2NII_Subject_SortASLVolumes(x, thisSubject.globalCounts, scanpath, scan_name, nii_files, scanFields.iSubject, scanFields.iVisit, scanFields.iSession, scanFields.iScan);
-
-    % Correct nifti rescale slope if parms.RescaleSlopeOriginal =~1 but nii.dat.scl_slope==1 (this can happen in case
-    % of hidden scale slopes in private Philips header, that is dealt with by xASL_bids_Dicom2JSON but not by dcm2niiX.
-    if ~isempty(nii_files) && exist('parms','var')
-        [TempLine, PrintDICOMFields] = xASL_imp_AppendParmsParameters(parms);
-        summary_line = [summary_line TempLine];
-    else
-        PrintDICOMFields = [];
```
jan-petr, line 189: It's wrong to remove PrintDICOMFields from here. And also since this is now inside another IF, you'll need to initialize PrintDICOMFields in the other branch as well.

Because it's an output variable and you will have it missing in certain cases -> error.

Modules/SubModule_Import/xASL_wrp_NII2BIDS_Subject.m
```diff
@@ -39,7 +39,7 @@
     % Subsequent code is based on having data per ASL scan, so we "fool" it
     % by renaming all runs into ASL_1 ASL_2 ASL_n and keeping the
     % unique runs only. Missing scans will issue a warning, not an error.
-    listRuns = xASL_adm_GetFileList(fullfile(x.modules.import.imPar.TempRoot,nameSubjectSession),'^(ASL|T1w|FLAIR).+$',false,[],true);
+    listRuns = xASL_adm_GetFileList(fullfile(x.modules.import.imPar.TempRoot,nameSubjectSession),'^(ASL|T1|T2|FLAIR).+$',false,[],true);
```
jan-petr, line 42: are you sure that this is a good idea?

Modules/SubModule_Population/xASL_wrp_CreateBiasfield.m
```diff
@@ -23,10 +23,27 @@ function xASL_wrp_CreateBiasfield(x)
 %              PM: add normalization of between-subjects SD as well.
```
jan-petr, line 23: I'm really confused. Why do we see all the edits here that we've done in a different issue? 
It looks like you have rebase this branch. In general, it is a really really bad idea to rebase a branch anytime before the PR is approved. If necessary to rebase, this has to be done in another way. Cause it makes then really complicated to asses the PR. It's OKish in this case, but otherwise we shouldn't do that?

Or is there another explanation why do I see edits here from another issue?

Functions/Import/xASL_imp_DCM2NII_Subject_StartConversion.m
```diff
@@ -38,18 +38,17 @@
...
-                xASL_Copy(scanpath, destfile, x.modules.import.imPar.bOverwrite, x.modules.import.imPar.bVerbose);
-                % gunzip if required
-                destfile = xASL_adm_UnzipNifti(destfile);
-                xASL_Move(destfile, first_match, x.modules.import.imPar.bOverwrite, x.modules.import.imPar.bVerbose);
+            first_match = fullfile(destdir, [scan_name '.nii.gz']);
+            % only copy this file if it doesn't exist yet, or if we want to
+            % overwrite
+            if x.modules.import.imPar.bOverwrite || ~xASL_exist(first_match,'file') % this will check both .nii & .nii.gz
+                xASL_Copy(scanpath, first_match, x.modules.import.imPar.bOverwrite, x.modules.import.imPar.bVerbose);
+                % gzip if required
```
jan-petr, line 51: Original version was unziping, now we are zipping - is that really what we want?

jan-petr (changes requested):
Changes needed...

Functions/Import/xASL_imp_DCM2NII_ConvertScan.m
HenkMutsaerts: You need this, otherwise in some cases, import won't work correctly. NIfTI2BIDS (or just before that) the code assumes that even T1's are in their own folders

Functions/Import/xASL_imp_DCM2NII_ConvertScan.m
HenkMutsaerts: Yes we probably need to do this. The problem is that it is confusing if we should use T1 or T1w in import, whereas T1w is the BIDS term (and we use both throughout). `scan_name` is not a BIDS keyword, it is something we use in our archaic token structure ;)

Functions/Import/xASL_imp_DCM2NII_ConvertScan.m
HenkMutsaerts: I actually think nothing should be initialized upfront; those defaults give problems.
Better to provide a default at the end, when an unknown `scan_name` is detected, with a warning. Initializing dependent variables upfront is asking for bugs, unless you assume that we always carefully think of every possible scenario when coding, which is not really possible in our convoluted code anymore ;)

Functions/Import/xASL_imp_DCM2NII_ConvertScan.m
HenkMutsaerts: I thought this was never used, this is why I removed it.

Modules/SubModule_Import/xASL_wrp_NII2BIDS_Subject.m
HenkMutsaerts: Well, this way, both T1w and T1 are supported. And T2 as well.
I would think it is the best if the first `dcm2nii` is very flexible, and the subsequent `nii2bids` should be restricted (and we can then solve problems in this stage).

Functions/Import/xASL_imp_DCM2NII_Subject_StartConversion.m
HenkMutsaerts: Yes, we don't want to unzip, when converting to BIDS, we need to zip. So I changed it in a way that `rawdata` will always contains `.nii.gz` (I think a few lines ahead, we run `xASL_adm_GzipAllFiles` to ensure that all are zipped, so this just saves time).

Modules/SubModule_Population/xASL_wrp_CreateBiasfield.m
HenkMutsaerts: No clue!

HenkMutsaerts:
@jan-petr Please go through these changes when looking at the test example that I added to the FlavorDatabase, then you understand some of my suggestions; otherwise, we should go through these together.

Functions/Import/xASL_imp_DCM2NII_ConvertScan.m
jan-petr: We use it - it's the function output. So this has to be reverted.

Functions/Import/xASL_imp_DCM2NII_ConvertScan.m
jan-petr: Actually, after some though. Yes - OK to use this as this will be renamed to BIDS that has a well defined name for it. 
And I have seen that you have made some further modifications later. 

However, two things have to be done:
1. We need to check how this behaves for T1c. Because previously we had T1w and T1c - which are quite distinct. Now with T1 and T1w, there might be issues that T1c is recognized incorrectly as T1.
2. Functions/BIDS/xASL_bids_Config.m - line 187 and 193 have to be adapted by adding the T1 and T2 types.

Functions/Import/xASL_imp_DCM2NII_ConvertScan.m
jan-petr: But previously. This worked rather correctly and you are now changing this behavior. That might be problematic.

Modules/SubModule_Import/xASL_wrp_NII2BIDS_Subject.m
jan-petr: You have convinced me - see the other comment that helps to fix this. Cause we have predefined inputs and assigned outputs. Enough to change this config file to make sure that things are solved on the `nii2bids` side.

jan-petr:
I've resolved some comments and checked new updates - only 3 issues left to be addressed.

Functions/Import/xASL_imp_DCM2NII_ConvertScan.m
HenkMutsaerts: Look up to the start of the function, I moved the initialization there, for readability. Just to ensure that we always initialize output parameters at the start.

I came across a debugging issue where I needed that.

Modules/SubModule_Import/xASL_wrp_NII2BIDS_Subject.m
HenkMutsaerts: Which other comment?

Functions/Import/xASL_imp_DCM2NII_ConvertScan.m
HenkMutsaerts: In general: we didn't test it anyway for `T1c`, but we can of course adapt regular expressions to exclude `T1c` when we mean `T1|T1w`.

Here, there is no different behavior for T1|T1w|T1c, this line only says that all structural stuff are `bAnatomical`.

`xASL_bids_Config` line 187 is correct (it has `T1w|T2w|T1c`)

Line 192: can it be `T1(|w)` under filename? This line was already incompatible with line 187, by not having `T1c`?

BTW: my change in `xASL_imp_DCM2NII_ConvertScan` just made sure that both `T1w` and `T1` will be interpreted as anatomical; doesn't necessarily mean that the whole import needs to be able to recognize T1 and T1w (although that would be nice).

Can you fix this further?

Functions/Import/xASL_imp_DCM2NII_ConvertScan.m
HenkMutsaerts: Well let's test it, it doesn't really work nicely, because `NII2BIDS` assumes folders for all scantypes, if I remember correctly. Perhaps especially if you have multiple T1 or multiple FLAIR scans. But there was something going wrong.

jan-petr: We need to check together then.

HenkMutsaerts:
@jan-petr I have resolved the last issues, if you can merge the NIfTI example in the Flavor Database (in branch feature 1300), then this can be closed.

Functions/Import/xASL_imp_DCM2NII_ConvertScan.m
jan-petr: OK. Makes sense.

jan-petr (approved):
OK

Issue #1103, opened 2022-04-29
<SUBJECTNAME> vs <SUBJECTNAME>_1


HenkMutsaerts:
### Description

When `/rawdata/<SUBJECTNAME>` `/derivatives/ExploreASL/<SUBJECTNAME>` both exist, `xASL_module_BIDS2Legacy` will recreate `/derivatives/ExploreASL/<SUBJECTNAME>_1` (failing to recognize that this is the same as the one without _1).

However, if `010_BIDS2LEGACY.status` exists for <SUBJECTNAME>_1, it will behave correctly.

Without `010_BIDS2LEGACY.status`, `/derivatives/ExploreASL/<SUBJECTNAME_1>` will be overwritten. 

It is probably nicest to add warnings, see proposed changes below.

### Tasks

- [x] If a folder in /derivatives/ExploreASL/ exists and will be overwritten, issue a warning
- [x] Recognize that <SUBJECTNAME> and <SUBJECTNAME>_1 are the same in `/derivatives/ExploreASL` (per legacy structure) so that this will also be overwritten and issue a warning

### How to test

clone `/ExploreASL/External/TestDataSet`, run `ExploreASL('path', 0, 0)`, rename `sub-Sub1_1` to `sub-Sub1` and rerun `ExploreASL('path', 0, 0)`

This will give the following two warnings. Note that the second warning and text will be shown if a folder already existed, without renaming sub-*** to sub-***_1. This only applies to the first visit/session, and only if the lock file `010_BIDS2LEGACY.status` doesn't exist for this subject.

<img width="694" alt="image" src="https://github.com/user-attachments/assets/28e8a9df-0e83-4ac4-9d40-6ef01f798d94">

### Release notes

Fix BIDS2Legacy renaming to SUBJECTNAME_1

PR #1305, opened 2023-01-31
Closes #1178 xASL_init_Iteration: fix diaryFile for LongReg

### LINKED ISSUE #1178: Symbol not defined: SESSION  -- when running Longitudinal Registration
### Description

Probably related to #1103 . In short, due to the "_#" approach, the SESSION symbol can't be resolved when it is requested to do so within `xASL_init_iteration` -> `runIteration`. Error happens around this block:

```matlab
% Optional diary logfile of command window
        if ~isempty(diaryFile)
            % NB. Don't overwrite diaryFile itself because it might be expanded
            % with different values in the next iteration.
            diaryFileEx = xASL_adm_ReplaceSymbols(diaryFile,x);
            diaryFileEx = fullfile(diaryFileEx);
```

### Tasks

- [ ] Fix SESSION field in `x` struct by the time `xASL_adm_ReplaceSymbols` is called before running longitudinal registration

### How to test

1) Checkout `develop` branch.

2) Run a dataset with multiple visits. Example given below, which successfully completed the Structural Module and is about to execute the LongReg module:

image

3) In dataPar.json, `x.modules.bRunLongReg` is set to 1.  Run with only the Structural module selected.

4) Breakpoint at line 304 in `xASL_init_Iteration` when `jobfn` is @xASL_module_LongReg. (workspace attached)

[LongRegBugSESSIONSymbol__xASL_init_Iteration.zip](https://github.com/ExploreASL/ExploreASL/files/9465340/LongRegBugSESSIONSymbol__xASL_init_Iteration.zip)

### Release notes

Required: Fixed a bug where the <SESSION> symbol was not interpreted correctly prior to beginning longitudinal regsitration.

HenkMutsaerts:
### Linked issue

Closes #1178

### DIFF
--- a/Functions/Initialization/xASL_init_Iteration.m
+++ b/Functions/Initialization/xASL_init_Iteration.m
@@ -129,8 +129,11 @@ function [bAborted, xOut] = xASL_init_Iteration(x, moduleName, dryRun, stopAfter
         dbSettings.diaryFile = fullfile('<ROOT>', 'derivatives', 'ExploreASL' ,'log', [moduleName '_sub-<SUBJECT>.log']);
     elseif ~isempty(regexpi(ModName,'(Structural|LongReg)', 'once'))
         dbSettings.diaryFile = fullfile('<ROOT>', 'log', [moduleName '_<SUBJECT>.log']);
+    elseif ~isempty(regexpi(ModName,'(ASL|func|dwi)', 'once'))        
+        dbSettings.diaryFile = fullfile('<ROOT>', 'log', [moduleName '_<SUBJECT>_<SESSION>.log']);
     else
         warning(['Unknown module name: ' moduleName]);
+        % default
         dbSettings.diaryFile = fullfile('<ROOT>', 'log', [moduleName '_<SUBJECT>_<SESSION>.log']);
     end
     
--- a/Functions/Initialization/xASL_init_Iteration.m
+++ b/Functions/Initialization/xASL_init_Iteration.m
@@ -130,7 +130,7 @@ function [bAborted, xOut] = xASL_init_Iteration(x, moduleName, dryRun, stopAfter
         dbSettings.diaryFile = ['<ROOT>/log/' moduleName '.log'];
     elseif ~isempty(regexp(ModName,'(Import|BIDS2Legacy)', 'once'))
         dbSettings.diaryFile = ['<ROOT>/derivatives/ExploreASL/log/' moduleName '_sub-<SUBJECT>.log'];
-    elseif ~isempty(regexp(ModName,'(Structural)', 'once'))
+    elseif ~isempty(regexp(ModName,'(Structural|LongReg)', 'once'))
         dbSettings.diaryFile = ['<ROOT>/log/' moduleName '_<SUBJECT>.log'];
     else
         dbSettings.diaryFile = ['<ROOT>/log/' moduleName '_<SUBJECT>_<SESSION>.log'];
--- a/Functions/Initialization/xASL_init_Iteration.m
+++ b/Functions/Initialization/xASL_init_Iteration.m
@@ -32,11 +32,13 @@ function [bAborted, xOut] = xASL_init_Iteration(x, moduleName, dryRun, stopAfter
     dbSettings.x                     = x;
     dbSettings.x.settings.RERUN      = false;
     dbSettings.x.settings.MUTEXID    = moduleName;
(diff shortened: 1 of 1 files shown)


jan-petr (changes requested):
This commit looks good. But please check and close all tasks from the main issue first.

HenkMutsaerts:
Done, see issue. Also addressed the request by Toshihiko, though he should test it (we don't have a TestDataSet for DARTEL or LongReg); btw: should he run LongReg or DARTEL?

jan-petr:
> Done, see issue. Also addressed the request by Toshihiko, though he should test it (we don't have a TestDataSet for DARTEL or LongReg); btw: should he run LongReg or DARTEL?

You mean a request that doesn't have an issue yet? Good. Thanks. BTW: Is issue #1103 also fixed by this?

HENK: no, read my response in the issue itself.

Functions/Initialization/xASL_init_Iteration.m
```diff
@@ -114,8 +114,11 @@
         dbSettings.diaryFile = fullfile('<ROOT>', 'derivatives', 'ExploreASL' ,'log', [moduleName '_sub-<SUBJECT>.log']);
     elseif ~isempty(regexpi(ModName,'(Structural|LongReg)', 'once'))
         dbSettings.diaryFile = fullfile('<ROOT>', 'log', [moduleName '_<SUBJECT>.log']);
+    elseif ~isempty(regexpi(ModName,'(ASL|func|dwi)', 'once'))
```
jan-petr, line 132: We don't want the local log? There's another place where this is done?

jan-petr (changes requested):
Looks good. One minor question only.

Functions/Initialization/xASL_init_Iteration.m
HenkMutsaerts: I don't understand your question. I have only done the following here:

1. Ensure that structural & LongRun don't get <SESSION> in their log filename (because they don't have SESSIONS, and it crashed because SESSIONS weren't defined)
2. Ensure that ASL|func|dwi do get SESSIONS in the log filename.
3. Use `fullfile` instead of `/`

I didn't change any location?

Functions/Initialization/xASL_init_Iteration.m
```diff
@@ -87,38 +89,37 @@
         ModName = moduleName;
     end
     
-    % DARTEL & LongReg settings
-    if  strcmp(ModName,'DARTEL') || strcmp(ModName,'LongReg')
-        dbSettings.x.dir.LockDir = [dbSettings.x.dir.LockDir '_' x.P.STRUCT ];
```
jan-petr, line 107: The lock-dir for LongReg is now different. And for DARTEL, no lockdir is now defined. Is that correct?

Functions/Initialization/xASL_init_Iteration.m
jan-petr: Sorry. I don't understand this neither anymore :) This what you mention is all fine. I just don't understand the lockdir change for DARTEL+LongReg - see the other comment.

jan-petr (changes requested):
I have clarified my question now. Maybe it is all OK - I'm just asking if this was done on purpose.

Functions/Initialization/xASL_init_Iteration.m
HenkMutsaerts: line 41 sets a general lockdir that is used for all non-specific cases; LongReg specifically needed a different lock folder, as this crashed for Maurice

jan-petr: Yes. So now DARTEL is using the default lockDir set on Line 41? And LongReg now uses the common lockdir set on line 96? Is that fine? I'm not saying it isn't ;)

jan-petr (changes requested):
See last questions in code. It's just to double-check, not necessarily wrong...

Functions/Initialization/xASL_init_Iteration.m
HenkMutsaerts: Yes, something like that. I tested it after I rewrote this; it should be fine:) One of the two expected a session-extension in the lockdir, which wasn't there, which made it crash for Maurice. Then I think I did a small improvement: DARTEL should have no extension, whereas LongReg should have a subject extension.

jan-petr (approved):
OK.

PR #1307, opened 2023-02-01
Optimize #1002 merge single multi pld quantification

### LINKED ISSUE #1002: Merge single- and multi-PLD quantification
### Description

Processing for single-PLD could in many locations of the pipeline be identical as multi-PLDs, by always processing multi-PLD where single-PLD is treated as a multi-PLD sequence with 1 PLD.

Currently, we divide single- and multi-PLD quantification in step 8 of the ASL module (Perform Quantification) and perform the preprocessing steps twice. For future code changes we need to merge this together and perform the specific single or multi PLD quantification steps a layer down (follow-up issue to #759). Furthermore, all scaling can be in separate function, so we might as well do this in a separate function.

### Requirements

- [X] Quantification: merge `xASL_quant_SinglePLD` and `xASL_quant_MultiPLD`
- [X] Check other location with `if bMultiPLD` and tweak likewise
HENK: NOTE THAT THIS LAST REQUIREMENT IS NOT FIXED, WE CAN MOVE THIS TO A LATER ISSUE

### Release notes

Merge single- and multi-PLD quantification to a single function.

jan-petr:
### Linked issue

Close #1002


Functions/Quantification/xASL_quant_CBF.m
```diff
@@ -41,39 +43,59 @@
...
+	error('Four input parameters required.');
+end
 
 if  xASL_stat_SumNan(M0_im(:))==0 && x.modules.asl.ApplyQuantification(5)
     error('Empty M0 image, something went wrong in M0 processing');
+elseif xASL_stat_SumNan(M0_im(:))==0
+    warning('Weird M0 image detected');    
 end
 
 if nargin<5 || isempty(bUseBasilQuantification)
```
HenkMutsaerts: Currently, we never check if a boolean really has a 0 or a 1. Shall we create an issue for Maarten, to check for all input parameters of ExploreASL, to verify their content (if they exist)? I wouldn't make this too complicated, we could later of course add dependencies, but for now it would already be nice to check that input parameters are correct before users get weird pipeline results down the line.

Functions/Quantification/xASL_quant_CBF.m
```diff
@@ -1,6 +1,6 @@
-function [ScaleImage, CBF] = xASL_quant_SinglePLD(PWI, M0_im, imSliceNumber, x, bUseBasilQuantification)
-%xASL_quant_SinglePLD % Perform a multi-step quantification
-% FORMAT: [ScaleImage[, CBF]] = xASL_quant_SinglePLD(PWI, M0_im, imSliceNumber, x)
+function [ScaleImage, CBF] = xASL_quant_CBF(PWI, M0_im, imSliceNumber, x, bUseBasilQuantification)
```
HenkMutsaerts: This name is not correct anymore, since we now also output ATT, Tex, and perhaps more in the feature, from this function. Should `xASL_quant_ASL` be more correct?

I know that our pipeline is called "ExploreASL" but we can also process other data than ASL :)

Functions/Quantification/xASL_quant_CBF.m
```diff
@@ -292,18 +298,48 @@
 fprintf('%s\n',' model with parameters:');
 
 if x.modules.asl.ApplyQuantification(3)
```
HenkMutsaerts: Seems to make code easier if you initialize PLDs & LDs before this part as `[]`, and give a warning instead when you cannot find them. Also, is LD an agreed upon abbreviation? Otherwise, `LabDur` could be clearer.

Functions/Quantification/xASL_quant_CBF.m
```diff
@@ -292,18 +298,48 @@
...
-            case 'casl'
-                fprintf('%s',['LabelingDuration = ' xASL_num2str(x.Q.LabelingDuration) ' ms, ']);
-                fprintf('%s',['PLD (ms) = ' xASL_num2str(x.Q.Initial_PLD)]);
-        end
-    else
-        fprintf('Labeling duration and initial PLD undefined...\n');
-    end
+	if modules.asl.bMultiPLD
+		% Prepare unique multi-PLD parameters
+		[PLDs, ind] = unique(x.Q.Initial_PLD, 'stable');
```
HenkMutsaerts: Does `ind` mean `index` or `individual`? Can you write it out?

Functions/Quantification/xASL_quant_CBF.m
```diff
@@ -292,18 +298,48 @@
...
-                fprintf('%s',['TI1 = ' xASL_num2str(x.Q.LabelingDuration) ' ms, ']);
-                fprintf('%s',['TI (ms) = ' xASL_num2str(x.Q.Initial_PLD)]);
-            case 'casl'
-                fprintf('%s',['LabelingDuration = ' xASL_num2str(x.Q.LabelingDuration) ' ms, ']);
-                fprintf('%s',['PLD (ms) = ' xASL_num2str(x.Q.Initial_PLD)]);
-        end
-    else
-        fprintf('Labeling duration and initial PLD undefined...\n');
-    end
+	if modules.asl.bMultiPLD
```
HenkMutsaerts: Can  you rewrite this part by treating this more similar for both multi- and single-PLD? In both cases, if you have multiple LabDur or multiple PLD you would do the same, only at the end you need a warning if singlePLD was selected but multiple PLD or LabDur detected right?

This would make the code more readable as well. Because the handling of multiple PLDs/LabDur is always the same. At the end before we quantify, we do all checks at a single place, including "warning: multiple PLDs or LabDur detected but singlePLD requested, choosing the max ... instead, this might be suboptimal"

Functions/Quantification/xASL_quant_CBF.m
```diff
@@ -51,6 +51,11 @@
 	fprintf('%s\n', 'Performing multi-PLD quantification');
 else
 	fprintf('%s\n', 'Performing single-PLD quantification');
+	if nargout > 2
+		warning('Cannot assign more than two output parameters for single-PLD quantification. Setting ATT and Tex to empty vectors.')
```
HenkMutsaerts: No this is wrong. Single-PLD can output `Tex`. Like below, I would treat single- and multi-PLD the same throughout the code, only at the end you do a check between them

Functions/Quantification/xASL_quant_CBF.m
```diff
@@ -153,12 +158,14 @@
 
     if bUseBasilQuantification
         % Here we perform FSL BASIL
-        PWI = xASL_quant_Basil(PWI, x);
+		if modules.asl.bMultiPLD
```
HenkMutsaerts: Also, this can be made single- and multi-PLD ignorant. Input is the same, and output can be the same as well, `xASL_quant_Basil` should also initialize `ATT` and `Tex` as empty at the start of its function.

HenkMutsaerts (changes requested):
### Sync a bit further
Nice, but it is only a small step further to make this function really multi- and single-PLD ignorant as much as possible, and only at the end do a check for dependencies.

This will help with the generalizability, and readability of our code. In many instances, you want the same for multi-PLD and single-PLD, or it is something that won't matter for single-PLD (like sorting or choosing which PLD or LabDur to use).

A dependency could be:
`warning: multiple PLDs detected but singlePLD quantification requested, using max PLD instead, this may give suboptimal results`

and later, we could replace this by:

`warning: multiple PLDs detected but singlePLD quantification requested, using multiPLD quantification instead`

OR

`warning: multiple PLDs detected but singlePLD quantification requested, now performing both using multiPLD and singlePLD quantifications`

### Only quantification

Also, you didn't do the second requirement in the issue list, you only merged the quantification right? I changed the title of the issue & PR accordingly, and we can make a new issue for merging this in other locations (or do we really not have multi-PLD specific processing elsewhere yet? Then we should make *that* a new issue, if we don't have it already ;) Anyway, in that case the title change would still be good, but my comment can otherwise be ignored

### Please pull
I did a small spelling correction & layout for readability, and I didn't want to annoy you by asking this so I did it myself :)

### Reviewers
Bea doesn't need to review this right?

Functions/Quantification/xASL_quant_CBF.m
jan-petr: Sorry, a mistake. I'm correcting that. Done in 827d3d9e

Functions/Quantification/xASL_quant_CBF.m
jan-petr: OK. Sure, question of style, but I have nothing against. Done here c147db9f

Functions/Quantification/xASL_quant_CBF.m
jan-petr: Yes, done here c147db9f

Functions/Quantification/xASL_quant_CBF.m
jan-petr: Yes true. Single-PLD, multi-TE. I didn't realize that. Done here 94202a18

Functions/Quantification/xASL_quant_CBF.m
jan-petr: sure. no prob. Done here 94202a18

jan-petr:
> ### Sync a bit further
> Nice, but it is only a small step further to make this function really multi- and single-PLD ignorant as much as possible, and only at the end do a check for dependencies.

Agree with this philosophy in general. And parts of the code should be like that. If not, please make specific comments or propose new issues. I would prefer to do only the basics here.
For example, we should make a new issue that multi-TE single-PLD are correctly crunched - lets do that with the newly added flavor that you have.

> This will help with the generalizability, and readability of our code. In many instances, you want the same for multi-PLD and single-PLD, or it is something that won't matter for single-PLD (like sorting or choosing which PLD or LabDur to use).
Here - the sorting will differ a bit necessarily. Not that it would have to be done differently in principle, but the implementation of the sorting for single and multi-PLD will slightly differ, so we need to account for it.
 
> A dependency could be: `warning: multiple PLDs detected but singlePLD quantification requested, using max PLD instead, this may give suboptimal results`
There's never a mutli-PLD detected, but single-PLD requested - we don't have an input option for that. We can open a new issue and create such an option.

 
> and later, we could replace this by:
> `warning: multiple PLDs detected but singlePLD quantification requested, using multiPLD quantification instead`
> OR
> `warning: multiple PLDs detected but singlePLD quantification requested, now performing both using multiPLD and singlePLD quantifications`

Yes - but see above.
 
> ### Only quantification
> Also, you didn't do the second requirement in the issue list, you only merged the quantification right? I changed the title of the issue & PR accordingly, and we can make a new issue for merging this in other locations (or do we really not have multi-PLD specific processing elsewhere yet? Then we should make _that_ a new issue, if we don't have it already ;) Anyway, in that case the title change would still be good, but my comment can otherwise be ignored

Yes - I only merged stuff.

> ### Please pull
> I did a small spelling correction & layout for readability, and I didn't want to annoy you by asking this so I did it myself :)

Perfect - that makes things more efficient. And I anyway check all new commits for errors.

> ### Reviewers
> Bea doesn't need to review this right?

Review code itself not, but test yes :D Hence the review request.

Functions/Quantification/xASL_quant_CBF.m
jan-petr: You mean validating of studyPar.json and dataPar.json? No we don't have that, we can create a new issue...

Functions/Quantification/xASL_quant_CBF.m
jan-petr: Very good point actually. There wasn't any reason here to separate single- and multi-PLD :) Rewritten and simplified here b0758c10

HenkMutsaerts (approved):
Looks good, I did a minor revamp of `xASL_quant_ASL`

jan-petr:
> Looks good, I did a minor revamp of `xASL_quant_ASL`

Looks all good. But Bea still needs to run it and test on both single/multi-PLD and Hadamard.

BeatrizPadrela (approved):
works for:
- single-PLD (Philips_PCASL_2DEPI_3.2.1.1_1)
- multi-PLD (Philips_PCASL_2DEPI_3.2.1.1_dummyMultiPLD_1)
- BBB-ASL (DZNE subject data)

BeatrizPadrela:
All good after testing TestDataSets
image

PR #1321, opened 2023-02-13
Closes #1320 loading is not zipping

### LINKED ISSUE #1320: Loading data should not be zipping data
### Description

When we load data when (re)running processing, `ExploreASL_Process`->`xASL_init_DataLoading` runs `xASL_adm_GzipAllFiles`. This is undesired behavior for BIDS2Legacy:

* `rawdata` should be untouched and read-only. NIfTIs don't need to be unzipped, all information is in the json-files. NIfTIs in `rawdata` are zipped already when running DICOM2BIDS.
* Legacy is a copy of `rawdata`, so already zipped.
* Zipping takes a lot of time, especially for files that will be opened later in the processing anyway.
* Loading data by definition should keep the data untouched, so even if we load data and not-reprocess it, we should not touch it.

### Tasks

- [x] Remove zipping when loading data

### Release notes

Avoid additional zipping of BIDS and Legacy data after import.

HenkMutsaerts:
### Linked issue

Closes #1320

### DIFF
--- a/Functions/Administration/xASL_adm_CleanUpX.m
+++ b/Functions/Administration/xASL_adm_CleanUpX.m
@@ -16,20 +16,9 @@ function [x] = xASL_adm_CleanUpX(x)
 % EXAMPLE:        x = xASL_adm_CleanUpX(x);
 %
 % __________________________________
-% Copyright 2015-2021 ExploreASL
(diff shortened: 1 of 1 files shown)


jan-petr (approved):
OK.

PR #1325, opened 2023-02-23
Feature #1214 siemens product multi pld

### LINKED ISSUE #1214: PLD=0 warning with PLD and labeling duration vector
### Description

Multi-PLD ASL (e.g., Siemens XA31 product) uses different labeling durations for different PLDs, for more efficient imaging. ExploreASL doesn't work with this yet (except for the specific Hadamard case). Also, Siemens uses inflow time, which is the sum of PLD and labeling duration, so their PLD vector will start with a few zeros.

### Tasks

- [x] Deal with the warning that some PLDs have the wrong magnitude. 1) The specific PLDs with the wrong values should be pointed out, to differentiate if these are the zeros or also other values (mainly to differentiate the BIDS seconds and common notation ms). 2) Is there a possibility that we detect that these are correct PLD=0? E.g., if PLD and LabDur are both vectors, and the specific combination of PLD 0 and labeling duration fits?

>> Warning: Field Initial_PLD in xASL structure has multiple values, from 0 to 2400, which is outside of the recommended range <10,10000> ms.  xASL_bids_parms2BIDS (line 227) xASL_adm_LoadParms (line 108) xASL_stat_GetDICOMStatistics (line 81)

JAN: Since we can have both 0 and also other small numbers, I have solved it by checking not only for min and max, but also for mean. If the mean value is within the range, then this is reported, but not as a warning.

- [x] Quantification per PLD should be corrected for the labeling duration differences. If BASIL doesn't do this, we could scale this ourselves.
- [x] New Siemens XA31 multi-PLD flavor was added. It has a several initial PLDs as 0. Moreover they are zero for several volumes, where Labeling duration changes. We shouldn't average over these PLDs. So we should check for unique for both LD and PLDs - merge somehow the result and skip only those that are really repeated.

### Release notes

Correct quantification for multi-timepoint ASL with differing labeling durations.

jan-petr:
### Linked issue

Closes #1214

### DIFF
--- a/Functions/BIDS/xASL_bids_parms2BIDS.m
+++ b/Functions/BIDS/xASL_bids_parms2BIDS.m
@@ -25,7 +25,7 @@ function outParms = xASL_bids_parms2BIDS(inXasl, inBids, bOutBids, bPriorityBids
 % EXAMPLE: outParms = xASL_bids_parms2BIDS(inXasl, inBids);
 %          outParms = xASL_bids_parms2BIDS(inXasl, [], 1, 0);
 % __________________________________
-% Copyright 2015-2021 ExploreASL
+% Copyright 2015-2023 ExploreASL
 
 
 %% Admin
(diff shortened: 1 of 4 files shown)


MDijsselhof (approved):
Looks good to me

PR #1328, opened 2023-03-01
Closes #1326 replaced checking of file system with checking of flavors.json

### LINKED ISSUE #1326: Flavor Test currently crashes at SIEMENS_PASL_2DEPI_VB15A_ADNI2_941_S_4764_1
### Description

Flavor SIEMENS_PASL_2DEPI_VB15A_ADNI2_941_S_4764_1 currently does not contain folders, which means that the FlavorTest crashes at this flavor.

### Tasks

- [x] Add a filter that ignores this specific flavor using the recently added bTesting=false flag. 

### How to test

Run the FlavorDatabase

### Release notes
Flavor-import testing allows to turn of certain dataset.

maartenhammer:
Closes #1326 

The added bTesting parameter in flavor.json is now used to check if a flavor needs testing. 

### How to test

Run Testing/xASL_test_Flavors.m

### Comments

I replaced a filesystem search with a check from flavors.json. 
Which means if a file is renamed in filesystem, but not in flavors.json it will not run in test_Flavors.m.
This can be fixed by adding the following check:
First a filesystem check xASL_adm_GetFileList(testConfig.pathFlavorDatabase, [], false, [], true);
Then a json read from the flavors.json with associated btesting value.
If both exists and btesting=true, then add it to a to be tested flavor
If flavor exists in either flavor.json or filesystem but not **both** throw a warning/error pointing the user to check it out, as this means either Flavor.json is incomplete, or the FlavorDatabase is incomplete.

### DIFF
--- a/Testing/xASL_test_Flavors.m
+++ b/Testing/xASL_test_Flavors.m
@@ -101,11 +101,6 @@ else
 	system(testConfig.cmdCloneFlavors);
 end
 
-% Check for flavor list
-if ~isfield(testConfig,'flavorList')
-	testConfig.flavorList = xASL_adm_GetFileList(testConfig.pathFlavorDatabase, [], false, [], true);
-end
-
 % Load database JSON
 if xASL_exist(fullfile(testConfig.pathFlavorDatabase,'flavors.json'),'file')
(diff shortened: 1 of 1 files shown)


jan-petr (approved):
OK.

jan-petr (approved):
OK to merge

PR #1332, opened 2023-03-05
Closes #1294 ge 3volumes flavor

### LINKED ISSUE #1294: Minor GE flavor bugs
### Description

Some minor issues I discovered when importing and processing the DV25.0 GE at 1.5T.

@jan-petr I have prepared this for you, just rebased with develop March 5th 2023, so should be easy for you to process.
The automatic splitting is now moved to #1331, for now I will just split the DICOMs and manually add the PostLabelingDelay.

### Tasks

- [x] HENK: Automatically discard a CBF volume in BIDS2Legacy (like we do with dummy volumes)? JAN: I would keep this in BIDS2Legacy and just make sure that it's copied correctly. And then when splitting ASL to ASL + ASL_source + M0, we get rid of CBF if we see we have M0 and ASL as well (so at the same time when we throw of dummy scans. See below
HENK: This now works with the correct parameters in `studyPar.json`, but this doesn't lead to correct BIDS rawdata, but it works. I thought is was very confusing that `xASL_module_ASL` line 266 > `xASL_io_SplitASL` uses `x.modules.asl.DummyScanPositionInASL4D`, but this wasn't initialized anywhere. I now see this is dealt with in BIDS2Legacy and have added a comment to acknowledge this.
- [x] GE vendor warning when sequence parameter exists and set to 3D spiral
- [x] Create import check for number of tokens in `folderHierarchy` and `tokenOrdering`, these should be the same, otherwise the import will continue but create wrong outputs (in my case it called the `ASL4D.nii` the `SubjectName.nii`)
- [x] `xASL_imp_DCM2NII_Subject_StartConversion` line 33 contains iSubject iVisit iSession iScan which are not defined (probably a copy-pasting issue).

### Release notes

Minor bug fixing for importing GE scans.

HenkMutsaerts:
### Linked issue

Closes #1294

### DIFF
--- a/Functions/Import/xASL_imp_CheckImportSettings.m
+++ b/Functions/Import/xASL_imp_CheckImportSettings.m
@@ -20,7 +20,7 @@ function [x] = xASL_imp_CheckImportSettings(x)
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
 % EXAMPLE:        n/a
 % __________________________________
-% Copyright 2015-2021 ExploreASL
+% Copyright 2015-2023 ExploreASL
 
 
     % Check bCheckPermissions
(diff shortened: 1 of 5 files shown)


Functions/Import/xASL_imp_CheckImportSettings.m
```diff
@@ -49,7 +49,31 @@
         fprintf('If you want to overwrite, first remove the full subject folder...');
     end
 
+    
+%% Check if folderHierarchy matches with tokenOrdering
+nLeftBrackets = 0;
```
jan-petr: This ignores the fact that a bracket can be added as a symbol as well.
In 90% of cases it's nice to report as a possible error. But in 10% of cases, this is a valid entry and shouldn't result in an error! Make this a warning only and a conditional one!

The counting of tokens is 100% wrong. There is no rule that the number of tokens defined has to match the number of tokens. We can make a warning if bracket number is smaller than the number of tokens. But inequality should be only reported as suspicious, never as an error!

Functions/Import/xASL_imp_DCM2NII_Subject_StartConversion.m
```diff
@@ -1,4 +1,4 @@
-function [globalCounts, x, summary_line, destdir, scanpath, scan_name, dcm2niiCatchedErrors, nii_files, first_match] = xASL_imp_DCM2NII_Subject_StartConversion(globalCounts, x, bSkipThisOne, summary_line, destdir, scanpath, scan_name, dcm2niiCatchedErrors)
```
jan-petr, line 1: scanfield has to be added to the parameter description and to the FORMAT.

jan-petr: but of course a nice catch finding out that scanfield is necessary. 
This bugs goes for Michael who splitted functions without controlling the variables and content.

Modules/xASL_module_ASL.m
```diff
@@ -261,6 +261,8 @@
 %% G1. Split ASL and M0 within the ASL time series
 % Run this when the data hasn't been touched yet
 % The first three states are here, because the first two are run only conditionally
+
+% NOTE THAT THIS PART IS BEING PHASED OUT, AND NOW PROCESSED IN `xASL_bids_parseM0`
```
jan-petr: I agree. But  we might write a bit more explaining that this is done in BIDS->Legacy conversion and we keep this here only when we need to work with data imported long time ago.

jan-petr (changes requested):
1. The bugs are correctly fixed. Very good.
2. The code checking for token and bracket numbers goes beyond what is the definition and will invalid also cases that are not incorrect - see further comments.
3. The issue with cbf scan is not dealt with correctly. The correct solution is not to throw out volumes set as "dummies" but to recognize that when we have both control/label/m0scan etc and cbf, then the CBF should be thrown out. At least for GE, I would do that. The thing is that GE nicely works out of DICOM without extra settings and I would be happier if this is done automatically and automatically deciding to throw CBF out.

Functions/Import/xASL_imp_CheckImportSettings.m
HenkMutsaerts: The only rule that I enforced here, is that the number of tokens in tokenOrdering equals the number of tokens in folderHierarchy, which should always be the case (provided it takes potentially empty sessions/visits into account), otherwise it will crash further down the code. If we only issue a warning here, the code continuous and can incorrectly use the wrong tokenOrdering token with the wrong folder in the folderHierarchy. This can then produce a result that doesn't crash the pipeline, but is incorrect and takes time to bugfix. So I would keep this check.

jan-petr: Yes, I'm fine with checking the number of tokens. BUT :) The number of brackets doesn't have to equal the number of tokens. There are other uses of brackets than for tokens...

Functions/Import/xASL_imp_DCM2NII_Subject_StartConversion.m
HenkMutsaerts: This should now be fixed

Modules/xASL_module_ASL.m
HenkMutsaerts: Done

jan-petr (changes requested):
The last issue about token numbers is still not addressed.

Functions/Import/xASL_imp_CheckImportSettings.m
HenkMutsaerts: You probably mean that you add e.g., (nii|nii\\.gz) but that should be illegal.

jan-petr: Nono. I mean that in regexp, there are other uses of the bracket character that is not defining tokens! So if we strictly follow regexp syntax, then we should rely that the number of brackets equals to the number of tokens.

HenkMutsaerts: That's what I'm saying. The folderHierarchy regexp doesn't allow () outside the use of defining tokens. Otherwise, we don't know if () is for either/or for defining tokens, right?

So when the number of brackets is not according to the number of tokens, I now added a hint to the error, for the user to understand that you can only use () for tokens.

HenkMutsaerts: See my last edit, does this clarify it?

HenkMutsaerts:
See my last edit, I have now added the clarification, that `()` is reserved only to define tokens (because otherwise, we don't know what is token or not in the interpretation, so we have to reinforce this with an error right?).

Unless you can come up with an example where you want to have `()` also for either/or, without making this illegal.
E.g., do you want to keep the possibility of doing instead of '(ASL|T1w)' '(ASL(01|02)|T1w)' ? Because '(ASL|T1w)(01|02)' would definitely not work (the 01 or 02 would become a token). In this situation, the user should just use '(ASL|T1w).*', because it otherwise is unclear, 01|02 could also be used for series? So "tokens within tokens" should be illegal, and () can only define tokens, right?

jan-petr (approved):
OK

PR #1346, opened 2023-03-18
Closes #577 bug #577 population sessions

### LINKED ISSUE #577: Incorrect volumetric output
Structural volumetric output (GM/WM/CSF_vol etc.) is not outputted into a _stats_ file at all times. If the session number does not start with 1, no output will be given. See below:

image

The printing of the stats files should recognise which session is the first in the subject and output volumetrics accordingly.

HenkMutsaerts:
### Linked issue

Closes #577

### How to test

### Comments

### DIFF
--- a/Functions/Administration/xASL_adm_GetPopulationSessions.m
+++ b/Functions/Administration/xASL_adm_GetPopulationSessions.m
@@ -50,7 +50,7 @@ else % If files found, continue with defining sessions from SessionList
         NewList{iSession,1} = SessionList{iSession}(1,IndexStart:IndexEnd); % create a cell array containings characters of all session numbers present
     end
     
-    SESSIONS = sort(unique(NewList)); % determine unique session numbers
+    SESSIONS = xASL_adm_SortStringNumbers(unique(NewList)); % determine unique session numbers
     % 4. define nSessions as highest unique session number
     nSessions = numel(SESSIONS);
     bSessionsMissing = 0;
--- a/Functions/Administration/xASL_adm_GetPopulationSessions.m
+++ b/Functions/Administration/xASL_adm_GetPopulationSessions.m
@@ -24,12 +24,6 @@ function [nSessions, bSessionsMissing, SESSIONS] = xASL_adm_GetPopulationSession
 % Copyright (C) 2015-2021 ExploreASL
 
 %% 1. Administration
-if ~isfield(x, 'SESSIONS')
-    SESSIONS = 'ASL_1';
-else
-    SESSIONS = x.SESSIONS;
-end
-
 currentSubjectRegExp = x.dataset.subjectRegexp;
 if strcmp(currentSubjectRegExp(1), '^')
     currentSubjectRegExp = currentSubjectRegExp(2:end);
@@ -37,8 +31,10 @@ end
 if strcmp(currentSubjectRegExp(end), '$')
     currentSubjectRegExp = currentSubjectRegExp(1:end-1);
 end
-% 2. Look for processed files from which to determine the amount of sessions present
-SessionList = xASL_adm_GetFileList(x.D.PopDir,['^' x.S.InputDataStr '_' currentSubjectRegExp '_ASL_\d+(\.nii|\.nii\.gz)$'], 'FPList', [0 Inf]);
+
+%% 2. Look for processed files from which to determine the amount of sessions present
+SessionList = xASL_adm_GetFileList(x.D.PopDir,['^' x.S.InputDataStr '_' currentSubjectRegExp '_ASL_\d+\.nii$'], 'FPList', [0 Inf]);
+
 
 %% 3. Obtain nSessions
 if isempty(SessionList) % If no files found, search for subject files instead of session files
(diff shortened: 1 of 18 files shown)


HenkMutsaerts:
@MDijsselhof Can you test this in your dataset(s) by rerunning the population module? The first columns in the stats files are created from `participants.tsv`, which is why you should first remove `participants.tsv`. Now `participants.tsv` should be created correctly and the rest should follow.

Functions/Administration/xASL_adm_GetPopulationSessions.m
```diff
@@ -54,13 +50,13 @@
         NewList{iSession,1} = SessionList{iSession}(1,IndexStart:IndexEnd); % create a cell array containings characters of all session numbers present
     end
     
-    UniqueSessions = unique(NewList); % determine unique session numbers
+    SESSIONS = sort(unique(NewList)); % determine unique session numbers
```
jan-petr: We have to be very careful about sorting! Because when sorting strings, ASL_12 will come before ASL_2, but we want to sort by the number - so ASL_12 after ASL_2

jan-petr: You have to sort as number not as string. Or perhaps this won't matter later in the code, but then the sorting might not be necessary. My point - if you need to sort, then you better do it properly ;)

Functions/BIDS/xASL_bids_Add2ParticipantsTSV.m
```diff
@@ -60,22 +60,32 @@ function xASL_bids_Add2ParticipantsTSV(DataIn, DataName, x, bOverwrite)
...
+    % Sessions found
+    if size(DataIn,2)<3 % A session column should exist
         warning('Session column missing, too few columns, skipping');
         return;
-    else
-        TempSessionNames = unique(DataIn(:,2));
-        if ~max(strcmp(TempSessionNames, 'ASL_1'))
-            warning('First session missing, might go wrong');
+    else % check if the sessions are the same in DataIn and x.SESSIONS
+        TempSessionNames = sort(unique(DataIn(:,2)))';
```
jan-petr: again - careful how you sort.

Functions/BIDS/xASL_bids_Add2ParticipantsTSV.m
```diff
@@ -202,4 +211,60 @@ function xASL_bids_Add2ParticipantsTSV(DataIn, DataName, x, bOverwrite)
...
+    %% Admin
+    if nargin<3 || isempty(bHasHeader)
+        bHasHeader = false;
+    end
+
+    numberColumns = size(MatrixIn,2); % Obtain the size of the matrix (number of data points)
+
+    %% Create the header (if needed)
+    if bHasHeader
+        MatrixOut(1, 1) = MatrixIn(1, 1); % subject header
```
jan-petr, line 241: careful -  you are using different brackets on each line. It is probably fine as MatrixIn has cells. I'm just asking to double check and you can then resolve without answering ;)

Functions/ImageProcessing/xASL_im_SplitImageLabels.m
```diff
@@ -1,4 +1,4 @@
-function xASL_im_SplitImageLabels(ImagePaths, LabelTable, OutputFolder, bOverwrite, ResampleDir, SubRegExp)
+function xASL_im_SplitImageLabels(ImagePaths, LabelTable, OutputFolder, bOverwrite, ResampleDir, SubRegExp, sessionFolder)
 %xASL_im_SplitImageLabels Extract individual label/regions from image(s)
 %
 % FORMAT: xASL_im_SplitImageLabels(ImagePaths, LabelTable[, OutputFolder, bOverwrite, ResampleDir, SubRegExp])
```
jan-petr: please add the new variable to this line as well

jan-petr (changes requested):
Some minor formal things to be fixed.

Functions/Administration/xASL_adm_GetPopulationSessions.m
HenkMutsaerts: Good point. At this time, nobody probably has used >9 sessions, Mathijs could have for NOMARED, would be good to check. I fixed this throughout now.

jan-petr (approved):
OK.

Functions/BIDS/xASL_bids_Add2ParticipantsTSV.m
```diff
@@ -1,7 +1,7 @@
 function xASL_bids_Add2ParticipantsTSV(DataIn, DataName, x, bOverwrite)
 %xASL_bids_Add2ParticipantsTSV Add data to participants.tsv
 %
-% FORMAT: xASL_bids_Add2ParticipantsTSV(DataIn, DataName, x, bOverwrite)
+% FORMAT: xASL_bids_Add2ParticipantsTSV(DataIn, DataName, x[, bOverwrite])
```
MDijsselhof, line 4: Should it be like this? With the [ ]?

MDijsselhof (changes requested):
I had one small question

Functions/BIDS/xASL_bids_Add2ParticipantsTSV.m
jan-petr: Yes. This is correct. Not in the real function declaration. But in the comment, it specifies that the parameters in [ ] are optional. So correct in this case - we do it like this in almost all other functions.

MDijsselhof (approved):

PR #1347, opened 2023-03-18
Closes #1120 motion correction parameters

### LINKED ISSUE #1120: Motion correction bug & feature requested by Anouk Schrantee EPOD
### Description

Anouk was using: Run EPOD candidate `029_1` with Matlab 2019a, development branch 2.0 on Radian.
EPOD ADHD kids move a lot, and their ASL is without background suppression.

Her motion correction exclusion parameters gave weird lines 

Also, they would like to account for motion parameters, and would like to get motion parameters both before and after exclusion of motion spikes.

### Tasks

- [x] Fix plotting exclusion line
- [x] Save motion parameters also after spike removal

### Release notes

Save motion parameters before and after spike removal.

BTW: @jan-petr if we want a nice motion correction challenge, we could use subject EPOD 006_1 ! This subject moves all over the place :)
BTW: one nice feature could be to remove motion pseudo-perfusion rim lines in images without background suppression

HenkMutsaerts:
### Linked issue

Closes #1120

### DIFF
--- a/Modules/SubModule_ASL/xASL_wrp_RealignASL.m
+++ b/Modules/SubModule_ASL/xASL_wrp_RealignASL.m
@@ -469,7 +469,26 @@ xASL_delete(rInputPath); % delete temporary image
 
 % Save results for later summarization in analysis module
 xASL_adm_CreateDir(x.D.MotionDir);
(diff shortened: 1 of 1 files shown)


jan-petr (approved):
OK

PR #1360, opened 2023-03-28
#1301 changed BASIL input parameters

### LINKED ISSUE #1301: BASIL ATT quantification settings are wrong
### Description

BASIL requires settings, currently defined in the `Basil_ModelOptions.txt` file, for ATT quantification. The `Include macrovascular component` setting is not included but should be to quantify ATT accurately.

### Tasks
- [x] Check if 'simple' masking helps speed-up BASIL processing
- [x] Parameters are incorrectly added to basil_options and should be added to the command line calling basil
- [x] Add dispersion parameter --disp=gamma at line 375, change code to [basiloptions xxx]
- [x] Add arterial component parameter --inferart at line 364, change code to [basiloptions xxx]. Default is on for multi-PLD?
- [x] Spatial regularlization is recommended, always turn on by default
- [x] iaff=diff not needed as BASIL only accepts PWI and other steps are performed before PWI is obtained https://asl-docs.readthedocs.io/en/latest/basilcmd_examples.html?highlight=iaf%3Ddiff. 
- [x] -@ basil_options.txt should be --optfile basil_options.txt, also true for FABBER https://fabber-core.readthedocs.io/en/latest/running.html
- [x] --save-model-fit is not an option in BASIL
- [x] --allow-bad-voxels not an option in BASIL, also we don't do structural space analysis right?
- [x]  Check if datapar options exist for BASIL settings
- [x] Add to documentation, at x.Q. etc.
- [x] Jan check documentation changes, 
- [x] Jan recompile and publish documentation.
- [x] Copy fblood to ABV map
- [x] Check how CBF/ATT/ABV is copied to ASL_x folder from BASIL output folder - use the latest step. | Latest step is searched for, so all good | 
- [x] Fix here also #558 Jan: Commits copied, testing will be done later.
- [x] Fix here also #1181 
- [x] Merge BASIL and FABBER and call the function xASL_quant_FSL
- [x] Make sure that FSL options for BASIL and FABBER are done the same way in xASL_sub_FSLOptions
- [x] Fix FABBER call with the correct options and options file
- [x] Delete previous BASIL/Fabber output -> here the output may be separate for calculating different maps (e.g., one may want to compute Tex twice, once with CBF and ATT maps from sequence A and once from sequence B)
- [x] Add option (configurable in dataPar.json) to turn off deleting FSL input/output to xASL_quant_FSL.
- [x] Check units and scalings for ABV.
- [x] --output and --data for FABBER - provide as absolute path.
- [x] For FABBER DEBBIE, we provide TIs (time from labeling to readout for PASL). Sometimes for PCASL , TIs are also provided instead of PLDs but TI=PLD+Tau. So I don't understand why we provide TI=PLD for FABBER for DEBBIE. Can @BeatrizPadrela and @amahroo explain that this is correct.
- [x] Keep PWI4D

### Release notes

Fix calling BASIL and FABBER.

jan-petr:
### Linked issue

Close #1301


Functions/Quantification/xASL_quant_ASL.m
```diff
@@ -15,6 +15,7 @@
 % ScaleImage        - image matrix containing net/effective quantification scale factor
 % CBF               - Quantified CBF image
 % ATT               - Estimated ATT map (if multi-PLD, otherwise empty)
+% ABV               - Estimated arterialBloodVolume map (if multi-PLD, otherwise empty)
```
MDijsselhof: Need some spaces?

Modules/SubModule_ASL/xASL_wrp_Quantify.m
```diff
@@ -431,11 +432,8 @@ function xASL_wrp_Quantify(x, PWI_Path, pathOutputCBF, M0Path, SliceGradientPath
 
 %% ------------------------------------------------------------------------------------------------
 %% 8.   Perform Quantification
-if ~x.modules.asl.bMultiPLD % single PLD quantification
-    [~, CBF, ATT, Tex] = xASL_quant_ASL(PWI, M0_im, SliceGradient, x, x.Q.bUseBasilQuantification); % also runs BASIL, but only in native space!
-elseif x.Q.bUseBasilQuantification
-    % perform BASIL multi-PLD quantification
-    [~, CBF, ATT, Tex] = xASL_quant_ASL(PWI, M0_im, SliceGradient, x, x.Q.bUseBasilQuantification); % also runs multi-PLD BASIL, but only in native space!
+if ~x.modules.asl.bMultiPLD || x.Q.bUseBasilQuantification % multi-PLD with BASIL or single-PLD
```
MDijsselhof, line 435: Nice, smart solution

Modules/SubModule_Population/xASL_wrp_CreatePopulationTemplates.m
```diff
@@ -284,6 +284,7 @@ function xASL_wrp_CreatePopulationTemplates(x, bSaveUnmasked, bCompute4Sets, Spe
     PreFixList{end+1}   = 'HEMI';                               TemplateNameList{end+1}  = 'HEMI';             SessionsExist(end+1)    = 0;    
     PreFixList{end+1}   = 'TASL';                               TemplateNameList{end+1}  = 'TASL';             SessionsExist(end+1)    = 0;        
     PreFixList{end+1}   = 'Tex';                                TemplateNameList{end+1}  = 'Tex';              SessionsExist(end+1)    = 1;        
+	PreFixList{end+1}   = 'ABV';                                TemplateNameList{end+1}  = 'ABV';              SessionsExist(end+1)    = 1;
```
MDijsselhof: A tab?

Modules/SubModule_Population/xASL_wrp_GetROIstatistics.m
```diff
@@ -126,6 +126,8 @@ function xASL_wrp_GetROIstatistics(x)
     x.S.unit = 'mL';
 elseif ~isempty(strfind(x.S.InputDataStr,'Tex'))
     x.S.unit = 's';
+elseif ~isempty(strfind(x.S.InputDataStr,'ABV'))
+    x.S.unit = 'mL/100g/min';
```
MDijsselhof: shouldn't this be mL or perhaps mL/min then?

Functions/Initialization/xASL_init_FileSystem.m
```diff
@@ -65,7 +65,7 @@
 FileDef{2} = {'y_ASL' 'ASL4D' 'ASL4D_RevPE' 'M0_RevPE'...
 	           'CBF' 'qCBF' 'qCBF4D' 'qCBF_untreated' 'qCBF_masked' 'despiked_ASL4D' ...
 			   'PseudoCBF' 'PWI' 'PWI4D' 'mean_PWI_Clipped' 'mean_PWI_Clipped_DCT' 'M0' 'mean_control' 'SD' 'SNR' 'SD_control'...
-			   'SNR_control' 'SliceGradient' 'SliceGradient_extrapolated' 'FoV' 'TT' 'ATT' 'Tex' 'PVgm' 'PVwm' 'PVcsf' 'PVwmh' 'CBFgm' 'CBFwm'}; 
+			   'SNR_control' 'SliceGradient' 'SliceGradient_extrapolated' 'FoV' 'TT' 'ATT' 'Tex' 'ABV' 'PVgm' 'PVwm' 'PVcsf' 'PVwmh' 'CBFgm' 'CBFwm'};
```
MDijsselhof, line 68: Why don't we call this CBV?

Functions/Quantification/xASL_quant_FSL.m
```diff
@@ -0,0 +1,573 @@
...
+	%             fprintf('BASIL: Mean maximum signal across brain: %f\n', brain_mag);
+	%             % This will correspond to whole brain CBF (roughly) - about 0.5 of GM
+	%             x.Q.BasilNoiseSD = sqrt(brain_mag * 2 / x.Q.BasilSNR);
+	%         end
+	%         fprintf('BASIL: Using a prior noise std.dev. of: %f\n', x.Q.BasilNoiseSD);
+	%         fprintf(FIDoptionFile, '--prior-noise-stddev=%f\n', x.Q.BasilNoiseSD);
+	%     end
+
+
+	%% 5. Extra BASIL fitting options
```
MDijsselhof: So we should do this by default

Functions/Quantification/xASL_quant_FSL.m
```diff
@@ -0,0 +1,573 @@
...
+
+	if x.Q.BASIL.bInferArt
+		fprintf('BASIL: Infer arterial BV and arrival time\n');
+		BasilOptions = [BasilOptions ' --inferart'];
+	end
+
+	switch (x.Q.BASIL.Exch)
+		case 'simple'
+			fprintf('BASIL Exchange model: Simple single compartment with T1 of blood, per white paper\n');
+			BasilOptions = [BasilOptions ' --exch=simple'];
```
MDijsselhof: This is default

Functions/Quantification/xASL_quant_FSL.m
```diff
@@ -0,0 +1,573 @@
...
+		case 'spa'
+			fprintf('BASIL Exchange model: A single pass approximation from St. Lawrence\n');
+			BasilOptions = [BasilOptions ' --exch=spa'];
+		otherwise
+			warning(['BASIL Exchange model: ' x.Q.BASIL.Exch ' not recognized.'])
+	end
+
+	switch (x.Q.BASIL.Disp)
+		case 'none'
+			fprintf('BASIL Dispersion model: none\n');
```
MDijsselhof: Default option

MDijsselhof: I suggest we put the default options there anyway, to make it clear for the user

Functions/Quantification/xASL_quant_FSL.m
```diff
@@ -0,0 +1,573 @@
...
+			BasilOptions = [BasilOptions ' --disp=gauss'];
+		case 'sgauss'
+			fprintf('BASIL Dispersion model: Spatial Gaussian dispersion kernel\n');
+			BasilOptions = [BasilOptions ' --disp=sgauss'];
+		otherwise
+			warning(['BASIL Dispersion model: ' x.Q.BASIL.Disp ' not recognized.'])
+	end
+
+	if isfield(x.Q,'BasilDebug') && x.Q.BasilDebug
+		BasilOptions = [BasilOptions ' --devel'];
```
MDijsselhof: In BASIL this is not an option as far as I can tell, and i can't find this for FABBER either

MDijsselhof (changes requested):
See comments

Functions/Quantification/xASL_quant_FSL.m
```diff
@@ -0,0 +1,573 @@
...
+	if bUseFabber
+		cmdlineOptions = [cmdlineOptions ' --output=' xASL_adm_UnixPath(pathFSLOutput, ispc)];
+		cmdlineOptions = [cmdlineOptions ' --data=', xASL_adm_UnixPath(pathFSLInput, ispc)];
+	else
+		cmdlineOptions = [cmdlineOptions ' -o ' xASL_adm_UnixPath(pathFSLOutput, ispc)];
+		cmdlineOptions = [cmdlineOptions ' -i ' xASL_adm_UnixPath(pathFSLInput, ispc)];
+	end
+	
+	% basil_options is a character array containing CLI args for the Basil/Fabber command
+	BasilOptions = xASL_sub_FSLOptions(pathFSLOptions, x, bUseFabber, PWI, pathFSLInput, dirFSLOutput);
```
MDijsselhof: This has more input options than allowed

Functions/Initialization/xASL_init_FileSystem.m
jan-petr: Michael calls this aBV - arterial blood volume... Though CBV is used for DSC and it isn't incorrect.

Functions/Quantification/xASL_quant_ASL.m
jan-petr: Yes.

Functions/Quantification/xASL_quant_FSL.m
jan-petr: Good catch. Fixed.

Functions/Quantification/xASL_quant_FSL.m
jan-petr: --spatial? Yes, Michael recommends it, so I would do this by default. See the start of the function where I set the defaults - do you agree with this choice? I have also renamed them a bit and I will fix that in the docu later.

Functions/Quantification/xASL_quant_FSL.m
jan-petr: See the start of the function where I define the defaults for x.Q.BASIL.Exch - I've set the default to 'simple' as you did.

Functions/Quantification/xASL_quant_FSL.m
jan-petr: Like in the other comments - defaults are set and printed on screen. See if you agree with that...

Functions/Quantification/xASL_quant_FSL.m
jan-petr: I can't find it either. No idea how it got there. I'm removing it.

Modules/SubModule_ASL/xASL_wrp_Quantify.m
jan-petr: Thanks. I'm keeping it then :)

Modules/SubModule_Population/xASL_wrp_CreatePopulationTemplates.m
jan-petr: Hopefully fine now...

jan-petr (commented):
Answering most comments. But I still need to check units for ABV and the correct quantificaiton.

Functions/Quantification/xASL_quant_FSL.m
```diff
@@ -0,0 +1,569 @@
...
+	fprintf(FIDoptionFile, '# Basil options written by ExploreASL\n');
+end
+
+%% 2. Basic model and tissue parameters
+% Basic model options
+if bUseFabber
+	fprintf(FIDoptionFile, '--method=vb\n');
+	fprintf(FIDoptionFile, '--model=asl_multite\n');
+	fprintf(FIDoptionFile, '--infertexch\n'); % Fit Tex
+	fprintf(FIDoptionFile, '--inferitt\n');   % Fit ATT
```
MDijsselhof, line 280: is this correct?

Modules/SubModule_Population/xASL_wrp_GetROIstatistics.m
jan-petr: Putting %, but we need to check the output to see that the units are really % and not ratio or ml/100g

Functions/Quantification/xASL_quant_FSL.m
jan-petr: that's what Bea got from Amnah. And her output really provides Tex and ATT, so I guess that these are the correct names in FABBER. ButBea should of course confirm.

BeatrizPadrela:
BASIL is working perfectly, so it is running for multiPLD (ASL_2 for DEBBIE)

image

BeatrizPadrela:
Okay, so between step 3 and step4 (where we do spatial regularization) the maps become weird, the FTISS output from FSL (used later on to obtain CBF.nii by dividing it by M0) becomes weird:

[step3]
WhatsApp Image 2023-04-07 at 11 47 32

[step4]
WhatsApp Image 2023-04-07 at 11 47 57


But this doesn't mean there's some kind of error, I think that all is working very well, maybe we just need to turn off the bSpai

Functions/Quantification/xASL_quant_FSL.m
MDijsselhof: @BeatrizPadrela ?

MDijsselhof (approved):

BeatrizPadrela:
Okay, for fabber quantification, it was giving this error:
Screenshot 2023-04-11 at 13 37 39

So, comparing to an older branch (#927), I realised that the fabber command line should be as simple as 'fabber_asl -@ txt_path ' and then all of the options should be inside the txt, including the output path and the PWI input path as well. I've now added a new input to the xASL_sub_FSLOptions:
image

... so that we can have these two commands inside the txt file:
image

Additionally, since we don't need nothing else than this orange command line, we can define these cmdlineOptions (in green) only for non-Fabber quantification:
image

PR #1369, opened 2023-04-05
#1318 Fix session count

### LINKED ISSUE #1318: Correctly convert session 2 in DCM2BIDS
### Description

Flavor Philips_PCASL_2DEPI_5.1.7.2_LookLocker_1has only a single sequence, that should be listed as sequence 2, but it is imported in BIDS as run_1
Basically, it correctly identifies sessions 1 and 2, but when session 1 is missing, then it gives session 1 even to the one that should be 2. And with multiple-context studypar, it also selects the wrong context.

### Tasks

- [x] Use the session string instead of session index number
- [x] Test in all flavors

### Release notes

Fix session import with missing sessions.

jan-petr:
### Linked issue

Close #1318

### DIFF
--- a/Functions/Import/xASL_imp_NII2BIDS_RunPerf.m
+++ b/Functions/Import/xASL_imp_NII2BIDS_RunPerf.m
@@ -30,13 +30,13 @@ function xASL_imp_NII2BIDS_RunPerf(imPar, bidsPar, studyPar, subjectSessionLabel
 % EXAMPLE:     xASL_imp_NII2BIDS_RunPerf(imPar, bidsPar, studyPar, subjectSessionLabel, inSessionPath, outSessionPath, listRuns, iRun);
 %
 % __________________________________
-% Copyright 2015-2022 ExploreASL
+% Copyright 2015-2023 ExploreASL
 
     %% 1. Define the pathnames
-	if (length(listRuns)>1) || (str2num(listRuns{1}(5))>1)
+	if (length(listRuns)>1) || (str2num(listRuns{iRun}(5:end))>1)
 		aslLegacyLabel = 'ASL4D';
 		subjectSessionLabel = ['sub-' subjectSessionLabel];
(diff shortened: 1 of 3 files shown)


Functions/Import/xASL_imp_NII2BIDS_RunPerf.m
```diff
@@ -30,13 +30,13 @@ function xASL_imp_NII2BIDS_RunPerf(imPar, bidsPar, studyPar, subjectSessionLabel
 % EXAMPLE:     xASL_imp_NII2BIDS_RunPerf(imPar, bidsPar, studyPar, subjectSessionLabel, inSessionPath, outSessionPath, listRuns, iRun);
 %
 % __________________________________
-% Copyright 2015-2022 ExploreASL
+% Copyright 2015-2023 ExploreASL
 
     %% 1. Define the pathnames
-	if (length(listRuns)>1) || (str2num(listRuns{1}(5))>1)
+	if (length(listRuns)>1) || (str2num(listRuns{iRun}(5:end))>1)
```
HenkMutsaerts: you can remove the ( ) || ( ), these are brackets you don't need in Matlab

HenkMutsaerts: should it be xASL_str2num?

HenkMutsaerts: This could be improved by using regexps, but if it works I'm fine of course :)

HenkMutsaerts (approved):

Functions/Import/xASL_imp_NII2BIDS_RunPerf.m
jan-petr: Fixed brackets and xASL_str2num. Keeping the fixed length as the keyword is "ASL_something" for sure.

PR #1383, opened 2023-04-13
Bug #1331 g eflavor cbf

### LINKED ISSUE #1331: GE flavor: CBF included in volumes
### Description
This relates to the new flavor GE_PCASL_3Dspiral_DV25.0_1.5T_1.

This is the same DV25.0 product sequence flavor as we see all the time, but now acquired at 1.5T and the three volumes are concatenated as [CBF deltaM M0scan]. In the dcm2NII stage, this will issue a LabelingDuration conflict warning, and it then stores: `"LabelingDuration":0.00145,`. Both `PostLabelDelay` and `InversionTime` are detected but removed in our nii2BIDS. 

Separate the [CBF] and [deltaM M0scan] in the DICOM stage fixes most issues, but the PostLabelingDelay is still not detected. And we should be able to handle a combination of [CBF deltaM M0scan], at least per BIDS.

NOTE FOR HENK: It works correctly now, though there are some smaller errors and warnings -> as noted above. I have added those to the list of minor flavor issues to fix. Added to #342 

>> Warning: Difference in field LabelingDuration between merged JSONs 
> In xASL_bids_MergeNifti>xASL_bids_MergeNifti_Merge (line 636)
  In xASL_bids_MergeNifti>xASL_bids_MergeNifti_GEASLFiles (line 258)
  In xASL_bids_MergeNifti (line 78)
  In xASL_imp_DCM2NII_Subject_SortASLVolumes (line 139)
  In xASL_imp_DCM2NII_ConvertScan (line 181)
  In xASL_wrp_DCM2NII_Subject (line 126)
  In xASL_wrp_DCM2NII (line 45)
  In xASL_module_Import (line 149)
  In xASL_init_Iteration>runIteration (line 382)
  In xASL_init_Iteration (line 125)
  In ExploreASL_Import (line 50)
  In ExploreASL (line 91) 

### Tasks

- [x] For the notes below, please load the ExploreASL branch belonging to this issue and load the flavor [#1294](https://github.com/ExploreASL/FlavorDatabase/tree/flavor-%231294_GE_3volumes)
- [x] Is the PLD automatically removed?
- [x] Correctly detect the `aslcontext.tsv`, now it detects [deltam deltam deltam] but this should be [cbf deltam m0scan]
- [x] If I specify the ASLcontext in `studyPar.json`, the "DICOM value" [deltam deltam deltam] will overwrite my value in `xASL_bids_BIDSifyASLJSON`. The `studyPar.json` values should always have preference over DICOM values (in general)? We only add stuff to `studyPar.json`if we are really sure that the parameters are missing or wrong in the DICOM. This should be fixed.
BTW: BIDS2Legacy works if I manually adapt the `ASLContext.tsv` in the rawdata.
- [x] If we have CBF in ASL when splitting ASL and we also have deltam + m0scan, or also control+label, then we treat CBF as if it would have been a dummy scan in `BIDS2Legacy`
- [x] The dummy scan position option is not working correctly. E.g., for `FlavorDatabase/Siemens_PCASL_multi_VE11C_1/rawdataReference/sub-101/ses-1/perf/sub-101_ses-1_run-1_asl.json` this is done incorrectly, as `DummyScanPositionInASL4D` is added to `studyPar.json` instead of `dataPar.json`, and thus ends up in the ASL sidecar (which is not according to BIDS). Moved this to #1062 

### Release notes

Remove cbf as dummy when deltam/control/label are present in the timeseries.

jan-petr:
### Linked issue

Close #1331

### DIFF
--- a/Functions/BIDS/xASL_bids_Dicom2JSON.m
+++ b/Functions/BIDS/xASL_bids_Dicom2JSON.m
@@ -132,6 +132,20 @@ function [parms, pathDcmDictOut] = xASL_bids_Dicom2JSON(imPar, pathIn, pathJSON,
 			else
 				seriesNumberList(iJSON) = 0;
 			end
+
+			imageTypeList{iJSON} = '';
+			if isfield(tmpJSON, 'ImageType')
+				if iscell(tmpJSON.ImageType)
+					for iIT = 1:length(tmpJSON.ImageType)
+						imageTypeList{iJSON} = [imageTypeList{iJSON} tmpJSON.ImageType{iIT}];
+					end
+				else
+					imageTypeList{iJSON} = tmpJSON.ImageType;
+				end
+				imageTypeList{iJSON} = regexprep(imageTypeList{iJSON},'[^a-zA-Z]','');
+			end
+
+
 		end
 		% Create a cell of parms
 		parms{iJSON} = struct();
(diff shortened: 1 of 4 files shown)


HenkMutsaerts (changes requested):
* I don't like the `xASL_sub_XXX` see discussion in group Whatsapp
* If I understand correctly, you also managed the import such that it creates the correct BIDS with `aslcontext` `[cbf deltam deltam]`?
* And you added the flavor to the flavor database?
* the release information is incomplete, you also managed the import right?

jan-petr:
- In a release, we do not comment on all flavors if they are not special. So commenting on associated code changes is enough.
- Flavor indeed added and it manages aslContext automatically.
- We can open a discussion about sub-functions, but not here - we have been implementing that also outside of this branch.

HenkMutsaerts (approved):
I did a last commit to improve readability, while I was checking `xASL_io_SplitsASL`. Also, we need to ensure that all vendor-specific changes are in specific vendor sections; if you go on a very long sabbatical we should still be able to easily read all this, and others should be able to (think about Siemens-specific discussions with Kamen, Chris Rorden, etc).
Otherwise, looks very nice of course

PR #1389, opened 2023-04-14
Bug #1388 import flavors

### LINKED ISSUE #1388: Token order error in flavors import
### Description

An error crashing several flavors probably related to merging some NII files.

The issues was the number of tokens in token order having to match the number of brackets in folderHierarchy. In regexp, ( ) is not reserved for tokens only and is also used for grouping. Therefore the number of tokens in tokenOrder must by equal or smaller than number of brackets, but not necessarily equal.

HENK: it should in our use be reserved for tokens only, otherwise it becomes tricky to interpret which brackets are tokens, and which brackets are used for grouping.

Either we switch the type of brackets, or we use the grouping bracket in the token definition, not in `folderHierarchy`. Let's discuss. But please provide the examples here that crash so I can check as well.

-> ALSO: we need to copy (can be empty) folders from the clinical scan sessions that we are not interested in, to ensure also that our `sourceStructure.json` works in those cases.

### Tasks

- [x] Fix bugs in previous commits
- [x] Add missing tags to the SPM version of DICOM read
- [x] Allow unused tokens, but print a warning
- [x] Allow token reusing, but print a warning
- [x] Allow missing tokenSessionAlias in sourcestructure.json

### Release notes

Allow unused and reused tokens in sourcestructure.

jan-petr:
### Linked issue

Close #1388

### DIFF
--- a/Functions/Import/xASL_imp_CheckImportSettings.m
+++ b/Functions/Import/xASL_imp_CheckImportSettings.m
@@ -71,11 +71,11 @@ function [x] = xASL_imp_CheckImportSettings(x)
 	fprintf('%s\n', [num2str(nTokens) ' tokens detected in tokenOrdering']);
 
 	if nLeftBrackets < nTokens
-		error('The number of tokens in folderHierarchy defined by the () symbol should be equal to or higher than the length of the tokenOrdering vector.');
+		error('The number of tokens in folderHierarchy and tokenOrdering should be equal. Note that the number of captured groups () should thus be equal to or higher than the number of tokens.');
 	elseif nLeftBrackets > nTokens
-		warning('The number of tokens in folderHierarchy defined by the () symbol is higher than the length of the tokenOrdering vector. This is allowed, but make sure that this is what you want.');
+		warning('The number of captured groups in folderHierarchy, defined by the () symbols, is higher than the number of tokens defined in tokenOrdering. Ensure that this is correct.');
 	elseif sum(x.modules.import.imPar.tokenOrdering>0) ~= nTokens
-		warning('Not all tokens from folderHierarchy appear to be used in the tokenOrdering vector. This is allowed, but make sure that this is what you want.')
+		warning('Not all captured groups in folderHierarchy are used as a token in tokenOrdering. Ensure that this is correct.')
 	end
 
 	%% Manage .nii vs .nii.gz extensions
--- a/Functions/Import/xASL_imp_CheckImportSettings.m
+++ b/Functions/Import/xASL_imp_CheckImportSettings.m
@@ -64,7 +64,7 @@ function [x] = xASL_imp_CheckImportSettings(x)
 		nRightBrackets = nRightBrackets + numel(rightBracket);
 	end
 
-	nTokens = sum(x.modules.import.imPar.tokenOrdering>0);
+	nTokens = max(x.modules.import.imPar.tokenOrdering);
 
 	% Report this
 	fprintf('%s\n', [num2str(nLeftBrackets) ' tokens detected in folderHierarchy']);
--- a/Functions/Import/xASL_imp_CheckImportSettings.m
+++ b/Functions/Import/xASL_imp_CheckImportSettings.m
@@ -70,8 +70,12 @@ function [x] = xASL_imp_CheckImportSettings(x)
 	fprintf('%s\n', [num2str(nLeftBrackets) ' tokens detected in folderHierarchy']);
 	fprintf('%s\n', [num2str(nTokens) ' tokens detected in tokenOrdering']);
 
-	if nLeftBrackets<nTokens
-		error('Number of tokens in folderHierarchy defined by () should be equal or higher than the length of tokenOrdering vector');
+	if nLeftBrackets < nTokens
+		error('The number of tokens in folderHierarchy defined by the () symbol should be equal to or higher than the length of the tokenOrdering vector.');
+	elseif nLeftBrackets > nTokens
+		warning('The number of tokens in folderHierarchy defined by the () symbol is higher than the length of the tokenOrdering vector. This is allowed, but make sure that this is what you want.');
+	elseif sum(x.modules.import.imPar.tokenOrdering>0) ~= nTokens
+		warning('Not all tokens from folderHierarchy appear to be used in the tokenOrdering vector. This is allowed, but make sure that this is what you want.')
 	end
 
 	%% Manage .nii vs .nii.gz extensions
--- a/Functions/Import/xASL_imp_CheckImportSettings.m
+++ b/Functions/Import/xASL_imp_CheckImportSettings.m
@@ -10,12 +10,14 @@ function [x] = xASL_imp_CheckImportSettings(x)
 %   x        - Struct containing pipeline environment parameters, useful when only initializing ExploreASL/debugging
 %
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
-% DESCRIPTION:    Basic import checks before execution.
+% DESCRIPTION:    Basic import checks before execution of dcm2nii
 %
-% - Check permissions for DCM2NII
-% - Get correct DCMNII version
-% - Define DCM Extension Filter
-% - Set default for skip subjects option
+% 1. Check permissions for DCM2NII
+% 2. Check correct DCMNII version
+% 3. Check DCM Extension Filter
+% 4. Check skip subjects option
+% 5. Check token definitions
+% 6. Manage .nii vs .nii.gz extensions
 %
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
 % EXAMPLE:        n/a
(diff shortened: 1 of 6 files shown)


Functions/Import/xASL_imp_CheckImportSettings.m
```diff
@@ -70,9 +70,8 @@
 	fprintf('%s\n', [num2str(nLeftBrackets) ' tokens detected in folderHierarchy']);
 	fprintf('%s\n', [num2str(nTokens) ' tokens detected in tokenOrdering']);
 
-	if nLeftBrackets~=nTokens
-		fprintf('Note that the () symbols are reserved to define tokens in folderHierarchy!');
-		error('Number of tokens of folderHierarchy and tokenOrdering should be the same');
+	if nLeftBrackets<nTokens
+		error('Number of tokens in folderHierarchy defined by () should be equal or higher than the length of tokenOrdering vector');
```
HenkMutsaerts: I see your point, but I want to see the examples to ensure that we cannot reinforce () == number of tokens. Because this would make it much easier and clearer for the user.

Also, you can practice the use of articles, which article is missing in your error message? ;)

Functions/Import/xASL_imp_CheckImportSettings.m
```diff
@@ -64,7 +64,7 @@
 		nRightBrackets = nRightBrackets + numel(rightBracket);
 	end
 
-	nTokens = sum(x.modules.import.imPar.tokenOrdering>0);
+	nTokens = max(x.modules.import.imPar.tokenOrdering);
```
HenkMutsaerts: This is incorrect, if I'm correct ;)

`tokenOrdering` is something like [1 0 5 3] (in the example that subject is the first folder level, there is no visit, the session is the 5th folder level (with some dummy folders in between that we want to keep as we want to change the `sourcedata` as minimally as possible), and scan as third folder layer) so for the number of tokens you need the sum of non-zero tokens. Note that the zeros should be appended before this part if they were omitted by the user (e.g., [1 5 3]

HenkMutsaerts (changes requested):

Functions/Import/xASL_imp_CheckImportSettings.m
jan-petr: Including those articles missing in what you wrote or excluding?

Functions/Import/xASL_imp_CheckImportSettings.m
jan-petr: No you're not correct :)
sum([1 0 5 3]>0) will return 3. But we need tokens 1, 5, and 3. So we need 5 tokens defined so that we skip 2 and 4. 

We use 3 tokens, but tokens 1 3 5, so maximum token number 5. So we need 5 () defined in folderhierarchy, not 3.

HenkMutsaerts: I would reserve tokens for those that have a meaning, they are a "token" for something. () without a meaning is not a token, but denotes a capturing group.

So () == brackets
tokens == defined in the aliases
capturing groups == all brackets (can be a group or a group+token)

Functions/Import/xASL_imp_CheckImportSettings.m
```diff
@@ -64,15 +64,18 @@
...
 
 	% Report this
 	fprintf('%s\n', [num2str(nLeftBrackets) ' tokens detected in folderHierarchy']);
 	fprintf('%s\n', [num2str(nTokens) ' tokens detected in tokenOrdering']);
 
-	if nLeftBrackets~=nTokens
-		fprintf('Note that the () symbols are reserved to define tokens in folderHierarchy!');
-		error('Number of tokens of folderHierarchy and tokenOrdering should be the same');
+	if nLeftBrackets < nTokens
+		error('The number of tokens in folderHierarchy defined by the () symbol should be equal to or higher than the length of the tokenOrdering vector.');
```
HenkMutsaerts: "The number of tokens in folderHierarchy and  tokenOrdering should be equal"
 ("length of vector" is too technical for the user).
AND
"Note that the number of captured groups () should thus be equal or more than the number of tokens."

Functions/Import/xASL_imp_CheckImportSettings.m
```diff
@@ -64,15 +64,18 @@
...
+	elseif nLeftBrackets > nTokens
+		warning('The number of tokens in folderHierarchy defined by the () symbol is higher than the length of the tokenOrdering vector. This is allowed, but make sure that this is what you want.');
```
HenkMutsaerts: "The number of captured groups in folderHierarchy, defined by the () symbols, is higher than the number of tokens defined in tokenOrdering. Ensure that this is correct."

Functions/Import/xASL_imp_CheckImportSettings.m
```diff
@@ -64,15 +64,18 @@
...
+	elseif sum(x.modules.import.imPar.tokenOrdering>0) ~= nTokens
+		warning('Not all tokens from folderHierarchy appear to be used in the tokenOrdering vector. This is allowed, but make sure that this is what you want.')
```
HenkMutsaerts: ('Not all captured groups in folderHierarchy are used as a token in tokenOrdering. Ensure that this is correct')"

Functions/InputOutput/xASL_io_DcmtkRead.m
```diff
@@ -239,19 +235,55 @@
 	end
 end
 
+if isfield(header, 'Private_0043_1083')
```
HenkMutsaerts: Shouldn't we do the captured groups vs tokens in a separate issue from these here?

Functions/InputOutput/xASL_io_DcmtkRead.m
```diff
@@ -239,19 +235,55 @@
...
+	header.AssetRFactor = header.Private_0043_1083;
```
HenkMutsaerts: Note that this is a GE-specific parameter. I would move the vendor-specific parameters into vendor-specific sections for readability.

Functions/InputOutput/xASL_io_DcmtkRead.m
```diff
@@ -239,19 +235,55 @@
...
+
+if isfield(header, 'Private_2001_1013')
+	header.MRSeriesEPIFactor = header.Private_2001_1013;
+end
+
+if isfield(header, 'Private_0019_1028')
+	header.BandwidthPerPixelPhaseEncode = header.Private_0019_1028;
+end
+
+if isfield(header, 'Private_0019_109c')
```
HenkMutsaerts, line 266: GE

Functions/InputOutput/xASL_io_DcmtkRead.m
```diff
@@ -239,19 +235,55 @@
...
+	header.GELabelingType = header.Private_0019_109c;
```
HenkMutsaerts, line 267: GE

Functions/InputOutput/xASL_io_DcmtkRead.m
```diff
@@ -239,19 +235,55 @@
...
+end
+
+if isfield(header, 'Private_0043_10a5')
+	header.GELabelingDuration = header.Private_0043_10a5;
+end
+
+if isfield(header, 'Private_0029_1020')
```
HenkMutsaerts, line 275: Siemens

Functions/InputOutput/xASL_io_DcmtkRead.m
```diff
@@ -239,19 +235,55 @@
...
+	header.PhoenixProtocol = header.Private_0029_1020;
+end
+
+if isfield(header, 'Private_0019_1029')
```
HenkMutsaerts, line 279: Siemens

Functions/InputOutput/xASL_io_DcmtkRead.m
```diff
@@ -239,19 +235,55 @@
...
+	header.SiemensSliceTime = header.Private_0019_1029;
+end
+
 % Keep only those fields below...
 listFieldsKeep = {'RepetitionTime', 'EchoTime', 'RescaleSlope', 'RescaleIntercept',...
         'NumberOfTemporalPositions', 'NumberOfAverages', 'AcquisitionTime',...
-        'PixelData', 'MediaStorageSOPClassUID', 'Manufacturer',  'MRScaleSlope',...
+        'PixelData', 'MediaStorageSOPClassUID', 'Manufacturer', 'ManufacturersModelName', 'MRScaleSlope',...
```
HenkMutsaerts, line 287: Philips

HenkMutsaerts (changes requested):
1) we need to correctly mention captured groups vs tokens in the comments and warnings.
2) all DICOM parameters from different vendors are mixed, these should be nicely into sections per vendor

HenkMutsaerts (approved):
I did a few extra clarifications in each instance where we have `folderHierarchy`, do you agree? This also creates extra warnings for Bea for e.g., using sessions double, but that is what we agreed upon (allowing but warning the user).

BTW: for your "revamp"
Should `xASL_imp_Initialize` line 146 -> token checks go to `xASL_imp_CheckImportSettings`?
Should `xASL_imp_ReadSourceData` line 74 also be moved to `xASL_imp_CheckImportSettings`?

And BTW: what do you think of my commit titles, is it not easier/faster to review (and especially if we would review it again 1 year later ;)

Functions/Import/xASL_imp_CheckImportSettings.m
```diff
@@ -64,21 +73,66 @@
...
+    fprintf('%s\n', [xASL_num2str(nGroupsTokenOrdering) ' captured groups () defined in tokenOrdering']);
+	fprintf('%s\n', [xASL_num2str(nTokensTokenOrdering) ' tokens () defined in tokenOrdering']);
+
+    fprintf('%s\n', [xASL_num2str(nVisitAliases) ' visits defined in tokenVisitAliases']);
+    fprintf('%s\n', [xASL_num2str(nSessionAliases) ' sessions defined in tokenSessionAliases: ']);
+    fprintf('%s\n', [xASL_num2str(nScanAliases) ' scans defined in tokenSessionAliases']);
+    
+    if nGroupsFolderHierarchy < nTokensTokenOrdering
+		error('The number of captured groups in folderHierarchy should >= the number of tokens in tokenOrdering');
+    elseif nGroupsTokenOrdering < nTokensTokenOrdering
```
jan-petr: [1 0 0 3] has nGroups 3 and nTokens 2, so it will create an error. But it should produce a warning.
 
So `elseif nGroupsTokenOrdering < nTokensTokenOrdering` condition should be removed

BeatrizPadrela:
All good!!

PR #1399, opened 2023-04-24
#1386 xASL_quant_FSL: Fix DEBBIE TIs

### LINKED ISSUE #1386: DEBBIE parameter TI
### Description

For FABBER DEBBIE, we provide TIs (time from labeling to readout for PASL). Sometimes for PCASL , TIs are also provided instead of PLDs but TI=PLD+Tau. So I don't understand why we provide TI=PLD for FABBER for DEBBIE. Can @BeatrizPadrela and @amahroo explain that this is correct.

Is it OK that TI=PLD in the code, or should this be changed to TI=PLD+Tau in the FABBER branch of xASL_quant_FSL.

HENK: In general we should use BIDS and OSIPI TF4.1 terminology. Amnah and Matthias specifically use `TI` as `inflow time` if I'm not mistaken in DEBBIE. Is inflow time really PLD+Tau? It should be PLD+Tau for the first labeled spin (in PCASL), for the other spins the inflow time is shorter. So it could also be PLD+0.5Tau or something similar (it is an integral in reality)?

### Tasks

- [x] Jan: In all flavors: update Labeling duration of all HAD8 to 0.4s.
- [x] Bea: In all datasets: update Labeling duration of all HAD8 to 0.4s.
- [x] JanL: PLDs are 0.5, 1.5, 2.5s for HAD4. TIs should be 1.5, 2.5, 3.5s for HAD4. TIs [1, 1.4, 1.8, 2.2, ... 3.4s] for HAD8 - fix code for BASIL options.
- [x] Bea: Test that it really works.

### Release notes

Fix TIs for DEBBIE quantification.

jan-petr:
### Linked issue

Close #1386

### DIFF
--- a/Functions/Quantification/xASL_quant_FSL.m
+++ b/Functions/Quantification/xASL_quant_FSL.m
@@ -377,7 +377,7 @@ switch lower(x.Q.LabelingType)
 
 			% Plotting the values into the doc (PLD=ti, LD=tau)
 			for iPLD = 1:length(PLDs)
-				fprintf(FIDoptionFile, '--ti%d=%.2f\n', iPLD, PLDs(iPLD));
+				fprintf(FIDoptionFile, '--ti%d=%.2f\n', iPLD, PLDs(iPLD) + LabDurs(iPLD));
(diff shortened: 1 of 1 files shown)


BeatrizPadrela (approved):
ok!

PR #1400, opened 2023-04-25
Bug #1366 import look locker

### LINKED ISSUE #1366: Fix import for Philips Look-Locker
### Description

Two things are not imported correctly in Philips Look-Locker.

### Tasks

- [x] Fix import of Flavors Philips_PASL_2DEPI_3.2.2.0_LookLocker_1 and Philips_PASL_2DEPI_5.1.7.2_LookLocker_1. These two things below are not fixed in flavors.
- [x] TR should be not taken as the shorter TR from DICOM - as this is TR for each volume and not the Cycle duration (ms) =		4150; (I put it to studyPar.json, but it should be read automatically)
- [x] TR for M0 are not correct, again, not 300 in total, but 300 ms is each step. Moreover, M0 seems to be an inversion recovery, so that has to be considered as well and somehow saved in BIDS (this is currently ignored completely).

### Release notes

Fix TR import for Philips Look-Locker ASL.

jan-petr:
### Linked issue

Close #1366

### DIFF
--- a/Functions/BIDS/xASL_bids_BIDSifyASLJSON.m
+++ b/Functions/BIDS/xASL_bids_BIDSifyASLJSON.m
@@ -27,15 +27,16 @@ function jsonOut = xASL_bids_BIDSifyASLJSON(jsonIn, studyPar, headerASL)
 % 8. Merge data from the Phoenix protocol
 % 9. Background suppression check
 % 10. SliceTiming check
-% 11. Check if length of vector fields match the number of volumes
-% 12. Reformat ASLcontext field
-% 13. Verify TotalAcquiredPairs against ASLContext
-% 14. Final field check
+% 11. Check for Look-Locker TR
+% 12. Check if length of vector fields match the number of volumes
+% 13. Reformat ASLcontext field
+% 14. Verify TotalAcquiredPairs against ASLContext
+% 15. Final field check
 %
 % EXAMPLE: n/a
 %
 % __________________________________
-% Copyright (c) 2015-2022 ExploreASL
+% Copyright (c) 2015-2023 ExploreASL
 
 %% 0. Admin
 if nargin < 1 || isempty(jsonIn)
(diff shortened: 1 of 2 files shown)


Functions/BIDS/xASL_bids_BIDSifyASLJSON.m
```diff
@@ -399,7 +400,39 @@
 	end
 end
 
-%% 11. Check if length of vector fields match the number of volumes
+%% 11. Check for Look-Locker TR
```
HenkMutsaerts, line 403: Please make up your mind if this is mainly Look-Locker specific (i.e., would other vendors with Look-Locker have the same problem?) or if this is Philips specific (this is not a general Look-Locker problem, but a problem on how Philips saves the data). In the latter case, please move this to the Philips section.

Also, please explain what you're doing here in a story under the header (the independent comments are great, but don't tell the story on what goes wrong)

Functions/BIDS/xASL_bids_BIDSifyASLJSON.m
```diff
@@ -399,7 +400,39 @@
...
+							jsonOut.RepetitionTimePreparation = (jsonOut.PostLabelingDelay(2)-jsonOut.PostLabelingDelay(1)) + max(jsonOut.PostLabelingDelay);
+						end
+					end
+				end
+			end
+		end
+	end
+end
+
+%% 12. Check if length of vector fields match the number of volumes
```
HenkMutsaerts, line 439: Can you make the error in line 441 clearer, e.g.,:
```
fprintf('%s\n', ['The length of the BIDS parameter ' XXXX ' doesn't fit with the number of NIfTI volumes: ' num2str(YYY)]);
```

Functions/BIDS/xASL_bids_BIDSifyASLJSON.m
```diff
@@ -508,7 +541,7 @@
 	end
 end
 	
-%% 13. Verify TotalAcquiredPairs against ASLContext
+%% 14. Verify TotalAcquiredPairs against ASLContext
```
HenkMutsaerts, line 548: `TotalAcquiredPairs` doesn't have to be equal to the number of volumes right? E.g., GE and Siemens average some of the volumes.

Functions/BIDS/xASL_bids_BIDSifyM0.m
```diff
@@ -78,6 +78,37 @@
 
 jsonOut.RepetitionTimePreparation = jsonOut.RepetitionTime;
 
+% Look-Locker acquisition for Philips has often TR given as the length of a single readout, not as the entire cycle
```
HenkMutsaerts: Is this the exact same addition as above? Then please move this to a separate function `xASL_BIDS_BIDSifyLookLocker`

HenkMutsaerts (changes requested):
Just some clarifications and code structure

Functions/BIDS/xASL_bids_BIDSifyASLJSON.m
jan-petr: Yes - will improve comments.
I have no idea how other vendors behave. And there's no Philips section. So I'll keep it here - sorry, but no point in overorganizing things. I'll keep it where it functionally belongs but of course clarify in comments that this is Philips specific. Even for vendor-specific things like this one, we have to organize things by field type and not by vendor, because there might be cases of cross-dependency of fields. E.g. for this action, you need to have PLDs already sorted out. So that's why it's better to order things by tag type, and within each task have vendor subsections. Rather than the other way around :)

So I understand your suggestion, but I have of course already though about it ;)

Functions/BIDS/xASL_bids_BIDSifyASLJSON.m
jan-petr: Sure. Done.

Functions/BIDS/xASL_bids_BIDSifyASLJSON.m
jan-petr: That is correct. That's why we say "verify" - this doesn't mean they have to be equal. I have expanded the comment to explain this.

Functions/BIDS/xASL_bids_BIDSifyM0.m
jan-petr: No. Not the same addition. The inputs and outcomes for M0 differ quite a lot. Only the first 3 IFs are equal. A separate function wouldn't work. I have updated comments though in the same way as I did for ASL.

HenkMutsaerts (approved):

PR #1404, opened 2023-04-29
Closes Feature #1387 masking basil

### LINKED ISSUE #1387: Simple masking in BASIL
### Description

Check if simple masking speeds up BASIL.

HENK: It should, but I would only do this if needed. There is a big benefit for "clinical" visual QC-ing for not masking: specific artifacts and noise etc can easier be appreciated when you don't mask, they may not be limited to the brain. And it is challenging to guarantee that masking is always perfect. So in practice it can be better if we don't need to mask.

IF we want to mask, it would be good to use the existing masking solutions that we use for defining the "analysis masks", or improving them.

In `xASL_wrp_CreateAnalysisMask` we create masks both in native and standard space:

`x.P.Path_FoV`
`x.P.Path_MaskVascular`
`x.P.Path_Pop_MaskSusceptibility` (only in standard space, and based on susceptibility templates)

I propose to add `x.P.Path_TissueMask` in the ASL folder, which would be 

If I'm correct, we still run BASIL both in native and standard space? I would create the following booleans: `bBASIL_nativeSpace` and `bBASIL_standardSpace`, the first one replacing our current BASIL boolean. And we default to the native space BASIL only.

JAN: No standardspace BASIL.

image
This is an example with the normal brainmask used for the analysis masks: `(pvGM+pvWM)>0.5`. But this is not an analysis mask but an image processing mask, where we want to be more inclusive.
image
This is what I would suggest as processing brainmask: `(pvGM+pvWM+pvCSF)>0.1`.

### Tasks

- [x] Add `x.P.Path_BrainMaskProcessing` to `xASL_init_FileSystem`
- [x] Consistency: `Path_Pop` -> `Pop_Path`
- [x] Save `x.P.Path_BrainMaskProcessing` in `xASL_wrp_CreateAnalysisMask`
- [x] Add the brain mask to `xASL_quant_FSL` (@MDijsselhof)
       Should this be on line 77 (where it says FIXME)?
       Should we then set the voxels outside the mask to zeroes? (i.e., does BASIL and FABBER use implicit masking?)
       This would be:
       Mathijs:
       For **BASIL**:

Image

      we should add this in `xASL_quant_FSL`, sub-function `xASL_sub_FSLOptions`, for example somewhere from line 461 
       `% 5. Extra BASIL fitting parameters onwards. We only have to put in `-m MaskPath` to FSL options for BASIL. We could also 
       add this at line 276 somewhere if we want it as default.

      For **FABBER**:
      
Image

      We add this to line 271 somewhere, as: `--mask=MaskPath` if we want it as default. 

```
BrainMask = xASL_io_Nifti2Im(x.P.Path_BrainMaskProcessing);
PWI(isnan(PWI)) = 0;
```
- [x] Also note that there are still unused options at the end of `xASL_quant_FSL` (for @jan-petr  to check) JAN: You mean those that are commented out? These might be used in the future. I'm not sure who added them anyway. But if nobody wants then, we can remove them. Bea? Mathijs?
HENK: they were added by Martin Craig I believe

Further tasks added by Jan:
- [x] Wait till #1323 is merged and then recreate this branch
- [x] Remove the standard space option and processing mask
- [x] Do not null any voxels in PWI (which should have been PWI4D anyway)
- [x] Pass the mask by default on FABBER
- [x] Pass the mask by default on BASIL
- [x] Write an option that allows to turn this masking OFF, make defaults for this, add to manual

### Release notes

Add default masking to FSL quantification.

HenkMutsaerts:
### Linked issue

Closes #1387

### How to test

Before testing starts - we need to complete all tasks Jan added to the issue.

Can you run BASIL & FABBER and see if:
for multi-PLD BASIL (@MDijsselhof):
- [x] it doesn't crash
- [x] output images still look good
- [x] processing is faster

for multi-TE FABBER (@BeatrizPadrela):
- [x] it doesn't crash
- [x] output images still look good
- [x] processing is faster

### Comments

### DIFF
--- a/Functions/Quantification/xASL_quant_FSL.m
+++ b/Functions/Quantification/xASL_quant_FSL.m
@@ -1,4 +1,4 @@
-% function [CBF_nocalib, ATT_map, ABV_map, Tex_map, resultFSL] = xASL_quant_FSL(PWI, x)
+function [CBF_nocalib, ATT_map, ABV_map, Tex_map, resultFSL] = xASL_quant_FSL(PWI, x)
 %xASL_quant_FSL Perform quantification using FSL BASIL/FABBER
 %
 % FORMAT: [CBF_nocalib, ATT_map, ABV_map, Tex_map, resultFSL] = xASL_quant_FSL(PWI, x)
--- a/Functions/Quantification/xASL_quant_FSL.m
+++ b/Functions/Quantification/xASL_quant_FSL.m
@@ -222,6 +222,11 @@ if ~bUseFabber
 		x.Q.BASIL = [];
 	end
 
+	if ~isfield(x.Q.BASIL,'bMasking') || isempty(x.Q.BASIL.bMasking)
+		fprintf('BASIL: Setting default option bMasking = true\n');
+		x.Q.BASIL.bMasking = true;
+	end
+
 	if ~isfield(x.Q.BASIL,'bSpatial') || isempty(x.Q.BASIL.bSpatial)
 		fprintf('BASIL: Setting default option bSpatial = false\n');
 		x.Q.BASIL.bSpatial = false;
@@ -278,6 +283,23 @@ else
 	% Path to input and output
 	FSLOptions = [FSLOptions ' -o ' xASL_adm_UnixPath(pathFSLOutput, ispc)];
 	FSLOptions = [FSLOptions ' -i ' xASL_adm_UnixPath(pathFSLInput, ispc)];
+end
+
+% Define masking
+if x.Q.BASIL.bMasking
+	% Check for uninitialized Mask variable or file
+	if ~isfield(x.P, 'Path_BrainMaskProcessing')
+		warning('BASIL masking set to TRUE, but the mask variable x.P.Path_BrainMaskProcessing is not initialized.');
+	elseif ~xASL_exist(x.P.Path_BrainMaskProcessing, 'file')
+		warning('BASIL masking set to TRUE, but the mask is missing: %s\n', x.P.Path_BrainMaskProcessing);
+	else
+		% Add the mask to the options file
+		if bUseFabber
+			fprintf(FIDoptionFile, '--mask=%s\n', xASL_adm_UnixPath(x.P.Path_BrainMaskProcessing, ispc));
+		else
+			FSLOptions = [FSLOptions ' -m ' xASL_adm_UnixPath(x.P.Path_BrainMaskProcessing, ispc)];
+		end
+	end
 end
 
 %% 2. Basic model and tissue parameters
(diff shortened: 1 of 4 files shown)


Functions/Quantification/xASL_quant_FSL.m
```diff
@@ -75,7 +75,13 @@
     
     %% 3. Write the PWI as Nifti file for Basil/Fabber to read as input
     % FIXME would be good to have a brain mask at this point -> PM: if this would be a brainmask as well, we can skip creating a dummy input image here
+    
     PWI(isnan(PWI)) = 0;
+
+    BrainMask = xASL_io_Nifti2Im(x.P.Path_BrainMaskProcessing); % load the brain mask used for processing
+    PWI(BrainMask==0) = 0; % set voxels outside the mask to zero
```
jan-petr: That's an incorrect approach IMHO. BASIL will have a specific input for brain mask. So this has to be passed as such to BASIL without trying to null the values ourselves.

Modules/SubModule_ASL/xASL_wrp_CreateAnalysisMask.m
```diff
@@ -112,6 +115,20 @@ function xASL_wrp_CreateAnalysisMask(x)
...
+
+%% 3C. Save brain mask for image processing (e.g., BASIL)
+% This mask can be used for fitting data, e.g., BASIL, fitting ATT, Tex,
+% etc. that is not useful in extracranial voxels, and beneficial to speed
+% up by only processing intracranial voxels. On the other hand, it can be
+% helpful to also process some voxels outside the brain for inspecting
+% artifacts, and registration errors shouldn't lead to accidentally
+% excluding intracranial voxels. Therefore, we use a rather inclusive mask.
+
+xASL_io_SaveNifti(x.P.Path_PWI, x.P.Path_BrainMaskProcessing, BrainMaskProcessingNativeSpace, 8, false);
```
jan-petr, line 142: Just to be sure that I don't overlook anything. This mask is only pGM+pWM+pCSF, right? Cause we want to keep the vascular masks and everything in.

What about tumors and so on. Are we sure that all irregular things and lesions are included as we want to process all these, just exclude outside of the brain.

Functions/Quantification/xASL_quant_FSL.m
jan-petr: btw: this won't work with PWI4D that is passed to BASIL anyway - so the entire thing here has to be deleted.

jan-petr (changes requested):
I fixed one minor mistake. Make sure that the mask includes also lesions etc. Note that we shouldn't null any PWI values. Instead, a proper mask has to be passed to BASIL. So Mathijs+Bea will need to complete this issue.

Lets continue working on this once Mathijs is done with Look-Locker quantification and once we merge everything. So that we can rebase the commits made so far.

Functions/Quantification/xASL_quant_FSL.m
HenkMutsaerts: I see. I have now commented this implicit masking, and provided an explanation on how to do explicit masking.

Modules/SubModule_ASL/xASL_wrp_CreateAnalysisMask.m
HenkMutsaerts: Yes, you're right. I have now explained all this in the header, is this to your liking?

jan-petr (approved):
OK

MDijsselhof (commented):
Code looks good, processing works and is very quick

PR #1410, opened 2023-05-15
#1408 Updating T2 values of blood and tissue

### LINKED ISSUE #1408: Minor issue - adjust the T2 of tissue and blood for quantification
### Description

The T2 times for tissue and blood are automatically defined by ExploreASL but Amnah suggested that we use these instead:

t2 = 0.085 s
t2b = 0.165 s

Ref: https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3554863/?report=reader#!po=42.5000 (Table 1) --> reference 30 of this paper – which is an ISMRM abstract [https://cds.ismrm.org/ismrm-2003/0131.pd]) – says T2b for 1.5T = 240ms, T2b at 3T = 165ms -> however these values are at 100% O2 level...)

**T2 of blood:**
https://onlinelibrary.wiley.com/doi/pdf/10.1002/mrm.21342, this paper's Table 1 says that the T2 of blood changes a lot with Hct:
Hct = 0.21, T2bl = 175 ms
Hct = 0.44, T2bl = 112 ms
Hct = 0.57, T2bl = 92 ms

--> We **_really_** need to have subject-specific Hct and a T2 mapto get sub-specific values of T2 in GM and WM

**T2 of tissue:**
 https://onlinelibrary.wiley.com/doi/10.1002/jmri.20356 says:
-> at 3T, T2 of GM = 88 ms and T2 of WM = 69 ms; 
-> at 1.5 T, T2 of GM = 99 ms and T2 of WM = 79 ms;

**T2 of tissue and blood:**
Other paper says very different values.. Stanisz GJ, Odrobina EE, Pun J et al. [T1, T2 relaxation and magnetization transfer in tissue at 3T. Magn Reson Med 2005; 54:507-512.](https://mri-q.com/uploads/3/4/5/7/34572113/t1_and_t2_at_1.5_and_3t_stanisz_et_al-2005-magnetic_resonance_in_medicine.pdf)
image

**My conclusion:**
Since we know that T2 of tissue is between 80-90, I htink using 0.085 as Amnah does is reasonable

### Tasks

- [x] Change the default values of T2art and T2star
- [x] Provide a reference for T2=85ms
- [x] Try to find the values to 1.5 and 7T
- [x] Check if it works

(I'm not sure where neither x.Q.T2art nor x.Q.T2 values come from, though - see below)
image

### How to test

Test quantification with Fabber with these values. I've done it and it works.

### Release notes

T2 timings of blood and tissue have been updated

BeatrizPadrela:
### Linked issue

#1408

### DIFF
--- a/Modules/SubModule_ASL/xASL_wrp_Quantify.m
+++ b/Modules/SubModule_ASL/xASL_wrp_Quantify.m
@@ -121,7 +121,10 @@ if ~isfield(x.Q,'T2star') || isempty(x.Q.T2star)
 end
 if ~isfield(x.Q,'T2') || isempty(x.Q.T2)
 	if x.MagneticFieldStrength == 3
-		x.Q.T2 = 85; % in ms - default for 3T (ref Johannes Gregori, JMRI 2013)
+		x.Q.T2 = 85; % in ms - default for 3T (ref Johannes Gregori, JMRI 2013) 88 for frontal GM, 79 for occipital GM (Lu et al, 2005 JMRI)
+		% Hct specific values are in 10.1002/mrm.21342
+	elseif x.MagneticFieldStrength == 1.5
+		x.Q.T2 = 95; % in ms - 99 for frontal GM, 90 for occipital GM (Lu et al, 2005 JMRI).
 	else
 		x.Q.T2 = 85;
 		fprintf('%s\n',['Warning: Unknown T2 for ' num2str(x.MagneticFieldStrength) 'T scanners, using 3T value']);
--- a/Modules/SubModule_ASL/xASL_wrp_Quantify.m
+++ b/Modules/SubModule_ASL/xASL_wrp_Quantify.m
@@ -34,6 +34,12 @@ function xASL_wrp_Quantify(x, PWI_Path, pathOutputCBF, M0Path, SliceGradientPath
 % EXAMPLE: xASL_wrp_Quantify(x);
 %
 % REFERENCES: 
+%     Gregori, Johannes et al. “T2-based arterial spin labeling measurements of blood to tissue water transfer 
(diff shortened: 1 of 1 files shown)


Modules/SubModule_ASL/xASL_wrp_Quantify.m
```diff
@@ -113,9 +113,9 @@ function xASL_wrp_Quantify(x, PWI_Path, pathOutputCBF, M0Path, SliceGradientPath
 end
 if ~isfield(x.Q,'T2') || isempty(x.Q.T2)
 	if x.MagneticFieldStrength == 3
-		x.Q.T2 = 180; % default for 3T (ref Jean Chen, MRM 2009)
+		x.Q.T2 = 85; % in ms - default for 3T (ref Johannes Gregori, JMRI 2013)
```
HenkMutsaerts: Shouldn't we add also the non-3T values, now we're at it?

HenkMutsaerts (changes requested):
Shouldn't we add also the non-3T values, now we're at it?

I would add the 1.5T and 7T values, otherwise this looks good!

Modules/SubModule_ASL/xASL_wrp_Quantify.m
jan-petr: I've added that for T2star and T2blood. 
@BeatrizPadrela can try to locate it for T2-tissue.

jan-petr (changes requested):
We need to add T2 of tissue for 1.5T. And we need to add the correct reference for T2-tissue and explain why it is 85ms and not 80ms.

jan-petr (approved):
OK

PR #1412, opened 2023-05-15
Bug #1411 fix multi tem0

### LINKED ISSUE #1411: Fix multi-TE M0 processing
### Description

For multi-TE M0s, we should take only the first shortest TE to be used for ASL scaling.
### Tasks

- [x] Identify multi-TE are being used.
- [x] Take only the shortest TE and calculate M0 image only from that TE.
- [x] Make sure that in the subsequent calculation, only that TE is used. line 250 in xASL_wrp_Quantify.
- [x] Test that DEBBIE HAD04 works now.
- [x] Take TE closest to ASL - this actually doesn't make sense as ASL and M0 are designed with short TEs. And the only reason to have longer TEs is when mutli-TE is done. So there's no reason to have M0 with multi-TE and having another TE than the shortest to be the closes to ASL. So this option is superfluous as it will always be fulfilled with shortest TE.
- [x] Make a warning that TEs are being stripped
- [x] Check that DEBBIE HAD04 CBF looks good.

### Release notes

Fix multi-TE M0 quantification by selecting shortest TE.

jan-petr:
### Linked issue

Close #1411

### DIFF
--- a/Modules/SubModule_ASL/xASL_wrp_ProcessM0.m
+++ b/Modules/SubModule_ASL/xASL_wrp_ProcessM0.m
@@ -56,7 +56,7 @@ function xASL_wrp_ProcessM0(x)
 %
 % EXAMPLE: xASL_wrp_ProcessM0(x);
 % __________________________________
-% Copyright (C) 2015-2020 ExploreASL
+% Copyright (C) 2015-2023 ExploreASL
 
 
 
@@ -88,19 +88,40 @@ elseif x.modules.asl.M0_conventionalProcessing == 1 && strcmpi(x.Q.readoutDim,'3
        warning('M0 conventional processing disabled, since this masking does not work with 3D sequences');
 end
 
-%% 0B) Scale PD between ASL & M0 if voxel sizes differ
-M0nii       = xASL_io_ReadNifti(x.P.Path_M0);
-M0size      = prod(M0nii.hdr.pixdim(2:4));
-ASLnii      = xASL_io_ReadNifti(x.P.Path_ASL4D);
-ASLsize     = prod(ASLnii.hdr.pixdim(2:4));
-M0ScaleF    = ASLsize/M0size;
+%% 0B) Scale PD between ASL & M0 if voxel volumes differ
+M0nii          = xASL_io_ReadNifti(x.P.Path_M0);
+M0voxelVolume  = prod(M0nii.hdr.pixdim(2:4));
+ASLnii         = xASL_io_ReadNifti(x.P.Path_ASL4D);
+ASLvoxelVolume = prod(ASLnii.hdr.pixdim(2:4));
+M0ScaleVolume  = ASLvoxelVolume/M0voxelVolume;
 
-% copy existing M0 for processing, in single precision
-% averaging if multiple frames, as we will blur later anyway,
-% we can skip the motion correction here
-xASL_io_SaveNifti(x.P.Path_M0, x.P.Path_rM0, mean(xASL_io_Nifti2Im(x.P.Path_M0).*M0ScaleF,4), 32, 0);
+% Copy existing M0 for processing, in single precision, we can skip the motion correction here
+% Load and scale the M0 for the ASL-M0 volume changes
+imM0 = xASL_io_Nifti2Im(x.P.Path_M0).*M0ScaleVolume;
 
-% Note that here we created rM0, which is averaged across 4th dimension, and adapted along this function
+% Try to load M0.json
+[pathM0json, filenameM0json,~] = xASL_fileparts(x.P.Path_M0);
+pathM0json = fullfile(pathM0json, [filenameM0json, '.json']);
+% Load JSON and locate TE part
+if xASL_exist(pathM0json, 'file')
+	jsonM0 = xASL_io_ReadJson(pathM0json);
+	if isfield(jsonM0, 'EchoTime')
+		% If there is more than 1 different nonzero TE, then we have to select the shortest ones
+		uniqueNonzeroTE = unique(jsonM0.EchoTime(jsonM0.EchoTime > 0));
+		if length(uniqueNonzeroTE) > 1
+			% Find the smallest positive nonzero TE
+			minTE = min(uniqueNonzeroTE);
+			% Average M0 image across all these minimal TEs
+			imM0 = mean(imM0(:,:,:,jsonM0.EchoTime == minTE), 4);
+		end
+	end
+end
+
+% Averaging if multiple frames, as we will blur later anyway. Unless this has been averaged across minimal TEs already
+imM0 = mean(imM0,4);
+
+% Save the resampled mean M0
+xASL_io_SaveNifti(x.P.Path_M0, x.P.Path_rM0, imM0, 32, 0);
 
 xASL_im_CreateASLDeformationField(x); % make sure we have the deformation field in ASL resolution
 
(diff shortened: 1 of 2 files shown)


Modules/SubModule_ASL/xASL_wrp_ProcessM0.m
```diff
@@ -88,19 +88,42 @@ function xASL_wrp_ProcessM0(x)
...
+pathM0json = fullfile(pathM0json, [filenameM0json, '.json']);
+% Load JSON and locate TE part
+if xASL_exist(pathM0json, 'file')
+	jsonM0 = xASL_io_ReadJson(pathM0json);
+	if isfield(jsonM0, 'EchoTime')
+		% If there is more than 1 different nonzero TE, then we have to select the shortest ones
+		uniqueNonzeroTE = unique(jsonM0.EchoTime(jsonM0.EchoTime > 0));
+		if length(uniqueNonzeroTE) > 1
+			% Find the smallest positive nonzero TE
+			minTE = min(uniqueNonzeroTE);
```
HenkMutsaerts, line 114: code says smallest, but comment says positive and non-zero? (which was selected on the previous sentence)

Modules/SubModule_ASL/xASL_wrp_ProcessM0.m
```diff
@@ -88,19 +88,42 @@ function xASL_wrp_ProcessM0(x)
...
+			fprintf('Multi-TE M0 present. Using the shortest TE=%.1f ms for M0 calibration.\n', minTE*1000);
+
+			% Average M0 image across all these minimal TEs
+			imM0 = mean(imM0(:,:,:,jsonM0.EchoTime == minTE), 4);
```
HenkMutsaerts: this should be the nanmean instead of mean.

Also, should we always do the shortest? And not the one closest to ASL?

Modules/SubModule_ASL/xASL_wrp_ProcessM0.m
```diff
@@ -88,19 +88,42 @@ function xASL_wrp_ProcessM0(x)
...
+		end
+	end
+end
 
-% copy existing M0 for processing, in single precision
-% averaging if multiple frames, as we will blur later anyway,
-% we can skip the motion correction here
-xASL_io_SaveNifti(x.P.Path_M0, x.P.Path_rM0, mean(xASL_io_Nifti2Im(x.P.Path_M0).*M0ScaleF,4), 32, 0);
+% Averaging if multiple frames, as we will blur later anyway. Unless this has been averaged across minimal TEs already
+imM0 = mean(imM0,4);
```
HenkMutsaerts: nanmean instead (because there may have been motion correction of M0 which gives some NaNs outside the FoV)

HenkMutsaerts: space behind the comma (code styleguide)

Modules/SubModule_ASL/xASL_wrp_Quantify.m
```diff
@@ -137,6 +137,12 @@ function xASL_wrp_Quantify(x, PWI_Path, pathOutputCBF, M0Path, SliceGradientPath
 PWI = xASL_io_Nifti2Im(PWI_Path); % Load CBF nifti
 ASL_parms = xASL_adm_LoadParms(x.P.Path_ASL4D_parms_mat, x);
 
+% Assigns the shortest minimal positive TE for ASL
+ASLshortestTE = [];
```
HenkMutsaerts, line 141: If the EchoTime field doesn't exist, this will create an error?
We need at least to add a warning below, if we need to do the M0 correction by T2 star and we don't know the `ASLshortestTE`

Modules/SubModule_ASL/xASL_wrp_Quantify.m
```diff
@@ -228,22 +236,21 @@ function xASL_wrp_Quantify(x, PWI_Path, pathOutputCBF, M0Path, SliceGradientPath
...
 	
+	% Assigns the shortest minimal positive TE for M0
+	M0shortestTE = [];
+	if isfield(M0_parms, 'EchoTime')
+		M0shortestTE = min(M0_parms.EchoTime(M0_parms.EchoTime > 0));
+	end
+
     % Check echo times
-    if  isfield(ASL_parms,'EchoTime') && isfield(M0_parms,'EchoTime')
+    if  ~isempty(ASLshortestTE) && ~isempty(M0shortestTE)
```
HenkMutsaerts, line 248: `else warning` missing

Modules/SubModule_ASL/xASL_wrp_Quantify.m
```diff
@@ -255,17 +262,16 @@ function xASL_wrp_Quantify(x, PWI_Path, pathOutputCBF, M0Path, SliceGradientPath
...
-            % Check if TE numbers match
-            if (numel(ASL_parms.EchoTime)==1) && (numel(M0_parms.EchoTime)==1)
-                M0_im = M0_im.*ScalingM0./ScalingASL;
-                fprintf('Delta TE between ASL %s ms & M0 %s ms, for %s, assuming %s decay of arterial blood, factor applied to M0: %s\n', ...
-                    num2str(ASL_parms.EchoTime),num2str(M0_parms.EchoTime),...
-                    x.Q.Sequence, CorrName, num2str(ScalingM0/ScalingASL));
-            else
-                fprintf('Warning: multi-TE processing still work in progress...\n');
-            end
+			% If the shortest TEs are unequal, then we have to compensate for this
```
HenkMutsaerts, line 267: This is the problem. Here, we assume that the shortest TE of both ASL and M0 is the one that should be equal between both, is that always the case?

E.g., if we have M0_TE = [10 20 30] and ASL_TE = 20, then we will have a problem.
So I would do the following:

1) establish the ASL_TE
2) establish the M0_TE
3) check which ones are overlapping
4) use the smallest value of those that are overlapping (3)
5) issue warning if the smallest overlapping value differs from the smallest ASL OR smallest M0 TE value

HenkMutsaerts (changes requested):
Here, we assume that the shortest TE of both ASL and M0 is the one that should be equal between both, is that always the case?

E.g., if we have M0_TE = [10 20 30] and ASL_TE = 20, then we will have a problem.
So I would do the following:

1) establish the ASL_TE
2) establish the M0_TE
3) check which ones are overlapping
4) use the smallest value of those that are overlapping (3)
5) issue warning if the smallest overlapping value differs from the smallest ASL OR smallest M0 TE value

BeatrizPadrela (approved):

Modules/SubModule_ASL/xASL_wrp_ProcessM0.m
jan-petr: Code has two parts - first selecting non-zero positive. and then selecting smallest of those. 
So all good and consistent, but I will fine-tune the comments.

Modules/SubModule_ASL/xASL_wrp_ProcessM0.m
jan-petr: Closest to ASL = we deal with that in reporting discrepancies. nanmean - implemented now.

Modules/SubModule_ASL/xASL_wrp_ProcessM0.m
jan-petr: yes. done.

jan-petr: done.

Modules/SubModule_ASL/xASL_wrp_Quantify.m
jan-petr: Good point. Adding errors on non-existeng TEs when needed.

Modules/SubModule_ASL/xASL_wrp_Quantify.m
jan-petr: Added

PR #1422, opened 2023-05-23
Bug #1407 flair import

### LINKED ISSUE #1407: Joint T1 and FLAIR import fails in DEBBIE
### Description

It only works separately, but together, T1w is called T1_1 in temp and FLAIR is called FLAIR_2 and isn't further converted in DCM2BIDS. This will hopefully be fixed in the other Import revamp+debug.

### Tasks

- [x] Session number is assigned based on the order of the session instead of on the number in the session identifier `ASL_1`.

### Release notes

Allow to load more than one structural scans (e.g., FLAIR and T1w).

jan-petr:
### Linked issue

Close #1407

### DIFF
--- a/Modules/SubModule_Import/xASL_wrp_DCM2NII_Subject.m
+++ b/Modules/SubModule_Import/xASL_wrp_DCM2NII_Subject.m
@@ -25,7 +25,7 @@ function [x, PrintDICOMFields, dcm2niiCatchedErrors] = xASL_wrp_DCM2NII_Subject(
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
 % EXAMPLE:     [x, PrintDICOMFields, dcm2niiCatchedErrors] = xASL_wrp_DCM2NII_Subject(x, matches, dcm2niiCatchedErrors);
 % __________________________________
-% Copyright 2015-2022 ExploreASL
+% Copyright 2015-2023 ExploreASL
 
 
     %% 1. Run DCM2NII for one individual subject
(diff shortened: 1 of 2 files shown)


Modules/SubModule_Import/xASL_wrp_DCM2NII_Subject.m
```diff
@@ -121,6 +121,14 @@
                 scanFields.iSession = iSession;
                 scanFields.iScan = iScan;
                 scanFields.name = thisRun.name;
+
+				% Create the session number based on the session name in format ASL_X
+				if isempty(thisRun.name) || isempty(regexpi(thisRun.name, '.*\d*$', 'once'))
+					% Set to 1 if session name cannot be identified
+ 					scanFields.numberSession = 1;
+				else
+					scanFields.numberSession = xASL_str2num(thisRun.name(5:end));
```
HenkMutsaerts, line 130: This doesn't fully match with the regexp on line 126. So this should be using xASL_str2num(regexp(...)) instead (we strictly speaking don't know if it is always 5:end).

Alternatively, the first regexpi on line 126 should match '^ASL_\d*$

by the way, it should be \d+ not \d* (the asterisk also allows zero instances of the digit, which we don't want). So it should be '^ASL_\d+$' then right? Unless we want to be less restrictive, but then line 130 should change to a regular expression

HenkMutsaerts (approved):
I found one minor regexp issue, otherwise this looks good (though it needs testing if you didn't mess up the numberSession vs iSession, but I guess that's what you did to detect the minor bug fix in the last commit)

Modules/SubModule_Import/xASL_wrp_DCM2NII_Subject.m
jan-petr: Yes '^ASL_\d+$

PR #1425, opened 2023-05-25
Bug #1424 population session2 only

### LINKED ISSUE #1424: Population module fails at session 2 if session 1 is absent
### Description

When we run the population module to calculate CBF on a dataset that has CBF only for session 2 only. Then it crashes. The issue is that number of sessions is 2, but the list of sessions has only `ASL_2`, so while the maximum number of sessions is 2, the effective number of sessions is 1. The maximum number of sessions and the effective number of sessions thus differ and this is a challenge for the current code.

### Tasks

- [x] Replace the `x.dataset.nSessions` with `length(x.SESSIONS)` as `x.SESSIONS` in population module locally reflect only sessions to be processed, and not necessarily all sessions available from 1 to N.
HENK: `x.dataset.nSessions` should do exactly the same, but it determines it at the point of starting the processing. During the processing, the number of sessions shouldn't change. The problem probably occurs because you concatenate and delete ASL_1? So the concatenation should happen but ASL_1 shouldn't be deleted?
But I agree, this shouldn't crash.
JAN: x.SESSIONs is locally changed IN the population module. The problem is only in the population module. It has nothing to do with concatenation as we didn't do any concatenation or deletion. Simply ASL_1 wasn't calculating CBF because I don't have FABBER. So NOT processing, but local problem in Population where x.SESSIONS is updated to process only those sessions that exist.

- [x] The list x.SESSIONS only reflects CBF. But if we also want to get `M0` in population module for ASL_1, then this will be ignored if CBF is on ASL_2 and M0 on ASL_1 and ASL_2.
HENK: when you reach the ROI processing, this shouldn't be an issue. For the `CreatePopulationTemplate` part, this will go wrong anyway with sessions, so we shouldn't change this. I made a specific remark there that this should be changed AFTER we use the new BIDS structure.
Anyway, this should be datatype-specific. Your solution may say that all subjects have missing data for ASL (because there isn't an ASL_1) and it will skip ASL. So this needs testing.
JAN: It IS an issue. Because it doesn't locate CBF for ASL_1 so it skips all ASL_1 sessions. But M0 exists for ASL_1 and we don't want it skipped. But this SESSION selecting function only took the first datatype. 

### HENK: OK THIS IS MY TAKE ON THIS:
xASL_adm_GetPopulationSessions: should give the amount of sessions (nSessions) and the specific names (x.SESSIONS) on each desired location in the population module.

Problem 1: in xASL_adm_CreatePopulationTemplates: xASL_adm_GetPopulationSessions is ran only once before, so it assumes that all datatypes have the same x.SESSIONS & 
nSessions. So solution: xASL_adm_GetPopulationSessions should run separately for each datatype inside xASL_adm_CreatePopulationTemplates.

Problem 2: in xASL_stat_GetRoiStatistics & xASL_stat_PrintStats:  xASL_adm_GetPopulationSessions is ran for each separate datatype, but inside xASL_adm_GetPopulationSessions it always uses CBF. Instead, it should use the input datatype (and default to CBF).

Note that we later have to do the same for visits…
And note that this may be easier with BIDS, because the nomenclature for visits & sessions will be standardised by then.

### Release notes

Fix population module - creation of templates when session 1 is missing but a second session does exist

jan-petr:
### Linked issue

Close #1424


Modules/SubModule_Population/xASL_wrp_CreatePopulationTemplates.m
```diff
@@ -293,9 +293,13 @@ function xASL_wrp_CreatePopulationTemplates(x, bSaveUnmasked, bCompute4Sets, Spe
 %% 2. Iterate over scan types & sessions
 for iScanType=1:length(PreFixList)
     UnAvailable = 0;
+
+	% Define sessions
+	x.S.InputDataStr = PreFixList{iScanType};
+	[~, ~, listSessions] = xASL_adm_GetPopulationSessions(x);
```
HenkMutsaerts: The first output is `nSessions`

Modules/SubModule_Population/xASL_wrp_CreatePopulationTemplates.m
```diff
@@ -293,9 +293,13 @@ function xASL_wrp_CreatePopulationTemplates(x, bSaveUnmasked, bCompute4Sets, Spe
...
     
     fprintf('%s\n', ['Searching ' TemplateNameList{iScanType} ' images:']);
-    for iSession=1:length(x.SESSIONS) % iterate over sessions
+    for iSession=1:length(listSessions) % iterate over sessions
```
HenkMutsaerts: Then we can replace `length(listSessions)` by `nSessions` throughout

Modules/SubModule_Population/xASL_wrp_CreatePopulationTemplates.m
```diff
@@ -334,8 +338,8 @@ function xASL_wrp_CreatePopulationTemplates(x, bSaveUnmasked, bCompute4Sets, Spe
             % ----------------------------------------------------------------------------------------------------
             %% 3. Check availability images          
             for iSubject = 1:x.dataset.nSubjects
-                SubjSess = (iSubject-1)*length(x.SESSIONS) + iSession;
-                xASL_TrackProgress(SubjSess, x.dataset.nSubjects*length(x.SESSIONS));
+                SubjSess = (iSubject-1)*length(listSessions) + iSession;
```
HenkMutsaerts: I like the replacement of x.SESSIONS by listSessions

Modules/SubModule_Population/xASL_wrp_CreatePopulationTemplates.m
```diff
@@ -294,9 +294,13 @@ function xASL_wrp_CreatePopulationTemplates(x, bSaveUnmasked, bCompute4Sets, Spe
 for iScanType=1:length(PreFixList)
     UnAvailable = 0;
 
-	% Define sessions
+	% Define sessions if relevant for the datatype
```
HenkMutsaerts, line 297: This defaulting to ASL_1 is something we need in every instance right, so I would remove it here and move it to `xASL_adm_GetPopulationSessions` (where I perhaps already programmed it)

HenkMutsaerts (changes requested):
Few minor changes, but this looks good!

Modules/SubModule_Population/xASL_wrp_CreatePopulationTemplates.m
jan-petr: Done

Modules/SubModule_Population/xASL_wrp_CreatePopulationTemplates.m
jan-petr: Done

Modules/SubModule_Population/xASL_wrp_CreatePopulationTemplates.m
jan-petr: ok ;)

Modules/SubModule_Population/xASL_wrp_CreatePopulationTemplates.m
jan-petr: Slightly different thing. In `xASL_adm_GetPopulationSessions` we are looking for `ASL_X` sessions. So things like `M0_ASL_2` or `CBF_ASL_3`. But it won't return any sessions for T1w images for example, as they are not bound to sessions.

So for T1w, `xASL_adm_GetPopulationSessions` will return an empty list of sessions and nSessions = 1. But here, we don't call this function, as we are not interested in session lists for dataTypes like T1w. So it's better that this stays here.

jan-petr: So while I agree with how you edit `xASL_adm_GetPopulationSessions` and will keep it. I would not call this function here. Because it will generate an unnecessary warning for missing session for a case where we are not even expecting that sessions exist.

jan-petr:
I accepted your first commit, but rejected the second because of conflicts. But most of it is in my version anyway - see comment.

HenkMutsaerts (approved):

PR #1427, opened 2023-05-26
Feature #1247 automated testing

### LINKED ISSUE #1247: Implement an automated testing script to run all tests in one go
### Description

xASL_test_RunAll.sh is a shell script that runs all four currently available tests (Test SPM, Unit tests, Test Pipeline and Test Flavor Database) in succession and concatenates the results in a single folder.
This script can be ran manually or added to cron if you want repeated tests every few days. 

### Tasks

- [x] Include bTesting = true to xASL_test_Flavors.m
- [x] Return all test directories back to a clean set after the test. (remove the derivatives or move files to a temp scratch location).
- [x] Change the output of test scrips to both .tsv and .mat
- [x] Concatenate all failed test result files to either a pdf or a summary table. 
- [x] Change the output of the test scripts to include date-time (ISO8601)
- [x] Change xALS_test_IgnoreFiles to xASL_test_IgnoreFiles
- [x] Separate xASL_test_ExploreASL into functions xASL_Test_SPM, xASL_Test_CompareReference and xASL_Summarize
- [x] Change the input to be a seperate JSON file
 
### How to test

tba.

### Release notes

#1247 Standardized Testing Script Results, Added extra fail saves to Testing scripts, added a test script that periodically runs all tests and logs the outcome.

maartenhammer:
Closes #1247


Testing/xASL_test_SPM.m
```diff
@@ -0,0 +1,126 @@
+function xASL_test_SPM(TestDirDest, bTestDataUsed)
```
jan-petr, line 1: I'm confused. This function used to be inside xASL_test_TestExploreASL. I'm fine having it independently, but then we should remove it from the other function to avoid having two versions.

Testing/xASL_test_CompareReference.m
```diff
@@ -0,0 +1,223 @@
+function [result] = xASL_test_CompareReference(pathReference, pathResults)
```
jan-petr: Is this copy pasted without changes from xASL_test_ExploreASL - then these should all be removed from there to avoid having two copies of the same code.

Testing/xASL_test_Summarize.m
```diff
@@ -0,0 +1,110 @@
+function [summary, summaryTable] = xASL_test_Summarize(ResultsDir)
```
jan-petr, line 1: A new function or copied from elsewhere?

jan-petr (changes requested):
See the individual comments.
Also - I see many merge commits, were there any issues with conflicts? I guess not because you didn't really change the testing functions itself, right? 
For merging, we should probably squash-merge this or cherry-pick to another branch to get rid of the distracting merge commits.

Testing/xASL_test_SPM.m
maartenhammer: I'll remove the one in xASL_test_TestExploreASL.

Testing/xASL_test_CompareReference.m
maartenhammer: Some changes were made, but I'll change xASL_test_ExploreASL to call this instead of something else.

Testing/xASL_test_Summarize.m
maartenhammer: This is a new Function.

maartenhammer:
Removed all redundant files and moved the weeklyTest main script to CustomScripts with a Json-like input document. 
Currently running xASL_test_TestExploreASL to ensure changes didn't break the test script itself.

jan-petr (approved):
OK.

PR #1432, opened 2023-06-05
Closes #1431 xASL_adm_GetPopulationSessions: regexp bugfix

### LINKED ISSUE #1431: xASL_adm_GetPopulationSessions regexp wrong
### Description

`xASL_adm_GetPopulationSessions` line 41: regexp searched for any occurrence of ASL_\d in any of the folder layers, which could be PCASL_3D giving ASL_3
(Whereas this should be ASL_1 if the path ends on ASL_1).

### Tasks

- [x] search for ASL_\d+\.nii
- [x] if multiple indices were found, use the latest

### Release notes

Fetch correct sessions in the population module.

HenkMutsaerts:
### Linked issue

Closes #1431

### DIFF
--- a/Modules/SubModule_Population/xASL_adm_GetPopulationSessions.m
+++ b/Modules/SubModule_Population/xASL_adm_GetPopulationSessions.m
@@ -51,7 +51,6 @@ if isempty(SessionList) % If no files found, search for subject files instead of
 else % If files found, continue with defining sessions from SessionList
     
     for iSession = 1:numel(SessionList)
-        [~, folderName] = fileparts(SessionList{iSession});
         [IndexStart, IndexEnd] = regexpi(SessionList{iSession}, 'ASL_\d+');
         NewList{iSession,1} = SessionList{iSession}(1,IndexStart(end):IndexEnd(end)); % create a cell array containings characters of all session numbers present
     end
--- a/Modules/SubModule_Population/xASL_adm_GetPopulationSessions.m
+++ b/Modules/SubModule_Population/xASL_adm_GetPopulationSessions.m
@@ -51,8 +51,9 @@ if isempty(SessionList) % If no files found, search for subject files instead of
 else % If files found, continue with defining sessions from SessionList
     
     for iSession = 1:numel(SessionList)
+        [~, folderName] = fileparts(SessionList{iSession});
         [IndexStart, IndexEnd] = regexpi(SessionList{iSession}, 'ASL_\d+');
-        NewList{iSession,1} = SessionList{iSession}(1,IndexStart:IndexEnd); % create a cell array containings characters of all session numbers present
+        NewList{iSession,1} = SessionList{iSession}(1,IndexStart(end):IndexEnd(end)); % create a cell array containings characters of all session numbers present
     end
     
     SESSIONS = xASL_adm_SortStringNumbers(unique(NewList)); % determine unique session numbers


Modules/SubModule_Population/xASL_adm_GetPopulationSessions.m
```diff
@@ -51,8 +51,9 @@
 else % If files found, continue with defining sessions from SessionList
     
     for iSession = 1:numel(SessionList)
+        [~, folderName] = fileparts(SessionList{iSession});
```
jan-petr: Where is the folderName being used?

Modules/SubModule_Population/xASL_adm_GetPopulationSessions.m
```diff
@@ -51,8 +51,9 @@
...
         [IndexStart, IndexEnd] = regexpi(SessionList{iSession}, 'ASL_\d+');
-        NewList{iSession,1} = SessionList{iSession}(1,IndexStart:IndexEnd); % create a cell array containings characters of all session numbers present
+        NewList{iSession,1} = SessionList{iSession}(1,IndexStart(end):IndexEnd(end)); % create a cell array containings characters of all session numbers present
```
jan-petr, line 55: It is good to go for the last occurrence. Cause in theory, there could be more occurrences. However, I don't understand how this solves what is written in the issue. Shouldn't you rather modify 'ASL_\d+' to '^ASL_\d+$'?

jan-petr (changes requested):
I do not understand why you mention ``PCASL_3D``. The part of your code that you have modified isn't about folderHiererachy but about session tokens. Session tokens can be only ASL_1 or ASL_3 or ASL_13 etc. But never PCASL_3D, unless somebody writes nonsense to sourcestructure.json.

Also - I do not understand why you mention last folder layers in the issue. The changes you made apply to sessions and not folderHierarchy.

I agree with the code changes, just want to clarify the issue etc. Or have I overlooked something??

HenkMutsaerts:
> I do not understand why you mention `PCASL_3D`. The part of your code that you have modified isn't about folderHiererachy but about session tokens. Session tokens can be only ASL_1 or ASL_3 or ASL_13 etc. But never PCASL_3D, unless somebody writes nonsense to sourcestructure.json.
> 
> Also - I do not understand why you mention last folder layers in the issue. The changes you made apply to sessions and not folderHierarchy.
> 
> I agree with the code changes, just want to clarify the issue etc. Or have I overlooked something??

Haha think about it. If you search for any ASL_3, there is an occurrence in PC ASL_3 D ;)

jan-petr:
> > I do not understand why you mention `PCASL_3D`. The part of your code that you have modified isn't about folderHiererachy but about session tokens. Session tokens can be only ASL_1 or ASL_3 or ASL_13 etc. But never PCASL_3D, unless somebody writes nonsense to sourcestructure.json.
> > Also - I do not understand why you mention last folder layers in the issue. The changes you made apply to sessions and not folderHierarchy.
> > I agree with the code changes, just want to clarify the issue etc. Or have I overlooked something??
> 
> Haha think about it. If you search for any ASL_3, there is an occurrence in PC ASL_3 D ;)

Yes. I am thinking about it :) It's a list of sessions, not folderHierarchy. The values can be only ASL_1 or ASL_2 etc. Where would PCASL_3D come from in that list? We are talking about a list of sessions here, not about folder names. If I get it right.

Also - check my other comments.

HenkMutsaerts:
> > > I do not understand why you mention `PCASL_3D`. The part of your code that you have modified isn't about folderHiererachy but about session tokens. Session tokens can be only ASL_1 or ASL_3 or ASL_13 etc. But never PCASL_3D, unless somebody writes nonsense to sourcestructure.json.
> > > Also - I do not understand why you mention last folder layers in the issue. The changes you made apply to sessions and not folderHierarchy.
> > > I agree with the code changes, just want to clarify the issue etc. Or have I overlooked something??
> > 
> > 
> > Haha think about it. If you search for any ASL_3, there is an occurrence in PC ASL_3 D ;)
> 
> Yes. I am thinking about it :) It's a list of sessions, not folderHierarchy. The values can be only ASL_1 or ASL_2 etc. Where would PCASL_3D come from in that list? We are talking about a list of sessions here, not about folder names. If I get it right.
> 
> Also - check my other comments.

What the code did: was checking for ANY occurrence of ASL_\d+ and taking the first instance, which would detect `/Siemens_PCASL_3D_Test/sourcedata/sub-001/ASL_1` as ASL_3 (from PCASL_3D).

I improved this now in two ways: 
1) ensuring it searches for ASL_d+\.nii
2) taking the last index only

You're right that (2) is not really necessary anymore, I just kept it there to be sure.

Modules/SubModule_Population/xASL_adm_GetPopulationSessions.m
HenkMutsaerts: You're right, this was a temporary thing I used but not anymore. Will delete now

Modules/SubModule_Population/xASL_adm_GetPopulationSessions.m
HenkMutsaerts: See my final attempt to explain below :)

jan-petr (approved):
OK.

Issue #1456, opened 2023-07-19
Add dummy field to ASLContext


jan-petr:
### Description

The dummy scan position option is not working correctly. E.g., for FlavorDatabase/Siemens_PCASL_multi_VE11C_1/rawdataReference/sub-101/ses-1/perf/sub-101_ses-1_run-1_asl.json this is done incorrectly, as DummyScanPositionInASL4D is added to studyPar.json instead of dataPar.json, and thus ends up in the ASL sidecar (which is not according to BIDS).

And we want to add a dummy scan already in import to BIDS!!! The Northeastern group doing TBI was also interested in that.

### Tasks

- [ ] Make sure that M0 and dummy parameters are correctly read from the ASL context in `xASL_io_SplitASL`.
- [ ] Having dummy in JSON, m0scan in context, it doesn't split both and M0 is ignored during the processing.
- [ ] Make sure that the splitting is always run from `xASL_bids_parseM0` and only inside check if it is necessary based on ASLcontext etc.
- [ ] Make sure that M0 and dummy parameters are correctly saved to the ASLContext and not elsewhere. And start phasing out to save them inside the JSON parameters, but only save them in ASLContext.tsv.
- [ ] Allow import to BIDS with a `dummy` scan in ASLcontext in `DCM2NII` and `NII2BIDS`.
- [ ] Update dummy in BIDS specification. Yes, you create an issue on https://github.com/bids-standard/bids-specification and if you want to solve it yourself, you fork the repository, create a branch, and do a PR from this branch to the original repository, and link it to this issue.
- [ ] Fix #748 
- [ ] Correct splitting of dummy scans from ASLcontext in BIDS2Legacy.
- [ ] Check adding and splitting dummy here: `xASL_bids_BIDSifyASLJSON`, `xASL_bids_parseM0`, `xASL_bids_BIDSifyASLNII`, `xASL_bids_MergeNifti`
- [ ] Adding `DummyScanPositionInASL4D` to studyPar.json shouldn't result in adding it to ASL.json (not according to BIDS) but instead adding it to the dataPar.json. See `Siemens_PCASL_multi_VE11C_1/rawdataReference/sub-101/ses-1/perf/sub-101_ses-1_run-1_asl.json` - this won't work for multi-sequence studies, so we really need to add that to the ASL4D.json and ASLContext.
- [ ] Inform Northeastern group about adding the dummy scan as a an extra entry.
- [ ] Properly link ASLContext and JSONs
[...]

PR #1457, opened 2023-07-19
#1062 Fix M0-split JSON parameters

### LINKED ISSUE #1062: Correct JSON when splitting M0
### Description

When splitting an M0 image from an ASL volumes, then we have to remove certain ASL tags when creating the M0.json as not all ASL tags are supposed to be there. 

Test on `Siemens_PCASL_3DGRASE_VE11C_3` and also find out why there is the `T1` tag, why the PLD and labdur tags are wrong, and why the TotalAcquiredPairs tag is wrong after import to BIDS.

When we are removing the dummyScans, we should make sure to reduce and check the TotalAcquiredPairs as well. For example in Siemens_PCASL_2DEPI_VE11C_MB_1, we have 45 pairs (including 2 dummy scans and 2 m0scans). We remove correctly both in BIDS2Legacy, but we reduce the TotalAcquired pairs only to 44 and not to 43.

### Tasks

- [x] When splitting M0 from ASL, run the BIDS function that strips the unnecessary ASL tags from the M0.json. Test on `Siemens_PCASL_3DGRASE_VE11C_3`. Check T1 tag, PLD and labdurs lengths.
- [x] Check `TotalAcquiredPairs` tag and adjust for removing dummy scans after splitting in `Siemens_PCASL_3DGRASE_VE11C_3`. and in `Siemens_PCASL_2DEPI_VE11C_MB_1`.
- [x] Check totalPairs and dummy in `xASL_bids_BIDSifyASLJSON`, `xASL_bids_parseM0`, `xASL_bids_BIDSifyASLNII`, `xASL_bids_MergeNifti`.

### How to test

- [x] Update `Siemens_PCASL_multi_VE11C_1` flavor.
- [x] RUn flavors test.
- [x] Do final flavor updates.

### Release notes

Fix TotalAcquiredPairs calculation with dummy scans.

jan-petr:
### Linked issue

Close #1062

### DIFF
--- a/Functions/BIDS/xASL_bids_BIDSifyASLJSON.m
+++ b/Functions/BIDS/xASL_bids_BIDSifyASLJSON.m
@@ -542,12 +542,23 @@ if isfield(jsonOut,'NumberOfAverages') && (max(jsonOut.NumberOfAverages) > 1)
 end
 
 % First count the number of controls, labels, and deltaMs
+% Check and exclude dummy scans first from ASLContext
+if isfield(studyPar, 'DummyScanPositionInASL4D') && ~isempty(studyPar.DummyScanPositionInASL4D)
+	for iDummy = 1:length(studyPar.DummyScanPositionInASL4D)
+		if length(ASLContextCell) >= studyPar.DummyScanPositionInASL4D(iDummy)
+			ASLContextCell{studyPar.DummyScanPositionInASL4D(iDummy)} = 'dummy';
+		end
+	end
+end
+
 ASLContextControlIndex = cellfun(@(x)~isempty(x),regexpi(ASLContextCell,'^control')); % Create a vector out of it
 ASLContextControlIndex = ASLContextControlIndex(1:dimASL(4)); % Remove the last empty field
 ASLContextLabelIndex = cellfun(@(x)~isempty(x),regexpi(ASLContextCell,'^label')); % Create a vector out of it
 ASLContextLabelIndex = ASLContextLabelIndex(1:dimASL(4)); % Remove the last empty field
 ASLContextDeltaMIndex = cellfun(@(x)~isempty(x),regexpi(ASLContextCell,'^deltam')); % Create a vector out of it
 ASLContextDeltaMIndex = ASLContextDeltaMIndex(1:dimASL(4)); % Remove the last empty field
+
+
 
 % If TotalAcquiredPairs is 1, but more control/label pairs od deltaMs are present, then set this to the correct
 % number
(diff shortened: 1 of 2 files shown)


Functions/BIDS/xASL_bids_BIDSifyASLJSON.m
```diff
@@ -542,13 +542,24 @@
 end
 
 % First count the number of controls, labels, and deltaMs
+% Check and exclude dummy scans first from ASLContext
```
HenkMutsaerts, line 545: Are we sure we also removed the dummy scans from the image volume, if we remove it from the ASLContext here? Or am I mixing things up

Functions/InputOutput/xASL_io_SplitASL.m
```diff
@@ -95,6 +95,9 @@ function xASL_io_SplitASL(inPath, iM0, iDummy)
 	% Check if both cbf and deltam/control/label volumes exist
 	if sum(~cellfun(@isempty,regexpi(aslContext,'deltam|control|label'))) && ...
 	   sum(~cellfun(@isempty,regexpi(aslContext,'cbf')))
+		% In this case, ignore CBF
+		warning('ASLContext has both a CBF image and delteM/control/label scans. CBF is thus ignored and discarded.')
```
HenkMutsaerts: delteM -> deltaM

I would change this warning to:

"ASLContext has both a CBF image and more rawer ASL images (e.g., deltaM, control, label). CBF is thus discarded."

And similar to above: are we sure that we sync our image volume changes with the ASLContext changes? They should probably always be done side-to-side to avoid making debugging difficult.

HenkMutsaerts (changes requested):

Functions/BIDS/xASL_bids_BIDSifyASLJSON.m
jan-petr: Yes. You are mixing things up :) Actually, a good comment. This part here is about counting controls/labels, so we are removing the dummy scans here only for the calculation so that we can calculate correctly TotalAcquiredPairs.

The real removal of dummy volumes are done in another function.

Functions/InputOutput/xASL_io_SplitASL.m
jan-petr: Yes - changing the comment. 

About syncing - you are right, this should be done and we have a new issue for this in 1.12.0. In 1.11.0 in this commit, we are only fixing the JSON information, which was incorrect. In the follow-up issue, we will do proper ASLContext, StudyPar, ASL.json synchronization.

Functions/BIDS/xASL_bids_BIDSifyASLJSON.m
HenkMutsaerts: OK but still we should link them in another issue right. And what is BIDS? TotalAcquiredPairs should normally be all volumes, including the dummy?

HenkMutsaerts (approved):

Functions/BIDS/xASL_bids_BIDSifyASLJSON.m
jan-petr: Will be all solved in #1456

PR #1475, opened 2023-08-14
Feature #1225 generate pdf

### LINKED ISSUE #1225: PDF Report restart
### Description

Restart of issue #984 as to keep the PDF report generated by Maarten separate from the one created by Micheal as well as creating an issue/branch to submit changes to. 

### Tasks

- [x] Have a meeting with Bas to look at the PDF reader.
- [x] Change import of images from qc.results to manual import. 
- [x] Allow images to be added to PDFGenerator
- [x] Allow scans to be added to PDFGenerator
- [x] Allow QC values to be added to PDFGenerator
- [x] Add a rudimentory form of Block forming with automated scaling. 
- [x] Implement a function that reads participants.tsv to include patient information if relevant. 
- [x] Clean up TemplateConfig.Json
- [x] Add information about the scanner and the ASL sequence type.
- [x] Change JSON objects to a unnamed Array instead
- [x] Change filePath to absolutePath or relativePath
- [x] More Complex Example
- [x] Create Markdown Manual
- [x] Add figure titles 

### How to test

- [x] Run MakeReport with complete data
- [x] Run MakeReport with incomplete data 
- [x] Run MakeReport with incomplete Config.json
- [x] Run MakeReport with incomplete Config.json and data
- [x] For anyone trying this in the test branch, a [small manual](https://github.com/ExploreASL/ExploreASL/blob/f5bb987cd9a04ddf6e7fe2a594a38af4cee17c0d/WorkInProgress/QualityControl/PDFGenerator.md ) has been written in addition to the example. Please check it out. 
- [x] Make a follow-up issue to read new version of x.output fields. Createdin #1552 
- [x] Move manual to docs before merging.

### Release notes

New, JSON-configurable PDF report, which can be used for QC output.

maartenhammer:
### Linked issue

Closes #1225

### How to test

You can run the pdf generator using [x, config] =  xASL_qc_GenerateReport(x[, subject]), with x being the ExploreASL s struct, if no subject is provided, it will default to the subject defined in x.SUBJECT.  
A [Small manual](https://github.com/ExploreASL/ExploreASL/blob/f5bb987cd9a04ddf6e7fe2a594a38af4cee17c0d/WorkInProgress/QualityControl/PDFGenerator.md) has been written that goes over the basics to creating a config report for the layout. 
If no configuration file is found in `<dataroot>/Derivatives/ExploreASL` it will use the default in the exploreASL root directory.

**Manual should be moved to the docs before merging!**


Functions/Administration/xASL_adm_GeneratePDFConfig.m
```diff
@@ -0,0 +1,83 @@
...
+%   jsonPath    - Path where json Should be placed  (OPTIONAL DEFAULT is .../derivatives/ExploreASL/ConfigReportPDF.json)
+%   bPrintInvisible - boolean specifying wether you want the config file to contain invisible parameters as well
+%
+% OUTPUT: n/a
+% OUTPUTFILE:
+%   xASL_Report_SubjectName.pdf - Json Structure containing info for the
+%   PDF generator
+%
+% -----------------------------------------------------------------------------------------------------------------------------------------------------
+% DESCRIPTION: This function checks output struct x and places the values
```
HenkMutsaerts: Please complete the description (chat GPT/Github colab can help ;) the title of this function suggests that this function generates the config file that is needed to generate the PDF? But that is not what is described here?

Functions/Administration/xASL_adm_GeneratePDFConfig.m
```diff
@@ -0,0 +1,83 @@
...
+% of the output in a Json file to be read by 
+% 
+% EXAMPLE: xASL_qc_GeneratePDFConfig(x);
+% __________________________________
+% Copyright (C) 2015-2021 ExploreASL
```
HenkMutsaerts: 2023

Functions/Administration/xASL_adm_GeneratePDFConfig.m
```diff
@@ -0,0 +1,83 @@
...
+
+%% -----------------------------------------------------------------------------------------------
+%  Admin
+
+
+if nargin <2 || isempty(jsonPath)
+    jsonPath = strcat(x.dir.xASLDerivatives, '/ConfigReportPDF.json');
```
HenkMutsaerts: better to use fullfile(XX, YY) to ensure that this also works in Windows, as fullfile chooses / or \ dependent on the OS

Functions/Administration/xASL_adm_GeneratePDFConfig.m
```diff
@@ -0,0 +1,83 @@
...
+% 
+% EXAMPLE: xASL_qc_GeneratePDFConfig(x);
+% __________________________________
+% Copyright (C) 2015-2021 ExploreASL
+
+%% -----------------------------------------------------------------------------------------------
+%  Admin
+
+
+if nargin <2 || isempty(jsonPath)
```
HenkMutsaerts: You first need to check the first input, if there is no X this function will crash. That is fine, but we can have a specific error message with something like "Did you forget to load ExploreASL?"

Functions/Administration/xASL_adm_GeneratePDFConfig.m
```diff
@@ -0,0 +1,83 @@
...
+    warning([PathX ' didnt exist, skipping xASL_qc_GeneratePDFConfig']);
+    return;
+end
+
+% Load Output from x.mat
+x = xASL_adm_LoadX(x, PathX, false); 
+ 
+try
+    %% Load Standard Configuration
+    sPath = strcat(x.opts.MyPath, '/Functions/QualityControl/templateConfigReportPDF.json');
```
HenkMutsaerts: fullfile as above

Functions/Administration/xASL_adm_GeneratePDFConfig.m
```diff
@@ -0,0 +1,83 @@
...
+    
+    if ~isfile(sPath)
```
HenkMutsaerts: isn't exist(sPath, 'file') better?

Functions/Administration/xASL_adm_GeneratePDFConfig.m
```diff
@@ -0,0 +1,83 @@
...
+        warning('Default PDF configuration not found, check the explore ASL directory.')
```
HenkMutsaerts: ExploreASL, not explore ASL ;)

be specific: what does it mean to "check the ... directory"? Does that mean that the user should go to that directory and check its contents? Or that the correct ExploreASL directory is provided as input to this function? Or what does "check" mean?

Functions/Administration/xASL_adm_GeneratePDFConfig.m
```diff
@@ -0,0 +1,83 @@
...
+        return
```
HenkMutsaerts: I think we already catch errors in ExploreASL's xASL_Iteration etc, so you might want to simply use error instead of warning&return

Functions/Administration/xASL_adm_GeneratePDFConfig.m
```diff
@@ -0,0 +1,83 @@
...
+    else 
+        config = xASL_adm_LoadPDFConfig(sPath);
+    end 
+
+    clear OutputFields
```
HenkMutsaerts: I did this previously as well but Paul warned me that using "clear" in Matlab functions is not a nice way to do it, better to replace the variable with an empty one or just ignore it.
In any case, here "clear" is redundant because you redefine it on the next line

Functions/Administration/xASL_adm_GeneratePDFConfig.m
```diff
@@ -0,0 +1,83 @@
...
+% -----------------------------------------------------------------------------------------------------------------------------------------------------
+% DESCRIPTION: This function checks output struct x and places the values
+% of the output in a Json file to be read by 
+% 
+% EXAMPLE: xASL_qc_GeneratePDFConfig(x);
+% __________________________________
+% Copyright (C) 2015-2021 ExploreASL
+
+%% -----------------------------------------------------------------------------------------------
+%  Admin
```
HenkMutsaerts: Nice function! You're lacking comments here and there. Also, there is an admin section here, but no other sections. I always try to improve readability by using sections and comments. You could just add another section after the admin (checking input arguments, paths etc)

Functions/Administration/xASL_adm_LoadPDFConfig.m
```diff
@@ -0,0 +1,54 @@
+function [Config] = xASL_adm_LoadPDFConfig(x)
+% xASL_adm_LoadPDFConfig Loads parameters from a .json config file
+% Essentialy an alias of spm_jsonread.
```
HenkMutsaerts: Essentialy -> Essentially

Functions/Administration/xASL_adm_LoadPDFConfig.m
```diff
@@ -0,0 +1,54 @@
...
+%
+% FORMAT: [config] = xASL_adm_LoadPDFConfig(x)
+%
+% INPUT:
+%   ConfigPath  - path to json REQUIRES
```
HenkMutsaerts: REQUIRES -> REQUIRED

Functions/Administration/xASL_adm_LoadPDFConfig.m
```diff
@@ -0,0 +1,54 @@
...
+%
+%   OUTPUT:
+%   config      - Struct containing all parameters of the jsonfile
```
HenkMutsaerts: REQUIRED or OPTIONAL?

Functions/Administration/xASL_adm_GeneratePDFConfig.m
```diff
@@ -0,0 +1,83 @@
+function xASL_adm_GeneratePDFConfig(x, jsonPath, bPrintInvisible)
+%xASL_qc_GeneratePDFConfig Print output QC
+%
+% FORMAT: xASL_qc_CreatePDF(x[, JsonPath, bVerbose])
+%
+% INPUT:
+%   x           - structure containing fields with all information required to run this submodule (REQUIRED)
+%   jsonPath    - Path where json Should be placed  (OPTIONAL DEFAULT is .../derivatives/ExploreASL/ConfigReportPDF.json)
```
HenkMutsaerts: comma missing between OPTIONAL, DEFAULT

Functions/Administration/xASL_adm_GeneratePDFConfig.m
```diff
@@ -0,0 +1,83 @@
...
+%   bPrintInvisible - boolean specifying wether you want the config file to contain invisible parameters as well
```
HenkMutsaerts: OPTIONAL or REQUIRED?

Functions/Administration/xASL_adm_LoadPDFConfig.m
```diff
@@ -0,0 +1,54 @@
...
+% xASL_qc_GenerateReport to make a report with defined parameters:
+%
+% 1. Load JSON file
+% 2. Deal with warnings
+% 3. return a struct containing the config parameters
+%
+% -----------------------------------------------------------------------------------------------------------------------------------------------------
+% EXAMPLE: [config] = xASL_adm_LoadPDFConfig(x);
+% __________________________________
+% Copyright 2015-2022 ExploreASL
```
HenkMutsaerts: 2023

Functions/Administration/xASL_adm_LoadPDFConfig.m
```diff
@@ -0,0 +1,54 @@
...
+
+
+
+%% ------------------------------------------------------------------------
+% 0. Argument check
```
HenkMutsaerts: Where is section 1? Now it jumps from 0 to 2

Functions/Administration/xASL_adm_LoadPDFConfig.m
```diff
@@ -0,0 +1,54 @@
...
+Config = struct; % default
```
HenkMutsaerts: Nice, this is better than `clear Config` (which you used above)

Functions/Administration/xASL_adm_LoadPDFConfig.m
```diff
@@ -0,0 +1,54 @@
...
+
+if nargin<1 || isempty(x)
```
HenkMutsaerts: Nice, this was missing above

Functions/Administration/xASL_adm_LoadPDFConfig.m
```diff
@@ -0,0 +1,54 @@
...
+    warning('x was not specified, aborting');
+    return
+end
+
+ConfigPath = strcat(x.dir.xASLDerivatives, '/ConfigReportPDF.json');
```
HenkMutsaerts: fullfile, as above

Functions/Administration/xASL_adm_LoadPDFConfig.m
```diff
@@ -0,0 +1,54 @@
...
+
+%% ------------------------------------------------------------------------
+%% 2. Load JSON file 
+if isfile(ConfigPath)
```
HenkMutsaerts: same as above, not exist?

Functions/Administration/xASL_adm_LoadPDFConfig.m
```diff
@@ -0,0 +1,54 @@
...
+    Config = spm_jsonread(ConfigPath);
+else
+    fprintf('configReportPDF.json does not exists, using default config \n');
```
HenkMutsaerts: I would print the full path that doesn't exist, not only the filename

Functions/Administration/xASL_adm_LoadPDFConfig.m
```diff
@@ -0,0 +1,54 @@
...
+    standardPath = strcat(x.opts.MyPath, '/Functions/QualityControl/templateConfigReportPDF.json');
```
HenkMutsaerts: fullfile

Functions/Administration/xASL_adm_LoadPDFConfig.m
```diff
@@ -0,0 +1,54 @@
...
+    Config = spm_jsonread(standardPath);
+end
+
+%% ------------------------------------------------------------------------
+%% 3. Deal with warnings
+if isempty(fields(Config))
+    warning('config struct empty, Check if config JSON is configured correctly.');
```
HenkMutsaerts: Always good to be specific. If the user gets this warning on the screen, there is no clue as to which config file this is. So better to specify e.g. "check that the configuration JSON of the PDF report is configured correctly"

Functions/QualityControl/templateConfigReportPDF.json
```diff
@@ -0,0 +1,338 @@
...
+        {
+            "type": "page",
+            "pageIdentifier": "Example_1",
+            "content": [
+                {
+                    "type": "text",
+                    "text": "This is an sample for the Explore ASL report Generator, See ExploreASL/Function/templateConfigReportPDF.json for the script generating this PDF."
+                },{
+                    "type": "scan",
+                    "name": "Pop_Path_rT1",
```
HenkMutsaerts: Nice! Only I would program the overlays with identical config layout as the "underlay", now the underlay has type (and overlay doesn't), underlay has the filename as value of the key "name" and overlay has the filename itself as key (with contour opacity color as children).

So I would rather say something like:
type: scan
position
size
slice
layer 1: {
name
contour
opacity
color}
layer 2: {

name
contour
opacity
color}
layer 3: {

name
contour
opacity
color}

etc. There is some redundancy in my approach, but having all layers identically set up as brother/sister layers makes programming easier I would say

Functions/QualityControl/templateConfigReportPDF.json
```diff
@@ -0,0 +1,338 @@
...
+                    },
+                    "overlay": {
+                        "Pop_Path_MaskVascular": {
+                            "contour": "true",
+                            "opacity": "0.3",
+                            "color": "r"
+                        }
+                    }
+                },{
+                    "type": "patients"
```
HenkMutsaerts: what type is patients?

Functions/QualityControl/templateConfigReportPDF.json
```diff
@@ -0,0 +1,338 @@
...
+                    "type": "text",
+                    "text": "These scans are referred to by their x.P.<name> in the ConfigReportPDF.json."
+                },{
+                    "type": "text",
+                    "text": "This is usually simply Path_Pop_<name>."
+                },{
+                    "type": "text",
+                    "settings": {
+                        "fontSize": "10",
+                        "fontWeight": "bold"
```
HenkMutsaerts: So, I guess these settings only apply to this text type, not the other ones?
And the other ones use the default settings?

Functions/QualityControl/templateConfigReportPDF.json
```diff
@@ -0,0 +1,338 @@
...
+                    },
+                    "text": "Below the Quality parameters of the Structural module will be printed."
+                },{
+                    "type": "text",
+                    "text": "These values can be found in derivatives/ExploreASL/<subject>/QC_collection_<subject>.json"
+                },{
+                    "type": "text",
+                    "text": "If the QC value is outside your specified range, it will display in red."
+                },{
+                    "type": "qc",
```
HenkMutsaerts: We need to define "type" and it is best to have a sensible classes/groups/levels

E.g., "text" and "image" (not scan) are datatypes
but "qc" and "patient" are not datatypes, but what are they?

Functions/QualityControl/templateConfigReportPDF.json
```diff
@@ -0,0 +1,338 @@
...
+                    "content": [
+                        {
+                            "type": "text",
+                            "text": "This is an example of such a content block."
+                        },{
+                            "type": "text",
+                            "text": "This block has text on the left and scans on the right."
+                        },{
+                            "type": "image",
+                            "popPath": "M0Check/Cor_M0_<SUBJECT>_<SESSION>.jpg",
```
HenkMutsaerts: Note that SESSION becomes RUN with BIDS (and visit becomes SESSION)

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
+function [x,config] = xASL_qc_GenerateReport(x, subject)
```
HenkMutsaerts: Why do you output config?

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
...
+    settings.fontWeight = 'normal';
+    settings.axesVisible = 'off';
+    settings.fontName = 'default';
+    settings.lineSpacing = 0.005;
+    settings.figurecount = 0;
+    settings.canvas = [0 0 1 1];
+
+    if ispc
+        settings.fontSize = 10;
+    elseif isunix || ismac
```
HenkMutsaerts: mac == unix, so you can remove ismac

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
...
+        settings.fontSize = 8;
+    end
```
HenkMutsaerts: else: "unknown OS, warning" is missing

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
...
+
+% Parse the entire Json Stack automatically making all the pages.
+[~] = xASL_sub_parseJson(config, x, [], [], settings);
+
+end
+
+
+%% ================================================================================================================================================
+%% ================================================================================================================================================
+function [settings] = xASL_sub_parseJson(json, x, figure, line, settings)
```
HenkMutsaerts: This is quite a long file so I would create a separate function and wrapper (xASL_qc_GenerateReport is the wrapper, and changes minimally with your PR right? That makes it nicely clear)

Also, I thought we decided to call this e.g., xASL_qc_GenerateReport_sub_parseJson because otherwise it is not clear where the subfunction is

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
...
+settings = xASL_sub_defaultSettings();
+
+% Load Pdf configuration
+config = xASL_adm_LoadPDFConfig(x);
+
+% Print the title
+fprintf('Printing ExploreASL PDF report:   \n');
+
+% Parse the entire Json Stack automatically making all the pages.
+[~] = xASL_sub_parseJson(config, x, [], [], settings);
```
HenkMutsaerts: You can remove the ~ output with Matlab

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
...
+% Delete existing xASL report files
+ExistingPrintFiles = xASL_adm_GetFileList(PrintDir, '^xASL_Report_.+$');
+if ~isempty(ExistingPrintFiles)
+    for iFile=1:size(ExistingPrintFiles)
+        xASL_delete(ExistingPrintFiles{iFile});
+    end
+end
+
+% Set defaults
+settings = xASL_sub_defaultSettings();
```
HenkMutsaerts: name of the subfunction, see below

Functions/QualityControl/xASL_qc_GenerateReport.m
HenkMutsaerts: json, figure, line are not defined in the main function.
Also, some may be protected Matlab variables, like figure, that are better to give a unique name

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
...
+            [settings, line] = xASL_sub_parseContent(currentField, x, figure, line, settings);
+        end 
+    end
+end
+
+
+function [settings, line] = xASL_sub_parseContent(currentField, x, figure, line, settings)
+
+    if ~isstruct(currentField)
+        return;
```
HenkMutsaerts: And then? A warning?

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
...
+    end
+
+    if ~isfield(currentField, 'type')
+        error([currentField, 'containts no field specifying the type of content']);
```
HenkMutsaerts: containts -> contains

This error lacks a hint to the user what could have gone wrong

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
...
+            line = xASL_sub_PrintQC(currentField, x, figure, line, settings);
+        case 'image'
+            settings = xASL_sub_PrintImage(currentField, x, figure, settings);  
+        case 'scan'
+            settings = xASL_sub_PrintScan(currentField, x, figure, settings);
+        case 'page'
+            xASL_sub_printPage(currentField, x, settings);  
+        case 'block'
+            xASL_sub_printBlock(currentField, x, figure, settings);
+        case 'settings'
```
HenkMutsaerts: shouldn't it be printSettings? for consistency

I like this function!

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
...
+            settings = xASL_sub_loadSettings(currentField, settings);  
+        case 'patients' 
+            line = xASL_sub_PrintPatient(x, figure, line, settings);
+    end  
+end
+
+function  xASL_sub_printPage(json, x, settings)
+
+    clear primaryFig ax
```
HenkMutsaerts: see above, better not clear but define empty defaults

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
...
+    end  
+end
+
+function  xASL_sub_printPage(json, x, settings)
+
+    clear primaryFig ax  
+    primaryFig = figure('visible', 'off','Units', 'centimeters','Position', [0 0 21 29.7]);
+
+    %% Print the Title
+    logoPath = fullfile(x.opts.MyPath, 'Design/ExploreASL_logoHeader.png');
```
HenkMutsaerts: / will crash on Windows, better to replace this by

logoPath = fullfile(x.opts.MyPath, 'Design', 'ExploreASL_logoHeader.png');

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
...
+    [~] = xASL_sub_PrintImage(logoPath, [], primaryFig, settings, [0 0.96 1 0]);
```
HenkMutsaerts: ~ not needed

by the way if we use variables like
figPrimary
pathLogo

it is easier to list them alphabetically than having 

primaryFig
logoPath

(if we have multiple figs paths etc)

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
...
+
+    %% Print the Footer
+    xASL_sub_PrintText('This report was automatically generated by ExploreASL', primaryFig, [0 0.02 1 0], settings);
+
+    %% Parse JSON and create the file
```
HenkMutsaerts: Parse which JSON and create which file? Be specific

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
...
+    [~] = xASL_sub_parseJson(json, x, primaryFig, [0 0.93 1 0], settings);
+
+    PrintFile = ['xASL_Report_', json.pageIdentifier];
```
HenkMutsaerts: Do we print each page as separate file?

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
...
+    PrintPath = fullfile(x.dir.xASLDerivatives, x.SUBJECT, PrintFile); 
+
+    print(primaryFig, PrintPath, '-dpdf', '-bestfit');
+
+    clear primaryFig ax
+end
+
+function xASL_sub_printBlock(input, x, pageFig, settings)
```
HenkMutsaerts: Also the subfunctions should have headers (at least to make them a bit readable)

and all functions are missing a lot of comments to make them readable

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
...
+    position = xASL_str2num(input.position);
+    size = xASL_str2num(input.size);
+    [settings.canvas, line] = xASL_sub_createNewCanvas(position, size, settings.canvas);
```
HenkMutsaerts: why is canvas in settings but position size line etc are not?

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
...
+function line = xASL_sub_PrintPatient(x, figure, line, settings)
+    %Check if participant data exists.
+    ExistsParticipants = xASL_adm_GetFileList(x.D.ROOT, 'participants.tsv');
+    PatientInfo = {};
+
+    if isempty(ExistsParticipants)
+        return
+    end
+
+    sParticipants = xASL_tsvRead(ExistsParticipants{1});
```
HenkMutsaerts: what does s mean?
n usually means number

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
...
+end
+
+function line = xASL_sub_PrintQC(json, x, figure, line, settings)
+    % Stop if module and field don't exists in output
+    if ~isfield(x.Output, (json.module)) && ~isfield(x.Output.(json.module), json.parameter)
+        return     
+    end
+
+    % Use field specific settings if they exists
+    settings.fontName = 'monospace';
```
HenkMutsaerts: I thought all settings were in the default config file, but some are in this code? Should we be consistent?

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
...
+        [range] = strsplit(json.range, '-');
+        if (TempValue < xASL_str2num(range{1})) || (TempValue > xASL_str2num(range{2}))
+            settings.color = 'r';
+        end
+        json.range = ['(' json.range ')'];
+    else
+        json.range = '';
+    end
+
+    if isnumeric(TempValue)
```
HenkMutsaerts: You can use xASL_num2str which deals with isnumeric and different formats of numbers

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
...
+            ImagePath = fullfile(x.dir.xASLDerivatives, x.SUBJECT, input.subjPath);
+        else
+            warning('xASL_sub_PrintImage didnt have a defined path') ;
+            return
+        end
+
+        if isfield(input, 'header')
+            header = input.header;
+        else
+            header = '';
```
HenkMutsaerts: Isn't it easier to default header to '' above, that avoids repetition

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
...
+        end
+    case 5
+        ImagePath = input;
+        header = '';
+    end
+
+    [canvas, ~] = xASL_sub_createNewCanvas(position(1:2), position(3:4), settings.canvas);
```
HenkMutsaerts: ~ can be removed. is only necessary when wanting to take the second or other output but ignoring the first

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
...
+    ax=axes('Position', canvas, 'Visible', settings.axesVisible, 'Parent', figure);
```
HenkMutsaerts: space missing with =

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
...
+    [img, ~, alphachannel] = imread(ImagePath);
```
HenkMutsaerts: Careful! imread and imshow are from the image processing toolbox. All ExploreASL functions should work without Matlab toolboxes

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
...
+    fg= imshow(img);
```
HenkMutsaerts: Best to avoid abbreviations, especially unreadable ones ;) Your code should be readable to others

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
...
+    [canvas, ~] = xASL_sub_createNewCanvas(position(1:2), position(3:4), settings.canvas);
+
+    if ~isfield(x.P, input.name)
+        warning (['could not print', input.name, 'check if nifti exists in ExploreASL/Derivatives/Population']);
+        return
+    end
+
+    ImIn = {x.P.(input.name)};
+    header = input.name;
+    ax=axes('Position', canvas, 'Visible', settings.axesVisible, 'Parent', figure);
```
HenkMutsaerts: space =

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
...
+    
+    if isfield(input, 'overlay')
```
HenkMutsaerts: I don't understand, a slice is a subset of a full image, and thus of an overlay. Does that mean that this function has multiple functions depending on its input? Because that would be confusing... Unless you have a single function that handles all inputs, but now it only handles overlay and slice inputs, we should avoid functions becoming convoluted.

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
...
+        [x.S.TraSlices] = xASL_sub_getSliceFromStruct(input.slice, 'TraSlice');
+        [x.S.CorSlices] = xASL_sub_getSliceFromStruct(input.slice, 'CorSlice');
+        [x.S.SagSlices] = xASL_sub_getSliceFromStruct(input.slice, 'SagSlice');       
+    else
+        x.S.TraSlices = [25, 50, 90];
+        x.S.SagSlices = [25, 50, 90];
+        x.S.CorSlices = [25, 50, 90];
+    end
+
+    img = xASL_vis_CreateVisualFig(x, ImIn, [], [], [], [], [], [], [], [], [], [], []);
```
HenkMutsaerts: avoid abbreviations

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
...
+function [slice] = xASL_sub_getSliceFromStruct(struct, name)
+
+    if isfield(struct, name) && ~isempty(struct.(name))
+        slice = str2num(struct.(name));
+    else 
+        slice = [];
+    end
+end
+
+function [newCanvas, line] = xASL_sub_createNewCanvas(position, size, oldCanvas)
```
HenkMutsaerts: should oldCanvas not be the first input argument, similar to newCanvas being the first output argument?

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
...
+
+function [slice] = xASL_sub_getSliceFromStruct(struct, name)
+
+    if isfield(struct, name) && ~isempty(struct.(name))
+        slice = str2num(struct.(name));
+    else 
+        slice = [];
+    end
+end
+
```
HenkMutsaerts: I would add some extra lines with %% ==================== between the sub functions to separate them clearer/more readable (at least I like to make functions readable :)

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
...
+function [newCanvas, line] = xASL_sub_createNewCanvas(position, size, oldCanvas)
+    newPosition = [oldCanvas(1) + position(1) * oldCanvas(3) oldCanvas(2) + position(2) * oldCanvas(4)];
```
HenkMutsaerts: There are comments lacking, e.g., what is a "canvas" in your code?

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
...
+            String = input.text;
+            if isfield(input, 'settings')
+                settings = xASL_sub_loadSettings(input.settings, settings);
+            end
+        otherwise
+            class(input)
+            error('xASL_sub_PrintText couldnt find string of printable text')
+    end
+
+    ax=axes('Position', line ,'Visible', settings.axesVisible, 'Parent', figure);
```
HenkMutsaerts: check spaces ax=axes

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
...
+end
+
+function [newCanvas, line] = xASL_sub_createNewCanvas(position, size, oldCanvas)
+    newPosition = [oldCanvas(1) + position(1) * oldCanvas(3) oldCanvas(2) + position(2) * oldCanvas(4)];
+    newSize = [oldCanvas(3:4) .* size];
+    newCanvas = [newPosition, newSize];
+    line = [newPosition(1)  newPosition(2) + newSize(2) newSize(1) 0];
+end
+
+function line = xASL_sub_PrintText(input, figure, line, settings)
```
HenkMutsaerts: above you have clearly defined types, and now you have an undefined first argument "input", which is handled differently depending on its content, I would avoid this, this makes for difficult to read code

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
...
+    ax=axes('Position', line ,'Visible', settings.axesVisible, 'Parent', figure);
+    text(0,0, String, 'Parent', ax, 'FontSize', settings.fontSize,'FontWeight', settings.fontWeight, 'Color', settings.color, 'FontName', settings.fontName, 'Interpreter', 'none', 'VerticalAlignment','top');
+    line = xASL_sub_NewLine(line, settings);
+end
+
+function line = xASL_sub_NewLine(line, settings)
+    line(2) = line(2) - (settings.lineSpacing) - (settings.fontSize * 0.001);
+
+    if line(2) < 0 
+        warning('No space left on page!');
```
HenkMutsaerts: On which page? Make warnings clear

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
...
+    elseif line(2) < settings.canvas(2) 
+        warning('Printing outside canvas, check block settings.');
```
HenkMutsaerts: Which canvas? Which block settings? Help the user

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
...
+function [figurecount] = xASL_sub_PrintHeader(header, figure, settings, position)
+
+    if isempty(header)
+        figurecount = settings.figurecount;
+        return
+    end
+
+    position(4) = 0;
+    settings.HorizontalAlignment = 'center';
+    figurecount = settings.figurecount + 1;
```
HenkMutsaerts: why is figureCount (check camelCase) defined twice, what is the role of "settings"? Make it readable

HenkMutsaerts (changes requested):
Very nice work! Some minor stuff, that are mainly about code, header, and comment consistencies, trying to keep everything readable. You currently have a lot of subfunctions where variables change names (e.g., "input") and depending on the content of the variable a different action will be triggered by a single function, that may not help readability and debugging. I would always go for simplicity, like you nicely do sometimes with things like

switch type
               A -> do this subfunction
               B -> do this subfunction

instead of doing the same subfunction for both A and B but inside the subfunction determining to what other subfunction A or B go

Another important issue: you use imshow and perhaps other functions from the Matlab image processing toolbox. Jan and I have put a lot of effort in not having any toolbox dependencies outside Matlab itself. This will ensure that most users can use ExploreASL (also those that didn't pay for the toolbox) and that the docker container stays lean.

Functions/Administration/xASL_adm_GeneratePDFConfig.m
jan-petr: OPTIONAL, DEFAULT = false.

Functions/Administration/xASL_adm_GeneratePDFConfig.m
jan-petr: Yes, it says what is done, but not for what purpose.

Functions/Administration/xASL_adm_GeneratePDFConfig.m
```diff
@@ -0,0 +1,83 @@
...
+        for iField=1:length(ModuleFields)   
+           if isfield(config, OutputFields{iModule}) && isfield(config.(OutputFields{iModule}), ModuleFields{iField})
+                OutputStruct(1).(OutputFields{iModule}).(ModuleFields{iField}) = config.(OutputFields{iModule}).(ModuleFields{iField});
+           elseif bPrintInvisible
+                OutputStruct(1).(OutputFields{iModule}).(ModuleFields{iField}) = StandardValue;
+           end
+        end
+    end
+
+    spm_jsonwrite(jsonPath, OutputStruct, struct('indent', ' '));
```
jan-petr: Please use xASL_io_WriteJson or xASL_io_ReadJson
the SPM version had several flaws and was also quite lenient in checking if JSONs were formally correct.
So that's why we wrote our wrapper function around jsonencode functions of Matlab.

Functions/Administration/xASL_adm_LoadPDFConfig.m
jan-petr: Is it just an alternative for jsonread? I guess it's much more than that. So this description should be improved. Also - lets not mention spm_jsonread, but just "reading JSON" or something like that.

Functions/Administration/xASL_adm_LoadPDFConfig.m
jan-petr: outputs are always optional ;) In a way, so no need to mention. Right?!

Functions/Administration/xASL_adm_LoadPDFConfig.m
jan-petr: it seems that "load JSON" section is 2 below instead of 1.

Functions/Administration/xASL_adm_LoadPDFConfig.m
jan-petr: it seems that "load JSON" section is 2 below instead of 1.

Functions/Administration/xASL_adm_LoadPDFConfig.m
```diff
@@ -0,0 +1,54 @@
...
+    warning('x was not specified, aborting');
+    return
+end
+
+ConfigPath = strcat(x.dir.xASLDerivatives, '/ConfigReportPDF.json');
+
+%% ------------------------------------------------------------------------
+%% 2. Load JSON file 
+if isfile(ConfigPath)
+    Config = spm_jsonread(ConfigPath);
```
jan-petr: replace with xASL_io_ReadJson

Functions/Administration/xASL_adm_LoadPDFConfig.m
```diff
@@ -0,0 +1,54 @@
...
+else
+    fprintf('configReportPDF.json does not exists, using default config \n');
+    standardPath = strcat(x.opts.MyPath, '/Functions/QualityControl/templateConfigReportPDF.json');
+    Config = spm_jsonread(standardPath);
```
jan-petr: xASL_io_ReadJson

Functions/QualityControl/templateConfigReportPDF.json
```diff
@@ -0,0 +1,338 @@
+{
```
jan-petr: please note that xASL_io_ReadJson strictly adheres to JSON standard - unlike spm_jsonread.

So don't be surprised if reading some JSONs throws an error.

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
+function [x,config] = xASL_qc_GenerateReport(x, subject)
+% xASL_qc_GenerateReport Print output QC
+%
+% FORMAT: xASL_qc_GenerateReport(x)
```
jan-petr: [, subject]) is missing here

jan-petr: also - is there is is there not output?

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
...
+%
+% INPUT:
+%   x           - structure containing fields with all information required to run this submodule (REQUIRED)
+%   subject     - subject name (OPTIONAL, default = x.SUBJECT)
+%
+% OUTPUT: n/a
```
jan-petr: function definition on the first line contains outputs.

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
...
+% EXAMPLE: xASL_qc_GenerateReport(x);
+% __________________________________
+% Copyright (C) 2015-2023 ExploreASL
+
+if ~usejava('jvm') % only if JVM loaded
+    fprintf('Warning: skipping PDF report, JVM missing\n');
+    config = NaN;
+    return;
+end
+
```
jan-petr, line 31: please add check for the first input parameter

Functions/QualityControl/xASL_qc_GenerateReport.m
jan-petr: Yes please. Cause not all users have them and hopefully, we will now be able to compile without toolboxed.

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,418 @@
...
+        [x.S.CorSlices] = xASL_sub_getSliceFromStruct(input.slice, 'CorSlice');
+        [x.S.SagSlices] = xASL_sub_getSliceFromStruct(input.slice, 'SagSlice');       
+    else
+        x.S.TraSlices = [25, 50, 90];
+        x.S.SagSlices = [25, 50, 90];
+        x.S.CorSlices = [25, 50, 90];
+    end
+
+    img = xASL_vis_CreateVisualFig(x, ImIn, [], [], [], [], [], [], [], [], [], [], []);
+    fg = imshow(img);
```
jan-petr: again imshow

jan-petr (changes requested):
Some minor changes needed.

Functions/Administration/xASL_adm_GeneratePDFConfig.m
maartenhammer: Old File removed

Functions/Administration/xASL_adm_GeneratePDFConfig.m
maartenhammer: Old File removed

Functions/Administration/xASL_adm_GeneratePDFConfig.m
maartenhammer: Old File removed

Functions/Administration/xASL_adm_GeneratePDFConfig.m
maartenhammer: Old File removed

Functions/Administration/xASL_adm_GeneratePDFConfig.m
maartenhammer: Old File removed

maartenhammer: But yes, replaced on different locations!

Functions/Administration/xASL_adm_GeneratePDFConfig.m
maartenhammer: Old File removed

Functions/Administration/xASL_adm_GeneratePDFConfig.m
maartenhammer: Old File removed

maartenhammer: But good to know for future projects

Functions/Administration/xASL_adm_GeneratePDFConfig.m
maartenhammer: Changed in other files

Functions/Administration/xASL_adm_LoadPDFConfig.m
maartenhammer: Description improved, indeed not just an alias. It has some Qc/warnings/Error handling.

Functions/Administration/xASL_adm_LoadPDFConfig.m
maartenhammer: Fixed descriptions

Functions/Administration/xASL_adm_LoadPDFConfig.m
maartenhammer: This was a little redundant, so I removed it instead. 
Used to be a safeguard as to not return an empty config, which is no longer possible.

Functions/Administration/xASL_adm_LoadPDFConfig.m
maartenhammer: Replaced and fixed

Functions/Administration/xASL_adm_LoadPDFConfig.m
maartenhammer: Fixed description and exists is faster and better!

Functions/Administration/xASL_adm_LoadPDFConfig.m
maartenhammer: Resolved and Fixed

Functions/Administration/xASL_adm_LoadPDFConfig.m
maartenhammer: Changed to be more Specific

Functions/Administration/xASL_adm_LoadPDFConfig.m
maartenhammer: Changed to be more Specific

Functions/QualityControl/xASL_qc_GenerateReport.m
maartenhammer: Debugging only, removed.

Functions/Administration/xASL_adm_LoadPDFConfig.m
maartenhammer: Replaced and resolved, works exellently.

Functions/Administration/xASL_adm_LoadPDFConfig.m
maartenhammer: same as Above

Functions/QualityControl/templateConfigReportPDF.json
maartenhammer: It didn't!

Functions/QualityControl/xASL_qc_GenerateReport.m
maartenhammer: Resolved and Added

Functions/QualityControl/xASL_qc_GenerateReport.m
maartenhammer: Yes, but qc-values specifically I prefer to use a different font to the standard default. 
Monospace ensures that the numbers line up better regardless of what comes before it. 
I have explained it better in the comments.

Functions/QualityControl/xASL_qc_GenerateReport.m
maartenhammer: XASL_tsvRead generates a struct from the tsv changed it to structParticipants

Functions/QualityControl/xASL_qc_GenerateReport.m
maartenhammer: Yes, limitation of matlab printing.

Functions/QualityControl/xASL_qc_GenerateReport.m
maartenhammer: Settings arent printed to the page, which is why theres a difference in name.

Functions/QualityControl/xASL_qc_GenerateReport.m
maartenhammer: fixed and added better hint

Functions/QualityControl/xASL_qc_GenerateReport.m
maartenhammer: This is just to ignore any nonStruct objects passed to this function this can happen in normal operation because the every top-level Json-field is at some point passed to this. I could change the caller to be more strict if wanted.

Functions/QualityControl/xASL_qc_GenerateReport.m
maartenhammer: I decided to check this, (discussing with Jan in Whatsapp) I have checked many times, but Imshow and Imread don't trigger any toolbox licenses. with `licence('inuse')` I regularly check if any toolboxes are triggered and this function does not trigger any. Using Matlabs built in dependency report you can also see no external or toolbox dependencies. 
image

Functions/QualityControl/templateConfigReportPDF.json
maartenhammer: Yes, if settings are specified for something they are only for that specific field. If you want to swap the defaults you should add a "type":"settings" json field.

Functions/QualityControl/xASL_qc_GenerateReport.m
maartenhammer: Added

Functions/QualityControl/xASL_qc_GenerateReport.m
maartenhammer: Changed, but this does make the function names very long.  Will resolve all other occurances of this issue to keep conversation condensed.

Functions/QualityControl/xASL_qc_GenerateReport.m
maartenhammer: I am not 100% sure what is meant here.
Do you mean change the variables to <type><Identifier> (figPrimary instead of primaryFig) or do you mean in order or definitions?
For both of these variables only one exists at a time.

Functions/QualityControl/xASL_qc_GenerateReport.m
maartenhammer: Moved all subfunctions to xASL_qc_ParsePdfConfig (maybe xASL_adm_ instead?) and renamed the subfunctions.

maartenhammer: Added description of the JsonParser, including defining undefined variables.

Changed figure to currentFigure, settings to settingsPDF.

Functions/QualityControl/xASL_qc_GenerateReport.m
maartenhammer: removed clear, replaced in certain instances with cla before defining images to have a clear figure.

Functions/QualityControl/xASL_qc_GenerateReport.m
maartenhammer: Yes

Functions/QualityControl/xASL_qc_GenerateReport.m
maartenhammer: This is mostly to avoid duplication, as of right now the watermark (This is generated by ExploreASL on the bottom of the page) is handled by the same function as the function that parses the pdf report. We can discuss this later.

Functions/QualityControl/xASL_qc_GenerateReport.m
maartenhammer: This isnt an abreviation but a matlab reserved function I believe, I don't ever use the ax/fg variables.

Functions/QualityControl/xASL_qc_GenerateReport.m
maartenhammer: As of right now this is information not know to the program in this subfunction, same of comment above.

Functions/Administration/xASL_adm_GeneratePDFConfig.m
maartenhammer: Fixed

Functions/QualityControl/xASL_qc_GenerateReport.m
maartenhammer: Added!

Functions/QualityControl/xASL_qc_GenerateReport.m
maartenhammer: Position and size are content specific parameters to draw new canvasses, added more context to it.

Functions/QualityControl/xASL_qc_GenerateReport.m
maartenhammer: Added

Functions/Administration/xASL_adm_LoadPDFConfig.m
jan-petr: OK

Functions/QualityControl/xASL_qc_GenerateReport.m
jan-petr: I confirm that - imread is not in Image processing toolbox anymore. Imshow is in the toolbox (or at least stated on Matlab webpages), but  mysteriously, this doesn't trigger the license demand. 

Both functions are in toolbox/matlab/images and not in toolbox/images

So it is all fine. We can use imshow and imread without restrictions!

Functions/QualityControl/templateConfigReportPDF.json
```diff
@@ -0,0 +1,338 @@
...
+                    "slice": {
+                        "TraSlice": "[25 75]",
+                        "CorSlice": "[25 75]",
+                        "SagSlice": "[]"
+                    },
+                    "overlay": {
+                        "Pop_Path_SliceGradient": {
+                            "contour": "true",
+                            "opacity": "0.3",
+                            "color": "r"
```
maartenhammer, line 54:
```
"type": "contentImage3D",
                    "name": "Pop_Path_rT1",
                    "position": "[0.70 0.35]",
                    "size": "[0.3 0.3]",
                    "slice": {
                        "TraSlice": "[25 75]",
                        "CorSlice": "[25 75]",
                        "SagSlice": "[]"
                    },
                    "overlay": [{
                            "name":"Pop_Path_SliceGradient",
                            "contour": "true",
                            "opacity": "0.3",
                            "color": "r"
                            },{
                            "name":"Pop_Path_<layer2>",
                            "contour": "true",
                            "opacity": "0.3",
                            "color": "r"
                            }]

```

Functions/QualityControl/templateConfigReportPDF.json
maartenhammer: Change to contentImage2D, contentImage3D, contentValues, contentPatients, metaPage, metaTextSettings, metaBox/metaBlock

Functions/QualityControl/templateConfigReportPDF.json
maartenhammer: Change Settings -> TextSettings for clarity

Functions/QualityControl/templateConfigReportPDF.json
maartenhammer: x.SUBJECT and x.SESSION (should be x.RUN in BIDS) and x.visit (should be x.SESSION in BIDS)

HenkMutsaerts:
Just tested this:

` [x, config] = xASL_qc_GenerateReport(x[, subject])` as you state above is not working, there is only a single x output.

Then I saw 2 minor bugs which I fixed.

Finally, the current config gives an example, can you replace the default config with one that creates a similar PDF as we had? Then we can improve it from there, but then at least I can already check how it works.

Also, now if there are any pre-existing xASL reports, it will delete them; this behavior should be disabled. Only, if there already is a PDF existing with an identical filename, it should perhaps rename the new file (I wouldn't touch the old/pre-existing file).

Also, this function now is dependent on `x.P`, so if a filename exists but is not defined in `x.P`, it will not be printed. I would instead just use relative filenames, not `x.P` fields (those fields are only useful within ExploreASL).

maartenhammer:
> Just tested this:
> 
> ` [x, config] = xASL_qc_GenerateReport(x[, subject])` as you state above is not working, there is only a single x output.
> 
> Then I saw 2 minor bugs which I fixed.
> 
> Finally, the current config gives an example, can you replace the default config with one that creates a similar PDF as we had? Then we can improve it from there, but then at least I can already check how it works.
> 
> Also, now if there are any pre-existing xASL reports, it will delete them; this behavior should be disabled. Only, if there already is a PDF existing with an identical filename, it should perhaps rename the new file (I wouldn't touch the old/pre-existing file).
> 
> Also, this function now is dependent on `x.P`, so if a filename exists but is not defined in `x.P`, it will not be printed. I would instead just use relative filenames, not `x.P` fields (those fields are only useful within ExploreASL).

Excellent, the changes added by you seem to improve stability when parameters are missing, as well as some householding data. 

I will also improve this by including [the definitions here](https://ars.els-cdn.com/content/image/1-s2.0-S1053811920305176-mmc1.pdf) as default for units and names if no alternatives are given.

jan-petr (approved):
Good work. No further comments.

maartenhammer:
Last conversations have been handled:
Translation layer added that translates ExploreASL Legacy to BIDS when turned on
Settings has been renamed to TextSettings for clarity
Image has been renamed to Image2D 
scan has been renamed to Image3D
A secondary categorizer has been added (instead of a json item being Type = image2D, we now have category = "content" , type = "image".
This is to distinguish metadata json-objects (pages, blocks, settings) and content json objects(like image2d, image3d, text, QCValues).
Added translation layer to always automatically translate [some of the qc parameters here](https://ars.els-cdn.com/content/image/1-s2.0-S1053811920305176-mmc1.pdf). Not all have been filled in, discuss with either Jan/Henk Jan which should be filled in. 
Changed default print to look more like the example given in [figure 3 of the ExploreASL paper](https://www.sciencedirect.com/science/article/pii/S1053811920305176) .
[New default now looks like this](https://github.com/ExploreASL/ExploreASL/files/13441564/xASL_Report_Example_1.pdf)

Edit: 
Discussion with Henk-Jan on thursday.
Possible additions:
Pdfconfig json generator that reads x.output and populates a PDFconfig.json instead of using a default template in Functions->Quality Controll, add it to wrapper.
Add text elipses (...) when qc_param exceeds certain length. (added)

maartenhammer:
While adding text wrapping I've noticed that xASL_num2str does not convert logical to str, which the matlab native num2str does.
A minor change to xASL_num2str is added in [3e1290d](https://github.com/ExploreASL/ExploreASL/pull/1475/commits/3e1290d25693845dacfbfd9814bbbd3439430f2d) which fixes this, as I assume this is a bug instead of intended behaviour.  (Discussed with Jan over whatsapp).

jan-petr (changes requested):
I am merely blocking the merging, because some commit messages need fixing ;)

Functions/QualityControl/qc_translation.tsv
```diff
@@ -0,0 +1,112 @@
+parameter	Alias	Unit	Range
```
HenkMutsaerts: As we now discussed, please add a qc_translation.txt with a brief explanation of what this file does, and do a spelling check. I can complete the aliases etc.

Functions/QualityControl/templateConfigReportPDF.json
```diff
@@ -0,0 +1,695 @@
+{
+    "globalSettings": {
+        "category": "metadata",
```
HenkMutsaerts: Can we create this dynamically by default, based on the QC values in x.output? To be overridden if the user provides their own version?

HenkMutsaerts (changes requested):
Nice work! Some questions as we discussed, to keep the current functionality of generating the PDF dependent on what was analyzed in ExploreASL.

E.g, if you have analyzed 2 ASL runs, you want to show 2 ASL runs. If you have analyzed only T1w and FLAIR, you want to only show them.

So while there might be a static part of the template/settings file, it would be nice if these things can also be generated dependently.

Functions/QualityControl/templateConfigReportPDF.json
maartenhammer: We can now.

Functions/QualityControl/qc_translation.tsv
maartenhammer: For these translation tables I'd rather keep it inside the exploreASL directory, as these wouldn't ever change. The only difference is which ones you use. Generating this in a subject/study directory would just generate a bunch of useless Tsv files for the user, and in the JSON you can override any and all of these anyway. 

Looking up BIDS definitions I do agree on.

Functions/Administration/xASL_adm_LoadPdfConfig.m
```diff
@@ -0,0 +1,58 @@
...
+end
+
+if nargin<2 || isempty(configPath)
+    configPath = fullfile(x.dir.xASLDerivatives, 'configReportPDF.json');
+end
+
+
+%% ------------------------------------------------------------------------
+%% 2. Load JSON file 
+if (exist(configPath, 'file') == 2)
```
HenkMutsaerts: () brackets are not required, and you don't have to test if the output of exist equals 2, this is what 'file' does

Functions/QualityControl/qc_translation.tsv
HenkMutsaerts: ? Did you do/check this?

Functions/QualityControl/templateConfigReportPDF.json
HenkMutsaerts: ?

Functions/Administration/xASL_adm_GeneratePdfConfig.m
```diff
@@ -0,0 +1,140 @@
+function [config] = xASL_adm_GeneratePdfConfig(x, subject, bOverwrite)
```
HenkMutsaerts: Would rename all QC-related functions to xASL_qc_ and move to the QualityControl folder, just easier to have these all sorted.

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,122 @@
...
+%   subject     - subject name (OPTIONAL, default = x.SUBJECT)
+%
+% OUTPUT: 
+%
+%   x           - structure containing fields with all information required to run this submodule
+% OUTPUTFILE:
+%   xASL_Report_SubjectName.pdf - printed PDF rapport containing QC images & values
+%
+% -----------------------------------------------------------------------------------------------------------------------------------------------------
+% DESCRIPTION:  This function generates a PDF report based on a predifined Json configuration file.
```
HenkMutsaerts: predefined

Functions/QualityControl/xASL_qc_GenerateReport.m
```diff
@@ -0,0 +1,122 @@
+function [x, settings] = xASL_qc_GenerateReport(x, subject)
+% xASL_qc_GenerateReport Generates a PDF report based on a predifined Json configuration file
```
HenkMutsaerts: predefined

Functions/QualityControl/xASL_qc_GenerateReport.m
HenkMutsaerts: Why did you resolve this conversation? Did you understand his comment (it is about line 4, not line 1)?

Functions/QualityControl/xASL_qc_ParsePdfConfig.m
```diff
@@ -0,0 +1,675 @@
...
+    if nargin < 4 || isempty(SymbolToFill)
+        SymbolToFill = ' ';
+    end
+
+    %% Create default string
+    resultText = repmat(SymbolToFill, 1, textWidth);
+    
+    % Pad spaces
+    if ~isempty(textToPrint)
+        textToPrint = [textToPrint ' '];
```
HenkMutsaerts: Here we sometimes have char arrays that are not only horizontal, because a parameter can have multiple values apparently, and then the values should be concatenated in a string/char or just the first reported (depending on the length).

E.g., I had here:

textToPrint =

  2×4 char array

    '1710'
    '3142'

First, I would never use char arrays, can we get rid of them?

Because if we have 

textToPrint2 = {'1710';'3142'}

textToPrint2 =

  2×1 cell array

    {'1710'}
    {'3142'}

this is easy to concatenate iterating over cells (2 elements)
With a char array, this example would be 8 elements.

Char arrays is something new of Matlab I think, at least Jan and I never used them.

Or if it becomes very tricky to get rid of char arrays, perhaps you can think of a different solution?

HenkMutsaerts (changes requested):
Some minor comments, and a bug while testing ExploreASL's TestDataSet

Functions/QualityControl/templateConfigReportPDF.json
HenkMutsaerts: Maarten made this dynamic already, so this is only an example json.
@maartenhammer then please rename this to append the suffix "example" to this file name, so this is clear.

Functions/QualityControl/templateConfigReportPDFAlternative.json
```diff
@@ -0,0 +1,457 @@
+{
```
HenkMutsaerts: Same for all files, if they are now dynamically created, please add the suffix "example" or something so this is clear.

HenkMutsaerts (changes requested):
Few additional things.

Functions/Administration/xASL_adm_GeneratePdfConfig.m
maartenhammer: Okay I will do that

Functions/Administration/xASL_adm_LoadPdfConfig.m
maartenhammer: Fixed

Functions/QualityControl/templateConfigReportPDF.json
maartenhammer: As discussed, this is now possible using the xASL_adm_GeneratePDFConfig function, which is called and the default behaviour of GenerateReport now. 
It essentially dumps all x.Output into a json configuration with minimal formatting, each series is it's own page due to matlab limitations. (Eg, an ASL pdf and T1 pdf are generated.)

Functions/QualityControl/xASL_qc_GenerateReport.m
maartenhammer: fixed

Functions/QualityControl/xASL_qc_GenerateReport.m
maartenhammer: fixed

Functions/QualityControl/qc_translation.tsv
maartenhammer: Yes, Bids does not have shortnames other than the standard bids names for parameters, I've added certain BIDS specific terms and renamed this to qc_glossary as a more "clear" name as for what this is. I've also spell checked and added a new row of "Descriptions", which can be used for longer descriptions, I've shortened the aliasses to make printing less ugly. 
Added a .txt file as well.

Functions/QualityControl/templateConfigReportPDFAlternative.json
maartenhammer: Changed this to be "example"

maartenhammer:
Two force pushed to incorporate two minor appends to the last commit.

Functions/QualityControl/xASL_qc_ParsePdfConfig.m
maartenhammer: I have decided to leave it as Char as it is, I've reworked the way it reads values, and it now horizontally concatenates them. Sadly for chars a simple var= var(:)' does not work due to indexing of chars, so I used append instead. This can become slow if a **very large** amount of values need to be concatenated.  This can be relieved by pre-defining the character array length and replacing values inside a predefined array, but realistically it won't ever matter for this purpose.

HenkMutsaerts (approved):
My test worked fine. I have added some wishes to the follow-up issue #1476 (in which we actually implement this in ExploreASL).

jan-petr (approved):
OK

PR #1487, opened 2023-08-30
#1177 xASL_dev_MakeStandalone removed -C tag for Singularity Compatib…

### LINKED ISSUE #1177: -C flag in compilation incompatible with Singularity
I was trying to run the xASL official Docker image in Singularity, which produced root access errors (Singularity runs as a non-root user). At runtime, MCR wants to extract the contents of xASL_latest.CTF into a new folder in the same parent directory, under /opt/, which is read-only.

This can be fixed by dropping the '-C' flag during tool compilation, which allows the user to select, via environment variable MCR_CACHE_ROOT, an extraction destination like /tmp/, which has user write access.

Is there any reason for the -C flag to be used in compilation? Could it be dropped in future official releases?

maartenhammer:
closes #1177


jan-petr (approved):
OK.

PR #1494, opened 2023-09-16
Closes #1483 xASL_quant_Hct2BloodT1: add warning for missing Hct

### LINKED ISSUE #1483: Provide study-specific default hematocrit values; missing hct warning
### Description

When a hematocrit value is missing (entered as n/a in participants.tsv), it treats this as a missing T1-blood and uses the default T1-blood value. This was fixed previously in #1481 , so missing values won't crash the processing.

However, this literature value of T1-blood is different from what our T1-blood-hct model would provide for a default Hct value. So we have to deal with this, or else subjects with missing Hct will have biased quantification.

There are two options:
1) We use the model for T1-blood calculation and use a default Hct value.
2) We calculate Hct as study or group Hct mean.
3) We just report this as a warning and tell users to provide a default value or we will enter a default T1-blood-Hct value as a parameter in dataPar.json

I would just add a warning and otherwise not change this behavior. It is a parameter added by the user in participants.tsv and creating study averages etc can be easily done by the user.

### Tasks

- [x] xASL_quant_Hct2BloodT1: Issue warning if hematocrit is detected but not a numerical usable value
I now added the following warning:
WARNING: POSSIBLE CBF BIAS

Please be warned that if you are correcting CBF values for hematocrit in your study,
this scan will NOT be corrected and ExploreASL will use the default blood T1 value
This can create a significant CBF bias

### Release notes

Issue a warning when Hct-correction is skipped due to an irregular Hct value.

HenkMutsaerts:
### Linked issue

Closes #1483

### DIFF
--- a/Functions/Quantification/xASL_quant_Hct2BloodT1.m
+++ b/Functions/Quantification/xASL_quant_Hct2BloodT1.m
@@ -42,19 +42,23 @@ if nargin<3 || isempty(B0)
     B0 = 3; % 3T field strength
 elseif length(B0)~=1
     warning('Incorrect B0 specified, skipping');
+    xASL_quant_Hct2BloodT1_printMissingHctWarning;
     return;
 end
 if nargin<2 || isempty(Y)
     Y = 0.97;
 elseif length(Y)~=1
     warning('Incorrect Y specified, skipping');
+    xASL_quant_Hct2BloodT1_printMissingHctWarning;
     return;
 end
 if nargin<1 || isempty(Hematocrit) || length(Hematocrit)~=1
     warning('Incorrect Hematocrit specified, so not using this');
+    xASL_quant_Hct2BloodT1_printMissingHctWarning;
     return;
 elseif isnan(Hematocrit)
     warning('Hematocrit value was missing (set to NaN), so not using this');
+    xASL_quant_Hct2BloodT1_printMissingHctWarning;
     return;
 end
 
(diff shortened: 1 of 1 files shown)


Functions/Quantification/xASL_quant_Hct2BloodT1.m
```diff
@@ -109,4 +115,22 @@
...
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+function xASL_quant_Hct2BloodT1_printMissingHctWarning(x)
+%xASL_quant_Hct2BloodT1_printMissingHctWarning We may print this message in
+%different cases, which is why this subfunction comes in handy
+
+
+fprintf('\n\n\n');
+fprintf('%s\n', 'WARNING: POSSIBLE CBF BIAS');
+fprintf('\n');
```
jan-petr, line 130: I would print the subject name?

jan-petr (changes requested):
All good, but do we want to print the name of the subject with the warning, so that the user doesn't have to go through the log to locate that?

Functions/Quantification/xASL_quant_Hct2BloodT1.m
HenkMutsaerts: No that's going to be too much hassle. It is printed in the same run, and `xASL_quant_Hct2Blood` is a function without `x`. Shouldnt' be too difficult for the user to find the subject name because it is printed all over right (folder and file names).

jan-petr (approved):
OK

PR #1498, opened 2023-09-23
Closes #1495 bids2 legacy parallelization

### LINKED ISSUE #1495: Code simplification for BIDS2Legacy parallelization
### Description

Simplification of code, used for revamping the BIDS2Legacy parallelization

### Tasks

- [x] `xASL_wrp_BIDS2Legacy`: instead of running all subject-session combinations from the BIDS table, process a single subject

- [x] Check what the clean-up does, should this move elsewhere? -> I checked this, was obsolete. This fixed the saving of log files at incorrect locations, but the log files are now already saved at the appropriate location in the import.
- [x] Move non-subject specific code from `xASL_wrp_BIDS2Legacy` to `xASL_init_BIDS2Legacy`. Added benefit is that variables go in `x.modules.bids2legacy`, which helps with provenance.
- [x] Code simplification: unlike other modules, `xASL_module_BIDS2Legacy` doesn't have multiple steps so there is no need for a separate wrapper, `xASL_wrp_BIDS2Legacy` could be fully embedded/wrapped in `xASL_module_BIDS2Legacy`.

- [x] Simplify ExploreASL_Initialize:
sorted subfunctions
improved comments
integrate separate xASL_init_*.m functions that are only used by ExploreASL_Initialize: 
  xASL_init_printSettings, 
  xASL_init_checkDatasetRoot, xASL_init_DetermineRequiredPaths
  xASL_init_DefineIndependentSettings
merge inside ExploreASL_Initialize:
xASL_init_DeployedModeHandling->xASL_init_GetMasterScript
xASL_init_GetMasterScript & xASL_init_GetMyPath -> xASL_init_GetMainExploreASLFolder

- [x] Simplify xASL_init_DataLoading
integrate separate xASL_init_*.m functions that are only used by ExploreASL_Initialize: 
xASL_init_DefineDataDependentSettings
xASL_init_LoadDataParameterFile.m
xASL_init_DefineStudyData -> THIS IS A BIG ONE
xASL_init_RemoveLockDirs.m
xASL_init_PrintCheckSettings.m
merged inside xASL_init_DataLoading:
xASL_init_DefinePaths.m -> xASL_init_DefineDataDependentSettings.m
xASL_init_Toolboxes.m -> xASL_init_DefineDataDependentSettings.m
PM: xASL_init_Toolboxes only loads SPM, and may go elsewhere as it is not data-dependent

- [x] Separate processing initialization functions:
It boils down to the addition of the import module, there is now a lot in common between the processing modules but not with import (and it shouldn't, to avoid bugs). The functionality of the processing and import modules are very different.
So I would propose to structure:
/Modules/Import
/Modules/Processing/Structural
/Modules/Processing/ASL
/Modules/Processing/Population
Because this allows to have initialization functions that are processing only, in the processing folder.

- [x] `xASL_imp_DetermineSubjectStructure` sort of creates a list of subjects and is used for both import and processing:
* import DICOM2NIFTI: `xASL_imp_DetermineStructureFromSourceData`
* import NIFTI2BIDS: `xASL_imp_DetermineStructureFromTempData`
* import BIDS2LEGACY: `xASL_imp_DetermineStructureFromRawData`

I have now moved the processing part out of this function.

* `xASL_imp_DetermineStructureFromRawData` -> replaces `xASL_imp_DetermineSubjectStructure` in `xASL_init_Process`

- [x] Move processing initialization functions to `/Modules/Module_Processing/`
* For BIDS2Legacy (run within ExploreASL_Process)
xASL_init_BIDS2Legacy.m
* For processing (run within ExploreASL_Process)
xASL_init_Process.m
* xASL_init_DataLoading & xASL_init_DefineStudyData
* xASL_init_LoadMetadata.m (in xASL_init_DefineStudyData & xASL_module_Population)
xASL_init_FileSystem.m
* xASL_init_LongitudinalRegistration.m
* xASL_init_MapsAndAtlases.m
xASL_init_VisualizationSettings.m
xASL_init_DefaultEffectiveResolution.m
xASL_init_LabelColors.mat (in Population module)

### Release notes

Report with #1275. Allow parallelization of BIDS2Legacy.

HenkMutsaerts:
### Linked issue

Closes #1495


Modules/xASL_wrp_BIDS2Legacy.m
```diff
@@ -1,196 +0,0 @@
...
-
-%% 6. Clean up
-xASL_bids_BIDS2Legacy_CleanUp(x.dir.DatasetRoot);
-
-
-end
-
-
-
-%% Clean-Up subfunction
```
jan-petr, line 162: Why do we delete this function without replacement? Don't we want to merge and save the log files?

Modules/xASL_wrp_BIDS2Legacy.m
```diff
@@ -65,12 +65,17 @@
...
-for iSubjSess=1:numel(x.modules.bids2legacy.BIDS.subjects) 
-    % Iterate over x.modules.bids2legacy.BIDS.subjects (indices that include both subjects & sessions)
-    % so  1 subject  6 session/visits, will give numel(x.modules.bids2legacy.BIDS.subjects)=6
-    % and 6 subjects 1 session/visits, will give numel(x.modules.bids2legacy.BIDS.subjects)=6
-    xASL_TrackProgress(iSubjSess, nSubjects);
+%% 2. Define Subject
+subjectBIDS = ['sub-' x.SUBJECT];
+
+
+indicesCurrentSubject = find(cellfun( @(y) strcmp(y, subjectBIDS), {x.modules.bids2legacy.BIDS.subjects.name}));
```
jan-petr: Just to understand. Previously, we converted all subjects here. But probably this function was called for each subject.

So previously - for all subjects we tried converting all subjects resulting this super-long runtimes.
Now - we run this function for all subject, but convert only one subject at a time.

Am I right? It is perfect that you fix this bug. I just want to make sure that this is the reason.

Modules/xASL_wrp_BIDS2Legacy.m
```diff
@@ -143,8 +145,18 @@
 if ~isempty(ListASL4D)
     for iList=1:numel(ListASL4D)
         xASL_bids_parseM0(ListASL4D{iList});
-        [~, currentNifti] = fileparts(ListASL4D{iList});
-        fprintf('M0 parsed for subject %s image %s ...\n', SubjectID, currentNifti);
+        [currentFolder, currentNifti] = xASL_fileparts(ListASL4D{iList});
+        [~, currentFolder] = fileparts(fileparts(currentFolder));
+
+        % Get session
+        sessionIndex = regexp(currentFolder, '_\d+');
```
jan-petr: Better look for '_\d+$'

jan-petr (changes requested):
First part of a review. Just checking if changes are displayed globally...

Modules/xASL_module_BIDS2Legacy.m
```diff
@@ -1,51 +1,81 @@
-function [result, x] = xASL_module_BIDS2Legacy(x)
-%xASL_module_BIDS2Legacy BIDS2LEGACY conversion script which calls xASL_wrp_BIDS2Legacy.
+function [result, x] = xASL_module_BIDS2Legacy(x, bOverwrite, bVerbose)
+%xASL_module_BIDS2Legacy Convert BIDS rawdata to ExploreASL legacy format
 %
-% FORMAT: [x] = xASL_module_BIDS2Legacy(x);
+% FORMAT: [x] = xASL_module_BIDS2Legacy(x [,bOverwrite]);
```
jan-petr: Please add bVerbose

Modules/xASL_module_BIDS2Legacy.m
```diff
@@ -1,51 +1,81 @@
...
         x.mutex.AddState('010_BIDS2LEGACY');
     elseif x.mutex.HasState('010_BIDS2LEGACY')
         fprintf('%s\n', ['BIDS2Legacy already done, skipping ' x.SUBJECT '.    ']);
     end
 
     
-    %% 2. Finalize and unlock mutex for this module
-    x.mutex.AddState('999_ready');
-    x.mutex.Unlock();
+    %% 6. Finalize and unlock mutex for this module
```
jan-petr, line 189: Why you add 6 here - that's not transparent at all. You have steps 1 and 6 in a function which is really confusing. I can undertstand that there are some lock-files in-between, but these are not here and the state 6 has no lock-file. So please change back to 2.

jan-petr: Now I get it. You copied the rest in a later commit :) Not sure if your tip to review commit by commit was worth it :)

Modules/Module_Processing/xASL_init_BIDS2Legacy.m
```diff
@@ -95,7 +98,31 @@
...
+    fprintf('%s\n', [x.dir.xASLDerivatives ' already exists']);
+    fprintf('%s\n', 'Note that all pre-existing derivative subject folders will be overwritten,');
+    fprintf('%s\n', 'unless BIDS2Legacy lock files already exist for a subject');
+else
+    xASL_adm_CreateDir(x.dir.xASLDerivatives);
 end
+
+
+%% 6. Load BIDS configuration for file renaming
+x.modules.bids2legacy.bidsPar = xASL_bids_Config;
```
jan-petr, line 117: OK. That is nice that you put bidsPar into a parameter within x. 
But then we have a lot of functions that use this bidsPar that get both bidsPar and x as an argument. So they will have doubled information.

The only solution is:
1. All functions that accept bidsPar as an input argument together with x will have to be rewritten.
2. The catch is that not all of these are inside bids2legacy, so we will have to check all calls of xASL_bids_Config and make sure that this is saved within the x-struct.

So that is a lot of extra work to be done. I can do that if you want but we have to be first careful that we don't leave this inconsistent code (cause we are cleaning, so we should clean properly) and we have to be careful that this is done correctly or it will leave us code that doesn't run...

Modules/xASL_wrp_BIDS2Legacy.m
```diff
@@ -151,46 +151,5 @@
 end
 
 
-%% 6. Clean up
```
jan-petr, line 154: What happened with the cleaning?

jan-petr (changes requested):
Some questions.

jan-petr (changes requested):
Unit tests fail in 
xASL_ut_master_xASL_ExploreASL line 442.

jan-petr:
> Unit tests fail in xASL_ut_master_xASL_ExploreASL line 442.

Then it crashes somewhere in the xASL_module_import.

jan-petr:
> > Unit tests fail in xASL_ut_master_xASL_ExploreASL line 442.
> 
> Then it crashes somewhere in the xASL_module_import.

Got it - if we run import as [0 0 1] = deface, then it somehow turns this into [1 0 0] for a weird reason and then it obviously crashes.

But I still have no idea why the import vector turns from [0 0 1] to [1 0 0].

HENK:
When I test this with our testdataset that has only `rawdata`, it keeps the x.opts.bImport at [0 0 1], even when I input [0 0 1 0] and it shortens.
It does crash because it doesn't have a studyPar, but let's assume users only run the import module completely :)

It is also works for me when running flavor GE_PCASL_2DEPI_HD23.0_1, but this flavor has no anat ;)

Then I tried Siemens_PASL_2DEPI_VB15A_ADNI2_941_S_4764_1 which crashed for me. I improved the warning of the last element of folderHierarchy (see last commit) and fixed `Siemens_PASL_2DEPI_VB15A_ADNI2_941_S_4764_1/sourceStructure.json`.

Now testing `Philips_PASL_2DEPI_5.4.1.0_ADNI3_006_S_6681_1`, with [1 1 1], this crashes on my MacBook (because the dcmTK compilation doesn't work and I have on purpose not installed the image processing toolbox) but runs fine on Luna.

It only crashes when you have already ran [1 1 0] and then rerun ExploreASL with [0 0 1].
I can fix this, but this potentially provides problems when running the import in parallel (which e.g., Maurice does with his GUI), because the subjects found in sourceStructure could differ from the subjects found in rawdata, and then problem A) only some are defaced but not all, problem B) the defacing finds subjects that don't have anatomical NIfTIs converted yet, and probably we can come up with more problems. In other words, defacing's dependency on the dcm2bids conversion should stay intact IMHO. If you are OK with my reasoning, I can add a warning for when [0 0 1] is called?

-> I have reverted xASL_imp_DetermineSubjectStructure, so it now again loads rawdata when import is ran with [0 0 1] (defacing). I also added a warning (last comment).

jan-petr:
So ExploreASL runs when you run defacing. Especially if you run it as a separate step.
The reason is that defacing is run on rawdata. And it can't find any subjects there and it crashes.

The real reason is that after DICOM->TEMP->NII, ExploreASL needs to read patient structure from rawdata so that it could anonymize the rawdata.

But you have moved the routines for reading patient structure from rawdata away and put it to the processing module.
So defacing can't work.

There are multiple solutions to that. 
1. Move defacing to processing -> I wouldn't do that.
2. Move the patient structure reading where it was.
3. Program the patient structure reading again and differently.

Up to you to decide. I would just put the function back.

Modules/Module_Processing/xASL_init_BIDS2Legacy.m
HenkMutsaerts: I see. This was mainly to make the code easier and avoid overlap between the import and processing modules. But more importantly, `bids_config` was now run at every xASL_module_BIDS2Legacy, and instead my edit runs it only once. This requires to put it in x (or to create many new bidsPar variables in `xASL_Iterate`. Just checked, and `xASL_module_BIDS2Legacy` provides this `x.modules.bids2legacy.bidsPar` instead of bidsPar to `xASL_bids_BIDS2Legacy_ParseModality`, which throughputs it to `xASL_bids_BIDS2Legacy_ParseScanType`. The latter two functions are unchanged. 

So the change for BIDS2Legacy's use of bidsPar is actually negligible. Are you OK with this?

Modules/xASL_wrp_BIDS2Legacy.m
HenkMutsaerts: See below the code that we then effectively remove.

To me it seems this code merges import log files, which is not something we want in processing-BIDS2Legacy, especially not because both could potentially be run at the same time. (converting to BIDS is completely separate from loading BIDS).

The other thing it does, is moving import log files to a derivatives folder. That is something that should only happen in the import right?

So to me it seemed that the deleted code (below), is rudimentary, a remnant of when BIDS2Legacy was part of the import module.

>>
%% Clean-Up subfunction
function xASL_bids_BIDS2Legacy_CleanUp(pathStudy)

    % Start with empty file list
    filesCleanUp = {};
    
    % Search for dcm2niix summary file
    dcm2niixImportLogFile = xASL_adm_GetFileList(pathStudy,'^import_log.+$');
    
    % Merge log file lists
    if ~isempty(dcm2niixImportLogFile)
        filesCleanUp = vertcat(filesCleanUp,dcm2niixImportLogFile);
    end
    
    % Move files to derivatives
    if ~isempty(filesCleanUp)
        for iFile = 1:size(filesCleanUp,1)
            sourceCleanUp = filesCleanUp{iFile};
            [~, fileCleanUp, extCleanUp] = xASL_fileparts(sourceCleanUp);
            destCleanUp = fullfile(pathStudy, 'derivatives', 'ExploreASL', [fileCleanUp extCleanUp]);
            if xASL_exist(sourceCleanUp,'file')
                xASL_Move(sourceCleanUp,destCleanUp);
            end
        end
    end
    
    % Print new line after track progress bar
    fprintf('   \n');

end

Modules/xASL_wrp_BIDS2Legacy.m
HenkMutsaerts: Yes!

Modules/xASL_wrp_BIDS2Legacy.m
HenkMutsaerts: This is outdated, because xASL_wrp_BIDS2Legacy is not there anymore (deleted in this issue).
BTW: I see now that we should then also delete `xASL_ut_function_xASL_wrp_BIDS2Legacy`?

Modules/xASL_wrp_BIDS2Legacy.m
jan-petr: Sure. We probably don't want to do that in BIDS2Legacy, but in import. But still, we want to take all the import logs and save them to derivatives. Cause we shouldn't save logs to rawdata and we save all logs to derivatives. So the question still comes - why it would help to not do this cleanup that saves the dcm2nii import logs to derivatives?

HENK: No, in the import module, the logs are created in the derivatives/log folder. This is just a remnant code that merges various import logs. Because BIDS2Legacy is not part of import anymore, we don't want to do this. BIDS2Legacy already puts it's log in derivatives/logs, because it is what `xASL_Iteration` does.

I tested some cases, this should work now.

Modules/xASL_wrp_BIDS2Legacy.m
jan-petr: not delete. but correctly rename. The function was testing some functionality and we should still keep testing it despite that you have changed the function names.

Modules/Module_Processing/xASL_init_BIDS2Legacy.m
jan-petr: Yes - I wasn't afraid that it would have consequences. And it's a useful edit. But the question is - now that we have improved it on one place, we should probably do the improvement everyhwere. I can do that if you want...

HenkMutsaerts: Sure, I'm fine with that. But that takes more effort. I would like to continue asap with cherry-picking my other stuff.

Modules/xASL_wrp_BIDS2Legacy.m
HenkMutsaerts: `xASL_wrp_BIDS2Legacy` loaded the bids configuration, check the existence of xASLDerivatives, and started 

`xASL_bids_BIDS2Legacy_ParseModality`
M0 parsing
and 
`xASL_bids_BIDS2Legacy_CleanUp`

The loading is gone to `xASL_init_BIDS2Legacy`, and the rest all moved to `xASL_module_BIDS2Legacy`.

So we could change `xASL_ut_function_xASL_wrp_BIDS2Legacy` to 
`xASL_ut_function_xASL_module_BIDS2Legacy`, but this is tricky because these modules work with the status/lock GoNoGo system; and it won't have `xASL_init_BIDS2Legacy`.

So I don't see how to easily unit test, because this is not a "unit" anymore :)

Modules/xASL_wrp_BIDS2Legacy.m
HenkMutsaerts: Same answer to your question on "Why do we delete this function without replacement? Don't we want to merge and save the log files?" above.

The cleaning was on combining the import and BIDS2Legacy log files, which we don't want anymore.

Modules/xASL_wrp_BIDS2Legacy.m
jan-petr: You are right. This code didn't do anything and could be deleted. What I couldn't understand was why the BIDS2Legacy output was so short now. 
Which I understand now - the bVerbose is set to false by default. But I'm not sure this is a good idea. 
Line 77 of xASL_module_BIDS2Legacy can indeed be suppressed. But line 178 should be reported. Can you revise the use of bVerbose there?

Modules/Module_Processing/xASL_init_BIDS2Legacy.m
jan-petr: OK. then can you create a new issue to finalize this later so that we don't have inconsistent code? We can do that later, but it would be good to do it in this release still.

Modules/xASL_wrp_BIDS2Legacy.m
jan-petr: This function tests if BIDS2Legacy was correctly executed and generated the correct import by running ExploreASL(path). So it wasn't testing any subpart of the module, it tested the module in its entirety. You simply need to rename the function correctly and it will work. Just rename it to `xASL_ut_function_xASL_module_BIDS2Legacy` without even changing a line of code inside. 

If this test function doesn't work, it is a sign that BIDS2Legacy doesn't work.

Even if you keep the original name, the whole test works. So the test is still running, useful, and correct, it just should be renamed to avoid confusion.

Modules/xASL_wrp_BIDS2Legacy.m
jan-petr: I see that this is useless now.

Modules/xASL_wrp_BIDS2Legacy.m
HenkMutsaerts: Done, sorry for my ignorance of our beautiful unit tests :)

jan-petr (approved):
All fixed.

Issue #997, opened 2021-12-13
Revamp PWI4D behavior


HenkMutsaerts:
### Description

WHY do we want to revamp/simplify/standardize the PWI4D behavior:
1. Reduction of if-statements, also with a single-PLD sequence, we use as much multi-PLD code that can be used for single-PLD as well (though superfluous as nPLD=1). This makes the code easier to bugfix and test.
2. Better performance with BASIL if we give the full PWI4D as input instead of PWI (see also #920)
2. For temporal calculations, e.g. spatial CoV or proximal-distance ratio for each control-pair in the case of 2D EPI, which could be a new "temporal cerebrovascular stability" biomarker of autoregulation
(See also #594)
4. For QCing. New QC NIfTIs/images/values needed are in #1317.

DISADVANTAGE:
1. Disk space usage increased (temporarily, could delete this intermediate NIfTI using the parameter `bSavePWI4D`)
2. RAM usage increased

SUGGESTION:
ASL4D = rawdata asl.nii, time decoded
PWI4D.nii = ASL4D subtracted, so N=0.5*nVolumes
PWI3D.nii is the PLD & TE-wise averaging of PWI4D
PWI is the single PWI volume (as it was)

#### Create new function `xASL_im_ASLSubtractionAveraging`
This is used by the following functions:
* xASL_wrp_RealignASL uses mean PWI to create a mask for ENABLE
* xASL_wrp_RegisterASL uses this to create single PWI for registration purposes
* xASL_wrp_ResampleASL uses this to save PWI4D PWI3D PWI, etc
* xASL_quant_ASL uses this to create the PWI from the last PWI4D image

- [x] Normalize ASL4D -> PWI4D -> PWI3D -> PWI for all multi-PLD, multi-TE, multi-LD sequences
- [x] move `xASL_wrp_Resample` section 8 subtraction & averaging to `xASL_im_ASLSubtractionAveraging`
- [x] Equalized handling quantification parameters, e.g., x.Q.EchoTime -> x.Q.EchoTime_PWI4D -> x.Q.EchoTime_PWI3D -> x.Q.EchoTime_PWI
- [x] Add motion application (spm_reslice) before resampling (based on existence of .mat sidecar)

#### Other
- [x] x.EchoTime -> x.Q.EchoTime
- [x] Handling x.Q.EchoTime x.Q.Initial_PLD x.Q.LabelingDuration at beginning xASL_module_ASL
[...]

PR #1499, opened 2023-09-29
#1096 Adding MergingList recognition and default to ASL module + conc…

### LINKED ISSUE #1096: FME_DEBBIE session merging
Automatize the DEBBIE quantification by merging sessions as defined in a dataPar.json

### Tasks (based on what was discussed on the 11th of September)
What is done already in the main version:
- [x] Make file merging a separate function
- [x] HAD_8 processing -> before quant we save PWI4D.nii
        -> xASL_quant_Basil / line 141 : 
- [x] In dataPar.json, a list of pairs of sessions is provided -> x.modules.asl.SessionMergingList:[["ASL_1", "ASL_2"], ["ASL_3", "ASL_4"]] -> meaning ASL1+ASL2 need to be concatenated before quantification as well as ASL3+ASL4.
- [x] In xASL_quant_FSL, if SessonMergingList is recognized, it checks if the PWI4D of the "pair" session exists (based on what was provided by the SessionMergingList), and it concatenates the two.  'bMergingTxt' is then considered TRUE in order to merge the FSL_ModelOptions.txt inside xASL_sub_FSLOptions (function written inside xASL_quant_FSL)
- [x] The two txts are merged and all TIs, TEs, and tau's are added to FSL Options before quantification
- [x] T1b should be 1.7306679
- [x] T1 = 1.33
- [x] It gives labeling duration of ASL_2 for both ASL_1 and ASL_2 in FSL_ModulesOptions.txt - this affects both Tau and TI
- [x] Check if order of TE, TI, Tau, and NII is correct in the automatic concatenation.
- [x] Debug and clean
- [x] Create a new issue for using PLD-PWI4D from issue #997 to merge sequence parameters in the same function as NIIs are merged. #1543
- [x] Go through the entire xASL_wrp_QUantify -> xASL_quant_ASL_ -> xASL_quant_FSL and make sure that we use the PWI4D throughout without unnecessary loading and saving all the time. Just simply use the PWI4D which is made for this ;) We will probably need to do this after merging this and #997 issue. In #1543
- [x] in the merging function, we can create a merged PLD vector based on PLD_PWI4D. However, we only have access to PLD_PWI4D from the current session. And PLD_PWI4D from the previous sessions is not available in x-struct.
The solution is to do this in resample_ASL, we save together with PWI4D a simplified JSON containing the important PLD and other parameters and load this later. #1543
- [x] Save the merged file to a special file called PWI4D_Merged.nii. That way, we are still able to quantify PWI4D and PWI4D_Merged if we want to. We can’t move this to resampling, because M0_quant is called only after resampling and we might need this for the merging. We add sessions to the name PWI4D_Merged_1_2 and then delete this after using. #1543
- [x] quant_FSL is parsing PLDs etc and saves them to a TXT. Currently, it is also handling the parsing of DEBBIE parameters to the proper form. It also handles the concatenation of the parameters. The ideal solution is to only do PLD->FSL parsin, but move creating of PLD vectors (e.g. related to Hadamard) to the resampling function. The resampling function will create PLD_PWI4D vectors and we will only use in quant_FSL these vectors irrespective of if they come from DEBBIE or multi-PLD. Quant-FSL will thus not really deal with how PLD vectors are created. Currently quant_FSL also does the merging. In the future, the merging will go to Resample function as well. However, we are editing this concurently in #997 and #1096. So only after merging these two issues, we will fix the creation of PLD vectors. Note that this won’t still be super easy, as multi-TE needs to be handeled in a very special way and we can’t simply paste the vector of TEs. But that is doable automatically. So this part will have to be done in a follow-up issue. in #1543 
- [x] If only 1 M0 scan in session1, copy it to current session (needs to be done anyway). This is done for the last and all intermediary sessions. M0 has to be in the first session. So if we have ASL1+2+3 and M0_1, then this M0 is copied and processing re-run for ASL_2 and ASL_3 and of course ASL_1+2+3. #1543
- [x] We remove the the bSessionMerge parameter from xASL_quant_FSL and we provide path to the correct file. #1543
- [x] 	(improve later) Motion correction: motion correct HAD8 to HAD4. How? Copy HAD4 from session 1 and copy the first volume of HAD4. run an extra co-registration with the first volume of HAD4 and HAD8. Run motion corr of HAD8. Then apply this motion correction between HAD4 and HAD8 @jan-petr this will be probably done in a separate issue -> this issue doesn't exist yet, so has to be created and then added to the DEBBIE Master list. #1543
- [x] Can we have 3 sessions merged together for quantification? ASL_1 + ASL_2 + ASL_3. Test if ASL_1+2+4 works. Test if ASL_2+3+4 works. Test double pair ASL_1+2 and ASL_4+5. Test different sizes ASL_2+3 and ASL_1+4+5. #1543
- [x] Once we are done, we add SessionMergingList and bAutomaticMerging (or their final names) to the Documentation (inside #1096)

- [x] Change merging to concatenation (provided that we always do A A A A B B B B)
- [x] Create a new issue for `bAutomaticASLSessionMerging`
- [x] JAN: CHECK in the code #1543 comments by Henk, that are there for the next issue.
- [x] THIS BRANCH CANNOT BE MERGED TO THE DEVELOP BRANCH, IT WILL BE MERGED ONLY WHEN #997 AND #1543 ARE MERGED AS WELL

### Release notes

Automatic concatenation of sessions based on a provided list.

BeatrizPadrela:
Close #1096 

### How to test

- [x] I've tested it with the option "x.modules.asl.SessionMergingList":[["ASL_1","ASL_2"],["ASL_3","ASL_4"] and it works!
- [x] I've tested it with ASL_1 = HAD8 and ASL_2 = HAD4, as well as ASL_1 = HAD4 and ASL_2 = HAD8

Tested in one subject of LCBC: perfect CBF!
image


Modules/xASL_module_ASL.m
```diff
@@ -189,6 +190,40 @@
     x.modules.asl.bMultiTE = false;
 	x.Q.NumberEchoTimes = 1;
 end
+%% D6. Recognizing automatic quantification (e.g. DEBBIE sequence)
```
jan-petr: I think we should remove this for now and only add it later, especially since it's not yet written too well. 
So either remove all of this code and add my comments to a new issue, or fix the following comments.

1. bAutomaticMerging should be an input parameter -> read from dataPar.json, so it needs to be added to adm_deprecatedfields
2. bAutomaticMerging should only be initialized `if ~isfield(x.modules.asl, 'bAutomaticMerging') || isempty(x.modules.asl.bAutomaticMerging)
3. We keep automaticMerging as it was from the input and we don't assign any new values -> because this is not about assigning automaticMerging, but about filling the Session.MergingList.
4. Update MergingList -> SessionMergingList
5. In case automaticMerging is TRUE and sessionMerginList filled, we don't do anything with bAutomaticMerging
6. In case bAutomatic is TRUE and sessionMergingList is empty, then that's the only case where we start filling the List. But we have to do it a bit more smart than you do it now. We don't only divide sessions by two, but we have to actively go through sessions and read JSONs a make the DEBBIE pairs.

So taken all together, this has to be far more complex than it is, so I propose you remove this whole code edit and make a new issue where you copy my comments here in the issue and add your code as a further comment.

Functions/Quantification/xASL_quant_FSL.m
```diff
@@ -95,8 +95,54 @@
     end
 
     %% 4. Create option_file that contains options which are passed to the FSL command
-	% FSLOptions is a character array containing CLI args for the BASIL/FABBER command
-	FSLOptions = xASL_sub_FSLOptions(pathFSLOptions, x, bUseFabber, PWI, pathFSLInput, pathFSLOutput);
+	bMergingTxt = 0; % only active if SessionMergingList + last session of list -> merge FSLOptions.txt
```
jan-petr: I would call it bMergingSessions

Functions/Quantification/xASL_quant_FSL.m
```diff
@@ -95,8 +95,54 @@
...
+    SessionsToMerge = {}; % To be used later during the txt merging
+    % For testing:
+    % x.modules.asl.SessionMergingList = {{'ASL_0','ASL_1', 'ASL_2'}; {'ASL_3', 'ASL_4','ASL_5'}};
```
jan-petr: I know it's an example only, but I would still remove ASL_0 which doesn't make sense.

Functions/Quantification/xASL_quant_FSL.m
```diff
@@ -95,8 +95,54 @@
...
+   
+    % 4.1 Check if necessary to merge sessions (and FSLOptions.txt)
+    if isfield(x.modules.asl,'SessionMergingList')
```
jan-petr: Actually, you should also check if the field is also not empty.
`&& ~isempty(x.modules.asl.SessionMergingList)

Also - we actually are checking all fields at the start of the ASL module and initializing them. So here you check for emptiness and at the start of the ASL module you check for existence and if it doesn't exist, you create an empty one...

Functions/Quantification/xASL_quant_FSL.m
```diff
@@ -95,8 +95,54 @@
...
+       index=0;
+            for V=1:numel(x.modules.asl.SessionMergingList)
```
jan-petr: V -> iSubList

Functions/Quantification/xASL_quant_FSL.m
```diff
@@ -95,8 +95,54 @@
...
+                if ismember(x.SESSION, x.modules.asl.SessionMergingList{V})
```
jan-petr: better to have `sum(ismember)`

Functions/Quantification/xASL_quant_FSL.m
jan-petr: Done.

Functions/Quantification/xASL_quant_FSL.m
jan-petr: Done.

Functions/Quantification/xASL_quant_FSL.m
jan-petr: Done

Functions/Quantification/xASL_quant_FSL.m
jan-petr: I mean - even if you want to test for non-existing sessions, you should not have names that cannot exist.

Functions/Quantification/xASL_quant_FSL.m
jan-petr: Done

Functions/Quantification/xASL_quant_FSL.m
jan-petr: done

Functions/Quantification/xASL_quant_FSL.m
```diff
@@ -95,8 +95,54 @@
...
+                    index=V;
+                    break;
+                end
+            end
+            CurrentSessionList = x.modules.asl.SessionMergingList{index};
+
+            if strcmp(x.SESSION,CurrentSessionList{end}) % If the SESSION is the last of the list, we want to merge the .txt and concatenate it with the previous sessions of the list
+
+                bMergingTxt = 1; % only active for the last session of a Merging list -> merge FSLOptions.txt
+                bUseFabber = 1;
```
jan-petr: We should actually not use this. We should in theory allow merging both for BASIL and for FABBER as in theory we could use this for other than DEBBIE sequences. We can of course now add a warning that this only works for FABBER, but this option should be removed.

Functions/Quantification/xASL_quant_FSL.m
```diff
@@ -95,8 +95,54 @@
...
+                
+                % Find PWI4Ds of the sessions to merge
+                CurrentSession = x.SESSION;
```
jan-petr: No need to define this parameter.

jan-petr: Done

Functions/Quantification/xASL_quant_FSL.m
```diff
@@ -95,8 +95,54 @@
...
+                    if xASL_exist (PairScanPath,'file')
+                        SessionsToMerge = [SessionsToMerge, ListSessions{nSession}]; % Creates a List of sessions that do exist
+                        CurrentScanIm = xASL_io_Nifti2Im(pathFSLInput);
+                        PairScanIm = xASL_io_Nifti2Im(PairScanPath);
+                        ConcatIm = cat(4,CurrentScanIm, PairScanIm);
+                        xASL_io_SaveNifti(pathFSLInput,pathFSLInput,ConcatIm)
+                        % Even if this loop runs more than 1 time, it will always
+                        % concatenate before [pair nii, current concatenated nii]
+                    else
+                        fprintf('Subject %s doesnt a PWI4D of the pair session %s  \n', x.SUBJECT, ListSessions{nSession})
```
jan-petr: I would just crash this step and won't provide results rather than to provide incomplete results.

jan-petr: Done

Functions/Quantification/xASL_quant_FSL.m
```diff
@@ -95,8 +95,54 @@
...
+                ListSessions = CurrentSessionList(1:end-1);
+                CurrentScanPWIdir = sprintf('%s/PWI4D_FSLInput.nii',CurrentSession);
+
+                for nSession = 1: numel(ListSessions)
+
+                    PairScanPWIdir = sprintf('%s/PWI4D.nii',ListSessions{nSession});
+                    PairScanPath = replace (pathFSLInput, CurrentScanPWIdir,PairScanPWIdir);
+
+                    if xASL_exist (PairScanPath,'file')
+                        SessionsToMerge = [SessionsToMerge, ListSessions{nSession}]; % Creates a List of sessions that do exist
```
jan-petr: As we are crashing for missing sessions, we don't need to rebuild this other parameter, but we can use the original list.

jan-petr: Done

Functions/Quantification/xASL_quant_FSL.m
```diff
@@ -95,8 +95,54 @@
...
+                bMergingTxt = 1; % only active for the last session of a Merging list -> merge FSLOptions.txt
+                bUseFabber = 1;
+                
+                % Find PWI4Ds of the sessions to merge
+                CurrentSession = x.SESSION;
+                ListSessions = CurrentSessionList(1:end-1);
+                CurrentScanPWIdir = sprintf('%s/PWI4D_FSLInput.nii',CurrentSession);
+
+                for nSession = 1: numel(ListSessions)
+
```
jan-petr: We can do this path replacement nicer using internal parameters in x. I will do that later.

Functions/Quantification/xASL_quant_FSL.m
```diff
@@ -95,8 +95,54 @@
...
+                    PairScanPWIdir = sprintf('%s/PWI4D.nii',ListSessions{nSession});
+                    PairScanPath = replace (pathFSLInput, CurrentScanPWIdir,PairScanPWIdir);
+
+                    if xASL_exist (PairScanPath,'file')
+                        SessionsToMerge = [SessionsToMerge, ListSessions{nSession}]; % Creates a List of sessions that do exist
+                        CurrentScanIm = xASL_io_Nifti2Im(pathFSLInput);
+                        PairScanIm = xASL_io_Nifti2Im(PairScanPath);
+                        ConcatIm = cat(4,CurrentScanIm, PairScanIm);
```
jan-petr: We should first concatenate all , and only save once.

Functions/Quantification/xASL_quant_FSL.m
```diff
@@ -95,8 +95,54 @@
...
+
+                for nSession = 1: numel(ListSessions)
+
+                    PairScanPWIdir = sprintf('%s/PWI4D.nii',ListSessions{nSession});
+                    PairScanPath = replace (pathFSLInput, CurrentScanPWIdir,PairScanPWIdir);
+
+                    if xASL_exist (PairScanPath,'file')
+                        SessionsToMerge = [SessionsToMerge, ListSessions{nSession}]; % Creates a List of sessions that do exist
+                        CurrentScanIm = xASL_io_Nifti2Im(pathFSLInput);
+                        PairScanIm = xASL_io_Nifti2Im(PairScanPath);
```
jan-petr: Before concatenation, we should check if dimensions X,Y,Z match and otherwise report an error - but a nice error explaining what was the problem.

Functions/Quantification/xASL_quant_FSL.m
```diff
@@ -224,7 +272,12 @@
 if ~isfield(x.Q.BASIL,'bMasking') || isempty(x.Q.BASIL.bMasking)
 		fprintf('BASIL: Setting default option bMasking = true\n');
 		x.Q.BASIL.bMasking = true;
-	end
+end
+
+% If we need to merge sessions, we will use Fabber
+if bMergingTxt
```
jan-petr: As above.

Functions/Quantification/xASL_quant_FSL.m
```diff
@@ -450,13 +557,20 @@
...
 		% Print labeling durations
-		if x.modules.asl.bMultiPLD
-			for iLabDurs = 1:length(LabDurs)
-				fprintf(FIDoptionFile, '--tau%d=%.2f\n', iLabDurs, LabDurs(iLabDurs));
-			end
-		else
-			fprintf(FIDoptionFile, '--tau=%.2f\n', LabDurs);
-		end
+        if x.modules.asl.bMultiPLD
+            for iLabDurs = 1:length(LabDurs)
```
jan-petr, line 530: I think this won't work nicely for merging three sessions - will fix that later.

Functions/Quantification/xASL_quant_FSL.m
```diff
@@ -430,6 +483,60 @@
 					fprintf(FIDoptionFile, '--te%d=%.3f\n', iTE, TEs(iTE));
 				end
 			end
+
```
jan-petr: WIll review this function later.

jan-petr (changes requested):
I did some fixes and some still need to be done.

I'll check the text merging later.

Functions/Quantification/xASL_quant_FSL.m
BeatrizPadrela: sure, I should have deleted this. It was just for my own testing tries

Functions/Quantification/xASL_quant_FSL.m
jan-petr: In general, I think that we should do this a bit more elegant. Especially since this will be more tricky for merging 3 files. I propose that in the FSL-options function we do the following steps:
1. We read fsloptions.txt at the start of the function and we read it for all previous sessions.
2. We parts all fsloptions and derive all important parameters from it - PLDs etc and put them in a separate character arrays.
3. We write a function that takes a character array and increases all numbers in a form PLD_XX etc by a given number so transforming PLD_1, PLD_2 to e.g. PLD_3, PLD_4. This function will only take the field format and the increment. We also write a function that obtains the last position in that field -> for PLD, it would return 4.
4. When merging, we always concatenate all the previous values, we do the necessary incrementing, and then we add the new values with a further increment.

All this will make it much easier and more transparent and will allow merging also 3 sessions - which is probably not possible now. Or at least I have only checked the code quickly, but haven't noticed any field increments.

Functions/Quantification/xASL_quant_FSL.m
jan-petr: Done

Functions/Quantification/xASL_quant_FSL.m
jan-petr: Done here with removing this options. But inside the options-function, we still need to check that this works for BASIL.

Functions/Quantification/xASL_quant_FSL.m
jan-petr: Done

Functions/Quantification/xASL_quant_FSL.m
jan-petr: Done

Modules/xASL_module_ASL.m
BeatrizPadrela: Yes, we don't need anything in xASL_module_ASL for now. I've created a new issue for this https://github.com/ExploreASL/ExploreASL/issues/1526

Functions/Quantification/xASL_quant_FSL.m
```diff
@@ -95,8 +95,57 @@
     end
 
     %% 4. Create option_file that contains options which are passed to the FSL command
-	% FSLOptions is a character array containing CLI args for the BASIL/FABBER command
-	FSLOptions = xASL_sub_FSLOptions(pathFSLOptions, x, bUseFabber, PWI, pathFSLInput, pathFSLOutput);
+	bMergingSessions = 0; % only active if SessionMergingList + last session of list -> merge FSLOptions.txt
```
HenkMutsaerts: This doesn't belong here, right. Merging should go earlier, at least at the beginning of the quantification but perhaps earlier. Merge everything, using a single function, at a single location

Functions/Quantification/xASL_quant_FSL.m
```diff
@@ -95,8 +95,57 @@
...
+			end
+		end
+
+		% If there are sessions to merge and the current session is the last of the list then we want to merge the .txt and concatenate it with the previous sessions of the list
+		if ~isempty(sessionsToMerge) && strcmp(x.SESSION, sessionsToMerge{end})
+
+			bMergingSessions = 1; 
+            bUseFabber = 1; 
+
+			% Find PWI4Ds of the sessions to merge and concatenate them
```
HenkMutsaerts: Very dirty, only merging within the FSL quantification function. This is asking for problems.

Functions/Quantification/xASL_quant_FSL.m
```diff
@@ -186,17 +235,19 @@
     
 end
 
-function [FSLOptions] = xASL_sub_FSLOptions(pathFSLOptions, x, bUseFabber, PWI, pathFSLInput, pathFSLOutput)
+function [FSLOptions] = xASL_sub_FSLOptions(pathFSLOptions, x, bUseFabber, PWI, pathFSLInput, pathFSLOutput, bMergingSessions, sessionsToMerge)
```
HenkMutsaerts: Shouldn't FSL be agnostic to a merged or non-merged sequence, we just have a single function earlier on that recreates PWI4D and settings in the JSON sidecar with it? (or the FSL settings with it)

Functions/Quantification/xASL_quant_FSL.m
HenkMutsaerts: Yes, and not do it here. Do it in a separate function earlier on.

There will be more sequences in the future in past that need merging, this needs to be more generic otherwise the merging/quantification/processing differs between FSL quantification and the rest of the pipeline, that's asking for problems.

Functions/Quantification/xASL_quant_FSL.m
```diff
@@ -430,6 +494,60 @@
...
 				end
 			end
+            
+            % So far e have all the info for the current scan.   Now we can
+            % add info of the scan concatenated
+
+            if bMergingSessions
+                CurrentASLJSONdir = fullfile(x.dir.SESSIONDIR,'/ASL4D.json');
+
+                for iSession = 1: numel(sessionsToMerge(2:end))
```
HenkMutsaerts: Nice! But move all to a separate function

Functions/Quantification/xASL_quant_FSL.m
HenkMutsaerts: Why merging three sessions, I thought we only need to merge 2? But nice indeed if we can merge as little and as many sessions as we want.

Modules/xASL_module_ASL.m
```diff
@@ -189,6 +190,34 @@
     x.modules.asl.bMultiTE = false;
 	x.Q.NumberEchoTimes = 1;
 end
+%% D6. Automatic session merging
```
HenkMutsaerts, line 225: Why don't we also do this in the separate merging function? This is only needed there, so it makes sense to only do it there. Plus, it here makes xASL_module_ASL busier for an exotic option.

HenkMutsaerts (changes requested):
Sorry I only now saw that I didn't press "finish your review" yet.
It looks great, I'm of course just waiting for it to go to a separate function ;)

Modules/SubModule_ASL/xASL_wrp_Quantify.m
```diff
@@ -571,3 +577,29 @@ function xASL_wrp_Quantify(x, PWI_Path, pathOutputCBF, M0Path, SliceGradientPath
 	end
 end
 end
+
+function PWI = xASL_wrp_Quantify_MergeSessions(x)
+% Merge several PWIs based on the list in x.modules.asl.sessionsToMerge
+PWI = [];
+for iSession = 1:numel(x.modules.asl.sessionsToMerge)
+	pathCurrentPWI4D = replace(x.P.Path_PWI4D, x.modules.asl.sessionsToMerge{end}, x.modules.asl.sessionsToMerge{iSession});
```
jan-petr: Check if this works for Nii and Nii.gz

jan-petr (changes requested):

Functions/Quantification/xASL_quant_FSL.m
jan-petr: Yes, that's for a possible future extension. And this should already work now for all cases.

Functions/Quantification/xASL_quant_FSL.m
jan-petr: So right now, I have improved xASL_quant_FSL, it's far from perfect, but it is able to work with a merged list of PLDs. Still, this merging happens inside xASL_quant_FSL, but after merging #997, we will have the PWI4D.jsons and the merged JSON and we'll read all the info from there. So it will all become really general.

Functions/Quantification/xASL_quant_FSL.m
jan-petr: Yes - as said above and also before. Merging of NII already happens in wrp_quantify. And merging of JSONs will also move there. So I'll do this in #1543

Functions/Quantification/xASL_quant_FSL.m
jan-petr: Yes. Will be finally fixed in #1543 . This is temporary.

Functions/Quantification/xASL_quant_FSL.m
jan-petr: The merging parameters will be removed, but we will need to use the name of the merged NII - this will be passed as an option.

Functions/Quantification/xASL_quant_FSL.m
jan-petr: I have cleaned that already. And this will go to the merging function once #997 is merged.

Modules/xASL_module_ASL.m
jan-petr: It's just checking params and making sure that the parsing list is OK. It's not doing any merging really.

Previously, you wanted to have all the general parameter checking at the start of ASL module...

HenkMutsaerts: Yes, for general parameters that are used at multiple locations, not exotic parameters.
Because these are only used in the merging function, I would move these there, that makes it cleaner.

My whole goal for a separate merging function, is that all merging stuff is there. This goal would be futile if we would put stuff here :)

Modules/SubModule_ASL/xASL_wrp_Quantify.m
```diff
@@ -571,3 +577,29 @@ function xASL_wrp_Quantify(x, PWI_Path, pathOutputCBF, M0Path, SliceGradientPath
 	end
 end
 end
+
+function PWI = xASL_wrp_Quantify_MergeSessions(x)
```
HenkMutsaerts: This still needs to go to a separate function, and it would be `xASL_im_MergePWI4D` right, because that is what we will do? (e.g., it could move elsewhere than quantification).

Can you ensure that all merging stuff is in this function, that would give fewer conflicts with #997 so saves us work.

HenkMutsaerts (changes requested):
We still need to move all to a separate function, to make the code nicer and save us time merging/rebasing

Modules/xASL_module_ASL.m
jan-petr: Again - you very specifically asked 1 year ago to move bMultiTE here and you strongly argued that you absolutely want to have all these parameters here and that we absolutely shouldn't have any parameters like bMultiTE anywhere else but at the start of ASL-module.

You have asked for bMultiTE and bHadamard and all these parameters to be specifically here. That is your decision from 1 year ago and you insisted on this.

It is not only used in merging function, but also in motion correction function, in quantification function and will be possibly also used elsewhere.

Modules/SubModule_ASL/xASL_wrp_Quantify.m
jan-petr: You have several times urged us to never create separate functions that are only called within a single function and that we should always keep them inside the function. 
This is exactly the case of this function. Having it as a separate function will make things more difficult as we will have to check inputs, while here we know the function have a single task.

This will not help with merging at all - because Github can efficiently deal with multiple edits done in the same file as long as they are not editing the same line. 

So moving the function out is
1) Against your previous recommendations.
2) Is not going to make merging with other commits more easy or more difficult.
3) Will make our life more difficult as we will have to work more on this single-purpose function.
4) You have very specifically asked to have this merging here. So it makes no sense arguing now that you want to move it elsewhere.

Functions/Quantification/xASL_quant_FSL.m
HenkMutsaerts: I would still separate the merging function now, that will save us time. The PWI4D.jsons will be done in #1543 indeed but the separate function should be done here.

jan-petr (changes requested):
Hopefully my comments will be submitted now...

Modules/SubModule_ASL/xASL_wrp_Quantify.m
jan-petr: Hopefully you can see this second comment.
I can put this as a separate function, but the motivation shouldn't be merging. Code added at the end of a function will be easily handled by GitHub and will not lead to conflicts. And call of the function is a single line and it will have to be there irrespective if we call internal or separate function.

You would have been absolutely right if the code would have been loosely added to xASL_wrp_QUantify, but it's in a subfunction, so that's not a problem for conflicts. And you always wanted to have one-time-use functions inside and not externally.

Functions/Quantification/xASL_quant_FSL.m
```diff
@@ -224,7 +230,16 @@
...
 		fprintf('BASIL: Setting default option bMasking = true\n');
 		x.Q.BASIL.bMasking = true;
-	end
+end
+
+if nargin<7 || isempty(bMergingSessions)
+	bMergingSessions = 0;
+end
+
+% If we need to merge sessions, we will use Fabber
```
HenkMutsaerts: bUseFABBER is already defined in parent functions

Functions/Quantification/xASL_quant_FSL.m
```diff
@@ -402,24 +462,28 @@
...
+			if ~bMergingSessions
+				TEs = round(x.EchoTime'/1000,3); % Convert Echo Times to seconds and keep 4 decimal digits
+				nTE = length(unique(x.EchoTime)); % Calculate the number of Echo Times
+				if (nPLD*nTE) ~= nVolume
+					error('The number of volumes %d does not match number of PLDs %d * number of TEs %d', nVolume, nPLD, nTE);
+				end
+				nTE = ones(size(PLDs))*nTE;
 			end
 
 			% Plotting the values into the doc (PLD=ti, LD=tau)
```
HenkMutsaerts: "plotting" -> "printing"
"the doc" -> "FSL options"

Modules/SubModule_ASL/xASL_wrp_Quantify.m
HenkMutsaerts: I see this now indeed, thanks!

1. Because this function can potentially be used at multiple locations (the function should be called xASL_im_MergePWI4D, and allow merging in general, perhaps best called before xASL_wrp_Quantify such that somebody can easily skip it and merge manually, or merge before or after applying M0 in the future etc) I believe this is not something that should be in a subfunction
2. I understand this now, I thought it was completely independent but you do need the merging of parameters in xASL_quant_FSL because there is no separate PWI4D.json yet. I didn't thought of this yet.
3. See (2)
4.  :) -> so my only argument here is that we now have 2 merging functions, one for the image matrix, and one for the metadata TE, PLD, LD. These will become a single function that ignorant of the type of ASL will merge PWI4D and PWI4D.json. So you could consider putting all in the single function that you call from xASL_quant_FSL. So I ask you to consider if merging the merging code in:

1) xASL_module_ASL
2) xASL_wrp_Quantify
3) xASL_quant_FSL

already now all to a single function, that is called from xASL_quant_FSL (in the future from xASL_wrp_Quantify or somewhere else).

That will make it cleaner I believe still (even though I understand now that you need it inside xASL_quant_FSL because we currently don't have a PWI4D.json)

Functions/Quantification/xASL_quant_FSL.m
HenkMutsaerts: You can close this. But I would generalize the session merging such that it iterates over all sessions defined in sessionMergingList (unless you have a good reason to do this in #1543).

Functions/Quantification/xASL_quant_FSL.m
```diff
@@ -386,13 +401,58 @@
 
 		% For Time-encoded, we skip the first volume per block
 		if x.modules.asl.bTimeEncoded
-			[PLDs, index] = unique(PLDs, 'stable');
-			LabDurs = LabDurs(index)';
+			% In case we are merging sessions, we have to read all JSONs again from files
+			if bMergingSessions
```
HenkMutsaerts, line 417: Note that session merging is not only for time encoded. Either you change that here or in #1543 (in which case it should be added to that list)

Functions/Quantification/xASL_quant_FSL.m
```diff
@@ -386,13 +401,58 @@
...
+				PLDs = [];
+				LabDurs = [];
+				nTE = [];
+				TEs = [];
+				for iSession = 1: numel(x.modules.asl.sessionsToMerge)
```
HenkMutsaerts, line 422: So here we allow merging 2 or more sessions?

Functions/Quantification/xASL_quant_FSL.m
```diff
@@ -386,13 +401,58 @@
 
 		% For Time-encoded, we skip the first volume per block
 		if x.modules.asl.bTimeEncoded
```
HenkMutsaerts, line 415: This loop now contains twice if bTimeEncoded and does twice the same, with and without merging. Can we not already simplify this here?

Functions/Quantification/xASL_quant_FSL.m
```diff
@@ -224,7 +230,16 @@
...
-	end
+end
+
+if nargin<7 || isempty(bMergingSessions)
+	bMergingSessions = 0;
+end
+
+% If we need to merge sessions, we will use Fabber
+if bMergingSessions
+    bUseFabber = 1;
```
HenkMutsaerts, line 251: Please can you separate session merging and the use of FABBER. E.g., Mervin will use session merging for all Singapore studies, where most are without DEBBIE.

Functions/Quantification/xASL_quant_FSL.m
```diff
@@ -95,8 +95,13 @@
     end
 
     %% 4. Create option_file that contains options which are passed to the FSL command
-	% FSLOptions is a character array containing CLI args for the BASIL/FABBER command
-	FSLOptions = xASL_sub_FSLOptions(pathFSLOptions, x, bUseFabber, PWI, pathFSLInput, pathFSLOutput);
+	% To be removed later when we are properly checking for merged multi-TE images
+	if x.modules.asl.bMergingSessions == 1
+		bUseFabber = 1;
```
HenkMutsaerts, line 108: Please can you separate session merging and the use of FABBER. E.g., Mervin will use session merging for all Singapore studies, where most are without DEBBIE.

Modules/xASL_module_ASL.m
HenkMutsaerts: Yes for bMultiTE and bHadamard, but no for everything that is used on a single location only. Why do we use bSessionMerging in motion correction? I would say that we either merge BEFORE motion correction etc or AFTER but the motion correction etc can be oblivious to session merging. Doing session merging at multiple locations makes for parallel pipeline routes which complicates debugging etc.

HenkMutsaerts (changes requested):
Some more :)

Modules/xASL_module_ASL.m
```diff
@@ -190,6 +191,54 @@
...
+x.modules.asl.sessionsToMerge = {};
+
+
+% Find the lists containing the current session
+% Note that these can be multiple lists
+for iList=1:nLists
+	if ~isempty(x.modules.asl.SessionMergingList{iList}) && sum(ismember(x.SESSION, x.modules.asl.SessionMergingList{iList}))
+		x.modules.asl.sessionsToMerge = x.modules.asl.SessionMergingList{iList};
+
+        if ~isempty(x.modules.asl.sessionsToMerge) && strcmp(x.SESSION, x.modules.asl.sessionsToMerge{end})
```
HenkMutsaerts: Here you see I simplified the code a bit, it didn't matter if `x.modules.asl.sessionsToMerge` is inside or outside the loop.

One thing that I don't understand yet, is that we allow multiple merging lists (the cell in cell structure you enforce above), but we don't deal with them properly. E.g., if we have merging lists:

ASL_1 ASL_3
ASL_2 ASL_3

then for ASL_3, we will only concatenate one of these lists in the current situation (irregardless of your version or this new version). I don't think we should already allow multiple mergings with the same last session, because that makes our life tricky, but we should at least issue a warning don't you think?

HenkMutsaerts (changes requested):
So I mostly replaced my reservations with code comments.
A few questions, some code explanations/ comments.

I wonder if we shouldn't use "concatenation" instead of "merging" as concatenation is more specific, "merging" can mean anything (e.g., intermittent ASL1_ASL2_ASL1_ASL2 would also be merging, but not concatenating).

Functions/Quantification/xASL_quant_FSL.m
jan-petr: It will indeed be removed. But simplifying would be double work. Cause not only this code will be removed from here. But we will not even really move it anywhere as the corresponding code is already now in xASL_wrp_Resample. And what we add to the merging function will not use this code. 
So no point in simplifying cause we won't be using it at all in the future.

Modules/xASL_module_ASL.m
```diff
@@ -190,6 +195,54 @@
...
+% SessionMergingList == total list of sessions that need concatenating (can be multiple lists for multiple concatenations)
+% Read the list of sessions to merge, defaulting to empty
+if ~isfield(x.modules.asl, 'SessionMergingList')
+	x.modules.asl.SessionMergingList = {};
+else
+	% The parameter should be a list of lists, but if a single list is provided, it converts it
+	if ~isempty(x.modules.asl.SessionMergingList) && iscell(x.modules.asl.SessionMergingList) && ~iscell(x.modules.asl.SessionMergingList{1})
+		x.modules.asl.SessionMergingList = {x.modules.asl.SessionMergingList};
+    end
+    nLists = numel(x.modules.asl.SessionMergingList);
```
jan-petr, line 238: We still have to check if the list is not empty... Because now we only know the variable exist but it might be empty

Modules/xASL_module_ASL.m
jan-petr: Good, but:
1. The reason for allowing several lists is that you can have a DEBBIE reproducibility study for example. You are right that we are concatenating lists for the last session and I definitively agree that multiple mergins with the same last session should never happen - I just didn't think of that happening. So we should never have two lists with the same end sessions. So 1+3 and 1+4 would be allowed. But 1+3 and 2+3 will not be allowed. I will simply write a function that checks this and then takes only the first session list and reports a warning. Or something like that.
2. Your code will crash on line 217. nList is not initialized if the list was not existing on line 197.
3. Related to point 1. You are right about this and we have to add an additional check. However, we already select the correct list on line 219. So line 223 has to be deleted.
4. Line 234 - you can simplify the code. for in Matlab works really nice for this
Instead of "n = numel(list), for i=1:n, print(list(i))" you can simply write for i=list, print(i) - that does exactly the same and is easier to read.

jan-petr (changes requested):
Few small bugs.

Functions/Quantification/xASL_quant_FSL.m
jan-petr: What's our final decision? Just to make sure? To keep as is now and more details will be added to the issue and code as comments? So we can resolve this now?

Functions/Quantification/xASL_quant_FSL.m
HenkMutsaerts: OK

Modules/xASL_module_ASL.m
HenkMutsaerts: Added nLists = 0 as dummy in 06081dd879b63b008be821d6baabc965cda9e5e5

Modules/xASL_module_ASL.m
HenkMutsaerts: 1. I agree, then shall we put that here then? I would write that already now, it won't take that much extra time.
2. Added nLists = 0 as dummy in 06081dd879b63b008be821d6baabc965cda9e5e5
3. We can't remove this because we need to check for x.SESSION being the latest index. I have now removed the first check which was unnecessary in 665dadb499182ba4edb67cd3fc57918492e0a51f
4. That's only the case in your example if list is already numerical right? In the example in this code, the list is not numerical but multiple cells containing ASL_1 ASL_2 etc. Otherwise, this would definitely not be easier to read (it would not make sense to make a for loop with string-cells as indices...

BTW: I'm happy for you to fix such things of course, I just wanted to simplify your code when I was trying to make it more readable by adding comments

HenkMutsaerts (changes requested):
Some answers

Modules/xASL_module_ASL.m
jan-petr: 1. Yes - we that already now. I'll do it, just waiting for you to agree.
2. Solved.
3. Yes, the first check wasn't necessary. But Line 229 (that was recently added) is identical to line 225 and should be removed.
4. Parameter `list2concatenate` should be removed as on line 238 and 240 we should be using `x.modules.asl.sessionsToMerge` instead. This works also for cells. Test this:
`c{1} = 'a'; c{2} = 'bc';for i=c,fprintf(['test ' i{1} '\n']),end`
You don't use string-cells as indices, you don't use indices at all. For simply goes across any array. 1:N is an array of integers, so you can have an array of strings - the exact same mechanism and you skip indexing completely.

BTW: Sure, no worries - your simplification helped. It's just that we need to finalize details. I can do point 1+3+4 now, I didn't want to touch things while you were working on it.

HenkMutsaerts: 3. Done in 69be7466ffa370657a2a5f3b23d7f21acef92fd2
4. Done in c08d02a902a9e99831f74d7a1d009b8f49cc8902

for goes across a list of indices, either internally (non-explicit to the user) or specified directly (clear to the user). Your example saves 1 line of code but makes it less easy readable. So I would keep it like we have it. Unless we want a new discussion again ;)

`c{1} = 'a'; c{2} = 'bc'; nI = numel(c); for i=1:nI,fprintf(['test ' i{1} '\n']),end`
will just cost a few words extra, but is more explicit. So I like it more. Note also that this behavior might vary between programming languages right?
BTW: note that `i` is not a good variable right because you can mistake it for the imaginary number, that's why we always use at least ii? Please don't always try to look at things from the ex-mathematical/physics student perspective ;)

jan-petr: I will still program the check for extra sequences - point 1.

I'm fine keeping it as it is as we removed the extra redundant code, though it's still better to add `numel` directly inside `for` without creating more variable.

jan-petr: Check done in 006d5958
I take the first eligible session. Issue a warning when a session fits in multiple lists. And a warning when at multiple endings. But a list ending with a current session has a priority.

Modules/xASL_module_ASL.m
```diff
@@ -246,8 +246,7 @@
 
 if x.modules.asl.bMergingSessions
     fprintf('%s', ' and will now concatenate the following sessions:')
-    nSessions = numel(x.modules.asl.sessionsToMerge);
-    for iSession=1:nSessions
+    for iSession = 1:numel(x.modules.asl.sessionsToMerge)
```
HenkMutsaerts, line 277: I thought we put nSessions here to reuse it also later

HenkMutsaerts (changes requested):
Two things:

A) Did you add the new parameters to the documentation?
B) Your current fixes are dependent on having multiple xASL_module_ASL runs getting one another's parameters, which is a dependency I would not want to implement. I would just detect current SESSION == in multiple list as END, issue a warning, and taking the first of that list. Your solution is perhaps elegant but creates unnecessary dependencies.

Modules/xASL_module_ASL.m
jan-petr: Yes. That was the entire point of my edit.
nSessions was only used here on this single occasion. And once you define such a variable, it increases the risk of a bug later. Because then somebody will see `nSession` and will reuse it without really knowing if that's number of sessions for this subject, or a number of sessions in a merging list etc. So this highly increases the risk of incorrect use later on through negligence.

jan-petr:
> Two things:
> 
> A) Did you add the new parameters to the documentation? B) Your current fixes are dependent on having multiple xASL_module_ASL runs getting one another's parameters, which is a dependency I would not want to implement. I would just detect current SESSION == in multiple list as END, issue a warning, and taking the first of that list. Your solution is perhaps elegant but creates unnecessary dependencies.

Careful to add comments as resolvable, or else we will overlook them in this long PR.
1. No - but there's a task for this in the issue and I'm waiting on fixing this name and will do that before closing this issue. Cause the code change is done anyway in another repository.
2. It doesn't create an interdependency between runs. All these parameters are newly checked for each ASL run completely independently on other ASL runs. It exactly implements what you want. If there's a dependency, then that was not my aim - and I do not really see that now.

Functions/Administration/xASL_adm_GetDeprecatedFields.m
```diff
@@ -111,6 +111,7 @@
     'M0PositionInASL4D'             'modules'   'asl'                                'M0PositionInASL4D';...
     'HadamardType'                  'modules'   'asl'                                'HadamardType';...
 	'ApplyQuantification'           'modules'   'asl'                                'ApplyQuantification';...
+    'MergeSessionList'              'modules'   'asl'                                'MergeSessionList';...
```
HenkMutsaerts: You just changed this name in the last commit, I don't know why, but if you change the name, I would consider the order of words. Starting with List makes it easier to sort variables by their type (==list), starting by Session makes it easier to sort variables by what they effect (==subject,session,etc) but starting with Merge to me has little advantages; but I'm sure you had an idea to change the variable name? :)

Modules/xASL_module_ASL.m
HenkMutsaerts: good point

HenkMutsaerts (changes requested):
Not sure why you changed the name, I added some reasoning. But otherwise fine to merge this

Functions/Administration/xASL_adm_GetDeprecatedFields.m
jan-petr: I can change this to any other name.

HenkMutsaerts (approved):

jan-petr (approved):
OK

jan-petr (approved):
OK.

PR #1505, opened 2023-10-07
Closes #870 sequence independent population

### LINKED ISSUE #870: Sequence-dependency population module
### Description

Previously, `xASL_module_Population` worked sequence-specific, because of sequence-specific masking.
Now, the sequence-specific susceptibility masking is performed in `xASL_module_ASL`->`xASL_wrp_CreateAnalysisMask`,
and (`xASL_module_Population`->) `xASL_im_CreateAnalysisMask` -> `xASL_im_CreateAnalysisMask_LegacySusceptibilityMasking` performs some extra thresholding to create group masks, based on 2D EPI & 3D GRASE templates.

### Requirements

- [x] 1. Test the ASL module susceptibility mask creation.
Using flavor Philips_PCASL_2DEPI_3.2.1.1_1, I tested the creation of the susceptibility mask inside the ASL module, which was a bit too lenient; around the susceptibility regions it was not correctly masked by the brainmask.
This is not a problem, because eventually. several masks are combined, but it is cleaner to mask extra by the brain mask, because the default susceptibility mask IS a brainmask (e.g., for sequences without a susceptibility mask, such as 3D spiral).

The susceptibility mask BEFORE & AFTER population module masking are nearly identical.

image
Brainmask
image
old susceptibility mask
image
new susceptibility mask

- [x] 2. This population-based part to create an analysis mask only runs when there are multiple subjects (e.g., more than 16).    To change this, we will now ALWAYS create templates and a group analysis mask, even if we have n=1. But if n<5, we will issue a warning that this may not work well.

- [x] 3. Check difference in masking only ASL module or also in population module:

image
image
image

Difference seems to be a small erosion around the original mask.
- [x] 4. Test the output of xASL_im_CreateAnalysisMask_LegacySusceptibilityMasking for 2D EPI & 3D GRASE. Move this extra masking that it does to `xASL_module_ASL` 
- [x] 5. Remove `xASL_im_CreateAnalysisMask` - `xASL_im_CreateAnalysisMask_LegacySusceptibilityMasking`
- [x] 6. Remove `xASL_module_Population`-`xASL_adm_DefineASLSequence`
- [x] 7. `xASL_stat_GetROIstatistics` uses `xASL_adm_DefineASLSequence` on line 154, but is not using x.Q.sequence, x.Q.readout* or x.Q.vendor, so we can remove this here. Same for `xASL_wrp_GetROIstatistics` line 87 has `xASL_adm_DefineASLSequence`, which can be removed.

### Release notes

Create sequence-specific mask in ASL module instead of the Population module.

HenkMutsaerts:
### Linked issue

Closes #870

### DIFF
--- a/Modules/Module_Processing/Module_ASL/xASL_wrp_CreateAnalysisMask.m
+++ b/Modules/Module_Processing/Module_ASL/xASL_wrp_CreateAnalysisMask.m
@@ -208,6 +208,7 @@ if DoSusceptibility
 
      MaskSusceptibility = BrainMask;
      MaskSusceptibility(MaskSuscept) = FinalMask(MaskSuscept);
+     MaskSusceptibility(~BrainMask) = 0;
 else
     MaskSusceptibility = BrainMask;
     % for e.g. 3D spiral the susceptibility mask is equal to the brain mask
--- a/Modules/Module_Processing/Module_ASL/xASL_wrp_CreateAnalysisMask.m
+++ b/Modules/Module_Processing/Module_ASL/xASL_wrp_CreateAnalysisMask.m
@@ -85,6 +85,10 @@ switch lower(x.Q.Sequence)
         error('Unknown ASL sequence!');
 end
 
+
+
+
+
 %% 1. Negative vascular signal
 NegativeMaskNative = xASL_im_MaskNegativeVascularSignal(x, 1); % native space
 NegativeMaskMNI = xASL_im_MaskNegativeVascularSignal(x, 2); % standard space
@@ -146,6 +150,10 @@ xASL_io_SaveNifti(x.P.Pop_Path_PWI, x.P.Pop_Path_BrainMaskProcessing, BrainMaskP
 xASL_io_SaveNifti(x.P.Path_PWI, x.P.Path_MaskVascular, MaskVascularNative, 8, false);
 xASL_io_SaveNifti(x.P.Pop_Path_PWI, x.P.Pop_Path_MaskVascular, MaskVascularMNI, 8, false);
 
+
+
+
+
 %% 5. Create susceptibility mask in standard space
 if DoSusceptibility
 	
@@ -206,12 +214,18 @@ if DoSusceptibility
      FinalThreshold = (ThresholdIntensity+SortedInt)/2; % average of 2 thresholds
      FinalMask = MixedIm>FinalThreshold;
 
-     MaskSusceptibility = BrainMask;
-     MaskSusceptibility(MaskSuscept) = FinalMask(MaskSuscept);
-     MaskSusceptibility(~BrainMask) = 0;
+     % %% Brief explanation what these masks here mean:
+     % BrainMask   => brain parenchyma (GM+WM) == 1, CSF & outside the brain == 0
+     % MaskSuscept => potential susceptibility regions, where artifacts may occur == 1, otherwise == 0
+     % FinalMask   => within MaskSuscept, voxels we want to include in the mask == 1, otherwise == 0
+
+     MaskSusceptibility = BrainMask; % we start with a wholebrain parenchyme mask
+     MaskSusceptibility(MaskSuscept) = FinalMask(MaskSuscept); % within potential susceptibility regions, we mask out susceptibility artifacts
+     MaskSusceptibility(~BrainMask) = 0; % we ensure that voxels without susceptibility artifacts, outside the brain, are still masked out
 else
     MaskSusceptibility = BrainMask;
-    % for e.g. 3D spiral the susceptibility mask is equal to the brain mask
+    % for e.g. 3D spiral the susceptibility mask is equal to the brain
+    % mask, effectively not masking out susceptibility artifacts
 end
 
 xASL_io_SaveNifti(x.P.Pop_Path_PWI, x.P.Pop_Path_MaskSusceptibility, MaskSusceptibility, [], false);
--- a/Modules/Module_Processing/Module_Population/xASL_im_CreateAnalysisMask.m
+++ b/Modules/Module_Processing/Module_Population/xASL_im_CreateAnalysisMask.m
@@ -49,10 +49,11 @@ PathSusceptibilityMask = xASL_adm_GetFileList(x.D.TemplatesStudyDir, ['^MaskSusc
 bSkipStandard = 0;
 
 if x.dataset.nSubjectsSessions<16
-    fprintf('%s\n', ['Too few subjects (' num2str(x.dataset.nSubjectsSessions) ') to create population-based analysis mask']);
-    x.S.MaskSusceptibility = xASL_im_IM2Column(ones(121,145,121),x.S.masks.WBmask);
-    x.S.VBAmask = x.S.MaskSusceptibility;
-    bSkipStandard = 1;
+    % With too small datasets, created templated are not reliable
+    fprintf('\n\n%s\n\n', ['Only n=' num2str(x.dataset.nSubjectsSessions) ' subject*runs, population-based analysis mask may not be useful']);
+    % x.S.MaskSusceptibility = xASL_im_IM2Column(ones(121,145,121),x.S.masks.WBmask);
+    % x.S.VBAmask = x.S.MaskSusceptibility;
+    % bSkipStandard = 1;
 elseif isempty(PathSusceptibilityMask)
     warning('Missing susceptibility maps, skipping...');
     bSkipStandard = 1;
--- a/Modules/Module_Processing/Module_Population/xASL_im_CreateAnalysisMask.m
+++ b/Modules/Module_Processing/Module_Population/xASL_im_CreateAnalysisMask.m
@@ -311,6 +311,11 @@ function [x,DoSusceptibility,TemplateMask,ThresholdSuscept,MaskSusceptibility] =
         % regions of air cavities)
         MaskSusceptibility = MaskSusceptibility > ThresholdSuscept;
 
+        [fPath, fFile] = xASL_fileparts(x.P.Pop_Path_MaskSusceptibility);
+        pathMaskSusceptibility2 = fullfile(fPath, [fFile '_AFTER.nii']);
+
+        xASL_io_SaveNifti(x.P.Pop_Path_PWI, pathMaskSusceptibility2, MaskSusceptibility, [], false);
+
         % Combine susceptibility & FoV
         MaskSusceptibility = MaskSusceptibility & MaskFoV;
     else
--- a/Modules/Module_Processing/Module_Population/xASL_im_CreateAnalysisMask.m
+++ b/Modules/Module_Processing/Module_Population/xASL_im_CreateAnalysisMask.m
@@ -114,11 +114,19 @@ if ~bSkipStandard
 	%% B) Create, combine & save vascular, susceptibity & FoV masks
 	MaskVascular = xASL_io_Nifti2Im(PathVascularMask)>=Threshold;
     MaskFoV = xASL_io_Nifti2Im(PathFoV)>=Threshold;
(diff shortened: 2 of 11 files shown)


Modules/Module_Processing/Module_ASL/xASL_wrp_CreateAnalysisMask.m
```diff
@@ -208,6 +208,7 @@ function xASL_wrp_CreateAnalysisMask(x)
 
      MaskSusceptibility = BrainMask;
      MaskSusceptibility(MaskSuscept) = FinalMask(MaskSuscept);
```
jan-petr: I am a bit confused by this notation. Because in the end - things that are zero on MaskSuscept and 1 on BrainMask but 1 on Final mask will turn out as ones in the end. 

Maybe you could write exactly in a comment what you really want to achieve.

That won't have a huge impact on the result, but we should check this for clarity.

Modules/Module_Processing/Module_Population/xASL_im_CreateAnalysisMask.m
```diff
@@ -49,10 +49,11 @@
 bSkipStandard = 0;
 
 if x.dataset.nSubjectsSessions<16
-    fprintf('%s\n', ['Too few subjects (' num2str(x.dataset.nSubjectsSessions) ') to create population-based analysis mask']);
-    x.S.MaskSusceptibility = xASL_im_IM2Column(ones(121,145,121),x.S.masks.WBmask);
-    x.S.VBAmask = x.S.MaskSusceptibility;
-    bSkipStandard = 1;
+    % With too small datasets, created templated are not reliable
+    fprintf('\n\n%s\n\n', ['Only n=' num2str(x.dataset.nSubjectsSessions) ' subject*runs, population-based analysis mask may not be useful']);
```
jan-petr, line 53: It might be easier to create the susceptibility mask so that we have it in all cases. No?

Modules/xASL_module_Population.m
```diff
@@ -102,11 +100,7 @@
     SusceptPath = fullfile(x.D.TemplatesStudyDir,['MaskSusceptibility' x.S.TemplateNumberName '_bs-mean.nii']);
 
     if ~xASL_exist(SusceptPath, 'file')
-        if x.dataset.nSubjects<6 % in this case, a group template doesn't make sense and no warning is required
-            fprintf('Susceptibility mask template was missing...\n');
-        else
-            warning('Susceptibility mask template was missing...');
-        end
+        warning('Susceptibility mask template was missing...');
```
jan-petr, line 103: It does not make sense to have this check here. The susceptibility mask is created only in function on line 125. 
So I would delete the entire code here from line 102 to 115. And put all reporting to the function for creating the analysis mask.

Modules/Module_Processing/Module_Population/xASL_im_CreateAnalysisMask.m
```diff
@@ -49,10 +49,11 @@
 bSkipStandard = 0;
 
 if x.dataset.nSubjectsSessions<16
-    fprintf('%s\n', ['Too few subjects (' num2str(x.dataset.nSubjectsSessions) ') to create population-based analysis mask']);
-    x.S.MaskSusceptibility = xASL_im_IM2Column(ones(121,145,121),x.S.masks.WBmask);
-    x.S.VBAmask = x.S.MaskSusceptibility;
-    bSkipStandard = 1;
+    % With too small datasets, created templated are not reliable
+    fprintf('\n\n%s\n\n', ['Only n=' num2str(x.dataset.nSubjectsSessions) ' subject*runs, population-based analysis mask may not be useful']);
+    % x.S.MaskSusceptibility = xASL_im_IM2Column(ones(121,145,121),x.S.masks.WBmask);
```
jan-petr, line 54: About the title of this entire function. Should we have a wrp_CreateAnalysisMask in ASL module and im_CreateAnalysisMask in population module?

Functionally, this is fine. But it might be slightly confusing to have an IM and WRP functions being disconnected. Perhaps we rename it to im_CreatePopulationMask or something?

Modules/Module_Processing/Module_Population/xASL_im_CreateAnalysisMask.m
```diff
@@ -49,10 +49,11 @@
...
-    fprintf('%s\n', ['Too few subjects (' num2str(x.dataset.nSubjectsSessions) ') to create population-based analysis mask']);
-    x.S.MaskSusceptibility = xASL_im_IM2Column(ones(121,145,121),x.S.masks.WBmask);
-    x.S.VBAmask = x.S.MaskSusceptibility;
-    bSkipStandard = 1;
+    % With too small datasets, created templated are not reliable
+    fprintf('\n\n%s\n\n', ['Only n=' num2str(x.dataset.nSubjectsSessions) ' subject*runs, population-based analysis mask may not be useful']);
+    % x.S.MaskSusceptibility = xASL_im_IM2Column(ones(121,145,121),x.S.masks.WBmask);
+    % x.S.VBAmask = x.S.MaskSusceptibility;
+    % bSkipStandard = 1;
 elseif isempty(PathSusceptibilityMask)
```
jan-petr: WHy are we checking for PathSusceptibilityMask - we are creating that here, so it should be always empty...

Modules/Module_Processing/Module_Population/xASL_im_CreateAnalysisMask.m
```diff
@@ -113,11 +114,19 @@
...
     MaskFoV = xASL_io_Nifti2Im(PathFoV)>=Threshold;
-        
+    
+
+
     % Legacy Susceptibility Masking
-    [x,DoSusceptibility, TemplateMask, ThresholdSuscept, MaskSusceptibility] = ...
-    xASL_im_CreateAnalysisMask_LegacySusceptibilityMasking(x, PathSusceptibilityMask, Threshold, MaskFoV);
-                    
+    MaskSusceptibility = xASL_io_Nifti2Im(PathSusceptibilityMask);
```
jan-petr: I do not understand. How can we load the mask that we will be creating only a few lines lower?

jan-petr (changes requested):
I like what you are trying to do and fully agree with that. I've tried to check everything to make sure the changes are perfectly correct. There are still some things that I do not understand or are wrong.

Modules/xASL_module_Population.m
HenkMutsaerts: No, `xASL_wrp_CreatePopulationTemplates` creates templates from all individual NIfTIs. MaskSusceptibility is an individual mask that every ASL run has, and `xASL_wrp_CreatePopulationTemplates` should create an average of this. In the past, it would not do this with too few subjects, because this would create less trustworthy templates. But now, it simply always creates templates. So SusceptPath should always exist.

`xASL_CreateAnalysisMask` on line 125 just creates a groupwise analysis mask based on multiple templates of vascular, FoV, Susceptibility masks, checking that at least 95% of the population have a certain voxel to include it, etc.

Does it help if I rename:
* `SusceptPath` -> `pathTemplateMaskSusceptibility`
* `xASL_im_CreateAnalysisMask` -> `xASL_im_CreateGroupAnalysisMask`
?

Modules/Module_Processing/Module_Population/xASL_im_CreateAnalysisMask.m
HenkMutsaerts: For clarification:

1) `xASL_module_ASL` -> `xASL_wrp_CreateAnalysisMask` now always creates all masks for all ASL sessions, for 3D GRASE this will be a dummy brain mask
2) `xASL_module_Population` -> `xASL_wrp_CreatePopulationTemplates` creates average template images of all NIfTIs in standard space, including all masks
2) `xASL_module_Population` -> `xASL_im_CreateAnalysisMask` takes the mask templates (of FoV, MaskSusceptibility, MaskVascular, etc) and generates a group analysis mask from them. Checking that at least 95% of the population have a certain voxel to include it, etc.

Does it help if I rename:

`xASL_wrp_CreateAnalysisMask` -> `xASL_wrp_CreateIndividualMasks`
`xASL_im_CreateAnalysisMask` -> `xASL_im_CreateGroupAnalysisMask`
?

So instead of creating an empty group susceptibility mask that we have it in all cases, with my new changes, we will here now skip this, because we will always use the individual susceptibility masks.

Modules/Module_Processing/Module_Population/xASL_im_CreateAnalysisMask.m
HenkMutsaerts: Yes, good point. See my suggestions above and below after reading your understandable confusion :)

Modules/Module_Processing/Module_Population/xASL_im_CreateAnalysisMask.m
HenkMutsaerts: No, here we load the template, and we create the group mask later. Shall I also change this variable name in this function to clarify this?

E.g.:

`PathSusceptibilityMask` -> `pathTemplateSusceptibilityMask`

Modules/Module_Processing/Module_Population/xASL_im_CreateAnalysisMask.m
HenkMutsaerts: Same as previous comment

Modules/Module_Processing/Module_ASL/xASL_wrp_CreateAnalysisMask.m
HenkMutsaerts: I have now improved the comments here, is this now sufficiently clear?

Modules/Module_Processing/Module_ASL/xASL_wrp_CreateAnalysisMask.m
```diff
@@ -206,11 +214,18 @@ function xASL_wrp_CreateAnalysisMask(x)
      FinalThreshold = (ThresholdIntensity+SortedInt)/2; % average of 2 thresholds
      FinalMask = MixedIm>FinalThreshold;
 
-     MaskSusceptibility = BrainMask;
-     MaskSusceptibility(MaskSuscept) = FinalMask(MaskSuscept);
+     % %% Brief explanation what these masks here mean:
```
HenkMutsaerts, line 217: @jan-petr Here, you had a comment that these masks where unclear. So I added some comments to clarify them.

For some strange reason, this conversation disappeared...

jan-petr: I guess that the issue is that you have to check the PR or the commit summary, not the individual comments. I see the orgiinal comment and I have resolved it already.

Modules/Module_Processing/Module_Population/xASL_im_CreateAnalysisMask.m
jan-petr: I did the rename as suggested. And agree with the rest. You can resolve this.

Modules/Module_Processing/Module_Population/xASL_im_CreateAnalysisMask.m
jan-petr: Agree. Can be resolved.

Modules/Module_Processing/Module_Population/xASL_im_CreateAnalysisMask.m
jan-petr: Agree

Modules/xASL_module_Population.m
jan-petr: Agree. Renaming helps and I did it.

jan-petr (approved):
OK.

PR #1507, opened 2023-10-07
Closes #1506 parallelization lock behavior

### LINKED ISSUE #1506: Parallelization BIDS2Legacy: Improve lock behavior
### Description

From cda98f7f349252dca80ed539c41fa3ad1cbcb954
to 8985da91e851d76f79b51bfca075515da593a59f

- [x] BIDS2Legacy module lock dir has only the subject name, without the 'sub-' prefix, unlike the other modules. This needs to be added for the BIDS2Legacy lock dir.

- [x] BIDS2Legacy uses BIDS notation without the '_1' suffix, but ExploreASL's processing uses the '_1' suffix, for comparison between lock dirs, this needs to be added.

- [x] Copy `dataPar.json` if it exists in studyRoot, to the derivatives folder

- [x] Before merging, fix commit messages - there were some minor errors.

- [x] When a single subject is locked for BIDS2Legacy, the pipeline shouldn't continue with the next module: this behavior needs to be tested. If this is not default behavior, we should lock the other processing modules during BIDS2Legacy and unlock at the end -> reconsidering, we don't need to test this. We just have to improve the parallelization, to ensure that no subjects are performed twice. Or should we? When the structural module is locked of one subject, and we rerun ExploreASL, will it run the ASL module? So all modules should get a "hierarchy", and they cannot run if the same subject is locked? Or if any of the "previous" modules are locked?

- [x] Improved the parallelization. Currently we had:
     nSubjPerWorker = ceil(nSubjects/nWorkers);
    iStartSubject = (iWorker-1)*nSubjPerWorker+1;
    iEndSubject = min(iWorker*nSubjPerWorker, nSubjects);

E.g., with nSubjects=10, and nWorkers=3, it used to be
worker 1 = 1:4
worker 2 = 5:8
worker 3 = 9:10

Now implemented (suggestion from Jan):
Rounding for each group. That is 10:3, then round, giving 
worker 1 = 1:3.33
worker 2 = 4.34:6.66
worker 3 = 7:67:10

after rounding this gives:
worker 1 = 1:3
worker 2 = 4:7
worker 3 = 8:10

### Release notes

Report with #1275. Allow parallelization of BIDS2Legacy.

HenkMutsaerts:
### Linked issue

Closes #1506


Functions/@xASL_GoNoGo/xASL_GoNoGo.m
```diff
@@ -84,7 +84,7 @@ function delete(obj)
                     state = regexprep(state,'\.status$','');
                 end
             else
-                % WARNING: not locked yet
```
jan-petr, line 87: Are you sure this warning is really meant to be printed for the user, or was this only an internal comment?

Functions/BIDS/xASL_bids_BIDSifyASLJSON.m
```diff
@@ -64,12 +64,21 @@
 jsonInMerged = xASL_bids_MergeStudyPar(jsonIn, studyPar, 'asl');
 	
 %% 3. Extract the scaling factors from the JSON header
+% Fujifilm fix
+if isfield(jsonInMerged, 'StationName') && ~isfield(jsonInMerged, 'Manufacturer')
```
jan-petr, line 68: Fine. But not very transparent to add unrelated changes to this issue. This should have been done and tested in a Fujifilm issue. Next time ;)

Functions/Initialization/xASL_init_InitializeMutex.m
```diff
@@ -1,4 +1,4 @@
-function [x] = xASL_init_InitializeMutex(x, ModuleName)
+function [x, bLocked] = xASL_init_InitializeMutex(x, ModuleName)
 %xASL_init_InitializeMutex Initialize mutex system for ExploreASL modules
 %
 % FORMAT: [x] = xASL_init_InitializeMutex(x, ModuleName)
```
jan-petr: Please modify format and output

Modules/Module_Processing/xASL_init_DefineStudyData.m
```diff
@@ -286,12 +302,18 @@
 % ------------------------------------------------------------------------------------------------
 %% 7) Parallelization: If running parallel, select cases for this worker
 if x.opts.nWorkers>1
-    nSubjPerWorker = ceil(x.dataset.nSubjects/x.opts.nWorkers); % ceil to make sure all subjects are processed
-    nSubjSessPerWorker = nSubjPerWorker*x.dataset.nSessions;
-    iStartSubject = (x.opts.iWorker-1)*nSubjPerWorker+1;
-    iEndSubject = min(x.opts.iWorker*nSubjPerWorker, x.dataset.nSubjects);
-    iStartSubjectSession = (x.opts.iWorker-1)*nSubjSessPerWorker+1;
-    iEndSubjectSession = min(x.opts.iWorker*nSubjSessPerWorker, x.dataset.nSubjectsSessions);
+    nSubjPerWorker = x.dataset.nSubjects/x.opts.nWorkers; % ceil to make sure all subjects are processed
```
jan-petr: Adapt comment as `ceil` was removed.

Modules/Module_Processing/xASL_init_DefineStudyData.m
```diff
@@ -286,12 +302,18 @@
...
+
+    % e.g., if nWorkers=3 & nSubjects=10
+    % iWorker 1 does [1 2 3]
+    % iWorker 2 does [4 5 6 7]
+    % iWorker 3 does [8 9 10]
+
+    iStartSubject = round((x.opts.iWorker-1)*nSubjPerWorker+1);
+    iEndSubject = min( round(x.opts.iWorker*nSubjPerWorker), x.dataset.nSubjects);
+
+    iStartSubjectSession = (iStartSubject-1) * x.dataset.nSessions + 1;
```
jan-petr, line 315: I would prefer to remove this SubjectSession parameter completely. This is only going to be a source of errors for studies with uneven number of subjects!

We should always processdata based on subjects and for each subjects we go through sessions.

but the subjectsession thing cannot ever work. Best to remove it here to make sure it's not used anywhere. Especially when we are doing this cleaning.

jan-petr: Also below - lets delete all mention of nSubjectSessions and SubjectSessions  everywhere.

Functions/Initialization/xASL_init_InitializeMutex.m
jan-petr: See my other comment and revert this added bLocked parameter.

Modules/xASL_module_ASL.m
```diff
@@ -58,7 +58,14 @@
 %% Admin
 result = false;
 [x] = xASL_init_SubStructs(x);
-x = xASL_init_InitializeMutex(x, 'ASL'); % starts mutex locking process to ensure that everything will run only once
+[x, bLocked] = xASL_init_InitializeMutex(x, 'ASL'); % starts mutex locking process to ensure that everything will run only once
```
jan-petr: I understand what are you trying to do and I agree that we should make sure that we don't execute processing even if another subpart of the subject is locked. 

BUT, you are not doing that correctly. You have your MUTEX function that should take care of all that. 

And that mutex should either create a nice lock and let you through. Or it should report an error and not let you through. This was nicely programmed by Paul. You have tried to hack it and it would work, but you are sort of delegating MUTEX functionality outside of the mutex. Which goes against how mutex should be used. The mutex should really control everything. But you are now taking its functionality outside and letting other functions to decide if they will respect that or not. 

So I agree with what you are doing in general, but the code has to change. This function should stay as it was, no bLocked. This should all be handled by the mutex itself. In the same way as if a specific module was locked - it will simply refuse to lock again and skip it.

jan-petr (changes requested):
See my comment in xASL_module_ASL first. I agree with what you are doing, but you have to do it differently.

Modules/xASL_module_BIDS2Legacy.m
```diff
@@ -56,8 +56,17 @@
         diary(x.dir.diaryFile); % Start diary
     end
     
-    x = xASL_init_InitializeMutex(x, 'BIDS2Legacy'); % Start Mutex
+    [x, bLocked] = xASL_init_InitializeMutex(x, 'BIDS2Legacy'); % Start Mutex
```
jan-petr: Same here - we shouldn't use bLocked parameter.

Modules/xASL_module_DARTEL.m
```diff
@@ -39,7 +39,15 @@
 %% -------------------------------------------------------------------------------------------------------
 %% Administration
 [x] = xASL_init_SubStructs(x);
-x = xASL_init_InitializeMutex(x, 'DARTEL'); % starts mutex locking process to ensure that everything will run only once
+[x, bLocked] = xASL_init_InitializeMutex(x, 'DARTEL'); % starts mutex locking process to ensure that everything will run only once
```
jan-petr: Same here - no bLocked.

Modules/xASL_module_LongReg.m
```diff
@@ -48,7 +48,15 @@
 %% --------------------------------------------------------------
 %% Administration
 [x] = xASL_init_SubStructs(x);
-x = xASL_init_InitializeMutex(x, 'LongReg' ); % starts mutex locking process to ensure that everything will run only once
+[x, bLocked] = xASL_init_InitializeMutex(x, 'LongReg'); % starts mutex locking process to ensure that everything will run only once
```
jan-petr: please remove bLocked.

Modules/xASL_module_Population.m
```diff
@@ -69,7 +70,15 @@
     warning('Detected no ASL scans, skipping ASL-specific parts of the Population module');
 end
 
-x = xASL_init_InitializeMutex(x, 'Population'); % starts mutex locking process to ensure that everything will run only once
+[x, bLocked] = xASL_init_InitializeMutex(x, 'Population'); % starts mutex locking process to ensure that everything will run only once
```
jan-petr: please remove bLocked.

Modules/xASL_module_Structural.m
```diff
@@ -42,7 +42,15 @@
 result = false;
 
 [x] = xASL_init_SubStructs(x);
-x = xASL_init_InitializeMutex(x, 'T1'); % starts mutex locking process to ensure that everything will run only once
+[x, bLocked] = xASL_init_InitializeMutex(x, 'T1'); % starts mutex locking process to ensure that everything will run only once
```
jan-petr: please remove bLocked.

Functions/Initialization/xASL_init_InitializeMutex.m
```diff
@@ -64,16 +64,80 @@
...
+	fprintf('\n');
+    fprintf('This means that this module is currently being parallel processed by another Matlab instance/worker\n');
 	fprintf('If this is not the case, the locked folder needs to be removed before proceeding\n');
 	fprintf('Also, check that there is no filesystem permission issue\n');
     fprintf('Otherwise this error can be ignored\n');
-	return;
-end
+    fprintf('\n');
+
+    bLocked = 1;
```
jan-petr: We sholdn't use this bLocked. 

The MUTEX has to work in exactly the same way as before. With the only exception that while creating a "lock", it will check if the entire subject is free of locks. 

But this should be done elsewhere.

Functions/Initialization/xASL_init_InitializeMutex.m
```diff
@@ -64,16 +64,80 @@
...
+else
+    
+    %% 4) Check if any module for this subject is locked
+    if ~strcmpi(x.ModuleName, 'population')
```
jan-petr, line 88: For the population module. I propose that it won't even try to lock the directory if it finds out it runs in parallel mode...

jan-petr: But all this still done elsewhere.

Functions/Initialization/xASL_init_InitializeMutex.m
```diff
@@ -64,16 +64,80 @@
...
+        % the population module is subject-independent
+        if ~isfield(x, 'mutex') || ~isprop(x.mutex, 'Root')
```
jan-petr, line 96: this all has to be done elsewhere.

Functions/BIDS/xASL_bids_CompleteBIDS2Legacy.m
```diff
@@ -106,7 +113,7 @@
 
     % Define file names
     pathParticipantsTSV = fullfile(x.dir.DatasetRoot, 'participants.tsv');
-    pathParticipantsTSVxASL = fullfile(fullfile(x.dir.DatasetRoot,'derivatives','ExploreASL'), 'participants.tsv');
+    pathParticipantsTSVxASL = fullfile(x.dir.DatasetRoot,'derivatives','ExploreASL', 'participants.tsv');
```
jan-petr, line 116: Who wrote the ugly original thing?! :D

Functions/BIDS/xASL_bids_CompleteBIDS2Legacy.m
```diff
@@ -130,7 +137,7 @@
 
 % Copy dataset_description JSON file
 xASL_Copy(fullfile(x.dir.DatasetRoot, 'rawdata', 'dataset_description.json'), ...
-	fullfile(fullfile(x.dir.DatasetRoot,'derivatives','ExploreASL'), 'dataset_description.json'));
+	fullfile(x.dir.DatasetRoot,'derivatives','ExploreASL', 'dataset_description.json'));
```
jan-petr, line 140: I hope it wasn't me :D

jan-petr (changes requested):
A few more comments.

Functions/BIDS/xASL_bids_BIDSifyASLJSON.m
HenkMutsaerts: You're right. It was just a very small fix, and you never know this is really Fujifilm-specific. Manufacturer could always be missing (we could issue a warning though).

Modules/Module_Processing/xASL_init_DefineStudyData.m
HenkMutsaerts: This is defined here for all processing modules. We need to establish this, for splitting the workers.
Why is this is a source of errors for studies with uneven number of subjects?

ExploreASL_Process (at the beginning) -> xASL_init_DataLoading -> xASL_init_DefineStudyData

So what do you suggest as alternative, that each (sub)module searches for their own subjects & sessions?

Functions/Initialization/xASL_init_InitializeMutex.m
HenkMutsaerts: True, this is just a sanety check. The population module won't run if ExploreASL has multiple workers.

Functions/BIDS/xASL_bids_CompleteBIDS2Legacy.m
HenkMutsaerts: Github copilot etc tells me this but I need to re-install this.

Functions/@xASL_GoNoGo/xASL_GoNoGo.m
HenkMutsaerts: No clue. This rewrite just seemed better

Modules/xASL_module_ASL.m
HenkMutsaerts: Are you sure? `xASL_GoNoGo` (== what you call the MUTEX function) is module- and subject-ignorant. It only handles the current module and folder.

So now `xASL_GoNoGo` is only the function for single module mutex handling.
`xASL_init_InitializeMutex` is the wrapper around this, dealing with subject and other modules.

Your suggestion to put both inside `xASL_GoNoGo` would combine the wrapper and function, to me this seems as more dirty programming (less layered/modular). Perhaps, there is something I miss that could be saving a bit code and debugging with the object-oriented programming of `xASL_GoNoGo`, which could be the benefit of expanding `xASL_GoNoGo`, is that what you mean?

But I would not mix up the "mutex function" `xASL_GoNoGo` and "mutex subject/module wrapper" `xASL_init_InitializeMutex`.

Or are you fine with the above, but mainly would like to add this information of other modules to `x.mutex`, and deal with this in `xASL_Iteration`?

Functions/@xASL_GoNoGo/xASL_GoNoGo.m
jan-petr: I would revert this one then ;)

Modules/Module_Processing/xASL_init_DefineStudyData.m
jan-petr: The SubjectSession parameter is generally used in a way that assumes we have the same number of sessions per subject - see line 315 for example. 

But this assumption won't hold anymore as we can have variable session numbers. So continuous counting of sessions across subjects won't match that formula anymore. Maybe this concept will still work and some SubjectSessions will be empty and skipped, or we might run into errors and mess up the processing order badly. We have previously seen some weird behavior and some scans not randomly processed - this could be the source of the error. 

In any case, we have argumented for Import and BIDS2legacy that we should split workers not for subjectsessions, but for subjects and then for each subject do all sessions. Even though it might happen that not all workers will then process the same amount of sessions.

So yes - I suggest to split workers to subjects. Then for each subject, we go across sessions/runs. And we definitively need to do this for BIDS2legacy and for import for various reasons. So to keep BIDS2legacy+processing in sync, we need to do this across subjects everywhere. This was supposed to be the major edit in this BIDS2legacy parallelization.

I'm of course happy to discuss this more.

Modules/xASL_module_ASL.m
jan-petr: Sorry. I might have missed on a few details. But I hope I got the overall message right. I didn't want to go into details of what is happening inside `xASL_GoNoGo` or inside `xASL_init_initializeMutex`. But my point was that some MUTEX function should either lock the folder or say it can't be locked, but all that locking should happen modularily inside it and it shouldn't output any variables outside to be kept and used by the user. So the whole process should be screened from the user. 

I do not know how this exactly works inside these functions, but seeing that you have `bLocked` that you have to deal with yourself outside of the MUTEX wasn't clean. So adding this to `x.mutex` sounds like a solution that will make me happy :)

I guess we are now on a similar note. Try to adapt the code and I'll then go step by step through it and verify it.

Functions/Initialization/xASL_init_InitializeMutex.m
jan-petr: OK. As a second level sanity check, this is fine. Could you please add that inside the code in a comment, so that we understand this later as well?

jan-petr (changes requested):
I provided a few clarifications. You can now proceed with code changes.

Functions/@xASL_GoNoGo/xASL_GoNoGo.m
jan-petr: So. I have checked what this really does. And one of the situations when this happens is when there is already the 'locked' directory. So in case it is locked (from previously or by another process), it prints this warning (which is incorrect, the directory is already locked and not unlocked. And then it doesn't assign the output parameter `b` so the entire thing crashes afterwards in the parent function because the function - e.g. function x.mutex.HasState doesn't return anything.

So this warning and behavior has to change.

HENK: can you then please fix this? I have never written this or changed this, and this object-orientated programming in Matlab is new to me. I now only changed the warning, which I'm happy for you to improve.

jan-petr (changes requested):
See the answer to my comment about xASL_GoNoGo

Modules/xASL_module_ASL.m
HenkMutsaerts: Now throughout `bLocked` -> `x.mutex.bAnyModuleLocked`

Functions/@xASL_GoNoGo/xASL_GoNoGo.m
HenkMutsaerts: Do you mind fixing this? I haven't touched this behavior, I only tweaked the warning.

Functions/Initialization/xASL_init_InitializeMutex.m
HenkMutsaerts: Now throughout bLocked -> x.mutex.bAnyModuleLocked

Modules/Module_Processing/xASL_init_DefineStudyData.m
HenkMutsaerts: Done

Modules/Module_Processing/xASL_init_DefineStudyData.m
HenkMutsaerts: That's exactly the plan. 

At the start of the processing, `xASL_init_DefineStudyData` takes the maximum number of sessions and indeed uses this for all subjects. Missing sessions will be skipped, I fixed that in the past.
If you want to tweak our current use of x.nSessions and x.SESSIONS, such that x.SESSIONS varies between subjects, this will be a major overhaul of ExploreASL. Instead, what I suggest is that we keep this behavior and simply skip unexisting sessions.

The parallelization into different workers was already only split between subjects, and I will keep it this way.
So e.g.,

sub-01 has ASL_1 ASL_2 ASL_3
sub-02 has ASL_2 only

nWorkers is 2.

Then iWorker == 1 will do sub-01 including ASL_1 ASL_2 ASL_3
iWorker == 2 will do sub-02 structural, skip ASL_1, do ASL_2, skip ASL_3

Is this OK for you?

Modules/xASL_module_BIDS2Legacy.m
HenkMutsaerts: Now throughout bLocked -> x.mutex.bAnyModuleLocked

Modules/xASL_module_DARTEL.m
HenkMutsaerts: Now throughout bLocked -> x.mutex.bAnyModuleLocked

Modules/xASL_module_LongReg.m
HenkMutsaerts: Now throughout bLocked -> x.mutex.bAnyModuleLocked

Modules/xASL_module_Population.m
HenkMutsaerts: Now throughout bLocked -> x.mutex.bAnyModuleLocked

Modules/xASL_module_Structural.m
HenkMutsaerts: Now throughout bLocked -> x.mutex.bAnyModuleLocked

Functions/Initialization/xASL_init_InitializeMutex.m
HenkMutsaerts: Now throughout bLocked -> x.mutex.bAnyModuleLocked.
Note that x.mutex.Locked still checks for the CURRENT module only if it is locked or not.
x.mutex.bAnyModuleLocked checks this for the CURRENT module PLUS all other modules if they are locked or not.
It would be cleaner perhaps to remove x.mutex.Locked, but I don't want to touch Paul's code.

Functions/Initialization/xASL_init_InitializeMutex.m
HenkMutsaerts: Done

Functions/Initialization/xASL_init_InitializeMutex.m
HenkMutsaerts: Per your other comments it seems that you agree now that we do this here right?
At the start of each module, `xASL_init_InitializeMutex` runs `xASL_GoNoGo`, which initializes `x.mutex.HasState` etc. `xASL_GoNoGo` also does a check if the current module is locked, but it doesn't check if other modules are locked (`xASL_GoNoGo` is essentially subject-agnostic, using its own module lock/status folder only)

I have now clarified this further in the comments

Modules/Module_Processing/xASL_init_DefineStudyData.m
jan-petr: OK. Makes sense. Having nSessions at max and skipping missing sessions is perfectly fine! The example described is then perfectly fine. OK for me. 

However, if it is like this, I do not understand what we need subjectSessions for. If we assume a unified number of sessions per subject and skipping of missing, then at each point of time, the conversion from subjectSession index to iSubject and iSession should be straightforward and the other way around. So the variable subjectSession has no added value and can only help us if added locally, but not globally.
Therefore, I would propose to get rid of it everywhere from the x-struct and at most keep at local function level or even remove from there.

Cause otherwise I'm afraid that it could be bug prone. Or do we need it for some other purpose?

Functions/@xASL_GoNoGo/xASL_GoNoGo.m
jan-petr: OK. Will investigate and fix this.

Functions/Initialization/xASL_init_InitializeMutex.m
jan-petr: Sounds good. Will check and test.

Modules/xASL_module_ASL.m
jan-petr: Sounds good. Will check and test in detail.

Modules/xASL_module_BIDS2Legacy.m
jan-petr: Sounds good. Will check and test in detail.

Modules/xASL_module_DARTEL.m
jan-petr: Sounds good. Will check and test in detail.

Modules/xASL_module_LongReg.m
jan-petr: Sounds good. Will check and test in detail.

Modules/xASL_module_Population.m
jan-petr: Sounds good. Will check and test in detail.

Modules/xASL_module_Structural.m
jan-petr: Sounds good. Will check and test in detail.

Functions/Initialization/xASL_init_InitializeMutex.m
jan-petr: Sounds good. Will check and test in detail.

Functions/Initialization/xASL_init_InitializeMutex.m
jan-petr: Sounds good. Will check and test in detail.

jan-petr (changes requested):
Please check my replies.

Modules/Module_Processing/xASL_init_DefineStudyData.m
HenkMutsaerts: We use x.datasets.nSubjectsSessions 32 times in various parts of the code, and this is always x.dataset.nSubjects * x.dataset.nSessions. So it just saves a line of coding 32 times :)

We could change this but that could then actually induce bugs. At least I wouldn't do this in this issue :) OK?

jan-petr: Indeed. A new issue #1537 created.

Functions/@xASL_GoNoGo/xASL_GoNoGo.m
```diff
@@ -106,7 +116,7 @@ function delete(obj)
                     delete(fullfile(obj.Root, obj.ID, [delStates{ii} '.status']));
                 end
             else
-                % WARNING: not locked yet
+                warning([someState ' not locked yet']);
```
jan-petr: I still have to check when this is executed etc. I will do this later.

Functions/Initialization/xASL_init_InitializeMutex.m
jan-petr: This works really well now. I mean the concept is nice, but there are some bugs :)
You can resolve this issue, but keep the others open for now.

Functions/Initialization/xASL_init_InitializeMutex.m
```diff
@@ -63,17 +63,99 @@
...
+						
+			% Go through all module that have lock dirs already and are thus potentially locked
+			checkModuleList = xASL_adm_GetFileList(lockDir, '.*', 'FPList', [], 1);
+            for iCheckModule=1:length(checkModuleList)
+                [~, checkModuleName] = fileparts(checkModuleList{iCheckModule});
 
+				% Verify that it is really a directory with an ASL module lock files
+                if strcmp(checkModuleName(1:12), 'xASL_module_')
+                    checkModuleName = checkModuleName(13:end); % Extract the name of the module
+                    if strcmpi(checkModuleName, x.ModuleName)
```
jan-petr, line 110: Note that for the structural module, the name isn't xASL_module_Structural but xASL_module_T1.

This function will still work at the end, but it's not really too good to have this different name. I wasn't sure where to change that and I also wasn't sure if this wouldn't create more problems. You know more about the history and reasons, so if you say that changing the name is fine and non-problematic, then I'll locate that and do the necessary changes.

Functions/Initialization/xASL_init_InitializeMutex.m
```diff
@@ -63,17 +63,99 @@
...
+							% List all visits
+							checkModulePath = xASL_adm_GetFileList(checkModuleList{iCheckModule}, ['^' subjectName '_\d+$'], 'FPListRec', [], 1);
+							for iVisit = 1:length(checkModulePath)
+								if ~isempty(xASL_adm_GetFileList(fullfile(checkModulePath{iVisit}, ['xASL_module_' checkModuleName]), '^locked$', 'FPListRec', [], 1))
+									bLockedFolders = true;
+								end
+							end
+						elseif ~strcmp(x.ModuleName, 'BIDS2Legacy') && strcmp(checkModuleName, 'BIDS2Legacy')
+							% In case we are in another module and checking for BIDS2Legacy, we have to remove the suffix
+							ivisitSuffix = regexp(subjectName, '_\d+$');
```
jan-petr, line 162: It was correct what you had in the code - to remove suffix. But only if the current sessions wasn't BIDS2Legacy and we are checking BIDS2Legacy session. Otherwise we don't need to remove the suffix.
I have fixed that, you can resolve.

Functions/Initialization/xASL_init_InitializeMutex.m
```diff
@@ -63,17 +63,99 @@
...
+                    if strcmpi(checkModuleName, x.ModuleName)
+                        % this is the current module, so we can continue if this is locked
+                    elseif strcmpi(checkModuleName, 'population') || strcmpi(checkModuleName, 'import')
+                        % for population or import modules, we can continue if other modules are locked
+                    else
+                        % For BIDS2Legacy, no visit suffix is given. For other modules, a suffix is given
+						bLockedFolders = false;
+						
+						if strcmp(x.ModuleName, 'BIDS2Legacy') && ~strcmp(checkModuleName, 'BIDS2Legacy')
+							% In case we are in BIDS2Legacy and are checking other modules, we need to add a suffix
```
jan-petr: In this case (we are in a BIDS2Legacy session but checking for possibly running structural module), we actually need to check all visits, so we need to go through all visits.
I have fixed that.
HOWEVER, this will be even more complicated for the ASL module. Because we have to check the lock folders inside the ASL sessions :) So we need to go through all visits, and for ASL also for all sessions. I haven't programmed that yet, but you or I need to do that. I can do that tomorrow.

Functions/Initialization/xASL_init_InitializeMutex.m
```diff
@@ -63,17 +63,99 @@
...
+							% In case we are in another module and checking for BIDS2Legacy, we have to remove the suffix
+							ivisitSuffix = regexp(subjectName, '_\d+$');
+							if  ~isempty(ivisitSuffix)
+								subjectName = subjectName(1:ivisitSuffix-1);
+							end
+							checkModulePath = fullfile(checkModuleList{iCheckModule}, subjectName, ['xASL_module_' checkModuleName]);
+							bLockedFolders = ~isempty(xASL_adm_GetFileList(checkModulePath, '^locked$', 'FPListRec', [], 1));
+						end
+                        
+                        if bLockedFolders
```
jan-petr, line 170: It was attempting to print all locked modules, but that's not possible because your path checking can only always check at a single location, not multiple modules at once.
I have fixed this, you can resolve.

jan-petr (changes requested):
See some justifications for code changes and also some things to be still done.

Functions/Initialization/xASL_init_InitializeMutex.m
HenkMutsaerts: That's easy right, by creating a recursive list of the lock folders?

jan-petr: As discussed. We don't want to load the entire directory tree, but only check at the specific folders. I have fixed that for structural and I will also (specifically) load the ASL directories for all sessions. Yes, easy to do, I just didn't have the extra 10 minutes to do that properly, so I only wrote it to a comment :)

I have now restructured what I did previously, added a proper check for multi-run modules. And I wrote it in a way that we will later be easily able to extend it for BIDS2Legacy having sessions.

A new issue for BIDS2Legacy having session is created #1538 

So this should be now nice and clear with the future possibility of further improving it.

Fixed here - 895e3a41
You can resolve it.

Functions/Initialization/xASL_init_InitializeMutex.m
jan-petr: It seems that the xASL_module_T1 name instead of xASL_module_Structural is given inside xASL_module_Structural.m and then it's only used in a few messages and inside this initi_InitializeMutex function - so we can safely change that to make the name consistent with the module name.

I did that already, so you can resolve this comment. 964adbea

jan-petr (approved):
OK

PR #1508, opened 2023-10-07
Closes #1486 whm segm empty

### LINKED ISSUE #1486: White matter lesion volume and count are incorrectly set to NaN
### Description

White matter lesion volume and count are incorrectly set to NaN, when there are actually no lesions detected. 

Solution is to create a dummy empty `WMH_SEGM.nii` in case of a FLAIR without WMH (currently, LST outputs no WMH_SEGM.nii in this case)

### Tasks
- [x] xASL_wrp_LST_Segment_FLAIR_WMH: at the end, if exist(FLAIR) && ~exist(WMH_SEGM), then create empty dummy WMH_SEGM.nii (and print warning)
- [x] If any WMH is found, their volumes should be added to participants.tsv; this depended on the last subject having WMH. New boolean created to keep track of any of the participants having a WMH volume

### How to test

@MDijsselhof 

### Release notes

Count FLAIR without lesions as having zero instead of NaN lesions.

HenkMutsaerts:
### Linked issue

Closes #1486

### DIFF
--- a/Modules/Module_Processing/Module_Population/xASL_stat_GetVolumeStatistics.m
+++ b/Modules/Module_Processing/Module_Population/xASL_stat_GetVolumeStatistics.m
@@ -29,6 +29,8 @@ x.S.unit = 'L';
 
 fprintf('%s\n',['Printing tsv-files with ' x.S.output_ID ' statistics...  ']);
 
+bAnyWMHFound = false; % by default we don't add WMH volumes to participants.tsv below,
+% only if these are found
 
 for iSubject=1:x.dataset.nSubjects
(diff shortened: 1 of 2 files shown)


jan-petr (approved):
Looks OK. Depends if Mathijs can confirm that it really works.

MDijsselhof:
MATHIJS: Something going wrong when adding WMH volumes with  `xASL_bids_Add2Participants.tsv`, which is done in for every subject/session for all derived values, with this note:

image

Using this location:

/home/mdijsselhof/lood_storage/divi/Projects/ExploreASL/ABBA/NORMENT/StrokeMRI/TestWMH/

MDijsselhof (approved):

PR #1512, opened 2023-10-12
Closes #1511 xASL_wrp_ProcessM0: Warning discrepancy M0 nVolumes ~ JSON nTE

### LINKED ISSUE #1511: Discrepancy warning for M0 volume ~ nTE
### Description

When the length of EchoTime differs from the number of M0 volumes, this does not give a clear warning.

Thanks to Mervin for noticing this.

### Tasks

- [x] xASL_wrp_ProcessM0: Warning discrepancy M0 nVolumes ~ JSON nTE

### Release notes

Issue a warning when there are more TEs in JSON than volumes in M0.

HenkMutsaerts:
### Linked issue

Closes #1511

### DIFF
--- a/Modules/Module_Processing/Module_ASL/xASL_wrp_ProcessM0.m
+++ b/Modules/Module_Processing/Module_ASL/xASL_wrp_ProcessM0.m
@@ -110,6 +110,12 @@ if xASL_exist(pathM0json, 'file')
 		% Select only non-zero TEs
 		uniqueNonzeroTE = unique(jsonM0.EchoTime(jsonM0.EchoTime > 0));
 		if length(uniqueNonzeroTE) > 1
+            % First check if the number of TEs equals the number of M0 volumes
+            if length(uniqueNonzeroTE)~=size(imM0,4)
+                warning('M0 JSON contains more echo times than number of M0.nii volumes');
+                fprintf('%s\n', 'Hint: check if your data are BIDS compatible');
+            end
+
 			% Find the smallest of the previously selected positive nonzero TEs
 			minTE = min(uniqueNonzeroTE);
 			fprintf('Multi-TE M0 present. Using the shortest TE=%.1f ms for M0 calibration.\n', minTE*1000);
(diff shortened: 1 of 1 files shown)


jan-petr (changes requested):
That warning does not really make sense. This is about Unique TEs. But it doesn't check for repetitions.

It's quite possible to have multiple TEs and repeated acquisition. If you will have 3TEs and repeated 2 times, then `uniqueNonzeroTE` will be 3, size of M0 will be 6, you will report a warning and everything will be fine and according to BIDS.

HenkMutsaerts:
> That warning does not really make sense. This is about Unique TEs. But it doesn't check for repetitions.
> 
> It's quite possible to have multiple TEs and repeated acquisition. If you will have 3TEs and repeated 2 times, then `uniqueNonzeroTE` will be 3, size of M0 will be 6, you will report a warning and everything will be fine and according to BIDS.

Are you really sure that you can have in BIDS a vector that is longer than 1, that is not according to the number of volumes? I remember that you like this, but as far as I know, with BIDS it can be either a scalar, or a vector with exactly the length of the number of volumes. Otherwise, you don’t know if the vector is incomplete or needs to be repeated.

If you are really sure about this, then I will change the condition to nTE>nM0 (this was the problem I believe).

jan-petr:
> > That warning does not really make sense. This is about Unique TEs. But it doesn't check for repetitions.
> > It's quite possible to have multiple TEs and repeated acquisition. If you will have 3TEs and repeated 2 times, then `uniqueNonzeroTE` will be 3, size of M0 will be 6, you will report a warning and everything will be fine and according to BIDS.
> 
> Are you really sure that you can have in BIDS a vector that is longer than 1, that is not according to the number of volumes? I remember that you like this, but as far as I know, with BIDS it can be either a scalar, or a vector with exactly the length of the number of volumes. Otherwise, you don’t know if the vector is incomplete or needs to be repeated.
> 
> If you are really sure about this, then I will change the condition to nTE>nM0 (this was the problem I believe).

This is not about BIDS. In BIDS, you indeed need to have either 1 or numberOfVolumes and nothing between that...
BUT, in the code, we take the TE vector and call unique on that which *could* result in having a vector that has a length in-between. But that is an internal variable in the function that we are working with, not the BIDS vector.

jan-petr (approved):
OK

PR #1515, opened 2023-10-17
Bug #1514 q cfor all asl sessions

### LINKED ISSUE #1514: QC.json is saved for the latest session and not for all ASL session
### Description

If we have several ASL sessions, then the QC.json and Report.pdf are only saved with the latest of the ASL sessions. And it doesn't even correctly indicate which session is that.
We need to correctly write the session to the PDF. And we need to add a list of sessions to the QC.json rather than listing only one non-specific session.

### Tasks

- [x] While collecting QC, we don't collect ASL_1 with hardcoded 1, but instead collect the current session. 20c15e1
- [x] While saving QC to json, we save different ASL sessions into x.Output.ASL.ASL_1 etc. 167e48f and 16f90ef
- [x] While saving QC.json, we save all that is in x.Output. But if we processing sessions 1 and 2. And then later add session 3 and process again, then sessions 1 and 2 won't be in x and the QC.json will only have session 3 with session 1 and 2 overwritten. I propose to always load old QC.json and updating it. Rather than deleting all previous stuff. It is of course possible that all that was previously in x.Output stayed there, so that reloading QC.json won't be necessary. I'll check both options. JAN: No change was necessary at the end.
- [x] For backwards compatibility - delete everything from x.Output.ASL that is not ASL_XX 5a0cae8
- [x] Update function to create PDF report 06c805f6
- [x] Should we update also `xASL_qc_WADQC_GenerateDescriptor`?? Because it also goes through ASL_1 only, but I'm not sure if this is a mistake or design this time :) @HenkMutsaerts 
       Yes, we should. This is a way to put QC from ExploreASL back in DICOMs, written partly by Joost Kuijer.
       Let's do this later, we'll find this out when using this again, we haven't used this for 5 years now :)
- [x] Create PDF report still only prints session ASL_1. Should we make a new issue for that? @HenkMutsaerts 
       HENK: I wouldn't bother. This code that I wrote just prints a single page so there is probably simply no space, even 
       though you already rewrote my code in this issue to allow more than one ASL session.
       When Maarten applies his PDF that allows multiple pages, and is more flexible, we can print multiple sessions.
       JAN: OK, I will make a follow-up issue then - in #1522 
- [x] We should write a session sub-structure to x.Output also for T1, func, and dwi. And add sessions to this.

### How to test

### Release notes

Save QC values and images for multiple runs

jan-petr:
### Linked issue

Close #1514

Check commits but also remaining tasks in the issue.


Functions/QualityControl/xASL_qc_CollectParameters.m
```diff
@@ -98,8 +102,10 @@
         x.Output.(ScanType).Version_ExploreASL = x.Output.SoftwareVersion.ExploreASL;
         x.Output.(ScanType).Version_Matlab = x.Output.SoftwareVersion.Matlab;
         x.Output.(ScanType).Version_SPM12 = x.Output.SoftwareVersion.SPM12;        
-    case {'ASL', 'dwi', 'func'}
+    case {'dwi', 'func'}
```
HenkMutsaerts: This should work equal. They can all have sessions (runs)

Functions/QualityControl/xASL_qc_CollectQC_ASL.m
```diff
@@ -192,6 +192,15 @@
         x.Output.('ASL').(SessionID) = xASL_qc_FillFields(x.Output.('ASL').(SessionID), ASL);
     end
 
+	% Keep only subfields in x.Output.ASL for each ASL session, but no other fields that might have been filled previously by mistake
+	listFields = fields(x.Output.('ASL'));
```
HenkMutsaerts: What is the difference here with:

listFields = fields(x.Output.ASL);

?

Functions/QualityControl/xASL_qc_CollectQC_ASL.m
```diff
@@ -192,6 +192,15 @@
...
+
+	for iField = 1:length(listFields)
```
HenkMutsaerts: I would at least add a warning when we remove pre-existing x.Output.ASL fields.

But for removing existing x.Output.ASL.<ses>,  you catch your own bugs right?
There is no other reason for those fields to exist that need removing (other than the current one)?

I mean: if field ASL_1 already exists in the old struct format, and we rerun ExploreASL with the new code for ASL_2 only, this should issue a warning because we don't know which field we remove (or do we, does the existing field include the name of the ASL session? We could check this in the TestDataSets where there are datasets where we test a rerun).

Also, I may have missed this, but how does this affect parameters that were added to the current ASL field throughout the ASL module?

Functions/QualityControl/xASL_qc_CollectQC_ASL.m
```diff
@@ -192,6 +192,15 @@
...
+		currentField = listFields{iField};
+		if isempty(regexp(currentField, 'ASL_\d+'))
+			x.Output.('ASL') = rmfield(x.Output.('ASL'), currentField);
```
HenkMutsaerts: Also here, why not simply x.Output.ASL?

Functions/QualityControl/xASL_qc_CreatePDF.m
```diff
@@ -111,8 +111,29 @@ function xASL_qc_CreatePDF(x, DoSubject)
             Ind = 1;
         else
             Ind = iSubjSess;
-        end
-        StrucFields{iField} = sort(fieldnames(x.Output.(OutputFields{iField})(Ind)));
+		end
```
HenkMutsaerts, line 114: Note that this change won't have any effect, or only temporarily, because Maarten will soon implement his PDF printing.

Functions/QualityControl/xASL_qc_CollectParameters.m
```diff
@@ -113,7 +113,26 @@
 %% -----------------------------------------------------------------------------------------------
 %% Save QC output
 QC_Path = fullfile(x.D.ROOT, x.SUBJECTS{iSubject}, ['QC_collection_' x.SUBJECTS{iSubject} '.json']);
-xASL_delete(QC_Path);
+
+% Load QC.json
+if xASL_exist(QC_Path, 'file')
```
HenkMutsaerts: Somehow this part is not collaborating with the part above, where QC_path is already loaded etc.

HenkMutsaerts (changes requested):
Few minor things. At least we should change this subfield behavior equally for the other non-structural data that can have sessions/runs.

We may also want to do this directly for structural as well, to save some code and make our future life easier, but you probably have a better overview if this would save coding now or later (or not at all ;) This would entail always using the subfield (so less if-statements), which defaults to 'ses-1' when not defined (e.g. for structural scans). For DWI and fMRI we now use dwi_1 and func_1 but would have to check this, for EPAD.

Functions/QualityControl/xASL_qc_CollectParameters.m
```diff
@@ -1,22 +1,23 @@
...
+function x = xASL_qc_CollectParameters(x, iSubject, ScanType, iSession)
 %xASL_qc_CollectParameters Collect all parameters from structural & ASL, spread over the derivative folders
 %
-% FORMAT: x = xASL_qc_CollectParameters(x, iSubject, ScanType, CollectQCFunction)
+% FORMAT: x = xASL_qc_CollectParameters(x, iSubject, ScanType, CollectQCFunction [, iSession])
 %
 % INPUT:
 %   x                   - structure containing fields with all information required to run this submodule (REQUIRED)
 %   iSubject            - index of current subject (REQUIRED)
 %   ScanType            - string for ScanType, options = 'Structural' 'ASL' 'func' 'dwi' (REQUIRED)
```
jan-petr, line 9: Test

jan-petr (commented):
Test

Functions/QualityControl/xASL_qc_CollectParameters.m
HenkMutsaerts: ?

Functions/QualityControl/xASL_qc_CollectParameters.m
```diff
@@ -1,22 +1,23 @@
...
+%   iSession            - index of current session (OPTIONAL, default 1, but RECOMMENDED for ScanType == 'ASL')
 %
 % OUTPUT:
 %   x                   - same as input
 %
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
-% DESCRIPTION: This function collects QC parameters for a module.
+% DESCRIPTION: This function collects QC parameters for a module for a given subject and for ASL for a given subject+session
```
HenkMutsaerts: Not only for ASL, at least I would do this also for func & dwi

Functions/QualityControl/xASL_qc_CollectParameters.m
```diff
@@ -1,22 +1,23 @@
...
 % 
-% EXAMPLE: x = xASL_qc_CollectParameters_Structural(x, 10, 'ASL');
+% EXAMPLE: x = xASL_qc_CollectParameters_Structural(x, 10, 'ASL', 1);
```
HenkMutsaerts: Perhaps create more than a single example then. If you don't add 1, it will default to 1 right?

Functions/QualityControl/xASL_qc_CollectParameters.m
```diff
@@ -34,37 +35,43 @@
     error('Missing iSubject...');
 end
 
+% It is recommended to provide session number for ASL
+if nargin<4 || isempty(iSession)
+	iSession = 1;
+	if strcmp(ScanType, 'ASL')
```
HenkMutsaerts: Not only ASL

Functions/QualityControl/xASL_qc_CollectParameters.m
```diff
@@ -34,37 +35,43 @@
...
     case 'ASL'
-        QCCollectFunction = @xASL_qc_CollectQC_ASL;
+		% Check for a session subfield and create when necessary
+		if ~isfield(x.Output.ASL, x.SESSIONS{iSession})
+			x.Output.ASL.(x.SESSIONS{iSession}) = struct;
+		end
+		x = xASL_qc_CollectQC_ASL(x, iSubject, iSession);
     case 'func'
-        QCCollectFunction = @xASL_qc_CollectQC_func;
+		x = xASL_qc_CollectQC_func(x, iSubject);
```
HenkMutsaerts: Why no session here? fMRI can have sessions too. Same for dwi.

Functions/QualityControl/xASL_qc_CollectParameters.m
```diff
@@ -34,37 +35,43 @@
...
+
 switch ScanType
     case 'Structural'
-        QCCollectFunction = @xASL_qc_CollectQC_Structural;
+        x = xASL_qc_CollectQC_Structural(x, iSubject);
     case 'ASL'
-        QCCollectFunction = @xASL_qc_CollectQC_ASL;
+		% Check for a session subfield and create when necessary
+		if ~isfield(x.Output.ASL, x.SESSIONS{iSession})
+			x.Output.ASL.(x.SESSIONS{iSession}) = struct;
```
HenkMutsaerts: Should this be here? This should go to the same location as the creation of the other structs I would say. Why don't we move this to `xASL_qc_Collect_QC_ASL`? That just makes it more "modularly" (because here we just checked the scantype, and scantype-specific code is handled in the individual functions).

Also, we need a warning and workaround when reprocessing existing data, i.e. x.Output.ASL could already exist (but perhaps it is always overwritten?)

Functions/QualityControl/xASL_qc_CollectParameters.m
```diff
@@ -34,37 +35,43 @@
...
     x.Output.(ScanType) = struct; 
 end
 
+%% -----------------------------------------------------------------------------------------------
+%% Collect subject-specific (i.e. structural/anatomical) QC results
+
 switch ScanType
     case 'Structural'
-        QCCollectFunction = @xASL_qc_CollectQC_Structural;
+        x = xASL_qc_CollectQC_Structural(x, iSubject);
```
HenkMutsaerts, line 59: Nice simplification of my code :) I can't remember why I used the @function saving, perhaps I thought of using this QC function at multiple locations in the code but I cannot remember seeing it anywhere else indeed :)

HenkMutsaerts (changes requested):
Does this now add to my first review?

Also, I now think you have changed the QC filling inside the registration wrapper/submodule, but there are many more that add QC parameters, they should be checked as well.

HenkMutsaerts (changes requested):
Sorry, just checked and dwi doesn’t allow sessions/runs currently in ExploreASL. Couldn’t check fMRI/func quickly, but I would assume the same. EPAD doesn’t have sessions/runs, and that is the only one that uses the dwi & func/fMRI modules. So you can ignore my suggestions for ASL if-statements in this PR. Although in the future we should perhaps still cater multiple sessions/runs for any scantype

Functions/QualityControl/xASL_qc_CreatePDF.m
jan-petr: Yes. I know, I was just applying changes consistently :)

Functions/QualityControl/xASL_qc_CollectParameters.m
jan-petr: Sorry - I was just testing if a comment is visible here and forgot to remove this. You can ignore this and resolve.

Functions/QualityControl/xASL_qc_CollectParameters.m
jan-petr: Actually - you had it written nicely. In switch, you assigned the function handle and then only later you have used it. But now that we require sessions for ASL, I preferred to write it without function handles to avoid any issues with the optional parameters. But this might further change with the required changes for `DWI` and `func` :D

jan-petr:
> Sorry, just checked and dwi doesn’t allow sessions/runs currently in ExploreASL. Couldn’t check fMRI/func quickly, but I would assume the same. EPAD doesn’t have sessions/runs, and that is the only one that uses the dwi & func/fMRI modules. So you can ignore my suggestions for ASL if-statements in this PR. Although in the future we should perhaps still cater multiple sessions/runs for any scantype

See my reply for that above.

Functions/QualityControl/xASL_qc_CollectParameters.m
jan-petr: Also with respect to the issue at the end:
OK. But dwi, func, structural will still have the SESSIONS and iSESSION parameters assigned. So while only a single session is allowed, I can already adapt the QC so that it would work with more sessions. Do you agree? If yes, I'll implement this now - rather than postponing it and making the code inconsistent.

I'm just asking to be sure that we both agree on this ;)
Don't resolve yet.

Functions/QualityControl/xASL_qc_CollectParameters.m
jan-petr: Taking all the comments together - I will add sessions also here - despite that only sessions 1 are allowed. So don't resolve yet ;)

Functions/QualityControl/xASL_qc_CollectParameters.m
jan-petr: Taking all together - agree. Will do, don't resolve yet.

Functions/QualityControl/xASL_qc_CollectParameters.m
jan-petr: Yes. Will do later.

Functions/QualityControl/xASL_qc_CollectQC_ASL.m
jan-petr: Agree. Fixed in c857e9b4

Functions/QualityControl/xASL_qc_CollectQC_ASL.m
jan-petr: No difference :) I was just copy-pasting previous code that had that. I have simplified this now. Fixed in c857e9b4

Functions/QualityControl/xASL_qc_CollectParameters.m
jan-petr: Agree. Done in c857e9b4

Functions/QualityControl/xASL_qc_CollectQC_ASL.m
jan-petr: Sorry. This was unclear.

Yes - we catch our own bug made in the previous version. So we are not catching a current bug, but cleaning data possibly created with older versions. I now clarify this.

This only concerns fields directly in x.Output.ASL. Fields inside x.Output.ASL.ASL_1 are always fine for any session.

This code won't delete anything for any newly and correctly added parameters.

See this commit for fixing the explanations a249bafb

Functions/QualityControl/xASL_qc_CollectParameters.m
jan-petr: Actually. Above (especially lines 57-67), we create the new QC parameters. 
And here, we load the old QC parameters from file so that we can merge old and new. Because previously, we were deleting old (all previous sessions) and saving new (the current session), so this had to be adapted to be able to keep more than the current ASL session.

Is that fine, shall I add explanatory comments in the code?

Functions/QualityControl/xASL_qc_CollectParameters.m
jan-petr: Good point. I moved it to the function.

I warning is not needed, as we are not overwriting x.Output.ASL, but always only overwritting x.Output.ASL.ASL_X and we want to overwrite by design ;).
See here d7ed87cd

Functions/QualityControl/xASL_qc_CollectParameters.m
HenkMutsaerts: Yes, I agree. This makes sense

Functions/QualityControl/xASL_qc_CollectParameters.m
HenkMutsaerts: Are you sure? In line 57-67 -> xASL_qc_CollectQC_ASL we load the new fields x.Output.ASL.(SessionID), but we also throw away the old field x.Output.ASL content (you wrote a warning there).

This assumes that we already loaded the QC_Path previously, which is true, we load the QC json at the beginning of the ASL module, then populate it within each wrapper/submodule, and at the end we write it again. This QC collection part here is a temporary solution, in fact all QC parameters should be written to memory x.Output.ASL.(SessionID) in the respective wrappers/submodules that create them. (just explaining why we should load QC json at the beginning of the ASL module and save at the end of the ASL module)

So by definition, we have already loaded the QC_Path, and here we only save it, for which we ensure to first delete it. So you have to revert this edit.
Makes sense?

Functions/QualityControl/xASL_qc_CollectParameters.m
HenkMutsaerts: So we keep the content of `x.Output.ASL` from previous runs and only add the new ASL_1 ASL_2 ASL_n fields? I thought that you just added a warning saying that the original content is removed :)

Functions/QualityControl/xASL_qc_CollectParameters.m
HenkMutsaerts: I see

Functions/QualityControl/xASL_qc_CollectParameters.m
jan-petr: Done in 0b7a798c

Functions/QualityControl/xASL_qc_CollectParameters.m
jan-petr: Done in 0b7a798c

Functions/QualityControl/xASL_qc_CollectParameters.m
jan-petr: Done in 0b7a798c

Functions/QualityControl/xASL_qc_CollectParameters.m
jan-petr: Done in 0b7a798c

Functions/QualityControl/xASL_qc_CollectParameters.m
jan-petr: Not really making sense. The problem is, if we keep everything in the memory, we will loose this memory after each session. And if we have processed sessions 1 and 2 previously. And now added session 3, we will need to load this from the path again.

Or your comment makes sense, but you don't realize that with each module run for a session, we clean x-struct, so that's why reloading is needed.

Or do I really get that wrong? I have been testing it and it seemed to be necessary to have this loading here - by a practical test.

Perhaps you can try to test it yourself and have some pre-processed and some new ASL sessions and see if all sessions would have been present. It seems to me they wouldn't be if I revert this edit...

Functions/QualityControl/xASL_qc_CollectParameters.m
jan-petr: Yes. `x.Output.ASL.ASL_1`, and `x.Output.ASL.ASL_2` are overwritten. All other tags within `x.Output.ASL` that don't have the format `x.Output.ASL.ASL_N` are removed - that was this warning about.

Is that fine?

Functions/QualityControl/xASL_qc_CollectParameters.m
HenkMutsaerts: We said that we would put all sessions to 1 right, irregardless of the scantype? So I would remove the scantype checking here, and just give a general warning. Or perhaps only checking for non-structural.

Functions/QualityControl/xASL_qc_CollectParameters.m
HenkMutsaerts: No, they should all have sessions except for structural here.

Functions/QualityControl/xASL_qc_CollectParameters.m
HenkMutsaerts: OK let me test this then. Removing old sessions, cleaning up, etc, should be done at the beginning of xASL_module_ASL, then it gets populated throughout all submodules, and at the end we save it.

But you're right that if ExploreASL crashes in the meantime, it should have everything saved to the disk. Not sure how we deal with this now; at the start all QC was indeed loaded and saved only in the last QC submodule. But we now do in fact output e.g. some Tanimoto parameters in registration, not sure how/if these get saved on disk properly.

HenkMutsaerts (changes requested):
* Everything other than structural should get 1 default session
* We have 1 pending discussion on the loading/saving on memory/disk

Functions/QualityControl/xASL_qc_CollectParameters.m
jan-petr: Yes. That is what I did. ASL and func have now sessions. Dwi doesn't have "sessions" in QC as dwi doesn't have a dedicated QC :)

Functions/QualityControl/xASL_qc_CollectParameters.m
jan-petr: I understand. Currently, we NEED to load and save QC from other session in each xASL_module_ASL. But you are right that we might not need to do that at the end, but that it could be done at the start of the module.

If the real issue is about loading at start and saving at the end. Instead of loading in the middle and saving at the end, then we can most likely move it without any consequences.

If you confirm that this is the problem, then I can investigate and change the code. Most likely, this is going to work, so no reason not to do that :)

Functions/QualityControl/xASL_qc_CollectParameters.m
jan-petr: Yes, that's more or less the case. 
1) We put all sessions to 1 - that is in the code already (in case that the session is missing). So that should be fine.
2) For structural, we don't really use session, so the checking is indeed not needed.

So I propose that I give a general warning, remove IF and print the warning using the `ScanType` variable. And we check if that is structural, we don't print the warning. Agree?

HenkMutsaerts: Yes exactly, if not structural, then we give this warning for scanType

Functions/QualityControl/xASL_qc_CollectParameters.m
HenkMutsaerts: True but func should also get a session (now it doesn't get it).

Functions/QualityControl/xASL_qc_CollectParameters.m
HenkMutsaerts: Yes that is what we need to check. I can also check it.

HenkMutsaerts (changes requested):

Functions/QualityControl/xASL_qc_CollectParameters.m
jan-petr: It definitively does. In the same way as ASL. I've coded that already in https://github.com/ExploreASL/ExploreASL/commit/0b7a798cc4352f2e26788e1e23f91521fcf3c66f. Or I miss something?

Functions/QualityControl/xASL_qc_CollectParameters.m
jan-petr: Agree. Done in 4f94b2c7

Functions/QualityControl/xASL_qc_CollectParameters.m
jan-petr: OK. All clear. I will check and implement it that way.

jan-petr (commented):
Some answers...

Functions/QualityControl/xASL_qc_CollectParameters.m
jan-petr: What to do
- [X] xASL_module_ASL before part C - add a function to clean old QC - move the code from the other function. 4301cee4  
- [X] xASL_module_ASL after part C - properly read the QC from old modules and add to x-struct. Don't do that in the collect function 2c8076ba
- [X] xASL_module_ASL part C - only remove the current session. 33a9de60
- [X] xASL_adm_CleanUpBeforeRerun - only clean QC of the current session. 33a9de60
- [x] Output_im should also have subsessions for ASL

jan-petr: All done. We just need to decide if we move Output_im for ASL to subsessions. We probably should.

Functions/QualityControl/xASL_qc_CollectParameters.m
HenkMutsaerts: Haha in this very line where we are discussing it (line 103), both dwi & func get 

x.Output.(ScanType).Version_FSL

instead of 

x.Output.ASL.(x.SESSIONS{iSession}).Version_FSL

Modules/xASL_module_ASL.m
```diff
@@ -101,14 +102,40 @@
...
 
-	if isfield(x,'Output') && isfield(x.Output,'ASL')
-		x.Output = rmfield(x.Output,'ASL');
+	if isfield(x, 'Output') && isfield(x.Output, 'ASL') && isfield(x.Output.ASL, x.SESSION)
+		x.Output = rmfield(x.Output.ASL, x.SESSION);
 	end
 	
 	% And saved the cleaned up version
 	xASL_adm_SaveX(x);
+
```
HenkMutsaerts: Nice, but this should go to `xASL_adm_CleanUpBeforeRerun`, which is a separate function handling what needs to be removed for a rerun.

Functions/QualityControl/xASL_qc_RemoveOutdatedQC.m
```diff
@@ -0,0 +1,42 @@
+function [x] = xxASL_qc_RemoveOutdatedQC(x)
```
HenkMutsaerts: nice!

HenkMutsaerts (changes requested):
Looks nice, though the stuff at the beginning should all go under `xASL_adm_CleanUpBeforeRerun`, that is a general function that checks what needs to be removed before a rerun, so we should keep all there nicely collected.

Another thing: there was a saving and reloading at the end that I disagreed with because loading should start at the beginning and saving at the end, did we end up fixing this? Or did we decide to move this to a new issue?

Modules/xASL_module_ASL.m
jan-petr: That's how you wrote that previously. And I didn't want to touch that. I just made the necessary modifications. 

I can of course change that, but then please say here explicitly which changes should I do.

Functions/QualityControl/xASL_qc_RemoveOutdatedQC.m
jan-petr: Thanks.

Modules/xASL_module_ASL.m
HenkMutsaerts: It should not be too hard. The way I wrote xASL_adm_CleanUpBeforeRerun, is that it iterates over modules, such that you can call it for each module/session/subject separately. And if you clean a whole study (e.g., by calling it outside of xASL_Iteration) it will just loop/run over all those.

So the cleanup you want to add for a session, should go there under the ASL module/session part. It should be easy to follow that.

Modules/xASL_module_ASL.m
```diff
@@ -101,14 +102,40 @@
...
 	xASL_adm_SaveX(x);
+	
+	% If we rerun completely, then we do not the reload QC for that session, otherwise we load it from QC.json if present.
+	bCompleteRerun = true;
+else
+	x = xASL_adm_LoadX(x, [], true); % assume x.mat is newer than x
+end
+
+% Load QC.json and save it to x-struct
+QC_Path = fullfile(x.D.ROOT, x.SUBJECT, ['QC_collection_' x.SUBJECT '.json']);
```
HenkMutsaerts: What you write here is needed every time you load the QC_collection JSON right, not when you loadX or anything else?

Then I would be in favor to move this to a separate function that deals with loading QC_Collection*.json (if it needs to be the same every time, irrespective of where we run it).

If this is something that we only need at the start of the ASL module, then we can leave it here.

BTW: why can't this loop go to xASL_qc_RemoveOutdatedQC?

HenkMutsaerts (changes requested):
So now I clicked on "Files changed (10)" and did a new review. Is there anything I missed? I thought this was the way to do reviews...

Modules/xASL_module_ASL.m
jan-petr: Yes. But your main argument was the QC json is loaded once at the start of the module and once at the end of the module. So that's why I didn't put any loading+checking in the middle. 
We can of course do loading+saving+checking also in the middle of a module, but I though you explicitly didn't want that.

Other than that, I moved the loop to the subfunction 4a9deb4f

Functions/QualityControl/xASL_qc_CollectParameters.m
jan-petr: Done in 43a10227

Modules/xASL_module_ASL.m
jan-petr: Not so easy because that function doestn' have a x-struct input or output. But I have moved it to the other QC cleaning function in 4a9deb4f

Functions/QualityControl/xASL_qc_CollectParameters.m
HenkMutsaerts: OK the only thing we have to be careful about, is that we hardcode func, ASL, dwi, whereas here we say for everything else than structural. This probably goes well, but I thought about not hardcoding, but everywhere saying "if is not structural"

Functions/QualityControl/xASL_qc_CollectParameters.m
HenkMutsaerts: I can't see the warning anymore quickly, but I would make this more clear perhaps (at least that is what my question above suggests, that it could be more clear :)

Modules/xASL_module_ASL.m
```diff
@@ -94,47 +94,14 @@
...
-			end
-		end
-	end
 end
 
+x = xASL_adm_LoadX(x, [], true); % assume x.mat is newer than x
+
 % Check and remove all outdated QC fields that are not used anymore
-x = xASL_qc_RemoveOutdatedQC(x);
+x = xASL_qc_CleanOldQC(x, 'ASL', bCompleteRerun);
```
HenkMutsaerts: Nice, but it still seems that this should go to the original cleanup function, see section 9.
Are you able to put it there?

Functions/QualityControl/xASL_qc_CleanOldQC.m
```diff
@@ -0,0 +1,89 @@
...
+% -----------------------------------------------------------------------------------------------------------------------------------------------------
+% DESCRIPTION: This function removes the old and outdated QC parameters. Specifically parameters for ASL and func which are now given in a subfield by session.
+%              It also loads further sessions from the disc and if necessary, it cleans the current session completely.
+%
+% EXAMPLE: x = xASL_qc_RemoveOutdatedQC(x, 'ASL');
+% __________________________________
+% Copyright (C) 2015-2023 ExploreASL	
+
+if nargin < 2
+	error('Two parameters are requied');
```
HenkMutsaerts: minor spelling error

HenkMutsaerts (changes requested):

Functions/QualityControl/xASL_qc_CollectParameters.m
HenkMutsaerts: yes, and the current new "old cleanup" function should be incorporated in xASL_adm_CleanUpBeforeRerun section 9

Modules/xASL_module_ASL.m
jan-petr: I am not sure if we should do that. 
Because  the CleanUp function is only run when re-runing the module.
This CleanQC function is executed every time to read and clean the QC.json on the disk.

So I could add it to the original cleanup function, but I would have to do certain hacks you probably wouldn't like.

So that's why I would prefer to keep it here as it is. But we can of course rename it.

Functions/QualityControl/xASL_qc_CleanOldQC.m
jan-petr: Good catch. Fixed and added to the next commit in the queue.

Functions/QualityControl/xASL_qc_CollectParameters.m
jan-petr: Output_im fixed in 0ff0f041
About integrating "old cleanup", see my comment below.

Functions/QualityControl/xASL_qc_CleanOldQC.m
```diff
@@ -0,0 +1,94 @@
...
+if nargin < 2
+	error('Two parameters are required');
+end
+
+if nargin < 3 || isempty(bRemoveCurrentSession)
+	bRemoveCurrentSession = false;
+end
+
+% All possible scantype lists
+ScanTypeList = {'func', 'ASL', 'dwi'};
```
HenkMutsaerts, line 33: We need a warning here if the second input parameter doesn't agree with one of these here.

Modules/xASL_module_ASL.m
HenkMutsaerts: I understand it now!

Functions/QualityControl/xASL_qc_CleanOldQC.m
```diff
@@ -0,0 +1,94 @@
...
+
+% Load QC.json and save it to x-struct
+QC_Path = fullfile(x.D.ROOT, x.SUBJECT, ['QC_collection_' x.SUBJECT '.json']);
```
HenkMutsaerts, line 38: Can you add headers to the code?
Something like:

1. Manage potentially old x.Output without sessions (here)
2. Fully remove current session (line 62)
3. Remove old x.Output without sessions (line 73)

but you might want to rename (1) because this is something different from (3) right?
And when we have these headers, it is also easier for me to check if the order is useful (e.g. perhaps if you only remove the current session, you don't have to run through the rest of the code)

Functions/QualityControl/xASL_qc_CleanOldQC.m
```diff
@@ -0,0 +1,94 @@
...
+		currentScanType = ScanTypeList{iScanType};
+		if isfield(oldOutput, currentScanType)
+			% Load all subfields for that scantype
+			listFields = fieldnames(oldOutput.(currentScanType));
+			for iField = 1:length(listFields)
+				% Copy only the subfields in ScanType_N
+				if ~isempty(regexp(listFields{iField}, [currentScanType '_\d+'], 'once'))
+					% But skip the current session unless bCompleteRun is false
+					if ~strcmp(listFields{iField}, x.SESSION) || ~bRemoveCurrentSession
+						x.Output.(currentScanType).(listFields{iField}) = oldOutput.(currentScanType).(listFields{iField});
```
HenkMutsaerts: Can't we do here exactly the same for x.Output_im as x.Output, because those are designed to have similar structure?

HenkMutsaerts (changes requested):
Few questions

Modules/xASL_module_ASL.m
jan-petr: Perfect! Yes, it also took me time to get into this. Can we thus resolve?

Functions/QualityControl/xASL_qc_CleanOldQC.m
jan-petr: This part of the code is about loading things from the QC.json, so we can't do that for images, these have to be kept in x-struct.

Functions/QualityControl/xASL_qc_CleanOldQC.m
jan-petr: Good point. The comments and explanation were not really clear. This is now improved in c523326a

Functions/QualityControl/xASL_qc_CleanOldQC.m
jan-petr: Sorry. That's just a confusion. We do two things in this function.
1) We load all in QC.json to x-struct so that we can save QC.json later on while having all the QC for other sessions inside and not only for the current session.
2) We clean outdated tags of format x.ASL.parameter instead of x.ASL.ASL_1.parameter.

Point 1 needs to be done for all possible scantypes in QC.json. 
Point 2 was done only for the current scantype in input. But we can actually do it also for all scan types :)

So I'm removing the scantype from the input and adding more info and comments and redoing slightly the code.

Fixed in c523326a

HenkMutsaerts: Nice!

HenkMutsaerts (changes requested):
Looks great for output/QC.json. But don't we want to do the same for x.output_im? I forgot your reason to do this, but this would make sense. Currently, we use the following general function for this: `xASL_vis_AddIM2QC` within the structural and ASL modules. So, you would only have to change this function accordingly.

Functions/QualityControl/xASL_qc_CleanOldQC.m
HenkMutsaerts: Looks great for output/QC.json. But don't we want to do the same for x.output_im? I forgot your reason to do this, but this would make sense. Currently, we use the following general function for this: `xASL_vis_AddIM2QC` within the structural and ASL modules. So, you would only have to change this function accordingly.

jan-petr: Not sure if I follow.
We can't do the same for output as for output_im. As output is saved in QC.json and we can't save images to QC.json. I'm happy to change things and I'm not disagreeing, I just don't understand what are you asking me now to do...

HenkMutsaerts: Both output and output_im are stored likewise in x.mat, with identical struct naming conventions. The difference is that x.output also goes to the QC.json. 

You changed the struct naming convention for x.output. I consider that we change the naming convention similarly for x.output_im.

HenkMutsaerts (approved):

PR #1518, opened 2023-10-20
Closes #1283 xASL_qc_SortBySpatialCoV: improve warnings

### LINKED ISSUE #1283: xASL_qc_SortBySpatialCoV.m does not sort the JPGs
### Description

In the population module, the function `xASL_qc_SortBySpatialCoV.m` does not select the proper index value to select the sCoV value for sorting the JPGS. The issue at line 85. Why is this the line `Index = find(cellfun(2(x) strcmp(x,NameSubjSess), subjectSessionList)) used; whereas I think we can use the iSubjSess created previously in the code.

image

This code takes the subjects and sessions from x.SUBJECTS and x.SESSIONS, which are lists throughout ExploreASL. Because we cannot know for sure that the sCoV values in the stats .tsv files are complete, and likewise for the ASL jpg images, we have to check if the subjectsession name exists in both cases, and only then we proceed in this iteration.
So, we have to skip subjects/sessions that for some reason didn't get a sCoV or didn't have their ASL jpg saved.

With which population did you have this problem? I would check why some ASL scans apparently weren't processed correctly, otherwise they would have both jpgs and sCoV values.

### Tasks

- [x] xASL_qc_SortBySpatialCoV: Improve warnings, to clarify why not all subject/sessions are sorted

### Release notes

Improve warnings when sortins scans by spatial CoV.

HenkMutsaerts:
### Linked issue

Closes #1283

### DIFF
--- a/Modules/Module_Processing/Module_Population/xASL_qc_SortBySpatialCoV.m
+++ b/Modules/Module_Processing/Module_Population/xASL_qc_SortBySpatialCoV.m
@@ -97,25 +97,32 @@ for iSubject=1:x.dataset.nSubjects
         % find current JPG
         JPGList = xASL_adm_GetFileList(x.D.ASLCheckDir, ['(?i)^Tra_qCBF_' NameSubjSess '.*'],'List', [0 Inf]);
         % find current sCoV
-        Index = find(cellfun(@(x) strcmp(x,NameSubjSess), subjectSessionList));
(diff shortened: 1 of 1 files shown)


jan-petr (approved):
Looks OK to me.

MDijsselhof (approved):
Looks good to me

PR #1532, opened 2023-10-29
Closes #1531 xASL_adm_DefineASLSequence: minor bugfixes & improvements

### LINKED ISSUE #1531: if readout == 2D, assume 2DEPI
### Description

if readout == 2D, assume 2DEPI. We needed this for running BIDS example asl004.
And some other improvements.

### Tasks

- [x] xASL_adm_DefineASLSequence: bugfixes and improvements
- [x] Add BIDS example asl004 as flavor: flavor-#1531_Siemens_2DEPI_multi_PLD

### Release notes

Improve defaulting of sequence type based on vendor and readout.

HenkMutsaerts:
### Linked issue

Closes #1531

### DIFF
--- a/Functions/Administration/xASL_adm_DefineASLSequence.m
+++ b/Functions/Administration/xASL_adm_DefineASLSequence.m
@@ -34,10 +34,6 @@ if isfield(x.Q,'Sequence')
         fprintf('%s\n', 'Trying to fix this');
         x.Q = rmfield(x.Q, 'Sequence');
 
-        if strcmpi(x.Q.readoutDim,'2D')
-            x.Q.Sequence = '2D_EPI'; % assume that 2D is 2D EPI, irrespective of Vendor
-            fprintf('%s\n', '2D readout detected, assuming 2D EPI');
-        end
     end
 end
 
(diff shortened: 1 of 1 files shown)


Functions/Administration/xASL_adm_DefineASLSequence.m
```diff
@@ -27,6 +27,21 @@
     warning('x.Q.readoutDim parameter missing');
 end
 
+% Check for illegal sequence definitions
+if isfield(x.Q,'Sequence')
+    if isempty(regexpi(x.Q.Sequence, '(2d_epi|3d_grase|3d_spiral)'))
+        warning('Unknown ASL sequence');
+
+        if strcmpi(x.Q.readoutDim,'2D')
```
jan-petr: You can improve this. If the readout is 3D then Philips and Siemens have 3D GRASE and GE will have 3D spiral ;)
But please also build a  isfield check for vendor.

jan-petr (changes requested):
One comment.

Functions/Administration/xASL_adm_DefineASLSequence.m
HenkMutsaerts: That is already there below, where we indeed also check for the vendor.
Here, we already assume 2D EPI even if there is no vendor specified.

But you're right that we should still do this check even if an unknown ASL sequence was detected.
So now explicitly deleting the `Sequence` field if an illegal field was detected, and trying to reset this field below.

jan-petr: Good point to delete the unknown sequence. And you are right that we assign the new sequence below. But actually, now we don't need to assign 2D_EPI here as this is done below. So we should remove 2D_EPI from here and only keep it below.

jan-petr (changes requested):
Better now - still one modif needed.

Functions/Administration/xASL_adm_DefineASLSequence.m
HenkMutsaerts: Not true, because we run the code below only if we know the vendor. Currenty, even if we don't know the vendor, we can assign 2D EPI. I will now delete this below so there is no confusion.

jan-petr: Yes - I know, it needed an extra edit ;)

jan-petr (approved):
OK

PR #1555, opened 2023-12-06
Closes #1547 bids2 legacy roo tfolder

### LINKED ISSUE #1547: Parallelize BIDS2Legacy: x.D.ROOT usage
### Description

This request started as wanting to parallelize BIDS2Legacy because it conflicted and took too long for processing (a lot of overhead.

To do this properly, this changed into changing the data loading, as follows:

* Import has nothing to do with BIDS2Legacy, or data loading
* At the start of processing, we load data, either from rawdata (BIDS2Legacy) or from derivatives (pre-existing)
* Then we define parallelization
* Then we define all other metadata accordingly

### Tasks
- [x] `x.D.ROOT` -> replaceD by `x.dir.RawData` or `x.dir.xASLDerivatives`
- [X] xASL_init_BIDS2Legacy & xASL_module_BIDS2Legacy use legacy format (subject_visit)
- [X] x.opts.bSkipBIDS2Legacy fully removed
- [x] Bugfix potential participants.tsv creation in parallel
- [x] MANAGE participants.tsv
xASL_init_DefineStudyData -> xASL_init_LoadMetadata -> xASL_bids_LoadParticipantsTSV
- [x] MANAGE dataset_description.tsv — when loading data
       These BIDS files should be in rawdata, and copied/merged to xASLDerivatives if there is not one already
       If this file exists in the datasetroot, issue a warning that we ignore this one
xASL_init_BIDS2Legacy -> checks availability dataset_description.json in rawdata
xASL_bids_CompleteBIDS2Legacy -> copies dataset_description from rawdata to derivatives
- [x] MANAGE DATAPAR — when starting processing (not needed for loading data, this is a processing settings file, but 
       can do before loading as well)
       This ExploreASL-specific settings file should be in datasetroot and copied to xASLDerivatives if there is not one 
       already
       If this file exists in rawdata, issue a warning  that we ignore this one
       xASL_init_BIDS2Legacy section 3 -> 
       xASL_bids_CompleteBIDS2Legacy 30 -> create dataPar if not existed
       xASL_init_DataLoading 372 -> read dataPar (and deal with backward compatibility)
- [x] MANAGE subjectRegexp
- [x] Major revamp initialization structure
- [x] Update headers

#### Fixes done for testing Jan
- [x] Revamp the order
Changes:
* Move xASL_init_LoadDataPar & xASL_init_RemoveLockDirs from AFTER subject loading to BEFORE subject loading
* Move xASL_init_PrintUserFeedback from AFTER subject loading to BEFORE subject loading
* Exchange names between xASL_init_DataLoading (BEFORE subject loading) & xASL_init_Process (AFTER subject loading)

Old order:

xASL_init_Process

    xASL_init_SubjectList
    xASL_init_Parallelization
    xASL_module_BIDS2Legacy
    xASL_init_CreateParticipantsTSV
    xASL_init_Session_TimePoint_Lists

xASL_init_DataLoading
>xASL_init_LoadDataPar
> xASL_init_Loaddataset_description
> xASL_init_DefineDataParDependentSettings
> xASL_init_DefineStudyStats
> xASL_init_RemoveLockDirs
> xASL_init_PrintCheckSettings

xASL_init_PrintUserFeedback

New order:

xASL_init_DataLoading

>xASL_init_LoadDataPar
> xASL_init_RemoveLockDirs

xASL_init_PrintUserFeedback

    xASL_init_SubjectList
    xASL_init_Parallelization
    xASL_module_BIDS2Legacy
    xASL_init_CreateParticipantsTSV
    xASL_init_Session_TimePoint_Lists

xASL_init_Process
> xASL_init_Loaddataset_description
> xASL_init_DefineDataParDependentSettings
> xASL_init_DefineStudyStats
> xASL_init_PrintCheckSettings

       

### Testing fixes non-specific to this issue
- [x] xASL_im_CreateGroupAnalysisMask: bugfix loading paths instead of cells
- [x] xASL_wrp_CreatePopulationTemplates: reduce verbosity for non-existing output
- [x] xASL_qc_CompareTemplate: bugfix too slow pinv
- [x] xASL_io_CheckDeprecatedFieldsX: only check existing fields

#### PM
HENK: I removed 2 obsolete tasks here. After you read this, you can remove this line :)
- [x] Move ExploreASL/Functions/Import -> ExploreASL/Modules/Module_Import JAN: Can be done now.
       HENK: done
- [x] Separated ExploreASL/Functions/BIDS into 
        ExploreASL/Modules/Module_Import/NII2BIDS 
        ExploreASL/Modules/Module_Processing/Module_BIDS2Legacy
        Testing/BIDS
### MOVED TO #1275 
- [x] What should happen with /Functions/BIDS/xASL_bids_CheckDatasetDescription.m? This file is currently not used.
- [x] Also, in ExploreASL_Initialize -> xASL_init_AddDirsOfxASL we now initialize all folders used in import, processing, testing. I think it would be nicer to move the import folders to the import initialization, the processing folders to the processing initialization, and the testing folders to the testing, don't you agree? Then we are sure that we are never mixing up code. But we can't really do this yet, because there are a few things we use for both import and processing. But we could already do this for testing right? The users don't need to see the testing folder?

- [x] Adapt documentation: remove old and add new parameters JAN: Are there any new/old parameters? Which ones?
HENK:
x.opts.bSkipBIDS2Legacy is removed
x.dataset.TimePoint*: some of these fields might be removed but these are internal fields only I guess?

- [x] BIDS2Legacy should use the rawdata subjects (ignore the derivatives subjects). JAN: Agree, and it is possible to run this with purely rawdata input without derivatives and sourcestructure, so that should work.

- [x] If we define subjects (`x.opts.subjectFolder`) from `x.dir.RawData`, then the problem is that our legacy folder doesn't have the BIDS names yet. So we should fix that (i.e., adding the _1 suffix).
image

### How to test
- [x] Flavors
- [x] TestDataSets
- [x] Update unit tests & test

### Release notes

Report with https://github.com/ExploreASL/ExploreASL/issues/1275. Allow parallelization of BIDS2Legacy.

HenkMutsaerts:
### Linked issue

Closes #1547


ExploreASL_Initialize.m
```diff
@@ -342,14 +342,6 @@
             x.opts.bLoadData = false;
         end
         
-        % We need to check if import/defacing and processing were run separately
-        if x.opts.bImportData && ~x.opts.bProcessData
-            x.opts.bSkipBIDS2Legacy = true;
-            x.opts.bLoadData = false;
```
jan-petr, line 348: On Line 817, we still need bLoadData. But that parameter has been removed here. So it crashes on Line 817.

jan-petr: And it's not only there - bLoadData is used on many places before being initialized. It also crashes then in ExploreASL.m line 96.
I do not know why you have removed this from here, but you have to add it then elsewhere.

I talk about bLoadData only, bSkipBIDS2Legacy has been removed everywhere...

jan-petr: Both these errors do not happen when I ran ExploreASL(path, 0, 1) but when running ExploreASL(path, 1, 0).

Modules/Module_Processing/Module_Population/xASL_adm_GetPopulationSessions.m
```diff
@@ -1,9 +1,10 @@
-function [nSessions, bSessionsMissing, SESSIONS] = xASL_adm_GetPopulationSessions(x)
+function [nSessions, bSessionsMissing, SESSIONS] = xASL_adm_GetPopulationSessions(x, bVerbose)
 % xASL_adm_GetPopulationSessions(x) obtain number of Sessions by determining amount of input files present in the Population folder
 % FORMAT: [nSessions, bSessionsMissing] = xASL_adm_GetPopulationSessions(x)
 %
 % INPUT:
 %   x                   - struct containing statistical pipeline environment parameters (REQUIRED)
+%   bVerbose            - boolean for verbosity (OPTIONAL, DEFAULT = true)
```
jan-petr, line 7: add bVerbose on line 3 as [,bVerbose]

jan-petr (changes requested):
Minor bugs with large consequences :)

Functions/Initialization/xASL_init_Iteration.m
```diff
@@ -105,17 +94,37 @@
     end
     
     % SUBJECT, SUBJECTDIR & LockDir
-    if ~isempty(regexpi(ModName,'(Import|BIDS2Legacy|Structural|ASL|func|LongReg|dwi)', 'once'))
+    if ~isempty(regexpi(ModName,'Import', 'once'))
         dbSettings.sets.SUBJECT = SelectedSubjects; % x.SUBJECTS
         dbSettings.x.dir.SUBJECTDIR = fullfile('<ROOT>', '<SUBJECT>');
         dbSettings.x.dir.LockDir = fullfile(dbSettings.x.dir.LockDir, '<SUBJECT>');
+    elseif ~isempty(regexpi(ModName,'BIDS2Legacy', 'once'))
```
jan-petr: Import and BIDS2Legacy options don't differ - should we merge those?

Modules/Module_Processing/Module_Population/xASL_im_CreateGroupAnalysisMask.m
```diff
@@ -72,6 +72,12 @@
     if isempty(pathTemplate)
         warning(['Skipping because of missing template: ' nameTemplate]);
         bSkipStandard = true;
+    elseif numel(pathTemplate)>1
+        warning(['Multiple templates found for ' nameTemplate x.S.TemplateNumberName '_bs-mean.nii']);
+        % This situation should not be able to occur
```
jan-petr: Please revise the comment "should not occur" ;)

Modules/Module_Processing/Module_Population/xASL_im_CreateGroupAnalysisMask.m
```diff
@@ -72,6 +72,12 @@
     if isempty(pathTemplate)
         warning(['Skipping because of missing template: ' nameTemplate]);
         bSkipStandard = true;
+    elseif numel(pathTemplate)>1
+        warning(['Multiple templates found for ' nameTemplate x.S.TemplateNumberName '_bs-mean.nii']);
+        % This situation should not be able to occur
+        eval([nameTemplate ' = ' nameTemplate '{1};']);
```
jan-petr: We shouldn't use eval. Just write nameTemplate = nameTemplate{1}.

1. It's confusing and has no advantage
2. We can't compile statements with `eval`. So no compiled version, no docker.

Modules/Module_Processing/xASL_init_DefineStudyStats.m
```diff
@@ -0,0 +1,146 @@
...
+% __________________________________
+% Copyright (c) 2015-2023 ExploreASL
+
+
+% ------------------------------------------------------------------------------------------------
+%% 1. Add sessions as statistical variable
+
+% Predefine SETS to avoid empty SETS & import predefined session settings as set settings
+x.S.SetsName{1} = 'session';
+x.S.SetsOptions = '';
```
jan-petr: You initialize as empty char array but later treat this as a cell. Best to just use cell(0), right?

Same with SetsOptions

Modules/xASL_module_BIDS2Legacy.m
```diff
@@ -77,113 +77,70 @@
...
-        % we use this below to see if the legacy subjectname gets _1 as visit suffix or not
-        [~, studyName] = fileparts(x.dir.DatasetRoot);
 
-        if bVerbose
-            fprintf('Converting from BIDS to Legacy: %s   \n', studyName);
-        end
 
         %% 2. Define Subject
-        indicesCurrentSubject = find(cellfun( @(y) strcmp(y, x.SUBJECT), {x.modules.bids2legacy.BIDS.subjects.name}));
+        iSubjSess = find(strcmp(x.SUBJECTS, x.SUBJECT));
```
jan-petr, line 70: I would strictly try to avoid the use of iSubjSess as a list of sessions. That is - I would always make length of SubjSess equal to nSessions*nSubject without skipping empty sessions. Not sure if you do that or not, but it's not really clear from the code.

That's just a comment, hopefully, the code is correct already.

Modules/Module_Processing/xASL_init_Session_TimePoint_Lists.m
```diff
@@ -0,0 +1,137 @@
...
+    x.SESSIONS = [];
+    SessionPathList = xASL_adm_GetFileList(x.dir.xASLDerivatives, '^(ASL|func)_\d+$', 'FPListRec', [0 Inf],1);
+    for iSess=1:length(SessionPathList)
+        [~, x.SESSIONS{end+1}]  = fileparts(SessionPathList{iSess});
+    end
+
+    if isempty(x.SESSIONS)
+        fprintf('%s\n', 'No sessions found, defaulting to a single ASL_1 session');
+        x.SESSIONS{1} = 'ASL_1'; % default session
+	else
```
jan-petr: check identation

Modules/Module_Processing/xASL_init_Session_TimePoint_Lists.m
```diff
@@ -0,0 +1,137 @@
...
+[~, TimePoint] = xASL_init_LongitudinalRegistration(x);
+
+for iT=unique(TimePoint)'
+    x.dataset.TimePointTotalSubjects{iT} = '';
+end
+
+for iSubj=1:x.dataset.nTotalSubjects
+    iSess=1;
+    iSubjSess = (iSubj-1)*x.dataset.nSessions + iSess;
+    x.dataset.TimePointTotalSubjects{TimePoint(iSubjSess)}{end+1} = x.dataset.TotalSubjects{iSubj};
```
jan-petr, line 80: I'm confused if TimePointTotalSubjects is a 1D or 2D cell array. Here it is 2D, below and above 1D.

Modules/Module_Processing/xASL_init_Session_TimePoint_Lists.m
```diff
@@ -0,0 +1,137 @@
...
+end
+
+
+% ------------------------------------------------------------------------------------------------
+%% 3. Create list of baseline & follow-up subjects (i.e. after exclusion)
+for iCell=1:length(x.dataset.TimePointTotalSubjects)
+    x.dataset.TimePointSubjects{iCell} = '';
+end
+    
+for iS=1:x.dataset.nSubjects
```
jan-petr, line 100: WHy do we have nSubjects here and nTotalSubjects above? Does it matter?

Modules/Module_Processing/xASL_init_Session_TimePoint_Lists.m
```diff
@@ -0,0 +1,137 @@
...
+    x.dataset.TimePointExcluded{iT} = '';
+end
+
+for iE=1:x.dataset.nExcluded
+    FoundE = 0;
+    iT = 0;
+    while ~FoundE % excluded subject not found in previous TimePoint
+        iT=iT+1;  % go to next TimePoint
+        
+        iS  = find(strcmp(x.dataset.TimePointTotalSubjects{iT}, x.dataset.ExcludedSubjects{iE}));
```
jan-petr: what if it doesn't ever find anything - wouldn't this while loop run endlessly?

jan-petr (changes requested):
A few more minor comments.

ExploreASL_Initialize.m
HenkMutsaerts: bLoadData is a parameter needed for Processing, not Import, which is why I cleaned this. But I see now that it needs proper initialization as false indeed. Fixed in 3572204f7c96a0ba01fba82ca0ac2faac5f3a635

Functions/Initialization/xASL_init_Iteration.m
HenkMutsaerts: I would opt to keep them separate, but I agree that this just simplifies reading here. Fixed in 9f57c9fb9a36450260a83fbd59cd9bfb9150cf49

Modules/Module_Processing/Module_Population/xASL_adm_GetPopulationSessions.m
HenkMutsaerts: Fixed in b911e3fbc806998efd455ba2a73018be999120bb

Modules/Module_Processing/xASL_init_Process.m
```diff
@@ -20,71 +28,76 @@
...
+    if ~xASL_exist(x.dir.RawData, 'dir')
+        x.opts.bReadRawdata = false;
+        fprintf('%s\n', ['rawdata folder missing: ' x.dir.RawData]);
+        
+        if ~xASL_exist(x.dir.xASLDerivatives)
+            fprintf('%s\n', ['derivatives ExploreASL folder missing: ' x.dir.xASLDerivatives]);
+
+            error('Both rawdata and derivatives folders missing, cannot continue processing');
+
+            x.opts.bLoadData = false;
```
jan-petr: This line can't be reached - there's an "error" function called above.

Modules/Module_Processing/xASL_init_Process.m
```diff
@@ -20,71 +28,76 @@
...
-    % EXAMPLE:        n/a
-    % __________________________________
-    % Copyright 2015-2022 ExploreASL
+    %% 5. Check basic directories
+    if isempty(x.dir.DatasetRoot)
+	    error('x.dir.DatasetRoot is a required parameter.');
+    end
     
+    % Verify that the rawdata subfolder exists
+    if ~exist(fullfile(x.dir.DatasetRoot,'rawdata'), 'dir')
```
jan-petr: Above - we have checked that we don't have rawdata folder. But we still try to check this, print a warning and then exit. 
We should skip this check completely when x.opts.ReadRawData is false.

Modules/Module_Processing/xASL_init_SubjectList.m
```diff
@@ -0,0 +1,181 @@
...
+
+% ------------------------------------------------------------------------------------------------
+%% 2. Advanced parameter to enforce a subject list, without querying folder names
+if isfield(x.dataset,'ForceInclusionList')
+    % This is an option if you want to select subjects yourself,
+    % instead of using all the subjects that comply with the regular expression
+    x.dataset.TotalSubjects = x.dataset.ForceInclusionList(:);
+    warning('Using custom list of subjects, on your own risk');
+
+    if ~isfield(x.dataset, 'subjectRegexp')
```
jan-petr, line 49: It's really weird that even-though I have dataPar.json in derivatives/ExploreASL directory, and it contains subjectRegexp, it reaches the code here without loading it.

Note that BIDS2Legacy copies dataPar.json from root to derivatives. But processing module is supposed to read derivatives/ExploreASL/dataPar.json

Modules/Module_Processing/xASL_init_Loaddataset_description.m
```diff
@@ -0,0 +1,102 @@
...
+% REFERENCES:  n/a
+% __________________________________
+% Copyright (c) 2015-2024 ExploreASL
+
+
+%% 1. Admin checks
+if ~isfield(x.dir,'dataset_description')
+	error('Missing dataset_description field...');
+end
+if isempty(x.dir.dataset_description)
```
jan-petr: If you call processing only of derivatives without rawdata, it has empty dataset_description and it crashes here.

Modules/Module_Processing/xASL_init_LoadDataPar.m
```diff
@@ -0,0 +1,133 @@
...
+    elseif bUseRoot
+        fFolder = x.dir.DatasetRoot;
+        listDatapar = listRoot;
+        fprintf('%s\n', ['dataPar.json found in ' fFolder]);
+    elseif bUseRawdata
+        fFolder = x.dir.RawData;
+        listDatapar = listRawdata;
+        fprintf('%s\n', ['dataPar.json found in ' fFolder]);
+    else
+        listDatapar = {};
```
jan-petr, line 84: In case no dataPar is discovered the it still tries to read it on line 97 and then crashes.

jan-petr (changes requested):
Few more minor bugs.

Modules/Module_Processing/Module_Population/xASL_im_CreateGroupAnalysisMask.m
HenkMutsaerts: 1: there is an advantage, nameTemplate is not a variable here containing the template information, but a dummy variable containing template names. Because this is a loop, evalling the same thing 7 times.

2. Tricky thing; I understand.

Now fixed in ad47b7f84584d9819f85850317d573152b7144a4 with a subfunction rather than eval.

Modules/Module_Processing/xASL_init_Session_TimePoint_Lists.m
HenkMutsaerts: Weird, the indentation looked fine in Matlab. Probably has to do with a tab being four spaces? Actually Matlab indeed complained that it was not aligned, but visually it was aligned. Removing empty spaces and adding tabs fixed it. This is a great example of something that automatic code improvement based on language models could save us time on.

Fixed in 52b2006df0326a4c88d729cb66965978697a61fb (also refactored this function)

Modules/Module_Processing/xASL_init_Session_TimePoint_Lists.m
HenkMutsaerts: Now added this in the comments. nTotalSubjects is before exclusion, nSubjects after.

x.dataset.exclusion is an optional cell list of subjects that you want to exclude from loading (either from rawdata or derivatives)

Fixed in 52b2006df0326a4c88d729cb66965978697a61fb

Modules/Module_Processing/xASL_init_Session_TimePoint_Lists.m
HenkMutsaerts: I didn't change this behavior, this was just moving code. Now did some major fixes in 52b2006df0326a4c88d729cb66965978697a61fb

Modules/Module_Processing/xASL_init_Session_TimePoint_Lists.m
HenkMutsaerts: Fixed in 47e11499aaa1c09451713efd75adfdfb7106f3be

Modules/xASL_module_BIDS2Legacy.m
HenkMutsaerts: You were confused with sessions/runs right? Here, session = a visit. I added a comment to clarify this. OK? Or do we rename some variables?

Fixed in 94563ac2dba3ecaa0f12dfea7912c55e84cc20c3

Modules/Module_Processing/Module_Population/xASL_im_CreateGroupAnalysisMask.m
HenkMutsaerts: Sure. Done in ad47b7f84584d9819f85850317d573152b7144a4

Modules/Module_Processing/xASL_init_DefineStudyStats.m
HenkMutsaerts: Good idea, I didn't know `cell(0)`. I did a quick search and we use this `= '';` initialization for cells quite a lot (probably my mistake). But in fact it doesn't seem to matter, see Matlab screenshot below. So it seems that an empty char array can be indexed by {} cell indices and filled and converted into a cell array. So unless you think initializing a char array is a problem, I would not go looking now to change these throughout the code. Do you agree? I have changed it here and will do when I see it.

    >> filelist = cell(0)
    filelist =
    0×0 empty cell array

    >> filelist{end+1} = 'jan'
    filelist =
    1×1 cell array
    {'jan'}

    >> filelist = ''
    filelist =
    0×0 empty char array

    >> filelist{end+1} = 'jan'
    filelist =
    1×1 cell array
    {'jan'}

Although:

    >> filelist = {''}
    filelist =
    1×1 cell array
    {0×0 char}
    >> filelist{end+1} = 'jan'
    filelist =
    1×2 cell array
    {0×0 char}    {'jan'}

Fixed in 9b26b9648f01ef5bf632e1b2247dda296e6c86e7 (and now also in other functions changed from now in this issue)

Modules/Module_Processing/xASL_init_Loaddataset_description.m
HenkMutsaerts: Sorry, this function was simply not finished. Fixed in decce6ef16de421b77588635b54d9b4956c90017

Modules/Module_Processing/xASL_init_Process.m
HenkMutsaerts: True. The question remains if x.opts.bLoadData and x.opts.bLoadableData are still usable after my refactoring. Fixed:

`x.opts.DataLoaded`: 509e2976f384163a8f70ac247415a7c4e29b5287
`x.opts.bLoadableData`: a62d07559eac39c6d486820d85acbeaddcf3b67f
`x.opts.bLoadData`: 419395dbfb1c8ae135dce3861dc5d9daa8d9d97e

Modules/Module_Processing/xASL_init_Process.m
HenkMutsaerts: Yes, now fixed in 419395dbfb1c8ae135dce3861dc5d9daa8d9d97e (or any of my recent other x.opts refactoring)

Modules/Module_Processing/xASL_init_SubjectList.m
HenkMutsaerts: dataPar.json is a processing settings file, so should be loaded for processing only. BIDS2Legacy doesn't copy dataPar.json anymore, this is now done in 

ExploreASL_Process>xASL_init_DataLoading>xASL_init_LoadDataPar

where it checks for dataPar.json at several locations, in order or preference, and loads it for further processing.

I now changed the order, see issue and new code. Done in latest commits

Modules/Module_Processing/xASL_init_LoadDataPar.m
HenkMutsaerts: Oops :)

Fixed in 2105779ffe8d77ebeef8f57150c086e5cbedfb1a

HenkMutsaerts (commented):
Some fixes!

Modules/Module_Processing/xASL_init_Process.m
```diff
@@ -10,81 +10,496 @@
...
+    
+            spm_get_defaults('cmd_line',true);
+    
+        end
+
+
 
+%% =======================================================================================================================
+%% =======================================================================================================================
+function [x] = xASL_init_RemoveLockDirs(x)
```
jan-petr: Careful. This can't be a subfunction as it is called also globally!
ExploreASL_Process.m:84:        x = xASL_init_RemoveLockDirs(x);

Modules/Module_Processing/xASL_init_DataLoading.m
```diff
@@ -44,608 +38,56 @@
...
+    %% 4. Which data to read
+    if ~isfield(x.opts, 'bReadRawdata')
+        x.opts.bReadRawdata = true;
+        %  by default we read rawdata
     end
     
-    % In a future release we could set all defaults based on a JSON here
-    if ~isfield(x.modules.structural, 'bSegmentSPM12')
-        x.modules.structural.bSegmentSPM12 = false;
+    if x.opts.bReadRawdata && ~xASL_exist(x.dir.RawData, 'dir')
```
jan-petr: If I run Processing module. And derivatives folder exists and rawdata folder is missing. It tries to read rawdata and crashes. Despite that rawdata are not really necessary here.

jan-petr: That's when any processing is executed...

jan-petr (changes requested):
Two small bugs.

Modules/Module_Processing/xASL_init_Process.m
HenkMutsaerts: Oops I didn't finish this. Fixed in 289f654680dff2b831fd25e67cfb0b9863510072

Modules/Module_Processing/xASL_init_DataLoading.m
HenkMutsaerts: Sorry for not testing this. Fixed in 136c839b51e258fb1c2b425409b7f0be4898e5ee

Modules/Module_Processing/xASL_init_DataLoading.m
```diff
@@ -44,608 +37,46 @@
...
+        % If the developer has not set this parameter, we try to detect whether we should read from /rawdata or /derivatives/ExploreASL 
+        % Usually, this parameter will not exist
+        if xASL_exist(x.dir.RawData, 'dir')
+            x.opts.bReadRawdata = true; %  by default we read rawdata
+            fprintf('%s\n', '/rawdata folder detected, trying to load subjects from BIDS');
+        elseif xASL_exist(x.dir.xASLDerivatives, 'dir')
+            x.opts.bReadRawdata = false;
+            fprintf('%s\n', '/derivatives/ExploreASL folder detected, trying to load subjects from previous ExploreASL processing');
+        else
+            error('No rawdata or derivatives found to load subjects from, first convert dicoms 2 BIDS');
```
jan-petr: Minor error - dicoms->DICOMs. You can immediately resolve after fixing without replying ;)

Modules/Module_Processing/xASL_init_DataLoading.m
```diff
@@ -44,608 +37,46 @@
...
-                x.ROOT = x.D.ROOT;
-            else
-                warning([x.D.ROOT ' didnt exist as folder, trying path of DataPar file']);
-                x.D.ROOT = pathstr;
-                x.ROOT = pathstr;
-            end
-        end
     
+    %% 4. Which data to read
+    if ~isfield(x.opts, 'bReadRawdata') || isempty(x.opts.bReadRawdata) || ~islogical(x.opts.bReadRawdata)
```
jan-petr: You probably need to be very careful here. I wouldn't be so strict with the logicals - x.opts.bReadRawdata can still be 0 or 1. Which is then still a boolean in the broad sense, but isn't "logical".

We can probably remove this entire condition as this is an internal variable always being true/false or 1/0 and not a string. Or am I missing something special?

Even for a parameter entered by the user in dataPar.json, I would not be too strict and let them use 1/0 instead of true/false.

Functions/Administration/xASL_adm_DefineASLSequence.m
```diff
@@ -1,4 +1,4 @@
-function [x] = xASL_adm_DefineASLSequence(x)
+function [x] = xASL_adm_DefineASLSequence(x, bVerbose)
```
jan-petr, line 1: Please add bVerbose to line 4, line 10. And if it is an optional parameter, it needs admin on line 23 as well.

jan-petr (changes requested):
A few small things.

Modules/Module_Processing/xASL_init_SubjectList.m
```diff
@@ -0,0 +1,183 @@
...
+
+    if isempty(x.dataset.TotalSubjects)
+        fprintf('%s\n', 'No subjects found, subjectRegexp in dataPar.json should match with subjectFolder');
+        fprintf(2,'This was %s ...\n\n', x.opts.subjectFolder);
+        error('No subjects defined...');
+    end
+
+    % And check that this doesn't accidentally contain pipeline directories
+    ListPipelineDir = logical([]);
+    for iSubject=1:x.dataset.nTotalSubjects
```
jan-petr, line 113: nTotalSubjects NOT defined at this line! It still has to be calculated from x.dataset.TotalSubjects.

jan-petr (changes requested):
One more small bug.

Modules/Module_Processing/xASL_init_DataLoading.m
HenkMutsaerts: Can be defined in dataPar indeed. But I see your point, will remove the islogical. Fixed in 9afb01be1e81c615a976a2ae525063d7e2e3c0c2

Functions/Administration/xASL_adm_DefineASLSequence.m
HenkMutsaerts: Sorry for missing this. Fixed in 9afb01be1e81c615a976a2ae525063d7e2e3c0c2

Modules/Module_Processing/xASL_init_SubjectList.m
HenkMutsaerts: Fixed in 9afb01be1e81c615a976a2ae525063d7e2e3c0c2

HenkMutsaerts (commented):
Fixed in 9afb01be1e81c615a976a2ae525063d7e2e3c0c2

Functions/Administration/xASL_adm_DefineASLSequence.m
jan-petr: Thanks - we still need to list it in the INPUT parameter list...

HenkMutsaerts: Done in 40c014963eebcc2c8d6d01a817980edd83077a5d

jan-petr (approved):
OK

Issue #1551, opened 2023-11-23
Master issue revamped subtraction


jan-petr:
### Description

Issues #1096 and #997 both edit the DEBBIE concatenation and how PWI subtraction is calculated. To finish these two tasks, there is a follow-up issues #1543 that can be only completed once these two branches are merged. 

There will be a provisional branch related to this issue and all issue mentioned here will be merged to this branch.
Note that steps have to be done sequentially, but issues within the same step can be done in parallel. 
- [x] Step 1 - Branch #1551 is based on `develop` after #1563 and #1577 are merged.
- [x] Step 2 - Branch #1096 will be merged into it and their conflicts resolved.
- [x] Step 3 - Branch #997  will be merged into it and their conflicts resolved.
- [x] Step 3 - UnitTest
- [x] Step 3 - FlavorTest
- [x] Step 4 - TestDataSets #1581 
- [x] Step 4 - #1585 
- [x] Step 4 - #1211 
- [x] Step 4 - #1601 
- [x] Step 5 - #1539 can be done anytime
- [x] Step 5 -  #1549    
- [x] Step 5 - #1603 is first part of the follow-up.
- [x] Step 6 - after merging all, rebase the branch with merged #1584, #1587, #1599,  #1596, #1590, #1610, #1597, #1535, #1613, #1081, #1071   
- [x] Step 7 - #1543 cannot be done in parallel with other issues.
- [x] Step 8 - Run and update values for TestDataSet
- [x] TestDataSet - TC ASL and TC registration doesn't work.
- [x] Step 9 - This branch will be then finally merged to develop. No more testing of it will be needed, as final testing can be done on #1543 , but we might do a last formal review of the code changes of this branch versus `develop`.

- [x] Expand the functionality of xASL_delete to also delete .mat & .json sidecars by default (integrating xASL_adm_DeleteFilePair), and apply this throughout the pipeline. JAN: Perhaps in a new issue later?
- [x] Add the scaling list parameter into the documentation. In documentation add SessionMergingScaling and change sessionMergingList to SessionMergingList. JAN will do this before merging.
[...]

PR #1556, opened 2023-12-06
Feature #1474 philips par rec bbb

### LINKED ISSUE #1474: Import and processing of Thijs's BBB ASL
### Description

Import PAR/REC BBB ASL data from Thijs. Make sure that the parameters are right and that all is set correctly for processing.

### Tasks

- [x] Add a flavor to a branch BBB-Thijs @HenkMutsaerts 
- [x] Create a new issue for XML/REC import #1510 
- [x] Insert everything into Flavors with XML/REC import deactivated for the moment
- [x] Adapt reading PAR parameters to new names
- [x] Add reading of `Scan Type` PAR tag
- [x] Import all data
- [x] Set all parameters
- [x] Make TE import work 
- [x] Fix a correct import of BBB PAR/REC to rawdata
- [x] dcm2nii reads each TE as a separate NII, so we need to interleave TEs again.
- [x] T2prep - repeat TEs correctly.
- [x] T2prep - make sure we have the correct number of totalacquired pairs
- [x] T2prep - create a correct PLD and TE and repetition ordering - change order TE/repetition/PLD to TE/PLD/repetition. Or make sure that this is correctly reorganized in the quantification. So we might need to move this for later.
- [x] Make sure the decoding goes well for Hadamard 4
- [x] Had8 is encoded with hadamard(8) from Matlab
- [x] check if old had8 from Philips still work
- [x] T2prep seems to interleave control/label in import and thus skipping the TE dimension - we have to avoid that.
- [x] T2prep looks like multi-TE, but acquired in a standard control/label fashion.
- [x] Special quantification of T2prep - any other parameters to read? The two sequences need to be added to the import
- [x] Make quantification work
- [x] Fix flavors:
- [x] 'Philips_PCASL_2DEPI_PARREC_1', 
- [x] 'Philips_PCASL_2DEPI_PARREC_2', 
- [x] Philips_PCASL_3DGRASE_DEBBIE_PARREC_1
- [x] 'Siemens_PCASL_3DGRASE_VD13A_DEBBIE_1', 
- [x] 'Siemens_PCASL_3DGRASE_VE11C_DEBBIE_1', 
- [x] 'Siemens_PCASL_3DGRASE_VE11C_DEBBIE_2' 
- [x] 'Siemens_PCASL_3DGRASE_VE11E_DEBBIE_1' 
- [x] 'Siemens_PCASL_3DGRASE_XA30_DEBBIE_1' 
- [x] Final test of Flavors

### How to test

Two flavors added `Philips_PCASL_3DGRASE_DEBBIE_PARREC_1` and `Philips_PCASL_3DGRASE_DEBBIE_XMLREC_1`, but the XML will only be fixed later.

ASL1 - T1, First HAD4_multiTE, M0
ASL2 - first T2prepHAD4
ASL3 - second T2prepHAD4
ASL4 - second HAD4_multiTE
ASL5 - HAD8 

For Multi-Echo:
The data dimensions will be: xdim ydim zdim 4 8 (I think it’s 4 and 8 or else 8 and 4, but you will see it). The ‘4’ denotes the hadamard4 matrix that was used to encode the data so these are four raw images needed to get 3 PLDs. The ‘8’ denotes the eight echoes that were obtained in the multi-echo GraSE readout.

For T2prep: The data dimensions will be: xdim ydim zdim 8 4. The ‘4’ denotes the hadamard4 matrix that was used to encode the data so these are four raw images needed to get 3 PLDs. The ‘8’ here denotes the dynamics which were used to acquire two rounds of T2preps: so [0, 40, 80, 160, 0, 40, 80, 160].

To perform Hadamard-4 decoding you shouldn’t use the Matlab hadamard(4) function because although it is a Hadamard-4 matrix, it isn’t exactly the same as what was implemented on our Philips scanner for encoding. You will need this one (and the first column is disregarded):
Hadamard4 = [1     1     1     1
             1     1    -1    -1
             1    -1     1    -1
             1    -1    -1     1];

Hadamard-8 is using Matlab's hadamard(8)

% Perform Hadamard decoding to subtract the data:   
      had_cnt =4;
echo_cnt =4;
Decoded=zeros(x_dim, y_dim, z_dim, echo_cnt, had_cnt-1); % assign empty variable for decoded ASL data
    for i=1:had_cnt-1
        for j=1:had_cnt
            Decoded(:,:,:,:,i)= Decoded(:,:,:,:,i)-(H(i+1,j)*inputdata(:,:,:,:,j));
        end
    end

It should look like perfusion data for three timepoints and over four echoes now.
The PLDs were 600, 1600, 2600
The LDs were 1000, 1000, 1000

### Release notes

Add new Philips Hadamard multi-TE flavor.

jan-petr:
### Linked issue

Close #1474


Functions/BIDS/xASL_bids_MergeNifti.m
```diff
@@ -346,14 +347,87 @@
 end
 
 
+%% ===========================================================================================================
+%% ===========================================================================================================
+function pathOut = xASL_bids_MergeNifti_PhilipsMultiTE(NiftiPaths)
```
HenkMutsaerts: You know for sure that this is specific for multi-TE Philips data, not just in general for Philips data with other parameter vectors?

Functions/Quantification/xASL_quant_HadamardDecoding.m
```diff
@@ -46,8 +46,8 @@
 
 % Specify the TimeEncodedMatrix
 if isfield(xQ,'TimeEncodedMatrix') && ~isempty(xQ.TimeEncodedMatrix)
-	if (size(xQ.TimeEncodedMatrix,1)+1) ~= size(xQ.TimeEncodedMatrix,2)
-		error('TimeEncodedMatrix must be square (N-1)xN ');
+	if (size(xQ.TimeEncodedMatrix,1)) ~= size(xQ.TimeEncodedMatrix,2)
```
HenkMutsaerts: You can remove the first double brackets

Functions/Quantification/xASL_quant_HadamardDecoding.m
```diff
@@ -46,8 +46,8 @@
...
+		error('TimeEncodedMatrix must be square NxN ');
```
HenkMutsaerts, line 50: You mean square MxN right?

Functions/Quantification/xASL_quant_HadamardDecoding.m
```diff
@@ -76,44 +76,49 @@
...
+                                    1 -1 -1  1  1 -1 -1  1;
+                                    1  1 -1 -1  1  1 -1 -1;
+                                    1  1 -1 -1 -1 -1  1  1;
+                                    1  1  1  1 -1 -1 -1 -1];
 		end
 		
 		% #### For Hadamard Decoding Matrix ####
-	elseif strcmp(xQ.TimeEncodedMatrixType,'Hadamard')
+	elseif strcmpi(xQ.TimeEncodedMatrixType,'Hadamard')
+		% An alternative Philips version xQ.TimeEncodedMatrix = [1,1,1,1;1,1,-1,-1;1,-1,1,-1;1,-1,-1,1];
```
HenkMutsaerts: I would put this here in the same format as the others (so not 1 continuous row but also a square)

Functions/Quantification/xASL_quant_HadamardDecoding.m
```diff
@@ -167,8 +172,8 @@
     for iTE = 1:xQ.NumberEchoTimes
         for iTI = 1:nDecodedTI
             
-            indexPositive = find(xQ.TimeEncodedMatrix(iTI,:)==1);
-            indexNegative = find(xQ.TimeEncodedMatrix(iTI,:)==-1);
+            indexPositive = find(xQ.TimeEncodedMatrix(iTI+1,:)==1);
+            indexNegative = find(xQ.TimeEncodedMatrix(iTI+1,:)==-1);
```
HenkMutsaerts, line 198: Why not keeping the matrix square until here, and then add an extra first row with positives only?
Or is that the whole point, that not everybody has this?

Modules/xASL_module_ASL.m
```diff
@@ -150,16 +150,45 @@
 
 %% D4. TimeEncoded parsing
```
HenkMutsaerts, line 152: I would rename this D4 to its final position (check our 997 & XX merging issue).

Also, if you make this this long, I would move it to a subfunction

Modules/xASL_module_ASL.m
```diff
@@ -150,16 +150,45 @@
...
+		x.Q.TimeEncodedMatrixSize = [];
+	end
+
+	if ~isfield(x.Q, 'TimeEncodedMatrix')
+		x.Q.TimeEncodedMatrix = [];
+	end
+
+	% Check for duplicate definitions
+	if ~isempty(x.Q.TimeEncodedMatrixType) && ~isempty(x.Q.TimeEncodedMatrix)
+		error('Both TimeEncodedMatrixType defined and TimeENcodedMatrix is provided. Please provide only one parameter.')
```
HenkMutsaerts: Bit weird error. The user could perfectly fine add the correct TimeEncodedMatrix that belongs to the MatrixType right.

BTW: these TimeEncoded parameters belong to the data (so to the ASL json) not to the dataPar/pipeline settings right?
And they are strictly a quantification parameter, why are they checked here?

This is a good example. Inside xASL_im_HadamardDecoding you could nicely check if TimeEncodedMatrix & MatrixType suit one another, instead of a quick and dirty check here.

Think of all the other data parameter checks that we would have to move here, to be consistent. This would be the full BIDS stuff, like checking that PASL doesn't have labeling duration. To understand the content, the same comments that are in the quantification functions need to be copied here...

As a temporary solution, what we could do, is create a separate `xASL_quant_ParameterCheck` function that checks all these parameters. This can then be called at the start of xASL_module_ASL as well as at the start of xASL_wrp_Quantify.

HenkMutsaerts (changes requested):
One major question, rest minor stuff

Functions/Quantification/xASL_quant_HadamardDecoding.m
jan-petr: Are you sure you don't want to google the definition of a square? :)

The matrix needs to be indeed a square (NxN), not a rectangle (MxN).

Functions/Quantification/xASL_quant_HadamardDecoding.m
jan-petr: Agree. Done in 63feb017

Functions/Quantification/xASL_quant_HadamardDecoding.m
jan-petr: Will do.

Functions/Quantification/xASL_quant_HadamardDecoding.m
jan-petr: While the first row/column is in most cases all 1, everybody still prints it in papers/code/etc. 
So more common, easier, and less confusing to have that line in here as well.

Modules/xASL_module_ASL.m
jan-petr: It makes no sense to rename things that are later renamed in another commit. This only creates unnecessary conflicts. 

Best to keep those edits in #997 only and this renaming will happen automatically when merging with the other issue. Trying to sync commits before merging manually only brings trouble.

Functions/BIDS/xASL_bids_MergeNifti.m
jan-petr: Yes. A very specific Philips thing, deeply connected to the PAR/REC format.

Modules/xASL_module_ASL.m
jan-petr: yes. studyPar.json. Good idea to check if it really differs, then use the matrix itself and issue a warning. I'm doing that now in the HadamardDecode as we need the correct matrix which is obtained there. Done in 63feb017

Other than that. This is not quick and dirty solution. You wanted to have these checks there. I just added the new custom matrix.

It can't be done in Wrp_Quantify because we need to have that already while doing the subtraction.
We can't do that in quant_ParamCheck as we can only do this later when we have the correct matrix.
We can't do this as part of BIDS check as this is not a BIDS parameter.

Functions/Quantification/xASL_quant_HadamardDecoding.m
jan-petr: Done in 63feb017

Modules/xASL_module_ASL.m
HenkMutsaerts: That's what I'm saying. We move all other parameter checks here, to be consistent. This will make this super-long (think of the Hct checks, blood T1 parameter checks, all other quantification checks).

So to fix this, I suggest that we create a separate `xASL_quant_ParameterCheck` function where we do all these checks, and we can call this at multiple locations.

Fine of course doing this in #1551

Modules/xASL_module_ASL.m
HenkMutsaerts: I disagree. It is only about the number in the header. I did the same in #997 and #1096. It will be much easier to merge if you have such true conflicting things (like the numbering of these section headers) in the issues themselves.

But your choice. You can resolve ;)

Functions/Quantification/xASL_quant_HadamardDecoding.m
HenkMutsaerts: I put two comments there, but for a strange reason this didn't go to my review...

Functions/Quantification/xASL_quant_HadamardDecoding.m
HenkMutsaerts: I find NxN square confusing, because in this case N~=N right (the contents of N are not the contents of N).
But I understand that MxN don't refer to its contents but simply the length. Now I understand. Never mind :)

Functions/BIDS/xASL_bids_MergeNifti.m
HenkMutsaerts: You're not answering my question.
My question was "are you sure that this is specific to multi-TE, not general Philips/PAR/REC?"
Then you have the choice of answering "yes this is specific to multi-TE. Other Philips or PAR/REC parameter never have this issue" OR "no this is a generic Philips or PAR/REC issue. You're right that this could occur to non-TE parameters as well. Let's rename this function.

Or am I missing something?

HenkMutsaerts (changes requested):
See in my responses

Functions/Quantification/xASL_quant_HadamardDecoding.m
jan-petr: Check this 
[link](https://github.com/ExploreASL/ExploreASL/pull/1556/commits/63feb0179603254c0b740de50f64244c6baccbb8)

Modules/xASL_module_ASL.m
jan-petr: I don't follow. Move where? 
Some checks need to be done at the start of ASL module, some later but before Quantify, some later but before subtraction.

So where you want to call this parameterCheck function? Cause you can't call all at the same time. you have to call some right at the start, but not later. ANd other check cannot be called at the start, but only later. 
So I don't see where you could call parameter check at the same time.

But of course - all parameter checks that are at the start of the ASL module will go to a sub-function. We have already agreed upon that.

Functions/Quantification/xASL_quant_HadamardDecoding.m
jan-petr: OK. So you can resolve...

Functions/Quantification/xASL_quant_HadamardDecoding.m
jan-petr: Two those things are fixed [here](https://github.com/ExploreASL/ExploreASL/pull/1556/commits/126a289a058a64b290bdbd4d59f4dacbb38d9e19)

Functions/BIDS/xASL_bids_MergeNifti.m
jan-petr: Sorry. Yes - specific to multi-TE Philips. The standard Philips merging could be done with the general function.

Though it's probably not necessary to keep multi-TE in the name. I have changed that.

Now I explain that we focus on Philips sequences and multi-TE specifically. Still, this function is supposed to deal with cases that cannot be fixed by general merging. And currently, this is only the multi-TE case. But for future generalizibility, I have removed the multi-TE mention from the function name.

See [here](https://github.com/ExploreASL/ExploreASL/pull/1556/commits/b2b5a5d7c49330706e127bca6b2ac3e490396021)

HenkMutsaerts (approved):

PR #1564, opened 2023-12-19
#1563 Minor fixes parallelization bugs

### LINKED ISSUE #1563: Processing and conversion error
### Description

There are several errors related to #1547 that appeared after merging it with other commits to develop.

### Tasks

- [x] DataPar.json is not read when loading separate derivatives folder before loading subjects. GE_3Dspiral_T1prefilled_Cyst_M0_noFLAIR_2ndRun_LoQ crashes because it loads subjects before loading dataPar.json with the correct subjectRegexp first. Fixed in 11f00201
- [x] GE_PCASL_3Dspiral_DV25.0_1.5T_1 during BIDS2Legacy, the subjectRegexp is missing in dataPar.json. Previously, we created the automatic one. But we don't need to create this as it is automatically created later. So we don't need to create x.dataset either. Skipping this in 11f00201 and updating all flavors.
- [x] Associated to above -> update reference dataPar.json in flavors.
- [x] Fix non-working derivatives - adding _1 in flavors to the subject name in derivatives.
- [x] **HENK:** xASL_init_SubjectList on line 28 it says it will use a default folder. But it doesn't initialize `subjectFolder` and will eventually crash on line 37.
**ANSWER HENK**: yes fixed in 25a6e2b1eca9fadee52f988828cc667ad69eba00, by changing the order (because the default folder depends on rawdata vs derivatives as input).
- [x] **JAN:** Does this processing run? `Philips_2DEPI_FLAIR_Segm_PWI4D_Quant_Parameters_LoQ`. 
      Fix1 35313a54
      Fix2 xASL_io_CheckDeprecated fields that creates field x.modules.asl = 1 that later crashes the code. The reason is the the CheckDeprecated fields finds a field in x-struct that matches the first two columns and starts renaming even if it doesn't match the third column. So x.modules.asl. To test this behavior, I have added a new function to UT. 8a13cd60
- [x] Cleanup warning bug in `Philips_2DEPI_FLAIR_Segm_PWI4D_Quant_Parameters_LoQ` - reach consensus on the solution. See my comment in https://github.com/ExploreASL/ExploreASL/pull/1564/commits/25a6e2b1eca9fadee52f988828cc667ad69eba00
- [x] Parameter TC_M0_Registration not assigned in the TestDataSet. The reason is that 'noSmooth_M0_' or 'mean_control_' were compared with the M0 template afterwards by the testing function. But the testing fuction is run after the population modules and we delete both files in the cleanup. Now, we moved the creation of this TC to the end of xASL_module_ASL. Note that we compared an individual M0 with a pre-existing template from another study (coming with ExploreASL), so we don't need to worry that this template doesn't exist.

- [x] xASL_qc_CleanOldQC: Bug when rerunning ExploreASL without pre-existing data: 088fad3be8e86dcfdfa69204479941add8f95f29

### How to test

- [x] Test Flavors
- [x] Test UnitTest
- [x] Test 10-TestDataSets

### Release notes

Report with #1547 .

jan-petr:
### Linked issue

Close #1563


Functions/Administration/xASL_adm_CleanUpBeforeRerun.m
```diff
@@ -324,10 +324,10 @@ function xASL_adm_CleanUpBeforeRerun(AnalysisDir, iModule, bRemoveWMH, bAllSubje
 
         RemoveFields = {};
         if ~isempty(find(iModule==1)) % if we remove the structural provenance
-            RemoveFields{end+1} = 'Structural';
+            RemoveFields{1} = 'Structural';
```
HenkMutsaerts: No this is incorrect. The list needs to be extended with structural and/or ASL. E.g., if there is no structural, ASL should go to position 1. And this doesn't give a bug. I reverted this now in 25a6e2b1eca9fadee52f988828cc667ad69eba00.

For an empty cell, {end+1} is the first position. So this works nicely when you iteratively extend a cell array.

HenkMutsaerts (commented):
There is a minor thing that I will now fix

HenkMutsaerts (approved):

Functions/Administration/xASL_adm_CleanUpBeforeRerun.m
jan-petr: In case iModule is [ 2], then RemoveFields == {'ASL'}.

But then later in the code, you use iModule to index RemoveFields, so you call RemoveFields{2}, which obviously isn't 'ASL' but returns an error. The correct solution that solves everything is to create RemoveFields with the iModule indices, instead as a growing list.
That means - revert back to the change I made. You can of course keep RemoveFields = cell(0) - that is fine.

jan-petr: Sorry, it wasn't an error. But it was an important warning.

Running cleanup on ASL only (that is [2] and not [1 2]) creates this warning and then skips a large part of this cleaning module.

`Cleaning up x-struct:100%Warning: Index exceeds the number of array elements. Index must not exceed 1. 
> In xASL_adm_CleanUpBeforeRerun (line 447)
In xASL_module_ASL (line 102)
In xASL_init_Iteration>runIteration (line 409)
In xASL_init_Iteration (line 146)
In ExploreASL_Process (line 138)
In ExploreASL (line 97) `

Either way - my fix fixes this.

HenkMutsaerts: Should now be fixed by ﻿﻿﻿6c7dd738416cb5a2eb60df9db5a48a3073920c1a

jan-petr: You are indeed missing the point.

You have iModule == [2] on the input, then you create RemoveFields{'ASL'} and then you try to index RemoveFields{'ASL'} with iModule and RemoveFields{2} throws an error...

While my solution RemoveFields{empty 'ASL'} is indexed only with 2, so the empty is effectively unused.

The problem is that this doesn't really have another solution than I propose because you need to use this iModule as index and by creating the incremental list, you loose any info about order. 

Just test this and you'll see it for yourself. I'm of course fine with any other more complicated solution. But the current version doesn't work.

HenkMutsaerts: Should be fixed now in 6c7dd738416cb5a2eb60df9db5a48a3073920c1a

jan-petr: Not fixed. it's even worse. Please read my comment. You can't create a list RemoveFields incrementally and then index this with iModule that has completely different indices.

PR #1565, opened 2023-12-20
#1418 xASL_quant_M0: Do not save BSup correction without JAVAVM

### LINKED ISSUE #1418: Bsup correction fails under -nojvm
### Description

When running the B-sup correction for M0 quantification in matlab with -nojvm option it crashes as it tries to display a figure.

`    'Error using figure
     This functionality is not supported under the -nojvm startup option.
     
     Error in xASL_quant_M0>xASL_quant_RevertBsupFxControl (line 333)
         FigureHandle = figure('visible','off');
     
     Error in xASL_quant_M0 (line 83)
         [M0IM, x] = xASL_quant_RevertBsupFxControl(M0IM, x);
`
### Release notes

Deactivate Bsup correction figure saving when JAVAVM is turned off.

jan-petr:
### Linked issue

Close #1418

### DIFF
--- a/Functions/Quantification/xASL_quant_M0.m
+++ b/Functions/Quantification/xASL_quant_M0.m
@@ -329,8 +329,17 @@ function [M0IM, x] = xASL_quant_RevertBsupFxControl(M0IM, x)
 			end
 	end
 	
-        
-    SignalPercentage = abs(xASL_quant_BSupCalculation(x.Q.BackgroundSuppressionPulseTime, ReadoutTime, x.Q.PresaturationTime, x.Q.TissueT1, SliceReadoutTime, x.D.M0CheckDir, 1));
+	if usejava('desktop') || usejava('jvm')
+		bCreateFigure = 1;
+		% Initialize the figure
+		FigureHandle = figure('visible','off');
+		subplot(2, 2, 1);
+	else
+		bCreateFigure = 0;
+		FigureHandle = [];
+	end
+	
+    SignalPercentage = abs(xASL_quant_BSupCalculation(x.Q.BackgroundSuppressionPulseTime, ReadoutTime, x.Q.PresaturationTime, x.Q.TissueT1, SliceReadoutTime, x.D.M0CheckDir, bCreateFigure));
     
     %% Obtain the slice-wise median Control signal before correction
     % First create a reasonable mask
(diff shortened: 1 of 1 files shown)


Functions/Quantification/xASL_quant_M0.m
```diff
@@ -364,31 +361,40 @@
...
+		xlabel('Slice (integer)');
+		% After correction
+		subplot(2, 2, 4);
+		plot(1:size(M0IM,3), SignalAfterCorrection,'r');
+		title('After correction');
+		ylabel('Pseudo-M0 (a.u.)');
+		xlabel('Slice (integer)');
+
+		% Save and close the figure
+		SavePath = fullfile(x.D.M0CheckDir, 'RevertBsupFxControl.jpg');
```
HenkMutsaerts: Actually, this is probably a bug somebody introduced before (not me ;)

This saves a single jpg that will be overwritten for each ASL subject/session right?
So it should probably be [x.SUBJECT '_' x.SESSION 'RevertBsupFxControl.jpg' ?
Or don't you want to check this for each subject (then still this will only show for the last subject/session)?

HenkMutsaerts (changes requested):
Minor comment and minor fix in my last comment

Functions/Quantification/xASL_quant_M0.m
jan-petr: Good point! Actually, this figure depends on the sequence parameters, so it won't really matter if this differs. But in the future it could, so I'm adding the session+name. 06496526

Functions/Quantification/xASL_quant_M0.m
```diff
@@ -387,7 +387,7 @@
 		xlabel('Slice (integer)');
 
 		% Save and close the figure
-		SavePath = fullfile(x.D.M0CheckDir, 'RevertBsupFxControl.jpg');
+		SavePath = fullfile(x.D.M0CheckDir, [x.SUBJECT '_' x.SESSION '_RevertBsupFxControl.jpg']);
```
HenkMutsaerts, line 395: Did you check that x.SUBJECT & x.SESSION are defined at this point in the code? That was just a guess from me, I haven't tested that.

HenkMutsaerts (approved):
Please check my question otherwise approved

Functions/Quantification/xASL_quant_M0.m
jan-petr: Aha ;) Will check and if this works, I will merge But I'm 99% sure it is define, why wouldn't it :)

jan-petr: Fixed

PR #1566, opened 2023-12-20
#1303 ExploreASL_Initialize fix spaces in path

### LINKED ISSUE #1303: Leading or trailing spaces in paths, initializing ExploreASL
### Description

Leading and trailing spaces in paths will crash ExploreASL. They should not crash ExploreASL but rather give a warning and continue (with the space removed). This will save us some user questions and thus time.

### Tasks

- [x] Detect spaces in paths on the first and last position. When detected, remove them and issue a warning.

### How to test

`ExploreASL(' /Users/henk/ExploreASL/FlavorDatabase/GE_PCASL_3Dspiral_DV22.0_1',1,0);`

### Release notes

Remove leading and trailing spaces in the path.

jan-petr:
### Linked issue

Close #1303

### DIFF
--- a/ExploreASL_Initialize.m
+++ b/ExploreASL_Initialize.m
@@ -600,6 +600,33 @@ end
                 end
             end
         
+			% Check and remove empty spaces at the start and end of the DatasetRoot
+			if ~isempty(x.opts.DatasetRoot)
+
+				% Remove the empty spaces at the start
+				if x.opts.DatasetRoot(1) == ' '
+					[~, ~, indexUniqueChars] = unique(x.opts.DatasetRoot, 'stable'); % Create a vector that gives for each character its unique identifier
+					indexCharacter2 = find(indexUniqueChars==2); % Find the indices of the second unique character
+					if ~isempty(indexCharacter2)
+						x.opts.DatasetRoot = x.opts.DatasetRoot(indexCharacter2(1):end);
+						warning(['x.opts.DataserRoot contains ' xASL_num2str(indexCharacter2(1)-1) ' empty spaces at the start. Fixing this']);
+					end
+				end
+
+				% Remove the empty spaces at the end
+				if x.opts.DatasetRoot(end) == ' '
+					% Repeat exactly the same proceedure, but on reversed character vector
+					x.opts.DatasetRoot = x.opts.DatasetRoot(end:-1:1);
+					[~, ~, indexUniqueChars] = unique(x.opts.DatasetRoot, 'stable'); % Create a vector that gives for each character its unique identifier
+					indexCharacter2 = find(indexUniqueChars==2); % Find the indices of the second unique character
+					if ~isempty(indexCharacter2)
+						x.opts.DatasetRoot = x.opts.DatasetRoot(indexCharacter2(1):end);
+						warning(['x.opts.DataserRoot contains ' xASL_num2str(indexCharacter2(1)-1) ' empty spaces at the end. Fixing this']);
+					end
+					x.opts.DatasetRoot = x.opts.DatasetRoot(end:-1:1);
+				end
+			end
+
             % Check if the user provided any path, and if this path exists
             if isempty(x.opts.DatasetRoot)
                 warning('Dataset root directory is not specified...');
--- a/ExploreASL_Initialize.m
+++ b/ExploreASL_Initialize.m
@@ -544,28 +544,20 @@ end
 			if ~isempty(x.opts.DatasetRoot)
 
 				% Remove the empty spaces at the start
-				if x.opts.DatasetRoot(1) == ' '
-					[~, ~, indexUniqueChars] = unique(x.opts.DatasetRoot, 'stable'); % Create a vector that gives for each character its unique identifier
-					indexCharacter2 = find(indexUniqueChars==2); % Find the indices of the second unique character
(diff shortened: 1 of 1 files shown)


ExploreASL_Initialize.m
```diff
@@ -600,6 +600,33 @@ function xASL_init_AddDirsOfxASL(MyPath)
                 end
             end
         
+			% Check and remove empty spaces at the start and end of the DatasetRoot
+			if ~isempty(x.opts.DatasetRoot)
+
+				% Remove the empty spaces at the start
+				if x.opts.DatasetRoot(1) == ' '
+					[~, ~, indexUniqueChars] = unique(x.opts.DatasetRoot, 'stable'); % Create a vector that gives for each character its unique identifier
```
HenkMutsaerts: This gives the rank right? "unique identifier" is very vague

When ranking 1000 2000 1500, you get 1 3 2, this is what you mean right?

ExploreASL_Initialize.m
```diff
@@ -600,6 +600,33 @@ function xASL_init_AddDirsOfxASL(MyPath)
                 end
             end
         
+			% Check and remove empty spaces at the start and end of the DatasetRoot
```
HenkMutsaerts: We have functions for this already, e.g., `xASL_adm_ReplaceSymbols` or `xASL_adm_UnixPath`, but they don't do exactly what you want here I believe.

I would at least make this a separate section in the code, or even a subfunction, to keep it all readable.

This could be as easy as replacing % with %%

HenkMutsaerts (changes requested):
Nice, some minor things

ExploreASL_Initialize.m
jan-petr: No. A unique unsorted identified. 
1000 2000 1500 gives 1,2,3

Explained in 3dc9978e

ExploreASL_Initialize.m
jan-petr: ReplaceSymbols and UnixPath indeed does a different thing.
I'm adding %%. Fine?
3dc9978e

ExploreASL_Initialize.m
HenkMutsaerts: I think you can simplify this using regular expressions that search for spaces. But fine with this implementation.

Just a minor note: if you want to explain what is happening, it is not only nice to explain what your code does technically (e.g., aacba gives 1 1 2 3 1) but also why you do this (e.g.,

% Remove empty spaces at the start of the path
% First, we check if the first index is a space
% Second, we check the position of the second unique character (the first unique character being a space).
% This position is where the path starts (there can be multiple spaces that need to be omitted)

ExploreASL_Initialize.m
```diff
@@ -600,7 +540,36 @@ function xASL_init_AddDirsOfxASL(MyPath)
...
+				if x.opts.DatasetRoot(1) == ' '
+					[~, ~, indexUniqueChars] = unique(x.opts.DatasetRoot, 'stable'); % Create a vector that gives for each character its unique identifier
+					indexCharacter2 = find(indexUniqueChars==2); % Find the indices of the second unique character
+					if ~isempty(indexCharacter2)
+						x.opts.DatasetRoot = x.opts.DatasetRoot(indexCharacter2(1):end);
+						warning(['x.opts.DataserRoot contains ' xASL_num2str(indexCharacter2(1)-1) ' empty spaces at the start. Fixing this']);
+					end
+				end
+
+				% Remove the empty spaces at the end
```
HenkMutsaerts, line 555: Same here

ExploreASL_Initialize.m
```diff
@@ -509,23 +462,10 @@ function xASL_init_AddDirsOfxASL(MyPath)
     
 end
 
-
```
HenkMutsaerts, line 512: I added these spaces for a reason I believe, that there is a lot of space between main code sections/subfunctions and less space between different code sections (not complaining on your edit, just explaining why I did this)

HenkMutsaerts (changes requested):
Minor comment, you can copy-paste my comment to explain your code.

ExploreASL_Initialize.m
jan-petr: Yesyes. I know - but then we added the double %%, which actually works now nicely, so I thought we can remove the space. If you agree?

ExploreASL_Initialize.m
jan-petr: Good idea. Actually, for PLDs, I had to write this code, which I then ported here. But you are right that with regexp, it is much easier, so I redid all that now with regexp.

The explanations is now also much more simple

HenkMutsaerts: Hmm you say that you redid everything with regexp, but that's not what I see in the code?

HenkMutsaerts (approved):
You say that you redid everything with regexp, but that's not what I see in the code? Perhaps you forgot to push?

Anyway, I'm fine keeping it like this, I added some extra comments to clarify what you're doing. I just kept 1 conversation open so you can see what I mean, but feel free to close that and merge this.

ExploreASL_Initialize.m
jan-petr: Sorry. Forgot to push indeed.

PR #1567, opened 2023-12-22
#1537 Remove nSubjectsSessions variable

### LINKED ISSUE #1537: Remove nSubjectSessions from x-struct
### Description

x.datasets.nSubjectsSessions is always x.dataset.nSubjects * x.dataset.nSessions, so this parameter can be removed from x-struct to avoid inconsistencies and errors.

### Release notes

Remove nSubjectsSessions parameter as redundant

jan-petr:
### Linked issue

Close #1537


Functions/Administration/xASL_adm_GetDeprecatedFields.m
```diff
@@ -22,7 +22,7 @@
 
 nameConversionTable = ...
     {'nSessions'                    'dataset'   'nSessions'                          '';...
-    'nSubjectsSessions'             'dataset'   'nSubjectsSessions'                  '';...
+    'nSubjects'                     'dataset'   'nSubjects'                          '';...
```
HenkMutsaerts, line 25: Nice. I just thought that we would have before:

nSessions
nSubjects
nSubjectsSessions

so are you sure that we didn't have nSubjects? Otherwise this could be duplicate.

did you check both nSubjectSessions & nSubjectsSessions?

Modules/Module_Processing/Module_Population/xASL_im_CreateGroupAnalysisMask.m
```diff
@@ -45,9 +45,9 @@
 end
 
 
-if x.dataset.nSubjectsSessions<16
+if x.dataset.nSubjects * x.dataset.nSessions < 16
```
HenkMutsaerts, line 48: do we need ( ) ? I know that * should probably take priority over <, but above you also used ( ) so why not make sure here as well

Modules/Module_Processing/Module_Population/xASL_wrp_CreatePopulationTemplates.m
```diff
@@ -161,9 +161,9 @@ function xASL_wrp_CreatePopulationTemplates(x, bSaveUnmasked, bCompute4Sets, Spe
     bSkipWhenMissingScans = false;
 end
 
-if ~x.bForceTemplates && x.dataset.nSubjectsSessions<6
+if ~x.bForceTemplates && (x.dataset.nSubjects * x.dataset.nSessions < 6)
```
HenkMutsaerts, line 164: brackets need to be around subjects & sessions (< 6 outside the brackets)

Modules/Module_Processing/Module_Population/xASL_wrp_CreatePopulationTemplates.m
```diff
@@ -716,10 +716,10 @@ function xASL_wrp_CreatePopulationTemplates4Sets(x, bSaveUnmasked, bRemoveOutlie
         try
             WithinGroup = SetID==UniqueSet(iU);
 
-            if ~SessionsExist(iScanType) && length(WithinGroup)==x.dataset.nSubjectsSessions
+            if ~SessionsExist(iScanType) && length(WithinGroup) == x.dataset.nSubjects * x.dataset.nSessions
                 % if no sessions exist, but "WithinGroup" definition was
                 % based on all subject/sessions, then correct this
-                CurrSess = [1:length(listSessions):x.dataset.nSubjectsSessions]' + (iSession-1);
+                CurrSess = (1:length(listSessions):x.dataset.nSubjects * x.dataset.nSessions)' + (iSession-1);
```
HenkMutsaerts, line 722: These brackets need to return to [ ] square brackets, because this is a list. At least, this was on purpose, not a mistake

Modules/Module_Processing/xASL_init_LoadMetadata.m
```diff
@@ -133,7 +133,7 @@
         end
 
         for iSession=1:x.dataset.nSessions
-            VarDataOri([iSession:x.dataset.nSessions:x.dataset.nSubjectsSessions-x.dataset.nSessions+iSession], 1) = x.SUBJECTS(:);
+            VarDataOri(iSession:x.dataset.nSessions:(x.dataset.nSubjects * x.dataset.nSessions - x.dataset.nSessions + iSession), 1) = x.SUBJECTS(:);
```
HenkMutsaerts, line 136: Same here, we may need the square brackets

HenkMutsaerts (changes requested):
Few minor points

Functions/Administration/xASL_adm_GetDeprecatedFields.m
jan-petr: This is all good. I think that you are only confused by this one edit.

This function just takes 
x.nSubjects and puts them into x.dataset.nSubjects

I have checked and `nSubjectSessions` does not exist.
nSubjects was used and is initialized everywhere, we just forgot to put it into this naming correction function. Which wasn't ever a problem, because we never had wrongly placed nSubjects.

Does this answer you?

Modules/Module_Processing/Module_Population/xASL_im_CreateGroupAnalysisMask.m
jan-petr: We don't really need () anywhere -  ( * ) < 16 or (* < 16) isn't necessary. 
But inserting them from time to time is useful to avoid mistakes.

So this here can stay as it is unless you want to have it to make it easier readable. Then I can add them...

Modules/Module_Processing/Module_Population/xASL_wrp_CreatePopulationTemplates.m
jan-petr: That was not necessary. [] brackets are needed for a list, but 1:2:4 is a list by itself and it doesn't need more brackets. So you only need normal brackets before the transposition.

The following is all correct:
[1,2,3]
1:3
[1:2,3]

Modules/Module_Processing/xASL_init_LoadMetadata.m
jan-petr: We don't . As above, we only need square brackets to construct a list from several variables.
The : sign constructs a list by itself and the [] brackets are not needed.

HenkMutsaerts (approved):

PR #1571, opened 2023-12-27
#1198 xASL_init_DataLoading: Skip Legacy conversion

### LINKED ISSUE #1198: BIDS to Legacy conversion starts unnecessary
### Description

While trying to re-run the population module, using `ExploreASL('dataroot', 0, [0 0 1], 0, 1, 1)` the import module starts the BIDS to Legacy conversion automatically.

image

### Release notes

Report with #1547.

jan-petr:
### Linked issue

Close #1198

### DIFF
--- a/Modules/Module_Processing/xASL_init_DataLoading.m
+++ b/Modules/Module_Processing/xASL_init_DataLoading.m
@@ -47,11 +47,6 @@ function [x] = xASL_init_DataLoading(x)
 
     
     %% 4. Which data to read
-	% In case we call population module only, then don't attempt to convert rawdata to Legacy, but only read from derivatives
-	if isfield(x.opts, 'bProcess') && isequal(x.opts.bProcess(:), [0;0;1]) && (~isfield(x.opts, 'bReadRawdata') || isempty(x.opts.bReadRawdata))
-		x.opts.bReadRawdata = false; 
-	end
-
     if ~isfield(x.opts, 'bReadRawdata') || isempty(x.opts.bReadRawdata)
         % If the developer has not set this parameter, we try to detect whether we should read from /rawdata or /derivatives/ExploreASL 
         % Usually, this parameter will not exist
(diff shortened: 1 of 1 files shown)


Modules/Module_Processing/xASL_init_DataLoading.m
```diff
@@ -47,6 +47,11 @@
 
     
     %% 4. Which data to read
+	% In case we call population module only, then don't attempt to convert rawdata to Legacy, but only read from derivatives
```
HenkMutsaerts: I don't get this at all.
**MAIN**
Why would we not want to attempt to convert rawdata to Legacy if we call population data only?
This is tricky:
A) Note that in the near future, we will "only" have rawdata (i.e., legacy subjects are a copy of rawdata). The option to not read rawdata is a legacy option, for old data.
B) xASL_module_Population still has group-wise wrappers that do not require xASL_module_Structural or xASL_module_ASL to have been run but only BIDS2Legacy, such as 4. print DICOM header parameters & acquisition time.
10. zipping 

**MINOR**
B) your isequal is a weird statement, why do you transpose both bProcess input and x.opts.bProcess)
C) You change something without reading the code. bReadRawdata not being specified was written 1 line lower :)

HenkMutsaerts (changes requested):
Not sure if this is a good idea. I also improved your code change.

Modules/Module_Processing/xASL_init_DataLoading.m
jan-petr: If we do population module only, then this will only work with data that we already processed in derivatives. Any newly converted BIDS2Legacy will be ignored anyway = hence no point converting them.

A) Yes, but in the future, for now, we still have the option bReadRawdata.
B) That I agree with.

MINOR B) This is not a weird statement, but a necessary statement isequal([0 0 1],[0;0;1]) == 0
MINOR C) I of course read the code. Please read mine and think why I added it this way ;) My code was dealing with special option of missing bReadrawdata and bprocess == [0 0 1] - it wouldn't be possible to seamlessly add that option to the code below.

But as said - I'm fine closing this issue without doing anything...

Modules/Module_Processing/xASL_init_DataLoading.m
```diff
@@ -59,6 +59,12 @@
         else
             error('No rawdata or derivatives found to load subjects from, first convert DICOMs 2 BIDS');
         end
+
+        if isequal(x.opts.bProcess, [0 0 1])
```
jan-petr, line 63: if somebody calls ExploreASL([0;0;1]) then this condition will not catch it. 
That's why I wrote bProcess(:) and [0;0;1] which is a universal solution

Modules/Module_Processing/xASL_init_DataLoading.m
```diff
@@ -59,6 +59,12 @@
...
+            x.opts.bReadRawdata = false;
```
jan-petr, line 64: if the directory rawdata exists - it will first report above that it will read from it and then it will switch this option OFF here. That's why I put this entire IF out of this block.

jan-petr (commented):
2 new comments.

HenkMutsaerts:
Closing this as obsolete

PR #1580, opened 2024-01-03
Bug #1577 group masks

### LINKED ISSUE #1577: Bug in group Analysis mask creation
### Description

There is something wrong with the population module that causes differences in CBF calculation compared to version 1.11.0. This is especially strong in certain datasets and especially in GM. Though not only in that. Once fixed, all datasets have to be tested again.

How to see the error (my entire though process, including identifying the difference, explaining why it happens, explaining if that is really an error or not :)):

1. Run ExploreASL in `Siemens3DGRASE_Structural_AlreadyDone_HiQ` on develop. If you re-run population module only in develop/main, then mean_qCBF_StandardSpace_TotalGM_n=1_29-Dec-2023_PVC0.tsv differs in Total_GM_B in ASL_1. The old version has ~44, the new version has incorrectly 39.

For some subjects, we have error<1%. For several subjects error ~5% and for this particular subject 20%. So all subjects are affected with a different intensity.

### Tasks

- [x] JAN: Use _Unmasked.nii [7a120ac](https://github.com/ExploreASL/ExploreASL/commit/7a120ac36444f213ac9008f43b5647ec3b548924)
- [x] JAN: Test if CBF and sCoV reproducibility is fixed
- [x] JAN: Implement pvGM/pvWM instead of pGM/pWM for masking
**HENK** I have implemented standard pvGM/pvWM templates in 2bd2abed8289eefeffcd0576d33af7931fa02112
- [x] JAN: See the influence on reproducibility.
- [x] MATHIJS: Is aggressive masking used in 1.11.0 for n>16? It looks good for all datasets, but best to check this carefully with Henk.
**HENK** Probably not. We can ask Mathijs to rerun the population module with this branch to check it?
- [x] JAN: Fix TC M0 - xASL_wrp_VisualQC_ASL.m, line 174 prioritizes noSmoothM0 instead of mean-control. We have to
A.  keep this and change reference
B. change priority here
C. create two QC entries and prioritize in the Testing script.

**HENK** This is a different issue?

- [x] JAN: Retest entire TestDataSet
- [x] JAN: Update the reference - create a new 1.12 reference already.

### Release notes

Bugfix, report with #807

jan-petr:
### Linked issue

Close #1577


Modules/Module_Processing/Module_Population/xASL_im_CreateGroupAnalysisMask.m
```diff
@@ -60,9 +60,6 @@
 [PathVascularMask, bSkipStandard] = xASL_sub_CheckTemplatePath('^MaskVascular', x, bSkipStandard);
 
 [PathT1, bSkipStandard] = xASL_sub_CheckTemplatePath('^T1', x, bSkipStandard);
-[PathpGM, bSkipStandard] = xASL_sub_CheckTemplatePath('^pGM', x, bSkipStandard);
```
jan-petr, line 63: Why are we using MNI space masks instead of using the individual masks? That defies the use of group mask completely, right?

I found the original version you wrote much better, we only had to change pGM to PV_pGM...

jan-petr (commented):
One comment in the code

HenkMutsaerts (approved):
Happy to approve my own (and your) changes :)

PR #1586, opened 2024-01-08
Closes #1585 xASL_io_SaveNifti: option to save Json sidecar

### LINKED ISSUE #1585: Save JSON along with NII
### Description

Modify xASL_io_SaveNiftii to allow saving along the JSON.

### Tasks

- [x] 1. Add an additional optional parameter bCopyJson (default = false) that takes the json from the reference image (if it exists) and saves its entire content together with the new NII
- [x] 2. Add a second additional optional parameter newJson (default = empty) that contains a json structure to be saved along with the NII
- [x] 3. If none of the parameters are provided, no JSON is saved
- [x] 4. If both parameters are provided, then NewJSON fields overwrite the fields from the reference NII.JSON.
- [x] I also noticed a bug with the temporary name, fixed this.
- [x] And I noticed a bug that if a previous .mat or .json sidecar exists for the new file, but we don't create a new one, this file remained. Fixed this now.

### Release notes

xASL_io_SaveNifti can now copy the JSON sidecar.

HenkMutsaerts:
### Linked issue

Closes #1585

### DIFF
--- a/External/SPMmodified/xASL/xASL_io_SaveNifti.m
+++ b/External/SPMmodified/xASL/xASL_io_SaveNifti.m
@@ -1,7 +1,7 @@
 function xASL_io_SaveNifti(pathOrigNifti, pathNewNifti, imNew, nBits, bGZip, changeMat, bCopyOrigJson, JsonFields)
 % Save a file to a Nifti format, while taking the parameters from another file
 %
-% FORMAT: xASL_io_SaveNifti(pathOrigNifti, pathNewNifti, imNew[, nBits, bGZip, newMat])
+% FORMAT: xASL_io_SaveNifti(pathOrigNifti, pathNewNifti, imNew[, nBits, bGZip, changeMat, bCopyOrigJson, JsonFields])
 %
 % INPUT:
 %   pathOrigNifti  Path to the original Nifti file to take parameters from (REQUIRED)
@@ -20,8 +20,8 @@ function xASL_io_SaveNifti(pathOrigNifti, pathNewNifti, imNew, nBits, bGZip, cha
 %   JsonFields     Save additional JsonFields to a new JSON sidecar (or merged with the original Json content)
 %                  These fields overwrite fields with the same fieldnames in the original Json
 %                  (OPTIONAL, DEFAULT = struct() empty)
-% Options are:
-% 1. bSaveJson     = false                       -> we don't save a new JSON sidecar
+% JSON saving options are:
+% 1. bCopyOrigJson = false & JsonFields = empty  -> we don't save a new JSON sidecar
 % 2. bCopyOrigJson = true  & JsonFields = empty  -> we copy the original JSON sidecar only
 % 3. bCopyOrigJson = false & JsonFields = filled -> we create a new JSON sidecar with new fields only
 % 4. bCopyOrigJson = true  & JsonFields = filled -> we merge the original JSON sidecar with the new fields
@@ -70,13 +70,9 @@ else
 	end
 end
 
-% JSON saving
-% Options are:
-% 1. bSaveJson     = false                       -> we don't save a new JSON sidecar
-% 2. bCopyOrigJson = true  & JsonFields = empty  -> we copy the original JSON sidecar only
-% 3. bCopyOrigJson = false & JsonFields = filled -> we create a new JSON sidecar with new fields only
-% 4. bCopyOrigJson = true  & JsonFields = filled -> we merge the original JSON sidecar with the new fields
-%                                                   the new fields have priority
+% JSON saving, see detailed explanation above.
+% If bCopyOrigJson is true OR JsonFields is not empty, we set bSaveJson to true
+% So we only save a JSON in either or both of these cases
 bSaveJson = false;
 
 if nargin < 7 || isempty(bCopyOrigJson)
@@ -86,7 +82,11 @@ else
 end
 
 if nargin < 8 || isempty(JsonFields)
-    JsonFields = struct([]);
+    JsonFields = struct;
+elseif ~isstruct(JsonFields)
+    error('JsonFields input variable should be a struct');
+elseif isempty(fields(JsonFields))
+    JsonFields = struct;
 else
     bSaveJson = true;
 end
@@ -279,14 +279,19 @@ end
 
 %% ====================================================================================
 %% 8. Create JSON sidecar
+% 0. Remove JSON sidecar if it already exists
+% Even if we don't save a new one, then we still don't want the wrong sidecar to a new NIfTI
+xASL_delete(pathNewJson);
+
 if bSaveJson
     % a. First, we load the reference sidecar JSON file, if it exists
     if bCopyOrigJson && xASL_exist(pathOrigJson, 'file')
         json = xASL_io_ReadJson(pathOrigJson);
     elseif bCopyOrigJson
         warning([pathOrigJson ' did not exist']);
+        json = struct;
     else
-        json = struct([]);
+        json = struct;
     end
 
     % b. Second, we add any provided JSON fields
--- a/External/SPMmodified/xASL/xASL_io_SaveNifti.m
+++ b/External/SPMmodified/xASL/xASL_io_SaveNifti.m
@@ -111,13 +111,13 @@ pathOrigJson = fullfile(PathOri, [NameOri '.json']);
 
 % Create temporary name for new NIFTI, since if pathOrigNifti & pathNewNifti
 % are the same, this will work better
-tempName = [pathNewNifti(1:end-4) '_temp.nii'];
-tempMat = [pathNewNifti(1:end-4) '_temp.mat'];
-newMat = [pathNewNifti(1:end-4) '.mat'];
+tempName = fullfile(PathNew, [NameNew '_temp.nii']);
+tempMat = fullfile(PathNew, [NameNew '_temp.mat']);
+newMat = fullfile(PathNew, [NameNew '.mat']);
 
 % Remove temp files in case they exist from a previous crash
 if xASL_exist(tempName)
-    warning(['Temporary file already existed, removing: ' tempName]);
+    warning(['Temporary file already existed, perhaps from a previous crash? Removing: ' tempName]);
     xASL_delete(tempName);
     xASL_delete(tempMat);
 end
@@ -221,6 +221,10 @@ newNifti.dat(:,:,:,:,:) = imNew;
 
 %% ====================================================================================
 %% 4. Remove redundant .mat orientation files
+% Start with removing any pre-existing destination .mat files
+% Even if we don't create a new one, we don't want the wrong .mat motion orientation file with a new NIfTI
+xASL_delete(newMat);
+
 if size(imNew,4)==1
     xASL_delete(tempMat);
 end
@@ -253,7 +257,7 @@ end
 xASL_Move(tempName,pathNewNifti,1,0);
 
 if exist(tempMat, 'file')
-    xASL_Move(tempMat,newMat,1,0);
+    xASL_Move(tempMat, newMat, 1, 0);
 end
 
 
--- a/External/SPMmodified/xASL/xASL_io_SaveNifti.m
+++ b/External/SPMmodified/xASL/xASL_io_SaveNifti.m
@@ -1,4 +1,4 @@
-function xASL_io_SaveNifti(pathOrigNifti, pathNewNifti, imNew, nBits, bGZip, changeMat, bSaveJson, JsonFields)
+function xASL_io_SaveNifti(pathOrigNifti, pathNewNifti, imNew, nBits, bGZip, changeMat, bCopyOrigJson, JsonFields)
 % Save a file to a Nifti format, while taking the parameters from another file
 %
 % FORMAT: xASL_io_SaveNifti(pathOrigNifti, pathNewNifti, imNew[, nBits, bGZip, newMat])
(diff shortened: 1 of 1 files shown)


External/SPMmodified/xASL/xASL_io_SaveNifti.m
```diff
@@ -16,6 +16,9 @@ function xASL_io_SaveNifti(pathOrigNifti, pathNewNifti, imNew, nBits, bGZip, cha
 %                  masks, since it cannot contain a large data range
 %   bGZip          Gzip the result (OPTIONAL, DEFAULT 1)
 %   changeMat      New orientation matrix 4x4 (OPTIONAL, DEFAULT same as previous)
+%   bSaveJson      Also save a JSON sidecar file with the NIfTI (OPTIONAL, DEFAULT = false)
```
jan-petr: This is not really about activating the saving, but about using the reference JSON. We should adapt the description.

The trouble is - what happens if we say bSaveJson == false and JsonFields = struct. I.e. we provide fields but say not to save them. This should throw an error?

I'd say that it's easier then to save new fields without loading the old JSON. This is then the option we don't really need, but it makes the input options more easy to use. Otherwise we would have two input options that could be in conflict and that's never good.

jan-petr: bSaveJson isn't about saving JSON sidecar, but about taking the JSON from pathOrigNifti.

External/SPMmodified/xASL/xASL_io_SaveNifti.m
```diff
@@ -16,6 +16,9 @@ function xASL_io_SaveNifti(pathOrigNifti, pathNewNifti, imNew, nBits, bGZip, cha
 %                  masks, since it cannot contain a large data range
 %   bGZip          Gzip the result (OPTIONAL, DEFAULT 1)
 %   changeMat      New orientation matrix 4x4 (OPTIONAL, DEFAULT same as previous)
+%   bSaveJson      Also save a JSON sidecar file with the NIfTI (OPTIONAL, DEFAULT = false)
+%   JsonFields     Additional JsonFields to be added to the new JSON sidecar, by default
```
jan-petr: you say that default is pathOriginNifti.json, but then you say DEFAULT = empty. 

Actually, this default should go to bSaveJson description.

But see the comment above. Note that I agree that option 2 is not necessary, it's just that the current implementation might be a bit confusing to the user.

External/SPMmodified/xASL/xASL_io_SaveNifti.m
```diff
@@ -60,18 +63,36 @@ function xASL_io_SaveNifti(pathOrigNifti, pathNewNifti, imNew, nBits, bGZip, cha
 	end
 end
 
+if nargin < 7 || isempty(bSaveJson)
+    bSaveJson = false;
+end
+
+if nargin < 8
```
jan-petr: I understand that you didn't add isempty. But actually, we should add || isempty as well. Cause if somebody provides [], then we should still replace that by an empty struct.

External/SPMmodified/xASL/xASL_io_SaveNifti.m
```diff
@@ -60,18 +63,36 @@ function xASL_io_SaveNifti(pathOrigNifti, pathNewNifti, imNew, nBits, bGZip, cha
...
-[pathstr, name0, ext0] = fileparts(pathOrigNifti);
-if isempty(pathstr)
+[PathOri, NameOri, ExtOri] = xASL_fileparts(pathOrigNifti);
+if isempty(PathOri)
+	% If a file-name only, then add the full current path to avoid ambiguity
+	pathOrigNifti = fullfile(pwd(), [NameOri ExtOri]);
+end
+
+pathOrigJson = fullfile(PathOri, [NameOri '.json']);
+pathNewJson = fullfile(PathOri, [NameNew '.json']);
```
jan-petr: PathOri -> PathNew

External/SPMmodified/xASL/xASL_io_SaveNifti.m
```diff
@@ -60,18 +63,36 @@ function xASL_io_SaveNifti(pathOrigNifti, pathNewNifti, imNew, nBits, bGZip, cha
...
+
+[PathOri, NameOri, ExtOri] = xASL_fileparts(pathOrigNifti);
```
jan-petr: Is this duplicate by mistake? Lines 92-95 seem equivalent to 83-87

External/SPMmodified/xASL/xASL_io_SaveNifti.m
```diff
@@ -234,4 +263,25 @@ function xASL_io_SaveNifti(pathOrigNifti, pathNewNifti, imNew, nBits, bGZip, cha
 end
 
 
+%% ====================================================================================
+%% 8. Create JSON sidecar
+if bSaveJson
```
jan-petr, line 289: Related to my comments above. This might change. Cause bSaveJson shjould be about loading the reference, not about saving. If we don't load reference, but still provide fields, we should anyway save....

jan-petr (changes requested):
A few small things.

External/SPMmodified/xASL/xASL_io_SaveNifti.m
HenkMutsaerts: Fixed in a768f82c2f482e9cdba3e7859302cf4b87dc2de3

External/SPMmodified/xASL/xASL_io_SaveNifti.m
HenkMutsaerts: Fixed in a768f82c2f482e9cdba3e7859302cf4b87dc2de3

External/SPMmodified/xASL/xASL_io_SaveNifti.m
HenkMutsaerts: Fixed in a768f82c2f482e9cdba3e7859302cf4b87dc2de3

External/SPMmodified/xASL/xASL_io_SaveNifti.m
HenkMutsaerts: Fixed in a768f82c2f482e9cdba3e7859302cf4b87dc2de3

External/SPMmodified/xASL/xASL_io_SaveNifti.m
HenkMutsaerts: Fixed in a768f82c2f482e9cdba3e7859302cf4b87dc2de3

External/SPMmodified/xASL/xASL_io_SaveNifti.m
HenkMutsaerts: Fixed in a768f82c2f482e9cdba3e7859302cf4b87dc2de3

External/SPMmodified/xASL/xASL_io_SaveNifti.m
```diff
@@ -234,4 +277,28 @@ function xASL_io_SaveNifti(pathOrigNifti, pathNewNifti, imNew, nBits, bGZip, cha
...
+%% ====================================================================================
+%% 8. Create JSON sidecar
+if bSaveJson
+    % a. First, we load the reference sidecar JSON file, if it exists
+    if bCopyOrigJson && xASL_exist(pathOrigJson, 'file')
+        json = xASL_io_ReadJson(pathOrigJson);
+    elseif bCopyOrigJson
+        warning([pathOrigJson ' did not exist']);
+    else
+        json = struct([]);
```
jan-petr: This won't work. 
try 
json = struct([])
json.test = 1
you have to change this back to struct or struct()

External/SPMmodified/xASL/xASL_io_SaveNifti.m
```diff
@@ -1,4 +1,4 @@
-function xASL_io_SaveNifti(pathOrigNifti, pathNewNifti, imNew, nBits, bGZip, changeMat)
+function xASL_io_SaveNifti(pathOrigNifti, pathNewNifti, imNew, nBits, bGZip, changeMat, bCopyOrigJson, JsonFields)
 % Save a file to a Nifti format, while taking the parameters from another file
 %
 % FORMAT: xASL_io_SaveNifti(pathOrigNifti, pathNewNifti, imNew[, nBits, bGZip, newMat])
```
jan-petr: Please update the FORMAT by adding the new parameters.
Note also that newMat -> changeMat

External/SPMmodified/xASL/xASL_io_SaveNifti.m
```diff
@@ -16,6 +16,16 @@ function xASL_io_SaveNifti(pathOrigNifti, pathNewNifti, imNew, nBits, bGZip, cha
 %                  masks, since it cannot contain a large data range
 %   bGZip          Gzip the result (OPTIONAL, DEFAULT 1)
 %   changeMat      New orientation matrix 4x4 (OPTIONAL, DEFAULT same as previous)
+%   bCopyOrigJson  Copy a JSON sidecar file (if existed for the original file) as sidecar of the new NIfTI (OPTIONAL, DEFAULT = false)
+%   JsonFields     Save additional JsonFields to a new JSON sidecar (or merged with the original Json content)
+%                  These fields overwrite fields with the same fieldnames in the original Json
+%                  (OPTIONAL, DEFAULT = struct() empty)
+% Options are:
+% 1. bSaveJson     = false                       -> we don't save a new JSON sidecar
```
jan-petr: please remove line bSaveJson. It's not an input parameter anymore.

External/SPMmodified/xASL/xASL_io_SaveNifti.m
```diff
@@ -60,20 +70,45 @@ function xASL_io_SaveNifti(pathOrigNifti, pathNewNifti, imNew, nBits, bGZip, cha
 	end
 end
 
+% JSON saving
```
jan-petr: That's a copy of what we have above. We can delete this. But we can of course introduce the bSaveJson variable.

External/SPMmodified/xASL/xASL_io_SaveNifti.m
```diff
@@ -60,20 +70,45 @@ function xASL_io_SaveNifti(pathOrigNifti, pathNewNifti, imNew, nBits, bGZip, cha
...
+% 2. bCopyOrigJson = true  & JsonFields = empty  -> we copy the original JSON sidecar only
+% 3. bCopyOrigJson = false & JsonFields = filled -> we create a new JSON sidecar with new fields only
+% 4. bCopyOrigJson = true  & JsonFields = filled -> we merge the original JSON sidecar with the new fields
+%                                                   the new fields have priority
+bSaveJson = false;
+
+if nargin < 7 || isempty(bCopyOrigJson)
+    bCopyOrigJson = false;
+else
+    bSaveJson = true;
```
jan-petr: That won't work. If you provide bCopyOrigJson == false, it will locate it and set bSaveJson = true :)

External/SPMmodified/xASL/xASL_io_SaveNifti.m
```diff
@@ -60,20 +70,45 @@ function xASL_io_SaveNifti(pathOrigNifti, pathNewNifti, imNew, nBits, bGZip, cha
...
+end
+
+if nargin < 8 || isempty(JsonFields)
+    JsonFields = struct([]);
```
jan-petr: struct([]) -> struct().
Cause this commands as it is creates an array and you can't add new fields to it - see also below.

External/SPMmodified/xASL/xASL_io_SaveNifti.m
```diff
@@ -234,4 +277,28 @@ function xASL_io_SaveNifti(pathOrigNifti, pathNewNifti, imNew, nBits, bGZip, cha
...
 
 
+%% ====================================================================================
+%% 8. Create JSON sidecar
+if bSaveJson
+    % a. First, we load the reference sidecar JSON file, if it exists
+    if bCopyOrigJson && xASL_exist(pathOrigJson, 'file')
+        json = xASL_io_ReadJson(pathOrigJson);
+    elseif bCopyOrigJson
+        warning([pathOrigJson ' did not exist']);
```
jan-petr, line 294: it will not really crash. but it would be better to also do json=struct() here

jan-petr (changes requested):
A few bugs

External/SPMmodified/xASL/xASL_io_SaveNifti.m
HenkMutsaerts: Fixed in f4672c87b7ea7fd1a9295d8a555b38979c718e09

External/SPMmodified/xASL/xASL_io_SaveNifti.m
HenkMutsaerts: Fixed in f4672c87b7ea7fd1a9295d8a555b38979c718e09

External/SPMmodified/xASL/xASL_io_SaveNifti.m
HenkMutsaerts: I don't get your comment. Are you sure?

External/SPMmodified/xASL/xASL_io_SaveNifti.m
HenkMutsaerts: Fixed in f4672c87b7ea7fd1a9295d8a555b38979c718e09

External/SPMmodified/xASL/xASL_io_SaveNifti.m
HenkMutsaerts: Fixed in f4672c87b7ea7fd1a9295d8a555b38979c718e09

External/SPMmodified/xASL/xASL_io_SaveNifti.m
HenkMutsaerts: Fixed in f4672c87b7ea7fd1a9295d8a555b38979c718e09

External/SPMmodified/xASL/xASL_io_SaveNifti.m
HenkMutsaerts: Fixed in f4672c87b7ea7fd1a9295d8a555b38979c718e09

External/SPMmodified/xASL/xASL_io_SaveNifti.m
jan-petr: You set the input parameter bCopyOrigJson == false, and JsonFields == [].
So at line 82, it will detect that bCopyOrigJson was assigned, so it will just to line 84 and 85 and set bSaveJson = true.

So despite bCopyOrigJson == false, it will attempt to save the JSON.
But of course later on - it will not load the reference JSON and you provided empty fields, so it will save an empty JSON.

The solution is to change line 84 to elseif bCopyOrigJson

HenkMutsaerts: Fixed in cd45719d06c30b9ccc2ef705ac80dc4e37562912

jan-petr (approved):
OK

PR #1589, opened 2024-01-09
Closes #1587 delete temporary files incompleted modules

### LINKED ISSUE #1587: Avoid deleting temporary files for not-completed modules
### Description

When the population module runs, at the end it will remove all temporal files irrespective of these files still being needed for reruns of incomplete modules.

### Tasks

- [x] Create a function that obtains a list of complete modules
- [x] run xASL_adm_DeleteManyTempFiles only for these completed modules

### How to test

Take four subjects with two sessions each, and run some modules fully, other modules break them halfway.
Then see if the temporary files are indeed removed only for those with the completed modules.

### Release notes

Report with #1560. Specify that we remove only if ASL module was completed.

HenkMutsaerts:
### Linked issue

Closes #1587

### DIFF
--- a/Functions/Administration/xASL_adm_GetListCompletedModules.m
+++ b/Functions/Administration/xASL_adm_GetListCompletedModules.m
@@ -1,0 +1,53 @@
+function [listStructDone, listASLDone] = xASL_adm_GetListCompletedModules(x)
+%xASL_adm_GetListCompletedModules This function obtains a list of completed modules
+%
+% FORMAT:       xASL_adm_GetListCompletedModules(x)
+% 
+% INPUT:        x               - x structure of ExploreASL
+%
+% OUTPUT:       listStructDone  - list of subjects for which the structural module is completed
+%               listASLDone     - list of sessions/runs for which the ASL module is completed
+%
+% 
+% -----------------------------------------------------------------------------------------------------------------------------------------------------
+% DESCRIPTION:  For both structural and ASL modules, we get a list of subject/sessions of which these modules are fully completed.
+%               This can be useful for subsequent group processing in e.g., the ExploreASL population module.
+%
+% -----------------------------------------------------------------------------------------------------------------------------------------------------
+% EXAMPLE:      xASL_adm_GetListCompletedModules(x);
+% __________________________________
+% Copyright 2015-2024 ExploreASL
+
+%% First, we create a list of subjects and sessions of which the structural and ASL modules, respectively, are completed
+
+dirLockRoot = fullfile(x.dir.xASLDerivatives, 'lock');
+dirLockStruct = fullfile(dirLockRoot, 'xASL_module_Structural');
+dirLockASL = fullfile(dirLockRoot, 'xASL_module_ASL');
+
+listStructSubjects = xASL_adm_GetFileList(dirLockStruct, '.*', 'List', [], 1);
+listStructDone = {};
+
+for iList = 1:length(listStructSubjects)
+    currentDir = fullfile(dirLockStruct, listStructSubjects{iList});
+    lockFile999 = fullfile(currentDir, 'xASL_module_Structural', '999_ready.status');
+    if exist(lockFile999, 'file')
+        listStructDone{end+1} = listStructSubjects{iList};
+    end
+end
+
+listASLSubjects = xASL_adm_GetFileList(dirLockASL, '.*', 'List', [], 1);
+listASLDone = {};
+for iList = 1:length(listASLSubjects)
+    currentDir = fullfile(dirLockASL, listASLSubjects{iList});
+    sessions = xASL_adm_GetFileList(currentDir, '^xASL_module_ASL_.*', 'List', [], 1);
+    for iSession=1:length(sessions)
+        currentDir = fullfile(dirLockASL, listASLSubjects{iList}, sessions{iSession});
+        lockFile999 = fullfile(currentDir, '999_ready.status');
+        if exist(lockFile999, 'file')
+            listASLDone{end+1} = [listASLSubjects{iList} '_' sessions{iSession}(length('xASL_module_ASL_')+1:end)];
+        end
+    end
+end
+
+
+end
--- a/Functions/Administration/xASL_adm_GetListCompletedModules.m
+++ b/Functions/Administration/xASL_adm_GetListCompletedModules.m
@@ -1,7 +1,7 @@
 function [listStructDone, listASLDone] = xASL_adm_GetListCompletedModules(x)
 %xASL_adm_GetListCompletedModules This function obtains a list of completed modules
 %
-% FORMAT:       xASL_adm_GetListCompletedModules(x)
+% FORMAT:       [listStructDone, listASLDone] = xASL_adm_GetListCompletedModules(x)
 % 
 % INPUT:        x               - x structure of ExploreASL
 %
@@ -14,7 +14,7 @@ function [listStructDone, listASLDone] = xASL_adm_GetListCompletedModules(x)
 %               This can be useful for subsequent group processing in e.g., the ExploreASL population module.
 %
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
-% EXAMPLE:      xASL_adm_GetListCompletedModules(x);
+% EXAMPLE:      [listStructDone, listASLDone] = xASL_adm_GetListCompletedModules(x);
 % __________________________________
 % Copyright 2015-2024 ExploreASL
 
@@ -28,8 +28,7 @@ listStructSubjects = xASL_adm_GetFileList(dirLockStruct, '.*', 'List', [], 1);
 listStructDone = {};
 
 for iList = 1:length(listStructSubjects)
-    currentDir = fullfile(dirLockStruct, listStructSubjects{iList});
-    lockFile999 = fullfile(currentDir, 'xASL_module_Structural', '999_ready.status');
+    lockFile999 = fullfile(dirLockStruct, listStructSubjects{iList}, 'xASL_module_Structural', '999_ready.status');
     if exist(lockFile999, 'file')
         listStructDone{end+1} = listStructSubjects{iList};
     end
@@ -38,11 +37,9 @@ end
 listASLSubjects = xASL_adm_GetFileList(dirLockASL, '.*', 'List', [], 1);
 listASLDone = {};
 for iList = 1:length(listASLSubjects)
-    currentDir = fullfile(dirLockASL, listASLSubjects{iList});
-    sessions = xASL_adm_GetFileList(currentDir, '^xASL_module_ASL_.*', 'List', [], 1);
+    sessions = xASL_adm_GetFileList(fullfile(dirLockASL, listASLSubjects{iList}, '^xASL_module_ASL_.*', 'List', [], 1);
     for iSession=1:length(sessions)
-        currentDir = fullfile(dirLockASL, listASLSubjects{iList}, sessions{iSession});
-        lockFile999 = fullfile(currentDir, '999_ready.status');
+        lockFile999 = fullfile(dirLockASL, listASLSubjects{iList}, sessions{iSession}, '999_ready.status');
         if exist(lockFile999, 'file')
             listASLDone{end+1} = [listASLSubjects{iList} '_' sessions{iSession}(length('xASL_module_ASL_')+1:end)];
         end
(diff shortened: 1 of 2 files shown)


Functions/Administration/xASL_adm_GetListCompletedModules.m
```diff
@@ -0,0 +1,53 @@
+function [listStructDone, listASLDone] = xASL_adm_GetListCompletedModules(x)
+%xASL_adm_GetListCompletedModules This function obtains a list of completed modules
+%
+% FORMAT:       xASL_adm_GetListCompletedModules(x)
```
jan-petr: add output parameters

Functions/Administration/xASL_adm_GetListCompletedModules.m
```diff
@@ -0,0 +1,53 @@
...
+% OUTPUT:       listStructDone  - list of subjects for which the structural module is completed
+%               listASLDone     - list of sessions/runs for which the ASL module is completed
+%
+% 
+% -----------------------------------------------------------------------------------------------------------------------------------------------------
+% DESCRIPTION:  For both structural and ASL modules, we get a list of subject/sessions of which these modules are fully completed.
+%               This can be useful for subsequent group processing in e.g., the ExploreASL population module.
+%
+% -----------------------------------------------------------------------------------------------------------------------------------------------------
+% EXAMPLE:      xASL_adm_GetListCompletedModules(x);
```
jan-petr: add output

Functions/Administration/xASL_adm_GetListCompletedModules.m
```diff
@@ -0,0 +1,53 @@
...
+dirLockRoot = fullfile(x.dir.xASLDerivatives, 'lock');
+dirLockStruct = fullfile(dirLockRoot, 'xASL_module_Structural');
+dirLockASL = fullfile(dirLockRoot, 'xASL_module_ASL');
+
+listStructSubjects = xASL_adm_GetFileList(dirLockStruct, '.*', 'List', [], 1);
+listStructDone = {};
+
+for iList = 1:length(listStructSubjects)
+    currentDir = fullfile(dirLockStruct, listStructSubjects{iList});
+    lockFile999 = fullfile(currentDir, 'xASL_module_Structural', '999_ready.status');
```
jan-petr: 31+32 can really go to a single line.

Functions/Administration/xASL_adm_GetListCompletedModules.m
```diff
@@ -0,0 +1,53 @@
...
+end
+
+listASLSubjects = xASL_adm_GetFileList(dirLockASL, '.*', 'List', [], 1);
+listASLDone = {};
+for iList = 1:length(listASLSubjects)
+    currentDir = fullfile(dirLockASL, listASLSubjects{iList});
+    sessions = xASL_adm_GetFileList(currentDir, '^xASL_module_ASL_.*', 'List', [], 1);
+    for iSession=1:length(sessions)
+        currentDir = fullfile(dirLockASL, listASLSubjects{iList}, sessions{iSession});
+        lockFile999 = fullfile(currentDir, '999_ready.status');
```
jan-petr: merge to a single line.

Functions/Administration/xASL_adm_GetListCompletedModules.m
```diff
@@ -0,0 +1,53 @@
...
+    lockFile999 = fullfile(currentDir, 'xASL_module_Structural', '999_ready.status');
+    if exist(lockFile999, 'file')
+        listStructDone{end+1} = listStructSubjects{iList};
+    end
+end
+
+listASLSubjects = xASL_adm_GetFileList(dirLockASL, '.*', 'List', [], 1);
+listASLDone = {};
+for iList = 1:length(listASLSubjects)
+    currentDir = fullfile(dirLockASL, listASLSubjects{iList});
```
jan-petr: merge to a single line

Modules/Module_Processing/Module_Population/xASL_adm_DeleteManyTempFiles.m
```diff
@@ -13,34 +13,76 @@ function xASL_adm_DeleteManyTempFiles(x)
...
+% Copyright 2015-2024 ExploreASL
 
-Files2Del = {'ATT_BiasField.nii' 'Mask_Template.nii' 'Mean_CBF_Template.nii' 'PseudoCBF.nii' 'RawTemplate.nii' 'VascularArtifact_Template.nii' 'mean_PWI_Clipped.nii' 'SliceGradient_extrapolated.nii' 'FoV.nii'};
+Files2DelNativeStructural = {}; % (currently no files here)
+Files2DelNativeASL = {'ATT_BiasField.nii' 'Mask_Template.nii' 'Mean_CBF_Template.nii' 'PseudoCBF.nii' 'RawTemplate.nii' 'VascularArtifact_Template.nii' 'mean_PWI_Clipped.nii' 'SliceGradient_extrapolated.nii' 'FoV.nii'};
+Files2DelStandardStructural = {'rT1_ORI_' 'rT1_' 'rFLAIR_' '(m|)rc\dT1_'};
+Files2DelStandardASL = {'noSmooth_M0_' 'mean_control_' 'PWI_' 'SliceGradient_' 'SNR'};
+
+%% First, we create a list of subjects and sessions of which the structural and ASL modules, respectively, are completed
+[listSubjects, listSessions] = xASL_adm_GetListCompletedModules(x);
```
jan-petr: In the fucntion you call these struct and ASL list. Here you call it differently. That's quite confusing.

Modules/Module_Processing/Module_Population/xASL_adm_DeleteManyTempFiles.m
```diff
@@ -13,34 +13,76 @@ function xASL_adm_DeleteManyTempFiles(x)
...
 
 if isfield(x,'D') && isfield(x.dir,'xASLDerivatives')
```
jan-petr: WHy checking for D and then looking into x.dir -> an old mistake?

Modules/Module_Processing/Module_Population/xASL_adm_DeleteManyTempFiles.m
```diff
@@ -13,34 +13,76 @@ function xASL_adm_DeleteManyTempFiles(x)
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
 % EXAMPLE:      xASL_adm_DeleteManyTempFiles(x);
 % __________________________________
-% Copyright 2015-2023 ExploreASL
+% Copyright 2015-2024 ExploreASL
 
-Files2Del = {'ATT_BiasField.nii' 'Mask_Template.nii' 'Mean_CBF_Template.nii' 'PseudoCBF.nii' 'RawTemplate.nii' 'VascularArtifact_Template.nii' 'mean_PWI_Clipped.nii' 'SliceGradient_extrapolated.nii' 'FoV.nii'};
+Files2DelNativeStructural = {}; % (currently no files here)
```
jan-petr, line 22: make a note that upon adding files, you have to update the code below.

Modules/Module_Processing/Module_Population/xASL_adm_DeleteManyTempFiles.m
```diff
@@ -13,34 +13,76 @@ function xASL_adm_DeleteManyTempFiles(x)
...
+    fprintf('Deleting temporary ASL files in native space folders:    ');
+
+    for iP=1:length(Files2DelNativeASL)
+        xASL_TrackProgress(iP, length(Files2DelNativeASL));
         
-        PathList = xASL_adm_GetFileList(x.dir.xASLDerivatives,['^' File2Search Fext '$'],'FPListRec',[0 Inf]);
-        for iL=1:length(PathList)
-            xASL_delete(PathList{iL});
+        for iSession=1:length(listSessions)
+            sessionIndex = regexp(listSessions{iSession}, 'ASL_\d*');
```
jan-petr: Note that this might work badly when there's ASL in the subject name - which is not really excluded. So we have to look for ASL_\d at the end of the string.

Modules/Module_Processing/Module_Population/xASL_adm_DeleteManyTempFiles.m
```diff
@@ -13,34 +13,76 @@ function xASL_adm_DeleteManyTempFiles(x)
...
+
+%% 3. Standard space — structural module files
+fprintf('Deleting temporary structural files in population folder:    ');
+
+for iPath=1:length(Files2DelStandardStructural)
+    xASL_TrackProgress(iP,length(Files2DelStandardStructural));
+
+    for iSubject=1:length(listSubjects)
+        % Delete any existing files
+        xASL_adm_DeleteFileList(x.D.PopDir, ['^' Files2DelStandardStructural{iPath} listSubjects{iSubject} '.*$'], 1, [0 Inf]);
```
jan-petr: the second parameter is a regexp, right? so that will work nicely, right?

jan-petr (changes requested):
minor fixes

Modules/Module_Processing/Module_Population/xASL_adm_DeleteManyTempFiles.m
```diff
@@ -13,34 +13,76 @@ function xASL_adm_DeleteManyTempFiles(x)
...
 % EXAMPLE:      xASL_adm_DeleteManyTempFiles(x);
 % __________________________________
-% Copyright 2015-2023 ExploreASL
+% Copyright 2015-2024 ExploreASL
 
-Files2Del = {'ATT_BiasField.nii' 'Mask_Template.nii' 'Mean_CBF_Template.nii' 'PseudoCBF.nii' 'RawTemplate.nii' 'VascularArtifact_Template.nii' 'mean_PWI_Clipped.nii' 'SliceGradient_extrapolated.nii' 'FoV.nii'};
+Files2DelNativeStructural = {}; % (currently no files here)
+Files2DelNativeASL = {'ATT_BiasField.nii' 'Mask_Template.nii' 'Mean_CBF_Template.nii' 'PseudoCBF.nii' 'RawTemplate.nii' 'VascularArtifact_Template.nii' 'mean_PWI_Clipped.nii' 'SliceGradient_extrapolated.nii' 'FoV.nii'};
+Files2DelStandardStructural = {'rT1_ORI_' 'rT1_' 'rFLAIR_' '(m|)rc\dT1_'};
+Files2DelStandardASL = {'noSmooth_M0_' 'mean_control_' 'PWI_' 'SliceGradient_' 'SNR'};
```
jan-petr, line 26: If we delete PWI after completing the entire ASL module. Then if we try to re-run ASL module from the "CreateMask" part, it crashes as it needs PWI. Do we really need to delete PWI?

jan-petr (changes requested):
One more comment.

Modules/Module_Processing/Module_Population/xASL_adm_DeleteManyTempFiles.m
HenkMutsaerts: Yes, xASL_adm_DeleteFileList has similar input variables as xASL_admGetFileList

Modules/Module_Processing/Module_Population/xASL_adm_DeleteManyTempFiles.m
HenkMutsaerts: We don't need to delete anything, but the more we delete the more data we save. We don't delete the native space PWI, only the standard space PWI.

The only reason I could think of, is that we may create the standard space PWI from ASL4D warped to standard space. If we re-create PWI by transforming the native space PWI, we may get more interpolation artifacts.

jan-petr: When creating individual analysis mask in ASL module in standard space, we are requiring to have this PWI in standard space. 

So if we re-run ASL module and remove lock files for 070_Create analysis mask, then it will crash, so we either stop deleting this PWI, or we have to re-run a bigger part of the ASL module.

HenkMutsaerts: This is the whole point why we now only delete such "temporary files" for completed modules. You give an example that you want for an incompleted module. So this situation will not occur. Unless you want to rerun part of a module, this is currently only possible if you haven't ran the population module.

I would keep it this way. We have the following cases in which we run the full population module
1. a module is not completed yet, so we don't delete the file (this solves your issue)
2. a module is fully completed, so we delete the file (you cannot redo half of a module, you always have to redo the whole module. unless you avoid running the population module.) Note that this is the case for many things. Even without deleting temporary files, you cannot arbitrary remove a few lock files and rerun a module; we have dependencies within a module. In fact we solve this for the case that you have only the first 2 lock files of a module done, if you then rerun a module it will also delete these first 2 lock files and repeat the full module.

3. If you are an advanced user and you want to repeat a module partly (e.g. for comparing different settings), you just don't run the population yet. The population module is the "end/finalizing" module.

jan-petr: I mostly talked about the scenario of an advanced user re-running half of a module after removing lock-files. But I agree that people shouldn't do this. So all good :D I'm resolving this and merging!

jan-petr (approved):
OK.

PR #1602, opened 2024-01-11
#1601 Avoid crash with missing parameters

### LINKED ISSUE #1601: Follow-up on 1211
### Description

After loading the ASL.json inside xASL_modules_ASL, the EchoTime was in x.EchoTime and the new function was looking for it inside x.Q.EchoTime.

We're now moving it there automatically by adding to the loading script that EchoTime belongs to the x.Q subfield.

Take this dataset as a testing case Philips_3DGRASE_MultiSessions_T1wAlreadyDone_LoQ

### Tasks

- [x] In Philips_3DGRASE_MultiSessions_T1wAlreadyDone_LoQ, ASL.JSON is somehow removed while executing. This was because saving to the same image as the reference image removed the JSON.
- [x] PWI3D and PWI4D.JSONs are all saved in Legacy format (ms) instead of BIDS format. Reading and writing of all JSONs have to go through our special conversion function.
- [x] Eventually, this JSON loading and saving messes up some TE, PLD or other parameters and screws up the quantification. To be fixed....

### How to test

- [x] Run TestDatSet on both Philips_3DGRASE datasets which were the two ones crashing due to this error.

### Release notes

Nothing. Report with #1585.

jan-petr:
### Linked issue

Close #1601


Modules/Module_Processing/Module_ASL/xASL_wrp_Quantify.m
```diff
@@ -162,13 +161,13 @@ function xASL_wrp_Quantify(x, PWI4D_Path, pathOutputCBF, M0Path, SliceGradientPa
 if ~exist(pathSidecar, 'file')
     warning('PWI4D JSON sidecar with quantification parameters was missing');
 else
-    JSON = spm_jsonread(pathSidecar);
-    x.Q.EchoTime_PWI4D = JSON.EchoTime_PWI4D;
-    x.Q.InitialPLD_PWI4D = JSON.InitialPLD_PWI4D;
-    x.Q.LabelingDuration_PWI4D = JSON.LabelingDuration_PWI4D;
+	JSON = xASL_io_ReadJson(pathSidecar); % Read the sidecar
```
HenkMutsaerts, line 164: This has to be moved elsewhere, as it is a new function, not a bugfix.

Modules/Module_Processing/Module_ASL/xASL_wrp_ResampleASL.m
```diff
@@ -246,19 +246,24 @@ function xASL_wrp_ResampleASL(x)
 
     % Save PWI4D (subtracted only)
     fprintf('%s\n', PathPWI4D{iSpace});
-    xASL_io_SaveNifti(PathASL4D{iSpace}, PathPWI4D{iSpace}, PWI4D, 32, false);
-    spm_jsonwrite(PathPWI4Djson{iSpace}, x.Q);
+	jsonFields.EchoTime = x.Q.EchoTime_PWI4D;
```
HenkMutsaerts, line 249: This is incorrect, we should save PWI4D as PWI4D, PWI3D as PWI3D, PWI as PWI A single JSON may have all three. Now you are renaming them, that doesn't work. Plus, there is no BIDS standard for derivatives yet, so moving to BIDS rawdata standard doesn't make sense for derivatives necessarily.

Modules/Module_Processing/Module_ASL/xASL_wrp_ResampleASL.m
```diff
@@ -246,19 +246,24 @@ function xASL_wrp_ResampleASL(x)
 
     % Save PWI4D (subtracted only)
     fprintf('%s\n', PathPWI4D{iSpace});
-    xASL_io_SaveNifti(PathASL4D{iSpace}, PathPWI4D{iSpace}, PWI4D, 32, false);
-    spm_jsonwrite(PathPWI4Djson{iSpace}, x.Q);
+	jsonFields.EchoTime = x.Q.EchoTime_PWI4D;
+	jsonFields.Initial_PLD = x.Q.InitialPLD_PWI4D;
+	jsonFields.LabelingDuration = x.Q.LabelingDuration_PWI4D;
+    xASL_io_SaveNifti(PathASL4D{iSpace}, PathPWI4D{iSpace}, PWI4D, 32, false, [], 0, xASL_bids_parms2BIDS(jsonFields, [], 1));
```
HenkMutsaerts, line 252: Why this parms2BIDS?

Modules/xASL_module_ASL.m
```diff
@@ -566,6 +566,8 @@
         warning(['Missing field x.Q.' parNames{iPar} ', this may be needed for quantification']);
         % After the warning, we default to single-parameter processing
         fprintf('%s\n', ['Defaulting to single-' parAbbreviation{iPar} ' ASL processing']);
+		x.Q.(parNames{iPar}) = [];
```
HenkMutsaerts, line 569: This should be illegal, every ASL flavor should have EchoTime, PLD, LD, because we will use them later in quantification. So we might as well make this an error instead of warning?

HenkMutsaerts (changes requested):
Some questions.

Especially, you squeeze a functionality change (loading & saving sidecars) into a "large bugfix" commit ;) That should go outside, or at least get the major attention. 

Better to hide a bugfix in a functionality change, which we can later find, then the other way around.

There are probably more locations for JSON sidecar loading and saving, and we should do this properly in an issue.

If you really want it here, I would rename this issue into JSON sidecar loading & saving, and we implement this for all cases throughout xASL_module_ASL.

HenkMutsaerts (approved):

PR #1604, opened 2024-01-12
Closes #1539 xASL_im_ASLSubtractionAveraging: weighted-average PWI3D -> PWI

### LINKED ISSUE #1539: Weighted-average PWI for improved registration
### Description

Currently, the PWI is the average of all PLDs, labeling durations, for the first echo time only.
Instead, I propose that we do a weighted sum.
So, we compute the SNR contribution of the TEs, which is an exponential contribution.

Likewise, we can compute the SNR contribution based on PLD and LD, but the location of the label needs to be taken into account so we could simply do a linear weighting where the earliest PLD counts the lowest and the latest PLD the most.

How to compute single PWI for registration purposes.
For this we still need to reach consensus. **JAN** Agree with 5 as the best option. And use 1-4 as input for the weighting.
1) we take longest PLD because it has most CBF weighting **JAN** Not really, PLD=3500 is pretty bad. 
2) we take longest LD because it has most SNR **JAN** Good idea if there are multiple LDs, then we first go for the longest LD and then select PLDs. Cause longest LD will probably also have the longest PLD.
3) we take the PLD closest to 2000 ms (Look Locker example) **JAN** Yes for Look-Locker
4) We take a combination
5) We take a weighted average (my preference) **JAN** Agree

### Proposal weighted-average
The goal of PWI is a CBF registration reference, but a very noisy CBF might be worse than a higher SNR combination between vascular and CBF, etc.

based on Alsop MRM 2014
https://pubmed.ncbi.nlm.nih.gov/24715426/

nsrPLD = exp(PLD/T1b) / (1-exp(-LD/T1b)) (nsr == 1/snr)
Single-compartment model uses this assuming that the label remains in the blood compartment.

nsrTE = exp(TE/T2(*))
-> this is T2 for 3D spiral & T2* for 3D GRASE or 2D EPI.

For CBF vs vascular weighting: we could say that a PLD > 2500 ms has 100% CBF weighting (2000 for 2D EPI because of multi-slice) and anything below 1000 ms has 1% CBF weighting, and that goes linearly.

pwiPLD =  max((min(PLD, 2500) - 1000) / 15,1); % how much perfusion-weighting we have in the image

So the contribution per volume would then be:

contribution = pwiPLD / (nsrTE * nsrPLD)

This gives a contribution sum for all volumes, which we scale linearly to 100%.

### Tasks

- [x] Base this branch of feature-#1551_RevampedSubtractionTempBranch
- [x] xASL_im_ASLSubtractionAveraging: Compute a weightings vector for all volumes that sums to 1
- [x] Test with DEBBIE (@jan-petr for some reason I couldn't test my DEBBIE sequence anymore, I got bugs in xASL_quant_HadamardDecoding) - **JAN** DEBBIE needs more fixes that are already done in develop. I will retest all once merging and rebasing all.
- [x] Move quantification parameter copy to start of xASL_module_ASL (blood T1, T2, etc) as separate function
- [x] What do we do for look-locker? Couldn't we do the same? **JAN** Mathijs has his Look-Locker issue where this is done. We'll fix that in that issue.

### How to test

- [x] Jan tests with DEBBIE sequence before merging (this gave Henk `xASL_quant_HadamardDecoding` bugs so he couldn't test it) **JAN** DEBBIE shouldn't be affected by this issue. All DEBBIE errors are related to previous errors - so we need to merge+rebase all previous issues that have likely fixed all the troubles.

### Release notes

Calculate mean PWI for registration as a weighted sum.

HenkMutsaerts:
### Linked issue

Closes #1539


Functions/ImageProcessing/xASL_im_ASLSubtractionAveraging.m
```diff
@@ -419,12 +413,112 @@
...
     end
     
-    %% Current quick&dirty fix was the average of all volumes with the lowest echo time
-    iMinTE = x.Q.EchoTime_PWI3D == min(x.Q.EchoTime_PWI3D(:));
-    PWI = xASL_stat_MeanNan(PWI3D(:, :, :, iMinTE), 4);
-    x.Q.EchoTime_PWI = min(x.Q.EchoTime_PWI3D(:));
-    x.Q.initialPLD_PWI = mean(x.Q.InitialPLD_PWI3D(:));
-    x.Q.LabelingDuration_PWI = mean(x.Q.LabelingDuration_PWI3D(:));
+    %% =================================================================================
+    %% PM: THIS PART ON T2STAR & T2 CAN BE ADDED TO THE START OF xASL_module_ASL INSTEAD
```
jan-petr: If this part is an exact copy from the Quantify module, then I fully agree to move this from both places to the start of ASL-module. In case you would modify something, then please do it in a separate issue. But that's probably not needed, so a pure move - and that's a good idea

jan-petr: Please only note that some of the references in the code are given at full in the function header. So please move/copy those as well to the new subfunction.

Functions/ImageProcessing/xASL_im_ASLSubtractionAveraging.m
```diff
@@ -419,12 +413,112 @@
...
+        T2_factor = x.Q.T2star;
+    end
+
+    nsrTE = exp(x.Q.EchoTime_PWI3D ./T2_factor); % for each volume, get the EchoTime weighting
+
+    % PLD & LD weighting
+    nsrPLD = exp(x.Q.InitialPLD_PWI3D ./ x.Q.BloodT1) ./ (1-exp(-x.Q.LabelingDuration_PWI3D ./x.Q.BloodT1));
+
+    % Perfusion-weighting (PW) vs vascular weighting
+    pwPLD = max((min(x.Q.InitialPLD_PWI3D, 2500) - 1000) ./ 15, 1) ./100;
```
jan-petr, line 438: The division by 100 isn't necessary and might be confusing. This is not percentage and nsrTE and nsrPLD isn't either. 

So we can remove this without any effects...

Functions/ImageProcessing/xASL_im_ASLSubtractionAveraging.m
```diff
@@ -419,12 +413,112 @@
...
+    
+    x.Q.EchoTime_PWI = 0;
+    x.Q.initialPLD_PWI = 0;
+    x.Q.LabelingDuration_PWI = 0;
+
+    for iVolume=1:size(PWI3D, 4)
+        PWI = PWI + contributionVolume(iVolume) .* PWI3D(:,:,:,iVolume);
+    end
+
+    x.Q.EchoTime_PWI = contributionVolume .* x.Q.EchoTime_PWI3D;
```
jan-petr: Don't forget the sum here ;)

jan-petr (changes requested):
Looks good. small things to fix...

Functions/ImageProcessing/xASL_im_ASLSubtractionAveraging.m
HenkMutsaerts: I disagree. The sum of contribution should be 1, so you want a fraction for each volume. See my comments as examples. I know it doesn't matter mathematically but nicer to follow perhaps.

Fixed in 6fda911643cea1f2783ea578410f3f6f5f94beae. Do you agree?

Functions/ImageProcessing/xASL_im_ASLSubtractionAveraging.m
HenkMutsaerts: Fixed in 6fda911643cea1f2783ea578410f3f6f5f94beae

Functions/ImageProcessing/xASL_im_ASLSubtractionAveraging.m
HenkMutsaerts: Done in 85bc34b1b6c07301f8aa5e2e261ce80425421cb5

Functions/ImageProcessing/xASL_im_ASLSubtractionAveraging.m
```diff
@@ -491,15 +491,19 @@
...
 
     nsrTE = exp(x.Q.EchoTime_PWI3D ./T2_factor); % for each volume, get the EchoTime weighting
+    % e.g., exp(17/47.3) = 1.4325
 
     % PLD & LD weighting
     nsrPLD = exp(x.Q.InitialPLD_PWI3D ./ x.Q.BloodT1) ./ (1-exp(-x.Q.LabelingDuration_PWI3D ./x.Q.BloodT1));
+    % e.g., exp(1525/1650) / (1-exp(-1650/1650) = 3.9865
 
-    % Perfusion-weighting (PW) vs vascular weighting
+    % Perfusion-weighting (PW) vs vascular weighting (range 0%-100%)
```
jan-petr: range is 1-100 not 0-100

Functions/ImageProcessing/xASL_im_ASLSubtractionAveraging.m
jan-petr: Exactly - mathematically it doesn't matter. Practically, people will be confused why there's a superfluous division by 100 that makes no difference.

You are calculating a weighted sum - there's no justification for having numbers <1.

I don't care that much, but you usually want to have things clean and explained ;) I'm just saying that this explanation doesn't really show much logic to me

Functions/Quantification/xASL_quant_DefineParameters.m
```diff
@@ -0,0 +1,164 @@
+function [x] = xASL_quant_DefineParameters(x)
```
jan-petr: Is that really the best name?
Shall we call them physiological parameters or quantification parameters

jan-petr (changes requested):
minor things

Functions/ImageProcessing/xASL_im_ASLSubtractionAveraging.m
HenkMutsaerts: It makes more sense to me (from the readability perspective) if each volume has a factor (e.g. 0.06) rather than an arbitrary number higher than 1.

The minimal standard for ExploreASL are not programmers or mathematicians Jan. You sometimes cannot think outside of that perspective ;) So please trust me when I say that I agree with you that mathematically and logically, it makes no sense. But for readability, it is nice if the example has factors around 1-10 or fractions as percentages (0-100). "Perfusion-weighted contribution was calculated as a percentage of full contribution, 100% (==full contribution) defined as 2500 ms ..." "

But of course it doesn't matter for functionality.

Functions/Quantification/xASL_quant_DefineParameters.m
HenkMutsaerts: I agree, but I thought that is implied in "quant". I will change it to xASL_quant_DefineQuantificationParameters. (BTW they are not physiological parameters but physics parameters right).

Done in 658253f56d3ca6eb68c3105233b571750ded66de

Functions/ImageProcessing/xASL_im_ASLSubtractionAveraging.m
HenkMutsaerts: Fixed in 658253f56d3ca6eb68c3105233b571750ded66de

Functions/ImageProcessing/xASL_im_ASLSubtractionAveraging.m
HenkMutsaerts: Please either close this conversation if you're happy with it, or respond something to me. Now I don't know if you have seen this. (But perhaps I did the same thing with your conversations :)

Functions/Quantification/xASL_quant_DefineQuantificationParameters.m
```diff
@@ -0,0 +1,164 @@
+function [x] = xASL_quant_DefineParameters(x)
```
jan-petr: Please change the name here as well

jan-petr: also on line 2, line 4, and line 46

Functions/ImageProcessing/xASL_im_ASLSubtractionAveraging.m
jan-petr: Sorry, I have overlooked this one...

Functions/ImageProcessing/xASL_im_ASLSubtractionAveraging.m
jan-petr: To be clear - I don't care about the division, I was just raising this point because you usually care. 
I know that they are not programmers - but especially for this reason, they will be confused that you have something in percentages, something between 0-1 and something as nsrPLD has no units at all. 

So somebody can be totally confused why are you calculating percentages for pwPLD and why.

But really no issue for me - I don't care about programmers or mathematicians. I was just saying that a normal simple user will not understand why do you have to divide by hundred and convert to percentages. If you think it is better for normal users, then I'm perfectly fine with it. Your decision really. My main point was - not necessary and potentially confusing, but you can keep it as far as I'm concerned as it has no influence.

jan-petr (changes requested):
Minor fix.

jan-petr (approved):
All fixed.

PR #1607, opened 2024-01-17
Closes #1603 despiked volumes json

### LINKED ISSUE #1603: Fix despiked number of volumes
### Description

We now carefully check for the number of volumes to match PLD. But this doesn't work with despiking removing some volumes. We need to fix this and test it on datasets that are failing in TestDataSet

All manipulations related to PWI merging, FSL quant, etc. will only be done in #1543. We are only fixing single-PLD and despiked problem.

### Tasks
- [x] check that we save JSON sidecars in each step that we recreate a new PWI volume, including despiked_ASL4D
- [x] Check if there are other instances where (despiked_)ASL4D is used, which should be replaced by PWI4D, e.g., in QC
- [x] xASL_io_ReadNifti & xASL_io_Nifti2Im read the json sidecar as well
- [x] Make sure that PLD_PWI4D and all other new vectors are correctly saved to JSON with NII. And correctly read from JSON, instead of being returned inside x.Q.
- [x] When we don't find a JSON sidecar, issue a warning & use x.Q
- [x] Create `xASL_im_ASLSubtractionAveraging` wrapper: `xASL_io_ASLSubtractionAveraging`  (which doesn't output to x). So for loading & saving json files along, we use `xASL_io_ASLSubtractionAveraging`, for only creating an image (e.g., temporary NIfTI for masking, we keep using `xASL_im_ASLSubtractionAveraging
- [x] Only save quantification parameters in new JSON sidecar, not copy existing x fields
- [x] Various bugfixes

### How to test
- [x] Flavors
- [x] UnitTest
- [x] test Philips_2DEPI_Segmentation_2D_T2_SPM_HiQ - large error 100%
- [x] test Philips_2DEPI_Lesion_NoFLAIR_NoM0_HiQ - large error 10-30%
- [x] test Philips_2DEPI_FEAST_Longitudinal_FLAIR_LGA_ASL_PartlyDone_HiQ - no working with despiked
- [x] test Siemens3DGRASE_Structural_AlreadyDone_HiQ - not working with despiked
- [x] test Siemens_2DEPI_noBsup_FLAIR_LoQ - not working with despiked
- [x] Full TestDataSet

### Release notes

Nothing. Follow-up of #997

HenkMutsaerts:
### Linked issue

Closes #1603

### DIFF
--- a/External/SPMmodified/xASL/xASL_io_ReadNifti.m
+++ b/External/SPMmodified/xASL/xASL_io_ReadNifti.m
@@ -1,7 +1,7 @@
-function [NiftiObject, pathOut] = xASL_io_ReadNifti(pathIn)
-% Wrapper around SPM nifti function to read Nifti file.
+function [NiftiObject, pathOut, json] = xASL_io_ReadNifti(pathIn)
+%xASL_io_ReadNifti Wrapper around SPM nifti function to read Nifti file
 %
-% FORMAT: [NiftiObject, pathIn] = xASL_io_ReadNifti(pathIn)
+% FORMAT: [NiftiObject, pathOut, json] = xASL_io_ReadNifti(pathIn)
 %
 % INPUT:
 %   pathIn - The path to the image. String or a single cell with a string. (REQUIRED)
@@ -14,14 +14,25 @@ function [NiftiObject, pathOut] = xASL_io_ReadNifti(pathIn)
 % DESCRIPTION: Read Nifti file given by the path. Return the NII object. And also return the actual path to the loaded 
 %              Nifti if by any reason the name changed during the function runtime (e.g. unzipping).
 %
-% EXAMPLE: 
-%     [NiftiObject, pathOut] = xASL_io_ReadNifti('/home/tmp/CBF.nii')
+% EXAMPLE:  NiftiObject = xASL_io_ReadNifti('/home/tmp/CBF.nii'); % for only loading the nifti
+% EXAMPLE2: [NiftiObject, ~, json] = xASL_io_ReadNifti('/home/tmp/CBF.nii'); % for also loading the json sidecar
+% 
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
 %
 % __________________________________
-% Copyright ?? 2015-2019 ExploreASL
+% Copyright ExploreASL 2015-2024 ExploreASL
 
 % Admin
+% Set defaults
+NiftiObject = [];
+pathOut = [];
+json = [];
+
+[fPath, fFile] = xASL_fileparts(pathIn);
+pathUnzipped = fullfile(fPath, [fFile '.nii']);
+pathZipped = fullfile(fPath, [fFile '.nii.gz']);
+pathJson = fullfile(fPath, [fFile '.json']);
+
 if iscell(pathIn)
 	if length(pathIn)==1
 		pathIn = pathIn{1};
@@ -42,14 +53,11 @@ end
 if ~xASL_exist(pathIn,'file')
 	error([pathIn ' does not exist as .nii[.gz]']);
 end
-    
-[SrcPath, SrcFile] = xASL_fileparts(pathIn);
-src_fileNII = fullfile(SrcPath, [SrcFile '.nii']);
 
 %% Checking whether .nii & .nii.gz are equal (if both exist)
-xASL_adm_UnzipNifti(src_fileNII);
-    
-pathIn = src_fileNII;
+xASL_adm_UnzipNifti(pathUnzipped);
+
+pathIn = pathUnzipped;
     
 xASL_adm_ManageMoCoMat(pathIn); % manage .mat orientation sidecars
 
@@ -61,16 +69,14 @@ catch ME
     try % try unzipping
         gunzip(pathIn);
 
-        [Fpath, Ffile] = fileparts(pathIn);
-        TempUnzippedFile = fullfile(Fpath, Ffile);
         warning('NIfTI seemed to be zipped, have unzipped this');
         try % if works, rename file
-            nifti(TempUnzippedFile);
-            xASL_SysMove(TempUnzippedFile, pathIn, true);                
+            nifti(pathUnzipped);
+            xASL_SysMove(pathUnzipped, pathIn, true);                
         catch % otherwise throw error
             warning(['Couldnt read ' pathIn]);
-            if exist(TempUnzippedFile,'file') && exist(pathIn,'file')
-                delete(TempUnzippedFile);
+            if exist(pathUnzipped, 'file') && exist(pathIn, 'file')
+                delete(pathUnzipped);
             end
             error(ME);
         end
@@ -79,4 +85,11 @@ end
 
 pathOut = [pathIn ImExt]; % put it back
     
+%% Load JSON sidecar if it exists
+
+if exist(pathJson)
+    json = xASL_bids_parms2BIDS([], xASL_io_ReadJson(pathJson), 0);
+end
+
+
 end
--- a/External/SPMmodified/xASL/xASL_io_ReadNifti.m
+++ b/External/SPMmodified/xASL/xASL_io_ReadNifti.m
@@ -8,14 +8,16 @@ function [NiftiObject, pathOut, json] = xASL_io_ReadNifti(pathIn)
 %
 % OUTPUT:
 %   NiftiObject - Nifti structure of the loaded file.
-%   pathOut   - Modified path to the image in the form that is valid for the current (possibly unzipped) file.
+%   pathOut     - Modified path to the image in the form that is valid for the current (possibly unzipped) file.
+%   json        - Loaded JSON sidecar, in ExploreASL legacy format
 %
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
 % DESCRIPTION: Read Nifti file given by the path. Return the NII object. And also return the actual path to the loaded 
 %              Nifti if by any reason the name changed during the function runtime (e.g. unzipping).
 %
-% EXAMPLE:  NiftiObject = xASL_io_ReadNifti('/home/tmp/CBF.nii'); % for only loading the nifti
-% EXAMPLE2: [NiftiObject, ~, json] = xASL_io_ReadNifti('/home/tmp/CBF.nii'); % for also loading the json sidecar
+% EXAMPLE:  
+%   NiftiObject = xASL_io_ReadNifti('/home/tmp/CBF.nii'); % for only loading the nifti
+%   [NiftiObject, ~, json] = xASL_io_ReadNifti('/home/tmp/CBF.nii'); % for also loading the json sidecar
 % 
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
 %
--- a/Functions/ImageProcessing/xASL_im_ASLSubtractionAveraging.m
+++ b/Functions/ImageProcessing/xASL_im_ASLSubtractionAveraging.m
@@ -92,6 +92,33 @@ elseif isempty(path_ASL4D) && ( (nargin>=3 && isempty(PWI4D)) || (nargin>=4 && i
 	error('Empty Path_ASL4D, PWI4D, and PWI3D');
 elseif (numel(path_ASL4D)~=1 && ~isnumeric(path_ASL4D) && numel(path_ASL4D)<1000)
     % this is a path, not an image or it is empty
+    
+    if xASL_exist(path_ASL4D, 'file')
+        % if this is a path, we will try to get the JSON sidecar
+        [pDir, pFile] = xASL_fileparts(path_ASL4D);
+        path_ASL4Djson = fullfile(pDir, [pFile '.json']);
+
+        if xASL_exist(path_ASL4Djson, 'file')
+            % we try to load this json and extract quantification parameters
+            json = xASL_io_ReadJson(path_ASL4Djson);
+            % if it has the two required quantification fields PLD & LD, we will use this
+            jsonFields = fields(json);
+            fieldPLD = find(contains(jsonFields, 'PostLabelingDelay'));
+            fieldLD = find(contains(jsonFields, 'LabelingDuration'));
+            fieldTE = find(contains(jsonFields, 'EchoTime'));
+
+            if ~isempty(fieldPLD) && ~isempty(fieldLD)
+                x.Q.LabelingDuration = json.LabelingDuration;
+                x.Q.Initial_PLD = json.PostLabelingDelay;
+
+                if ~isempty(fieldTE)
+                    x.Q.EchoTime = json.EchoTime;
+                else
+                    x.Q.EchoTime = [];
+                end
+            end
+        end
+    end
 else
     % Path_ASL4D is a matrix. Then we delete it and move it to ASL4D
 	ASL4D = path_ASL4D;
--- a/Functions/ImageProcessing/xASL_im_ASLSubtractionAveraging.m
+++ b/Functions/ImageProcessing/xASL_im_ASLSubtractionAveraging.m
@@ -69,9 +69,9 @@ function [PWI, PWI3D, PWI4D, x, Control, Control3D, Control4D] = xASL_im_ASLSubt
 % quantification. So the user can, after QC, potentially remove some volumes from PWI4D, but should be aware that the information in the JSON sidecar
 % (TE, PLD, LabDur) should be adapted accordingly.
 %
-% EXAMPLE used by xASL_wrp_RegisterASL: PWI = xASL_im_ASLSubtractionAveraging(x, ASL_im);
-% EXAMPLE used by xASL_wrp_ResampleASL: [PWI, PWI3D, PWI4D, x] = xASL_im_ASLSubtractionAveraging(x, ASL4D, PWI4D, PWI3D);
-% EXAMPLE used by xASL_quant_ASL: [PWI] = xASL_im_ASLSubtractionAveraging(x, [], PWI4D);
+% EXAMPLE:
+%     Inside xASL_wrp_RegisterASL: [PWI, ~, ~, ~, Control] = xASL_im_ASLSubtractionAveraging(x, x.P.Path_despiked_ASL4D);
+%     Inside xASL_quant_ASL: [PWI] = xASL_im_ASLSubtractionAveraging(x, [], PWI4D);
 % __________________________________
 % Copyright (C) 2015-2024 ExploreASL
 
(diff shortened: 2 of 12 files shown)


Functions/ImageProcessing/xASL_im_ASLSubtractionAveraging.m
```diff
@@ -92,6 +92,33 @@
...
     % this is a path, not an image or it is empty
+    
+    if xASL_exist(path_ASL4D, 'file')
+        % if this is a path, we will try to get the JSON sidecar
+        [pDir, pFile] = xASL_fileparts(path_ASL4D);
+        path_ASL4Djson = fullfile(pDir, [pFile '.json']);
+
+        if xASL_exist(path_ASL4Djson, 'file')
+            % we try to load this json and extract quantification parameters
+            json = xASL_io_ReadJson(path_ASL4Djson);
```
jan-petr, line 103: Add this line:
json = xASL_bids_parms2BIDS([], json, 0); % Convert it from BIDS to Legacy

jan-petr: Note that these are in line with what we have discussed in the issue.

Functions/ImageProcessing/xASL_im_ASLSubtractionAveraging.m
```diff
@@ -92,6 +92,33 @@
...
+            % if it has the two required quantification fields PLD & LD, we will use this
+            jsonFields = fields(json);
```
jan-petr: You can remove these lines and looking for fields. We have converted to Legacy so we know that PLD has to be either at
JSON.Q.Initial_PLD or it isn't there.

So we simply remove lines 105-119 and have instead this. With the changes above, we'll already be in correct Legacy notation and format...
if isfield(json, 'Q') && isfield(json.Q, 'LabelingDuration')
   x.Q.Initial_PLD = json.Q.Initial_PLD;
   end 
   
   Same for EchoTime and LabelingDuration

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
```diff
@@ -506,30 +506,22 @@ function xASL_wrp_RealignASL(x, bASL)
...
-
         save(x.P.Path_despiked_ASL4D_mat, 'mat');
 
         % Also change parameter vectors
-        x.Q.EchoTime = x.Q.EchoTime(~exclusion);
-        x.Q.Initial_PLD = x.Q.Initial_PLD(~exclusion);
-        x.Q.LabelingDuration = x.Q.LabelingDuration(~exclusion);
-
-        % #997 Need to save this in json sidecar for further usage in the pipeline
+        jsonFields.EchoTime = x.Q.EchoTime(~exclusion);
```
jan-petr: The function parms2BIDS will be able to handle that. But still, it's better to use the correct format and write
jsonFields.Q.EchoTime instead of jsonFields.EchoTime - and the same for the other two fields.

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
```diff
@@ -506,30 +506,22 @@ function xASL_wrp_RealignASL(x, bASL)
...
+        jsonFields.Initial_PLD = x.Q.Initial_PLD(~exclusion);
+        jsonFields.LabelingDuration = x.Q.LabelingDuration(~exclusion);
 
+        xASL_io_SaveNifti(x.P.Path_ASL4D, x.P.Path_despiked_ASL4D, NewIm, 32, 0, [], 1, xASL_bids_parms2BIDS(jsonFields, [], 1));
```
jan-petr, line 527: the conversion parms2bids is all good. But in SaveNifti, do we really want to set the parameter bCopyOrigJson == true? I though we only want to save the three new fields without copying the entire JSON.

jan-petr (changes requested):
Some comments related to the decision from the issue.

Functions/ImageProcessing/xASL_im_ASLSubtractionAveraging.m
HenkMutsaerts: Done this in ...
It issues a warning though that the PLD values are not within the optimized range, but that the average is. We should suppress these warnings if we reload later in the ASL module right, shall we add a verbosity boolean in xASL_bids_parms2BIDS for this?

Functions/ImageProcessing/xASL_im_ASLSubtractionAveraging.m
HenkMutsaerts: No we still have to check. The code actually becomes a bit longer, because we have to check for the existence of the field Q. Otherwise, the code is identical.

Functions/ImageProcessing/xASL_im_ASLSubtractionAveraging.m
jan-petr: We can indeed add verbosity. Or even better, change the range, because PLD 0 is now common for certain multi-PLD sequence. I would prefer the latter.

Functions/ImageProcessing/xASL_im_ASLSubtractionAveraging.m
jan-petr: I mean a different thing. It makes no sense to look for general PLD field as it **has** to be in Q.PLD and nowhere else.
Also - you need to change the notation as mentioned above. The parms2BIDS conversion changes the notation to Legacy and you are looking for BIDS fields.

Functions/ImageProcessing/xASL_im_ASLSubtractionAveraging.m
HenkMutsaerts: Done in b4fab938c054d67bafcf53983698c413a682ea9b

Functions/ImageProcessing/xASL_im_ASLSubtractionAveraging.m
HenkMutsaerts: Fixed

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
HenkMutsaerts: Done in 6608345d8747fad9f022083eed435441de6c2bb5

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
HenkMutsaerts: Done in 9f7dc90711ec73ab569ce02e709c319e5897e2da

External/SPMmodified/xASL/xASL_io_ReadNifti.m
```diff
@@ -1,7 +1,7 @@
-function [NiftiObject, pathOut] = xASL_io_ReadNifti(pathIn)
-% Wrapper around SPM nifti function to read Nifti file.
+function [NiftiObject, pathOut, json] = xASL_io_ReadNifti(pathIn)
+%xASL_io_ReadNifti Wrapper around SPM nifti function to read Nifti file
 %
-% FORMAT: [NiftiObject, pathIn] = xASL_io_ReadNifti(pathIn)
```
jan-petr, line 4: Please add "json" to the OUTPUT description.

External/SPMmodified/xASL/xASL_io_ReadNifti.m
```diff
@@ -14,14 +14,25 @@
 % DESCRIPTION: Read Nifti file given by the path. Return the NII object. And also return the actual path to the loaded 
 %              Nifti if by any reason the name changed during the function runtime (e.g. unzipping).
 %
-% EXAMPLE: 
-%     [NiftiObject, pathOut] = xASL_io_ReadNifti('/home/tmp/CBF.nii')
+% EXAMPLE:  NiftiObject = xASL_io_ReadNifti('/home/tmp/CBF.nii'); % for only loading the nifti
+% EXAMPLE2: [NiftiObject, ~, json] = xASL_io_ReadNifti('/home/tmp/CBF.nii'); % for also loading the json sidecar
```
jan-petr: Not sure about the style Michael did, but best to have "EXAMPLE" and then several lines, instead of "EXAMPLE2".

Functions/ImageProcessing/xASL_im_ASLSubtractionAveraging.m
```diff
@@ -92,6 +92,39 @@
...
+                fieldTE = find(contains(jsonFields, 'EchoTime'));
+
+                if ~isempty(fieldPLD) && ~isempty(fieldLD)
+                    x.Q.LabelingDuration = json.Q.LabelingDuration;
+                    x.Q.Initial_PLD = json.Q.Initial_PLD;
+    
+                    if ~isempty(fieldTE)
+                        x.Q.EchoTime = json.Q.EchoTime;
+                    else
+                        x.Q.EchoTime = [];
```
jan-petr: Careful, x.Q.EchoTime = [] is only assigned when the Q field is there.

Functions/ImageProcessing/xASL_im_ASLSubtractionAveraging.m
```diff
@@ -92,6 +92,39 @@
...
+
+        if xASL_exist(path_ASL4Djson, 'file')
+            % we try to load this json and extract quantification parameters
+            json = xASL_io_ReadJson(path_ASL4Djson);
+            json = xASL_bids_parms2BIDS([], json, 0); % Convert it from BIDS to Legacy
+
+            % if it has the two required quantification fields PLD & LD, we will use this
+            if isfield(json, 'Q')
+                jsonFields = fields(json.Q);
+                fieldPLD = find(contains(jsonFields, 'Initial_PLD'));
```
jan-petr: isfield(json.Q, 'Initial_PLD') seems easier to use than this construct.

Functions/ImageProcessing/xASL_im_ASLSubtractionAveraging.m
```diff
@@ -92,6 +92,39 @@
...
+                fieldLD = find(contains(jsonFields, 'LabelingDuration'));
+                fieldTE = find(contains(jsonFields, 'EchoTime'));
+
+                if ~isempty(fieldPLD) && ~isempty(fieldLD)
```
jan-petr: Why assigning PLD and LD only when we have both?
In the end, it won't matter that hugely, but we should simply check for PLD -> Assign PLD; check for LD -> assignLD etc.

Functions/InputOutput/xASL_io_ASLSubtractionAveraging.m
```diff
@@ -0,0 +1,184 @@
...
+    error('x input missing');
+end
+if nargin<2 || isempty(saveWhichNifti)
+    error('Second input argument missing');
+end
+if nargin<3 || isempty(bCopyOrigJson)
+    bCopyOrigJson = true;
+end
+
+% pathNifti     = {path_PWI               path_PWI3D               path_PWI4D                 path_Control                path_Control3D                  path_Control4D};
```
jan-petr: Shall we deleted this commented code?

Functions/InputOutput/xASL_io_ASLSubtractionAveraging.m
```diff
@@ -0,0 +1,184 @@
...
+% imageInput = {ASL4D PWI4D PWI3D Control4D Control3D}
+% imageOutput = {PWI PWI3D PWI4D Control Control3D Control4D}
+
+% [PWI, PWI3D, PWI4D, x, Control, Control3D, Control4D] = xASL_im_ASLSubtractionAveraging(x, ASL4D [, PWI4D, PWI3D, Control4D, Control3D])
+[imageOutput{1}, imageOutput{2}, imageOutput{3}, x, imageOutput{4}, imageOutput{5}, imageOutput{6}] = xASL_im_ASLSubtractionAveraging(x, imageInput{1}, imageInput{2}, imageInput{3}, imageInput{4}, imageInput{5});
+ 
+
+%% 3. =====================================================================================
+%% 4. Save the NIfTI & JSON files
+fieldNamesPLD = {'InitialPLD_PWI'       'InitialPLD_PWI3D'       'InitialPLD_PWI4D'         'InitialPLD_Control'        'InitialPLD_Control3D'          'InitialPLD_Control4D'};
```
jan-petr, line 95: Do we have Initial_PLD or InitialPLD notation elsewhere?

Functions/InputOutput/xASL_io_ASLSubtractionAveraging.m
```diff
@@ -0,0 +1,184 @@
...
+    if ~exist(pathJson, 'file')
+        warning(['Reverting to x.Q memory: JSON sidecar missing, : ' pathJson]);
+    else
+        json = xASL_io_ReadJson(pathJson); % load the json-file
+        json = xASL_bids_parms2BIDS([], json, 0); % Convert it from BIDS to Legacy
+
+        % if it has the two required quantification fields PLD & LD, we will use this
+        if isfield(json, 'Q')
+            jsonFields = fields(json.Q);
+            fieldPLD = find(contains(jsonFields, 'Initial_PLD'));
```
jan-petr: Same comment as above in another fucntion on the same code.

Functions/InputOutput/xASL_io_ASLSubtractionAveraging.m
```diff
@@ -0,0 +1,184 @@
...
+
+%% ========================================================================================
+%% 2. Load NIfTIs & JSONs
+for iNifti=1:5
+    if nargin<(iNifti+3)
+        path2load{iNifti} = [];
+    else
+        path2load{iNifti} = varargin{iNifti};
+    end
+    [x, imageInput{iNifti}] = xASL_io_ASLSubtractionAveraging_sub_LoadPWI_JSON(x, path2load{iNifti});
```
jan-petr: I'm confused. you return x-struct, which is being overwritten through the cycle. And since we don't differentiate between x.Q.EchoTime for PWI or PWI4D, then this is again going to be overwritten. So it makes no sense to return x-struct.

Functions/InputOutput/xASL_io_ASLSubtractionAveraging.m
```diff
@@ -0,0 +1,184 @@
...
+
+%% 3. =====================================================================================
+%% 4. Save the NIfTI & JSON files
+fieldNamesPLD = {'InitialPLD_PWI'       'InitialPLD_PWI3D'       'InitialPLD_PWI4D'         'InitialPLD_Control'        'InitialPLD_Control3D'          'InitialPLD_Control4D'};
+fieldNamesLD  = {'LabelingDuration_PWI' 'LabelingDuration_PWI3D' 'LabelingDuration_PWI4D'   'LabelingDuration_Control'  'LabelingDuration_Control3D'    'LabelingDuration_Control4D'};
+fieldNamesTE  = {'EchoTime_PWI'         'EchoTime_PWI3D'         'EchoTime_PWI4D'           'EchoTime_Control'          'EchoTime_Control3D'            'EchoTime_Control4D'};
+
+for iNifti=1:size(saveWhichNifti, 1)
+    n2save = saveWhichNifti{iNifti,1};
+    xASL_io_ASLSubtractionAveraging_sub_SavePWI_JSON(x, saveWhichNifti{iNifti,2}, imageOutput{n2save}, fieldNamesPLD{n2save}, fieldNamesLD{n2save}, fieldNamesTE{n2save}, bCopyOrigJson);
```
jan-petr, line 101: This seems to be wrong by design. We load to x.Q.Initial_PLD previously, but here we save to x.Q.Initial_PLD_PWI etc.

jan-petr (changes requested):
Some smaller comments.

External/SPMmodified/xASL/xASL_io_ReadNifti.m
HenkMutsaerts: Done in 256794409e13b6504044300e9cf128b7642e8509

External/SPMmodified/xASL/xASL_io_ReadNifti.m
HenkMutsaerts: Done in 256794409e13b6504044300e9cf128b7642e8509

Functions/ImageProcessing/xASL_im_ASLSubtractionAveraging.m
HenkMutsaerts: Because of their interdependence.
Fine to change it to what you suggest. My thinking was that, we have to deal with the situation:
x.Q contains something else then
json

then only if JSON is complete, we will overwrite x.Q

Because PLD and LD are not independent right? E.g., imagine a multi-PLD sequence, LD may differ between volumes with different PLDs, so we cannot load one without the other.

Happy to adapt if you have a better idea ;) Of course this would be a bad situation, but we need to hope for the best and plan for the worst. Perhaps add a warning :)

Functions/ImageProcessing/xASL_im_ASLSubtractionAveraging.m
HenkMutsaerts: Fixed in 256794409e13b6504044300e9cf128b7642e8509

Functions/InputOutput/xASL_io_ASLSubtractionAveraging.m
HenkMutsaerts: Done in 3c55ce996809257f2dcdbb45f4915692a8e86709

Functions/InputOutput/xASL_io_ASLSubtractionAveraging.m
HenkMutsaerts: Fixed in 3c55ce996809257f2dcdbb45f4915692a8e86709. You're right, this was a bug :)

Functions/InputOutput/xASL_io_ASLSubtractionAveraging.m
HenkMutsaerts: Perhaps we do, but we need to redefine here, because we save the JSON sidecars with the BIDS notation. So we are ignorant which file contains which PLD/LD/TE unless we redefine this here.

Functions/InputOutput/xASL_io_ASLSubtractionAveraging.m
HenkMutsaerts: Fixed in 3c55ce996809257f2dcdbb45f4915692a8e86709

Functions/InputOutput/xASL_io_ASLSubtractionAveraging.m
HenkMutsaerts: Yes, see my latest fix in 3c55ce996809257f2dcdbb45f4915692a8e86709. It is a wrapper based on what we already had, and probably there is a more efficient way to do this. But fine for now?

HenkMutsaerts (commented):
One residual major design question

jan-petr (approved):
OK

PR #1608, opened 2024-01-17
#1549 quant_FSL set options for Quality==0

### LINKED ISSUE #1549: Accelerate BASIL/FABBER testing
### Description

ExploreASL has Quality==0 for testing stuff. But BASIL/FABBER still uses high quality then which takes a lot of time.
Would be helpful if we can downtune BASIL/FABBER quality parameters based on this ExploreASL-wide parameter.

**JAN** I based it on #1551, because we had some changes in the BASIL parameter names that I want to use. And doing that on develop would be possible, but could result in name consistency later.

### Tasks

- [x] With low quality, set the following BASIL/FABBER options
-- infert1 (off)
-- inferart (off)
-- exch=simple
-- disp=none

### Release notes

Faster low-quality mode for FSL.

jan-petr:
### Linked issue

Close #1549


HenkMutsaerts (approved):
Great, though this doesn't help us with testing BBB right? Can't we use similar parameters for FABBER as well?

PR #1609, opened 2024-01-17
#1597 init_DataLoading warnings for rawdata and population

### LINKED ISSUE #1597: When running Population only skip BIDS2legacy
### Description

bReadRawdata is true by default, unless the rawdata folder is missing. This is needed because by default we process the original scans in rawdata, not any derivatives. But in the case when a user does not want to process anything but just wants to do group analyses, running through BIDS2Legacy (even for checking that this was already done) may unnecessary slow things down.

In this case, the user can set bReadRawdata to false, which will skip BIDS2Legacy. We can automate this.

### Tasks

- [x] When rawdata exists, and the user gives [0 0 1] as processing booleans (== only running the population module), we give a warning that we will skip BIDS2Legacy and set bReadRawdata to false.
- [x] We give a warning if the user had bReadRawdata == true and processing [0 0 1], if the user for whatever reason wanted to both do BIDS2Legacy and running the population module only. 
----->>>>> the reason for this, is that the population module does a few things that DON'T NEED the structural & ASL modules. E.g., creating templates of the rawdata T1w, M0, or giving an overview of quantification parameters. But this probably won't happen a lot, which is why I propose this solution with a warning.

### Release notes

Don't run BIDS2Legacy conversion for re-running the population module.

jan-petr:
### Linked issue

Close #1597

### DIFF
--- a/Modules/Module_Processing/xASL_init_DataLoading.m
+++ b/Modules/Module_Processing/xASL_init_DataLoading.m
@@ -50,10 +50,24 @@ function [x] = xASL_init_DataLoading(x)
     if ~isfield(x.opts, 'bReadRawdata') || isempty(x.opts.bReadRawdata)
         % If the developer has not set this parameter, we try to detect whether we should read from /rawdata or /derivatives/ExploreASL 
         % Usually, this parameter will not exist
-        if xASL_exist(x.dir.RawData, 'dir')
-            x.opts.bReadRawdata = true; %  by default we read rawdata
-            fprintf('%s\n', '/rawdata folder detected, trying to load subjects from BIDS');
+        if xASL_exist(x.dir.RawData, 'dir') && xASL_exist(x.dir.xASLDerivatives, 'dir')
+			% Both rawdata and derivatives exist
+
+			% By default we read rawdata
+			if ~isequal(x.opts.bProcess(:), [0; 0; 1])
+				x.opts.bReadRawdata = true; %
+				fprintf('%s\n', '/rawdata folder detected, trying to load subjects from BIDS');
+			else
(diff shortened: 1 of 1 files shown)


Modules/Module_Processing/xASL_init_DataLoading.m
```diff
@@ -63,6 +77,9 @@
         % Optionally, developers have set this parameter to force reading subjects from /rawdata
         warning(['rawdata folder missing: ' x.dir.RawData]);
         error('rawdata did not exist, if you want to read subjects from /derivatives/ExploreASL, set x.opts.bReadRawdata to false in dataPar.json');
+    elseif x.opts.bReadRawdata && isequal(x.opts.bProcess(:), [0; 0; 1])
```
HenkMutsaerts, line 78: This is not true right?
If the user specifically set bReadRawdata to true, we don't change it

HenkMutsaerts (changes requested):
Single suggestion, and I shortened the if-statement list a bit

Modules/Module_Processing/xASL_init_DataLoading.m
jan-petr: If use sets explicitly bReadRawdata to true, we don't change it. But I don't chante it :) I just issue a warning like agreed :)

All good? :)

HenkMutsaerts: I think so. Did you check my last commit?

jan-petr: Yes. It's all good and you can approve.

HenkMutsaerts (approved):

PR #1614, opened 2024-01-29
Closes #1613 xASL_init_LoadMetadata: fix continuous numerical data loading

### LINKED ISSUE #1613: Hematocrit not loading correctly from participants.tsv
### Description

It seems that continuous data may not be loaded correctly. This has precluded the hematocrit-correction in ExploreASL. I fixed this now, but we need to test if it doesn't break anything else (I don't think so).

This was a bug noticed by @MDijsselhof in 1.11.0 main release, he will test if it now works and @jan-petr can then merge this to 1.12.0 (if it wasn't already fixed in the develop branch).

### Tasks

- [x] xASL_init_LoadMetadata: fix loading continuous numerical data

### Release notes

Fix hematocrit loading from participants.tsv.

HenkMutsaerts:
### Linked issue

Closes #1613


jan-petr (approved):
OK

PR #1627, opened 2024-02-27
#1626 Fix PhoenixAnalyzer for VEPCASL sequence

### LINKED ISSUE #1626: Incorrectly calculate PLD for PCASL
### Description

ExploreASL automatically calculates PLD from TI - inversion time. But it does this also for PCASL even when the correct PLD is provided in studyPar.json.

### Release notes

Fix DICOM Phoenix conversion for VEPCASL implementation.

jan-petr:
### Linked issue

#1626


HenkMutsaerts (approved):
You're now asking 3 instead of 2 conditions to be true before you scale with 1 million, I assume this is true and you have tested this with other Phoenix/Siemens flavors?

PR #1633, opened 2024-02-28
Feature #1631 phoenix vepcasl

### LINKED ISSUE #1631: Correct DICOM-Phoenix parsing of a VEPCASL implementation
### Description

Implement Phoenix parsing for Siemens VEPCASL sequence.

### Tasks

- [x] Remove unused BIDS tags that are not even defined in BIDS
- [x] Remove incorrectly named tags converted by dcm2nii that do not correspond to the correct BIDS tag names
- [x] Identify this sequence in Phoenix parsing and convert all tags
- [x] Fix Import summary bug
- [x] Locate all other similar sequences in Flavors
- [x] Make sure they all work
- [x] Add VEPCASL to flavors
- [x] Final flavor test

### Release notes

Phoenix-DICOM parsing for a sequence to_ep2d_vepcasl.

jan-petr:
### Linked issue

Close #1631


Modules/Module_Import/NII2BIDS/xASL_bids_PhoenixProtocolAnalyzer.m
```diff
@@ -14,198 +14,256 @@
...
 
+	%% 2. Obtain basic information about the sequence
 	bidsPar = struct();
-    %% Analyze parameters                                                   % Exemplary dataset in ExploreASL flavor library
-	if ~isempty(strfind(sourcePar.tSequenceFileName,'SiemensSeq'))
-		bidsPar.SequenceType = 'Siemens'; % Not really a BIDS field
-	elseif ~isempty(strfind(sourcePar.tSequenceFileName,'CustomerSeq'))
-		bidsPar.SequenceType = 'Customer';
+    % Find out if it's a product sequence or a patch
+	if ~isempty(regexp(sourcePar.tSequenceFileName,'SiemensSeq', 'once'))
```
HenkMutsaerts: Don't you want regexpi in these instances as well, like below?

Modules/Module_Import/NII2BIDS/xASL_bids_PhoenixProtocolAnalyzer.m
```diff
@@ -14,198 +14,256 @@
...
-	if isfield(bidsPar,'SoftwareVersions') &&...
-			(~isempty(regexpi(bidsPar.SoftwareVersions,'N4_VB15A'))||...
-			 ~isempty(regexpi(bidsPar.SoftwareVersions,'N4_VB17A'))||...
-			 ~isempty(regexpi(bidsPar.SoftwareVersions,'N4_VB19A')))
-		if isfield(sourcePar,'sGroupArraysPSatdGap') && ~isempty(sourcePar.sGroupArraysPSatdGap)
-			bidsPar.LabelingDistance = sourcePar.sGroupArraysPSatdGap;
+
+	%% 4. Reading a default sequence
+	if ~bSequenceIdentified
+		%TODO
```
HenkMutsaerts: We used PM (pro memoria) previously. I don't care what we do, but it would be nice to have a searchable single word we use :)

Modules/Module_Import/NII2BIDS/xASL_bids_PhoenixProtocolAnalyzer.m
```diff
@@ -14,198 +14,256 @@
...
-			lastEchoTimeCount = lastEchoTimeCount + 1;
-			echoTimePar = getPhoenixParameters({['alTE[' num2str(lastEchoTimeCount) ']']},parameterList,0);
-			if isempty(echoTimePar{2})
-				lastEchoTime = 0;
-			else
-				lastEchoTime = str2num(echoTimePar{2})/1000/1000;
-				bidsPar.EchoTime(lastEchoTimeCount) = lastEchoTime;
+
+		% N4_VD13D and N4_VE11C have also slab_thickness_mm parameter defined and it is unclear if this is imaging FOV or labeling slab thickness...
+		if isfield(bidsPar,'SoftwareVersions') &&...
```
HenkMutsaerts, line 194: Lots of if statements throughout the whole function, that probably only you understand :) I don't care, but switch-case constructions could be more readable for yourself and others (including comments if helpful)

Modules/Module_Import/NII2BIDS/xASL_bids_PhoenixProtocolAnalyzer.m
```diff
@@ -14,198 +14,256 @@
...
+		% The function of this parameter is unclear
+		%if isfield(sourcePar,'Slab_thickness_mm') && ~isempty(sourcePar.Slab_thickness_mm)
+		%	bidsPar.LabelingSlabThickness = sourcePar.Slab_thickness_mm;
+		%end
+
+		if isfield(sourcePar,'alTE0') && ~isempty(sourcePar.alTE0)
+			bidsPar.EchoTime = sourcePar.alTE0 / 1000 / 1000;
+			lastEchoTime = bidsPar.EchoTime;
+			lastEchoTimeCount = 1;
+			while lastEchoTime > 0
```
HenkMutsaerts, line 216: no risk ending up in an infinite loop?

echoTimePar{2} is always empty in the last case, cannot be an empty struct/cell array that matlab sees as non-empty?

HenkMutsaerts (changes requested):
Few minor things to validate. But more or less I'm fine of course

Modules/Module_Import/NII2BIDS/xASL_bids_PhoenixProtocolAnalyzer.m
jan-petr: Actually, this was there before. And is not associated with any issue. 
I'd still prefer TODO as this is easier searcheable than PM. Unless we of course have a specific issue, then PM #issuenumber
OK?

Modules/Module_Import/NII2BIDS/xASL_bids_PhoenixProtocolAnalyzer.m
jan-petr: That was the old code, which will be gradually phased out and rewritten. I just moved that part, not edited it.

Modules/Module_Import/NII2BIDS/xASL_bids_PhoenixProtocolAnalyzer.m
jan-petr: Exactly - no risk. This works and was tested.
It reads alTE[XX] from Phoenix. It ends when there are zeros or when the field ends. Absolutely no risk this ends in an infinite loop.

HenkMutsaerts (approved):

Modules/Module_Import/NII2BIDS/xASL_bids_PhoenixProtocolAnalyzer.m
jan-petr: Fixed.

PR #1644, opened 2024-03-12
#1536 Improve headers

### LINKED ISSUE #1536: Improve header of xASL_im_Smooth3D
### Description

Explain that input is in voxels, not mm. And that it is sigma and not FWHM and how to calculate FWHM.

### Tasks

- [x] Explain.
- [x] I would then also check if you want to "propagate" this explanation in the functions that you call.
- [x] check if there are other smoothing functions that have this issue, like our usages of `xASL_im_ndnanfilter` - a new issue was created #1643 

### Release notes

Improve comments in smoothing functions.

jan-petr:
### Linked issue

Close #1536


HenkMutsaerts (approved):
Nice! (though if you only change comments we don't need a review, do we?

Issue #58, opened 2020-07-24
Improve the code for the native space processing


jan-petr:
Native space processing
ToDo Jan: add header explanations & comments, e.g. to xASL_wrp_PreparePV (6th step)

xASL_im_CreateAnalysisMasks.m
I would vote for having bROIStats_NativeSpace & bROIStats_StandardSpace, instead of bSkip* for xASL_im_CreateAnalysisMask & bNativeSpaceAnalysis:
* bNativeSpaceAnalysis -> too vague, we want to specify that this is the ROIStats (e.g. not native space creation of CBF.nii or whatsoever)
* bSkip* -> better to have a bRun than a bDontRun :)
(I realize I’m probably stating the obvious :)

* Native space masks -> we have these already after the xASL_module_ASL->xASL_wrp_CreateAnalysisMask.m.
Do we rename them, or do they keep the same name in native space? I would prefer keeping the same name for transparancy.

* GZIP: I would remove the zipping & deleting, and I will put on the "ExploreASL Wishlist” to use the .nii.gz / .mat option
Note that at the end of the pipeline (Population module) everything is zipped.

* Interpolation -> not sure if we need to do the presmooth>spline>p>0.5, seems a lot of computation time for minor change.
Perhaps include the x.Quality=0 option, in which this is simply nearest-neighbour interpolated for quicker testing
This should be there in general (but I see you use the x.Quality)

* NativeSpaceAnalysis -> should go to the xASL_wrp_GetROIstatistics

xASL_module_Population -> the atlases MNI_Structural, Hammers & Harvard-Oxford take a lot of time, which is why they were disabled by default (and they are easy to run, as shown in the tutorial, that you provide the x.S.InputAtlasPath etc yourself). So I would keep only the TotalGM, DeepWM and WholeBrain as default. Or we can switch this with the x.Quality as well.

xASL_stat_GetROIstatistics -> 
* x.LeftMask -> make space-independent: you always need a x.LeftMask input (which can be a 4D list of path names)
So the stat function always loads the LeftMask through xASL_io_Nifti2Im, and the wrp function prepares this. It is a cell list with path names or images, so in case of standard space this would be an image (created in wrp
as (1:60,:,:) = 0) and for native space this is a list of temporary mask NIfTIs. Note that xASL_io_Nifti2Im allows both a image matrix or path input.

[...]

PR #1653, opened 2024-03-19
#1648 DefaultEffectiveResolution fix

### LINKED ISSUE #1648: Sequence-dependency native-space population module bug
### Description

This was fixed for the population analysis in #870. But it still doesn't work when the native-space analysis was on.

### Tasks
- [x] Load x.Q.Sequence from NII+JSON when missing.
- [x] Pass changes from #870 to native space.
- [x] Use the group mask from standard space and transfer that to the native space as well along with the other templates.

### Release notes

Report with #870

jan-petr:
### Linked issue

Close #1648

### DIFF
--- a/Modules/Module_Processing/xASL_init_DefaultEffectiveResolution.m
+++ b/Modules/Module_Processing/xASL_init_DefaultEffectiveResolution.m
@@ -35,9 +35,14 @@ function [EffectiveResolution] = xASL_init_DefaultEffectiveResolution(PathASL, x
 %% ----------------------------------------------------------------------------------------
 %% Admin
 tIM = xASL_io_ReadNifti(PathASL);
+[fPath, fName] = xASL_fileparts(PathASL);
+tJson = xASL_io_ReadJson([fullfile(fPath, fName) '.json']);
+tJson = xASL_bids_parms2BIDS([], tJson, 0);
 NativeResolution = tIM.hdr.pixdim(2:4);
-if ~isfield(x.Q,'Sequence')
-    warning(['Setting x.Q.Sequence missing, skipping']);
+if ~isempty(tJson) && isfield(tJson.Q, 'Sequence')
+	x.Q.Sequence = tJson.Q.Sequence;
+elseif ~isfield(x.Q, 'Sequence') 
+    warning('Setting x.Q.Sequence missing, skipping');
     return;
 end
 
--- a/Modules/Module_Processing/xASL_init_DefaultEffectiveResolution.m
+++ b/Modules/Module_Processing/xASL_init_DefaultEffectiveResolution.m
@@ -30,18 +30,38 @@ function [EffectiveResolution] = xASL_init_DefaultEffectiveResolution(PathASL, x
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
 % REFERENCES: Petr, 2018 MAGMA; Vidorreta 2013 Neuroimage
 %
-% Copyright 2015-2020 ExploreASL
+% Copyright 2015-2024 ExploreASL
 
 %% ----------------------------------------------------------------------------------------
 %% Admin
 tIM = xASL_io_ReadNifti(PathASL);
 [fPath, fName] = xASL_fileparts(PathASL);
+
+% Read JSON of the ASL scan
 tJson = xASL_io_ReadJson([fullfile(fPath, fName) '.json']);
 tJson = xASL_bids_parms2BIDS([], tJson, 0);
+
+% Obtain the resolution based on the image matrix
 NativeResolution = tIM.hdr.pixdim(2:4);
-if ~isempty(tJson) && isfield(tJson.Q, 'Sequence')
-	x.Q.Sequence = tJson.Q.Sequence;
-elseif ~isfield(x.Q, 'Sequence') 
+
+% Load the x.Q fields from the JSON
+if ~isempty(tJson) && isfield(tJson, 'Q')
+	if isfield(tJson.Q, 'Vendor')
+		x.Q.Vendor = tJson.Q.Vendor;
+	end
+	if isfield(tJson.Q, 'Sequence')
+		x.Q.Sequence = tJson.Q.Sequence;
+	end
+	if isfield(tJson.Q, 'readoutDim')
+		x.Q.readoutDim = tJson.Q.readoutDim;
+	end
+end
+
+% Verify the x.Q.Sequence and fill in based on Vendor and ReadoutDim
+x = xASL_adm_DefineASLSequence(x, false);
+
+% If sequence is still missing we skip this function
+if ~isfield(x.Q, 'Sequence') 
     warning('Setting x.Q.Sequence missing, skipping');
     return;
 end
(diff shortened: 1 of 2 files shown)


HenkMutsaerts (changes requested):
This needs more checking.

1. The JSON sidecars don't have a "Sequence" field right? They have different fields from which we disentangle the sequence, as we do in `xASL_adm_DefineASLSequence`
2. You say that this crashes the population module, but the population module should be sequence-ignorant, since a recent change I did, to allow multiple sequences in the population module.
The population module used the 

xASL_init_DefaultEffectiveResolution -> xASL_adm_DefineASLResolution is used in:

ASL module:
* xASL_im_CreateASLDeformationField
* xASL_module_ASL
* xASL_qc_CollectQC_ASL
* xASL_wrp_PreparePV

Population module:
* xASL_im_CreateGroupAnalysisMask

For the ASL module, your changes won't work.
For the population module, there should not be sequence-specific information.
So the question is:
A. is my change already merged in the develop branch and I forgot to fix this?
B. my change still needs merging to the develop branch

jan-petr:
> 1. The JSON sidecars don't have a "Sequence" field right? They have different fields from which we disentangle the sequence, as we do in `xASL_adm_DefineASLSequence`

They do. x.Q.Sequence is created from all JSONs through BIDS2Legacy. This works, I've checked.

HENK: 
x.Q.Sequence is a field containing either:
"3D_GRASE" "3D_spiral" "2D_EPI"

In BIDS2Legacy, you take "PulseSequenceType" which may contain these fields, but may not exist (e.g., see our ExploreASL/external/ example dataset) or may contain something like "EPI" without "2D". So this may seem to work but it doesn't work correctly. We have a separate function xASL_adm_DefineASLSequence for this instead.

> 2. You say that this crashes the population module, but the population module should be sequence-ignorant, since a recent change I did, to allow multiple sequences in the population module.

Yes, you did and it still crashes :) That's why I'm fixing it...

HENK: Nope we didn't communicate optimally about this. I'm repeatingly telling you that the population-module should be sequence-ignorant (so without ever knowing x.Q.Sequence, for the very reasons we discussed; that group analyses should be sequence-ignorant) but you built in an extra exotic reason to know x.Q.Sequence for the native space processing, which I didn't know. I moved all the x.Q.Sequence instances to the ASL module, except for the one in the native space processing part, which I think we discussed and you were fine with needing to fix this later.


>    The population module used the
> xASL_init_DefaultEffectiveResolution -> xASL_adm_DefineASLResolution is used in:
Yes - exactly - this is exactly the function I had to fix. What you haven't realized with your fix is that when running a ASL module, you save x.Q.Sequence. And this remains till the Population. So this was then used and it seemed to you it works. But running population module independently didn't have x.Q.Sequence and crashed. 

So my fix fixes that and loads the specific info for each subject.

HENK: See above, the population module is a group-level, standard-space module that should not contain session-specific processing. Probably best to move this to the ASL module like we discussed previously. So you fixed a symptom it seems :)


> Population module:
> 
> * xASL_im_CreateGroupAnalysisMask
> 
> For the ASL module, your changes won't work.
Why do you think?

HENK: With this remark I meant that PulseSequenceType cannot always translate to x.Q.Sequence. Sometimes you may be lucky when you test this, but it won't work 100%, it was not designed to. See above.

> For the population module, there should not be sequence-specific information. 
See above - there shouldn't be, but it was. You have overlooked something.

Which doesn't matter because we should never query x.Q.Sequence in the population module (except for the native space case which we would fix when we clean this up).

> So the question is: A. is my change already merged in the develop branch and I forgot to fix this? B. my change still needs merging to the develop branch
Yes - all was merged. You have overlooked something. You have thought you are skipping specific sequence resolution, but instead it looked into the obsolete value from ASL module.

I searched quite a while for the problem:

So it seems that you still want to know the sequence type:
xASL_module_Population -> xASL_im_CreateGroupAnalysisMask -> line 140 xASL_adm_DefineASLResolution.


So we have the following options for this:
1. Construct the native space atlas/masks in the ASL module (this would have my preference, because we have a dedicated ASL-run-specific mask wrapper for this already, and it avoids having native space/session-code in a standard space/population-level module)
2. Replacing the code correctly as I explained above. But if you decide this, it still seems quick and dirty unless we can make a specific function or wrapper for it. I'm really in favor of finishing these loose ends before taking up new tasks, we need to be careful because you can already see that this stupid bug costs me half an hour, and this would have been avoided if we would have fixed the native space thingy before.

So I worry that we will get more and more less efficient if we keep our code "dirty", we've improved a lot with this recently with the PWI4D etc.

jan-petr:
I checked again the code and everything.

I agree to remove the sequence-wise processing from the Masks in the standard space. However, to do the population module in the native space, we take the global standard space masks and transform them to the native space of each subject for further processing. While transferring, we need to adapt the masks to the resolution of the native space - we can use the nominal resolution of native space, but best to use the approximate effective resolution of the native space.

I know but ideally the population-module is metadata-independent.

HENK: Construct the native space atlas/masks in the ASL module (this would have my preference, because we have a dedicated ASL-run-specific mask wrapper for this already, and it avoids having native space/session-code in a standard space/population-level module).
JAN: We can't do the native space masks in ASL module. Because for the final analysis, we need the population-wise analysis masks that are only created in the population module. So only after processing all subjects, we create the final masks. So we have to transfer them back to the native space in the population module.

Yes, can we store the effective resolution with the CBF.json sidecar, so it cannot make an error. -> I moved this to #58

HenkMutsaerts (changes requested):
I still don't get it.

xASL_init_DefaultEffectiveResolution -> calls -> xASL_bids_parms2BIDS
xASL_bids_parms2BIDS converts PulseSequenceType -> Sequence

note that in some flavors PulseSequenceType == x.Q.Sequence (e.g. 2D_EPI)
in some flavors PulseSequenceType == 'GRASE' (incomplete)
in some flavors PulseSequenceType == fully missing

You told me that xASL_bids_parms2BIDS calls xASL_adm_DefineASLSequence, but xASL_adm_DefineASLSequence is only called in the beginning of the ASL module, afaik.

(xASL_adm_DefineASLSequence takes readoutdim & vendor etc and creates 2D_EPI 3D_GRASE etc in x.Q.Sequence). 

So this still needs fixing, as we cannot assume that xASL_bids_parms2BIDS can generate x.Q.Sequence (even though in certain flavors it can!)

Modules/Module_Processing/xASL_init_DefaultEffectiveResolution.m
```diff
@@ -30,14 +30,39 @@
...
+% Read JSON of the ASL scan
+tJson = xASL_io_ReadJson([fullfile(fPath, fName) '.json']);
+tJson = xASL_bids_parms2BIDS([], tJson, 0);
+
+% Obtain the resolution based on the image matrix
 NativeResolution = tIM.hdr.pixdim(2:4);
-if ~isfield(x.Q,'Sequence')
-    warning(['Setting x.Q.Sequence missing, skipping']);
+
+% Load the x.Q fields from the JSON
```
HenkMutsaerts, line 47: This is nice but I would add this to 

xASL_bids_parms2BIDS

Because PulseSequenceType -> Sequence

inside xASL_bids_parms2BIDS (which may not be correct) and there is more flexibility to do this list in parms2BIDS rather than only here. 

Here we initialize the default effective resolution. We don't translate parms2BIDS (sequence, readoutdim, vendor, is all translation).

HenkMutsaerts (changes requested):
Sorry I thought I explained this clearly but I probably didn't?

PR #1656, opened 2024-03-25
#1652 Import TR-vector for M0

### LINKED ISSUE #1652: Automatic reading of TR for a built-in M0
### Description

Flavors tgse_pcasl_ve11c with built-in M0 have a special problem. This particular sequence sets M0 TR to 2s, but the TR of ASL is around 4s. Either way, the import ignores the possibility toaumatically detect M0 posiitons and scaling.

This specifically concerns Flavors Siemens_PCASL_3DGRASE_VE11C_3, PCASL_multiSequence_VE11C_1, PCASL_multiSequence_VE11C_2 
And casl_818 sequences have TR 4s in Phoenix

### Tasks

- [x] Read and save TR for each volume
- [x] Test all flavors working
- [x] Read TR from Phoenix and save to the M0scan for casl818 nad pcasl_ve11c
- [x] Make sure that TR is correctly splitted in BIDS2Legacy
- [x] Implement TR reading for PCASL_VE11c
- [x] Fix GE-RX28 Image type
- [x] Test all flavors working
- [x] Inform Catherine Morgan upon merging

### Release notes

Import Siemens with built-in M0 with different TR.

jan-petr:
### Linked issue

Close #1652


jan-petr:
COMPARISON TABLE:]
                    flavor                     dataset                name                                                  message                                   
    _______________________________________    ________    __________________________    _____________________________________________________________________________

    {'GE_PCASL_3Dspiral_RX28.0_multiPLD_1'}    {'Both'}    {'Different file content'}    {'Different file content: /sub-sub1/perf/sub-sub1_run-6_asl.json'           }
    {'GE_PCASL_3Dspiral_RX28.0_multiPLD_1'}    {'Both'}    {'Different file content'}    {'Different file content: /ExploreASL/sub-sub1_1/ASL_6/ASL4D.json'          }
    {'GE_PCASL_3Dspiral_RX28.0_multiPLD_1'}    {'Both'}    {'Different file content'}    {'RMSE of NIFTIs above threshold: /ExploreASL/sub-sub1_1/ASL_6/ASL4D.nii.gz'}
    {'GE_PCASL_3Dspiral_RX28.0_multiPLD_1'}    {'Both'}    {'Different file content'}    {'Matrix dimensions do not agree: /ExploreASL/sub-sub1_1/ASL_6/ASL4D.nii.gz'}
    {'GE_PCASL_3Dspiral_RX28.0_multiPLD_1'}    {'Both'}    {'Different file content'}    {'Different file content: /ExploreASL/sub-sub1_1/ASL_6/ASL4D_Source.json'   }
    {'GE_PCASL_3Dspiral_RX28.0_multiPLD_1'}    {'Both'}    {'Different file content'}    {'RMSE of NIFTIs above threshold: /ExploreASL/sub-sub1_1/ASL_6/M0.nii.gz'   }
    {'GE_PCASL_3Dspiral_RX28.0_multiPLD_1'}    {'Both'}    {'Different file content'}    {'Matrix dimensions do not agree: /ExploreASL/sub-sub1_1/ASL_6/M0.nii.gz'   }

Modules/Module_Import/NII2BIDS/xASL_bids_BIDSifyASLJSON.m
```diff
@@ -526,9 +526,15 @@
 ASLContextM0Index = cellfun(@(x)~isempty(x),ASLContextM0Index); % Create a vector out of it
 ASLContextM0Index = ASLContextM0Index(1:dimASL(4)); % Remove the last empty field
 
+% If ASLContext contains M0-scans, and repetitionTimePreparation vector is a scalar, and RepetitionTimeM0 is defined, then we create a repetitionTimePreparation vector
```
HenkMutsaerts, line 529: No I would always create a vector, for all parameters for all DICOMs. And only if it is unique/the same for all, we use it in the old way, otherwise we create a vector in BIDS. Would that not be more general (less new surprises for us)?

Modules/Module_Import/NII2BIDS/xASL_bids_PhoenixProtocolAnalyzer.m
```diff
@@ -151,6 +153,10 @@
         bidsPar.ScanType = 'M0';
 	end
     
+	if isfield(sourcePar,'sProtConsistencyInfotBaselineString') && ~isempty(sourcePar.sProtConsistencyInfotBaselineString)
```
HenkMutsaerts, line 156: All nice, just a comment, you know that nobody is able to read your code right?
I mean I can understand a bit what you're doing, but I don't know what each field consists. So if you want you could add a comment that tells the reader what each field does, so others could read this code as well if you would become ill or decide to stop working ;)

Just a comment, you can resolve

HenkMutsaerts (approved):
Please check my 2 comments, this is perhaps something you would want to implement later.

The import **should always assume** that M0 has **different** parameters than ASL, even if they are in the same acquisition.

Moreover: the best/most generic way, would be to check each DICOM file separately for its settings/parameters. I know that this will take more time, but theoretically a control file can have different settings than a label file. And if you have a multi-PLD PCASL file where each control-label pair has different PLD settings, then you check those as well right?

Modules/Module_Import/NII2BIDS/xASL_bids_PhoenixProtocolAnalyzer.m
jan-petr: Good point. Comments added.

PR #1657, opened 2024-03-26
Feature #1655 new isnear

### LINKED ISSUE #1655: Implement xASL_stats_isnear
### Description

Implement an own version of isnear and replace all occurences.

### Tasks

- [x] Create a new function xASL_stat_Isnear
- [x] Replace all occurences
- [x] Remove isnear and all directories related

### Release notes

A new function for checking near equivalence of numerical arrays.

jan-petr:
### Linked issue

Close #1655


Functions/Statistics/xASL_stat_Isnear.m
```diff
@@ -0,0 +1,56 @@
...
+
+
+% Admin
+if nargin < 2 || isempty(x) || isempty(y)
+    error('Two non-empty inputs are required')
+end
+
+% Default tolerance
+if nargin < 3 || isempty(tol)
+   tol=1e-8;
```
HenkMutsaerts: needs a space

Functions/Statistics/xASL_stat_Isnear.m
```diff
@@ -0,0 +1,56 @@
...
+end
+
+% Check if X and Y inputs are the same size or scalars
+if ~isequal(size(x), size(y)) && numel(x) > 1 && numel(y) > 1
+   error('Inputs X and Y must be the same size or either of them must be a scalar');
+end
+
+% Check if tol is a scalar or has the same size as X and Y
+if numel(tol) > 1 && ( (numel(x) > 1 && ~isequal(size(tol), size(x))) || (numel(y) > 1 && ~isequal(size(tol), size(y))) )
+	error('Input TOL must have the same size as X or Y or be a scalar');
```
HenkMutsaerts: I would turn the error around 1) tol needs to be a scalar (default) OR 2) tol has equal size (exotic version)

Functions/Statistics/xASL_stat_Isnear.m
HenkMutsaerts: Also, should the default tolerance not be eps? That seems less arbitrary? Or is eps 1^e-8 :)

HenkMutsaerts (changes requested):
Minor stuff.
More importantly: I would tweak nomenclature so it is less obvious that this is a copy of the original. E.g., 

x -> inputA (or a)
y -> inputB (or b)
tol -> tolerance

isnear -> approximates OR isCloseBy OR isNearby

Functions/Statistics/xASL_stat_Isnear.m
jan-petr: I removed EPS to remove similarities. So 1e-8 is the value isnear was really using.
Space added.

jan-petr:
> Minor stuff. More importantly: I would tweak nomenclature so it is less obvious that this is a copy of the original. E.g.,
> 
> x -> inputA (or a) y -> inputB (or b) tol -> tolerance

I use X and Y because the original used A and B :)
tol is a common keyword, so no need to use tolerance.
 
> isnear -> approximates OR isCloseBy OR isNearby

Agree, but those names sound lame. I would use "IsEqualTol"

Functions/Statistics/xASL_stat_Isnear.m
jan-petr: Done

HenkMutsaerts (approved):

PR #1661, opened 2024-03-27
Feature #1503 aggregate lesion atlases

### LINKED ISSUE #1503: ROI/lesion statistics creates too many tsv files
### Description

When there are ROI/lesion NIfTI files, instead of a single TSV file with ROI statistics, multiple TSV files will be created, once for every participant. We need to create one TSV per Lesion|ROI.

So `xASL_im_Lesion2Mask` takes `(ROI|Lesion)_(T1|FLAIR).nii` and creates several ROIs: 1. Intralesional, 2. Perilesional, 3. Rest of the hemisphere minus (1) and (2). 4—6 the same as 1—3 but then in the contralateral hemisphere.

These are then stored in the Population folder, for each subject. And they go to the `xASL_wrp_GetROIStatistics` which considers them as atlases/ROIs. This function iterates over all subjects, so for each ROI/lesion, a TSV file is created for all subjects.

The easiest would be to create a separate function that merges these TSV files afterwards. More difficult, but perhaps cleaner, would be to have a subject-specific masking option in `xASL_stat_GetROIStatistics`, which would then read the atlas name for a subject-ID, and searches the data-list for the same subject-ID. The latter would require changing `xASL_stat_AtlasForStats` to iterate over subjects as well.

Jan & Henk agreed:
1. xASL_stat_AtlasForStats into a xASL_stat_LesionForStats
2. xASL_stat_GetROIStatistics is called once for each subject with the specific lesion

### Tasks
- [x] Remove xASL_adm_OutputDifferentAtlasResults as we don't use it (Michael created this)
- [x] Create a testing scenario
- [x] Modify xASL_stat_AtlasForStats instead of create new xASL_stat_LesionForStats. xASL_stat_AtlasForStats has many functions and it would be very difficult to maintain two functions. So instead, we modify this function that it is either called as 'default' - map a new atlas. Or 'update' which doesn't run the overhead, but it replaces the current lesion atlas with new Lesion for the given subject.
HENK: I would instead have a subject-specific atlas option, where it reads 1 atlas per subject in the 4rd dimension (currently the 4rd dim are the different ROIs) or 5th dimension (so 4rd dim can stay intra/peri/etc) -> 5th dim is perhaps easiest
- [x] xASL_stat_AtlasForStats is run only once for all subjects. But for Lesions, we will update xASL_stat_AtlasForStats 
       for each subject -> see modif above.
- [x] So xASL_wrp_GetROIstatistics can stay virtually unchanged
- [x] xASL_stat_GetROIstatistics will be modified for Lesions|ROIs to always update the atlas for each subject (or report 
       zero if an atlas is missing).
-> do we need a bSubjectSpecificROIs boolean also here? Would be good to have the same boolean throughout the 
       population module (or in xASL_wrp_GetROIstatistics at least)
- [x] name of the ROI should be "intralesional" instead of "intratumoral" V
- [x] values should not be only GM, but also WM -> yes the whole ROI
        BTW: somebody made an x.S.bMasking, I cannot remember this. I like the idea, but I need to check if it is 
        implemented well with the way I set it up
- [x] Use ROI names from xASL_im_Lesion2Mask
- [x] Skip missing ROI for P03 in T1
- [x] Find differences in Lesion_T1 P01 and P02 - why are most ROIs 0? Something wrong with left-sided ROIs??
- [x] Lesion by default doesn't use GM, ROIs do.
- [x] Use proper Lesion_FLAIR name

### How to test

- [x] Test & fix this for a single lesion
- [x] Test & fix this for multiple lesion/ROIs

### Release notes

ROI/Lesion masks produce a single stat-file per mask.

jan-petr:
### Linked issue

Close #1503


Modules/xASL_module_Population.m
```diff
@@ -256,6 +256,7 @@
 
             % ROI statistics (default: standard space)
             x.S.InputNativeSpace = 0;
+			x.S.bSubjectSpecificAtlas = false;
```
HenkMutsaerts: I would call it "SubjectSpecificROIs" rather than atlas

Modules/xASL_module_Population.m
```diff
@@ -265,25 +266,25 @@
             end
         end
     
-        % Check if we should do the same for Lesion or ROI masks (i.e.
-        % individual "atlases") -> NB not yet developed/tested in native space
-        LesionROIList = xASL_adm_GetFileList(x.D.PopDir, '(?i)^r(Lesion|ROI).*\.nii$', 'FPList', [0 Inf]);
+        % Check if we should do the same for Lesion or ROI masks (i.e. individual "atlases")
+		% PM: not yet developed/tested in native space
```
HenkMutsaerts, line 276: The native space (quick & dirty) fix is subject-specific inverse application of y_ASL.nii anyway, so it should be relatively easy to fix this, if the "ROI path" becomes a regular expression as I said before.

So for native space, you have ROIdir & ROIregexp (extension is always .mat OR .nii.gz).
ROIregexp would be either '^MindBoggle$' OR '^rLesion_sub-.*'.

But we want to be sure that we only process the subjects that we loaded as folders, not all the subjects that exist?

(BTW: we could easily move from getting x.SUBJECTS from foldernames to just the first column of participants.tsv, that would make things a lot easier for both us and the users (instead of all the exclusion and forceLoading parameters. Do you agree? Then we can make this an issue.)

Modules/xASL_module_Population.m
```diff
@@ -265,25 +266,25 @@
...
         end
     
-        % Check if we should do the same for Lesion or ROI masks (i.e.
-        % individual "atlases") -> NB not yet developed/tested in native space
-        LesionROIList = xASL_adm_GetFileList(x.D.PopDir, '(?i)^r(Lesion|ROI).*\.nii$', 'FPList', [0 Inf]);
+        % Check if we should do the same for Lesion or ROI masks (i.e. individual "atlases")
+		% PM: not yet developed/tested in native space
+        
+		% Read the names of the lesion files
+		LesionROIList = xASL_adm_GetFileList(x.D.PopDir, '(?i)^r(Lesion|ROI).*\.nii$', 'List', [0 Inf]);
```
HenkMutsaerts: should be

`LesionROIList = xASL_adm_GetFileList(x.D.PopDir, '(?i)^r(Lesion|ROI)_(T1|FLAIR)_\d_.*\.nii$', 'List', [0 Inf]);`

Modules/xASL_module_Population.m
```diff
@@ -265,25 +266,25 @@
...
+			LesionROIList{iLesion} = LesionROIList{iLesion}(iStart:iEnd);
+		end
+		% Obtain a unique list of lesion names without the subject name
+		LesionUniqueROIList = unique(LesionROIList);
+
         x.S.InputNativeSpace = 0;
-        for iAtlas = 1:length(LesionROIList)
-            x.S.InputAtlasPath = LesionROIList{iAtlas};
+		x.S.bSubjectSpecificAtlas = true;
+        for iAtlas = 1:length(LesionUniqueROIList)
```
HenkMutsaerts: Also here, iROI is more correct than iAtlas

Modules/xASL_module_Population.m
```diff
@@ -265,25 +266,25 @@
...
-        % Check if we should do the same for Lesion or ROI masks (i.e.
-        % individual "atlases") -> NB not yet developed/tested in native space
-        LesionROIList = xASL_adm_GetFileList(x.D.PopDir, '(?i)^r(Lesion|ROI).*\.nii$', 'FPList', [0 Inf]);
+        % Check if we should do the same for Lesion or ROI masks (i.e. individual "atlases")
+		% PM: not yet developed/tested in native space
+        
+		% Read the names of the lesion files
+		LesionROIList = xASL_adm_GetFileList(x.D.PopDir, '(?i)^r(Lesion|ROI).*\.nii$', 'List', [0 Inf]);
+		% Go through the lesions and remove the subject names
+		for iLesion = 1:length(LesionROIList)
```
HenkMutsaerts: Simplification:

[~, iEnd] = cellfun(@(y) regexp(y, '^r(Lesion|ROI)_(T1|FLAIR)_\d'), LesionROIList);

for iLesion = 1:length(LesionROIList)
    LesionROIList = LesionROIList{iLesion}(1:iEnd);
end

Modules/Module_Processing/Module_Population/xASL_wrp_GetROIstatistics.m
```diff
@@ -87,10 +92,29 @@ function xASL_wrp_GetROIstatistics(x)
...
-x = xASL_stat_AtlasForStats(x); % check all atlases that are requested (see InputAtlasPath above)
-% check them, check their ROI names, and make them ready for use below
-
+if ~x.S.bSubjectSpecificAtlas
+	x = xASL_stat_AtlasForStats(x); % check all atlases that are requested (see InputAtlasPath above)
+	% check them, check their ROI names, and make them ready for use below
 
+	% Skip atlas loading here for subject-specific atlases
+else
+	% In case of a subject specific atlas, load the subject-specific atlas for any random existing subject to be able to create the ROI names
```
HenkMutsaerts: In the case of subject-specific ROIs/lesions

and remove the "random existing subjects" :)


More importantly:
It seems to me that this is too complicated.

xASL_stat_AtlasForStats does:
1. Loading atlas names, we don't have this for the lesions/ROIs, this becomes much to convolved, and these are user-specified anyway, so the user should know what ROI 1 2 3 4 etc stand for. So we don't need this for subject-wise
2. Deal with memory-mapping: we don't use memory-mapping anymore
3. Resample atlas to 1.5 isotropic, we don't need this because we have transformed the lesions & ROIs to standard space ourselves
4. Convert atlas to 4D binary image, we don't need this, each ROI or lesion is done separately (see iAtlas loop in xASL_module_Population)
5. Convert/compress masks into columns -> we currently need this, but you are removing this necessity in the other issue. So if this is the only requirement for running xASL_stat_AtlasForStats subject-wise, I would do this other issue first so we don't have to run xASL_stat_AtlasForStats.
6. Print atlas-overview image: we don't need this, because we already print a QC overview subject-wise elsewhere


CONCLUSION: 
1. I would fix the other issue first, in which you optimize the population-module NIfTI loading, and remove the conversion to a column. Then we need to remove the conversion to column as well throughout the whole xASL_module_Population in that issue, including the atlases/ROIs (because the ROIs need to be in the same format as the data anyway).
2. Then we don't need xASL_stat_AtlasForStats for subject-wise ROIs, which makes it a lot easier

HenkMutsaerts (changes requested):
1. I would fix the other issue first, in which you optimize the population-module NIfTI loading, and remove the conversion to a column. Then we need to remove the conversion to column as well throughout the whole xASL_module_Population in that issue, including the atlases/ROIs (because the ROIs need to be in the same format as the data anyway).
2. Then we don't need xASL_stat_AtlasForStats for subject-wise ROIs, which makes it a lot easier

(see inline explanations)

jan-petr:
> 1. I would fix the other issue first, in which you optimize the population-module NIfTI loading, and remove the conversion to a column. Then we need to remove the conversion to column as well throughout the whole xASL_module_Population in that issue, including the atlases/ROIs (because the ROIs need to be in the same format as the data anyway).
> 2. Then we don't need xASL_stat_AtlasForStats for subject-wise ROIs, which makes it a lot easier
> 
> (see inline explanations)

0. BTW: Best to make all comments in the code so that they are easier to reply.

1. Conversion to a column - yes, lets make that as a separate issue and then we simplify things. But we need to fix this now and can work on the columns later.

2. AtlasForStats will indeed get much simpler, but we will always have to load atlases/ROIs anyway. 

As a summary - I agree with your solution. But lets finish this and immediately make a new issue for the rest that we will still do in 1.12.0.

Modules/xASL_module_Population.m
jan-petr: Done. I used ROI instead of ROIs, because in every iteration, we do a single ROI.

Modules/xASL_module_Population.m
jan-petr: About the native space - yes. It should be easy and we scheduled this in a new issue. First I fix #58 and then native ROIs. 

Processing only those subject which have ROIs is another point in the issue indeed.

About getting x.SUBJECTS - not disagreeing, but perhaps not understanding fully. Can you make that an issue and describe more details?

Modules/xASL_module_Population.m

Modules/xASL_module_Population.m
jan-petr: We already have iLesion, you might have looked into an older commit.

Modules/xASL_module_Population.m
jan-petr: Done partly. Cause in your example, there would be a variable iEnd for different atlas types.

Modules/Module_Processing/Module_Population/xASL_wrp_GetROIstatistics.m
jan-petr: will check later....

jan-petr: > In the case of subject-specific ROIs/lesions
> and remove the "random existing subjects" :)
Agree. Done.

> More importantly: It seems to me that this is too complicated.
> 
> xASL_stat_AtlasForStats does:
> 
> 1. Loading atlas names, we don't have this for the lesions/ROIs, this becomes much to convolved, and these are user-specified anyway, so the user should know what ROI 1 2 3 4 etc stand for. So we don't need this for subject-wise
> 2. Deal with memory-mapping: we don't use memory-mapping anymore
> 3. Resample atlas to 1.5 isotropic, we don't need this because we have transformed the lesions & ROIs to standard space ourselves
> 4. Convert atlas to 4D binary image, we don't need this, each ROI or lesion is done separately (see iAtlas loop in xASL_module_Population)
> 5. Convert/compress masks into columns -> we currently need this, but you are removing this necessity in the other issue. So if this is the only requirement for running xASL_stat_AtlasForStats subject-wise, I would do this other issue first so we don't have to run xASL_stat_AtlasForStats.
> 6. Print atlas-overview image: we don't need this, because we already print a QC overview subject-wise elsewhere
> 
> CONCLUSION:
> 
> 1. I would fix the other issue first, in which you optimize the population-module NIfTI loading, and remove the conversion to a column. Then we need to remove the conversion to column as well throughout the whole xASL_module_Population in that issue, including the atlases/ROIs (because the ROIs need to be in the same format as the data anyway).
> 2. Then we don't need xASL_stat_AtlasForStats for subject-wise ROIs, which makes it a lot easier

Regarding 1-6 and 1-2. I fully agree. Except for the order. We have this issue fixed already and we need it for several tumor studies ASAP. So I would finalize, test, and merge this. And then I'll immediately start working on removing the columns and all you mention above. Do you agree?

Modules/xASL_module_Population.m
HenkMutsaerts: x.SUBJECTS -> #1669

Modules/xASL_module_Population.m
HenkMutsaerts: Yes sorry, should have been:

LesionROIList = LesionROIList{iLesion}(1:iEnd(iLesion));

Modules/xASL_module_Population.m
jan-petr: So this can be resolved, right?

Modules/xASL_module_Population.m

Modules/xASL_module_Population.m
jan-petr: I did some tests and it crashed with some empty ROIs, so going through them 1-by-1 seems to be easier to debug.

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
```diff
@@ -178,7 +177,9 @@
 	end
 
 else
-	x.S.InputMasks = logical(x.S.InputMasks);
+	if ~x.S.bSubjectSpecificROI
```
HenkMutsaerts: Why do we need this? Quick and dirty...

Atlas processing should be the same for single or multiple masks. Otherwise, you could move this down, where the mask needs to be logical. Or you add the similar %%% here as well :)

HenkMutsaerts: You can resolve this

HenkMutsaerts (approved):
See my last comment, and I added some comments for the future issues to improve the code

Issue #1637, opened 2024-03-07
Remove the functionality _parms.mat


jan-petr:
### Description

Legacy _parms.mat is not really used already in several versions. So for several years, we haven't created a _parms.mat and it is only there for backwards compatibility. Currently, there is extremely small probability of having a dataset with _parms.mat.

### Tasks

- [ ] We convert the original _parms.mat files into json early in the ASL module, and then remove the _parms_mat. We issue a warning. 
- [ ] Then everywhere in the code, we remove the _parms.mat functionality so that any conversion will happen only once at the start and then only JSONs are used.
- [ ] in xASL_io_ASLSubtractionAveraging we will only use JSONs and we phase out LoadParmsMat as something we don't use anymore.
- [ ] DefineASLSequence is hidden inside xASL_module_ASL_ParseParameters -> LoadParms -> BIDS2Parms -> DefineASLSequence. Which makes it hard to follow. We need to simplify this while getting rid of the parms.mat reading.

HENK: I would remove the warning and do this in version 2.0.0, just as a clean up and not having backward compatibility. Anouk Schrantee is the only one who had this afaik, and also she is now using the new BIDS import, so we should be safe.
HENK2: Let's save the conversion code (parms.mat->json) though for old users. Or they can just take an older version and run xASL_bids_parms2BIDS or something.

### How to test

Optional: insert description about how to test the code changes here

### Release notes

Required: summarize the changes for the release notes here

PR #1665, opened 2024-03-28
Bug #1648 sequence dep native

### LINKED ISSUE #1648: Sequence-dependency native-space population module bug

jan-petr:
### Linked issue

Close #1648 

I apply the function xASL_adm_DefineASLSequence to all BIDS2Legacy conversions. so x.Q.Sequence should always be available.

### DIFF
--- a/Functions/Administration/xASL_adm_DefineASLSequence.m
+++ b/Functions/Administration/xASL_adm_DefineASLSequence.m
@@ -1,44 +1,44 @@
-function [x] = xASL_adm_DefineASLSequence(x, bVerbose)
+function [xQ] = xASL_adm_DefineASLSequence(xQ, bVerbose)
 %xASL_adm_DefineASLSequence Obtain ASL sequence type for sequence-specific
 %image processing
 %
-% FORMAT: [x] = xASL_adm_DefineASLSequence(x, bVerbose)
+% FORMAT: [xQ] = xASL_adm_DefineASLSequence(xQ, bVerbose)
 %
 % INPUT:
-%   x                - x structure containing all input parameters (REQUIRED)
-%   x.Q.readoutDim   - dimensionality of readout (2D or 3D) (OPTIONAL)
-%   x.Q.Vendor       - Either 'GE', 'Philips', 'Siemens' (OPTIONAL)
+%   xQ                - x.Q structure containing all input parameters (REQUIRED)
+%   xQ.readoutDim   - dimensionality of readout (2D or 3D) (OPTIONAL)
+%   xQ.Vendor       - Either 'GE', 'Philips', 'Siemens' (OPTIONAL)
 %   bVerbose         - verbose output (OPTIONAL, DEFAULT=true)
 %
 % OUTPUT:
-%   x               - x structure containing all output parameters
-%   x.Q.Sequence    - sequence type (readout)
+%   xQ               - x structure containing all output parameters
+%   xQ.Sequence    - sequence type (readout)
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
 % DESCRIPTION: This ExploreASL function tries to check what ASL sequence is
-% being processed, if this was not already defined in x.Q.Sequence.
+% being processed, if this was not already defined in xQ.Sequence.
 % It does so by checking known combinations of readout dimensionality
-% (x.Q.readoutDim) and Vendor, knowing the product sequences of the Vendors.
+% (xQ.readoutDim) and Vendor, knowing the product sequences of the Vendors.
 %
-% EXAMPLE: x = xASL_adm_DefineASLSequence(x);
+% EXAMPLE: xQ = xASL_adm_DefineASLSequence(xQ);
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
-% Copyright 2015-2023 ExploreASL
+% Copyright 2015-2024 ExploreASL
 
 
-%% Check quantification fields readoutDim & x.Q.Sequence
+%% Check quantification fields readoutDim & xQ.Sequence
 if nargin<2 || isempty(bVerbose)
     bVerbose = true;
 end
 
-if ~isfield(x.Q, 'readoutDim') || isempty(x.Q.readoutDim)
-    warning('x.Q.readoutDim parameter missing');
+if ~isfield(xQ, 'readoutDim') || isempty(xQ.readoutDim)
+    warning('xQ.readoutDim parameter missing');
 end
 
 % Check for illegal sequence definitions
-if isfield(x.Q,'Sequence')
-    if isempty(regexpi(x.Q.Sequence, '(2d_epi|3d_grase|3d_spiral)'))
-        warning(['Unknown ASL sequence: ' xASL_num2str(x.Q.Sequence)]);
+if isfield(xQ,'Sequence')
+    if isempty(regexpi(xQ.Sequence, '(2d_epi|3d_grase|3d_spiral)'))
+        warning(['Unknown ASL sequence: ' xASL_num2str(xQ.Sequence)]);
         fprintf('%s\n', 'Trying to fix this');
-        x.Q = rmfield(x.Q, 'Sequence');
+        xQ = rmfield(xQ, 'Sequence');
 
     end
 end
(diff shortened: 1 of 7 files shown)


Functions/BIDS/xASL_bids_parms2BIDS.m
```diff
@@ -345,6 +345,10 @@
     outParms = xASL_io_CheckDeprecatedFieldsX(outParms);
 end
 
+% If we convert to legacy, we make sure to properly initialize the field Q.Sequence
```
HenkMutsaerts: should be "the field x.Q.Sequence"

Functions/Administration/xASL_adm_DefineASLSequence.m
```diff
@@ -1,44 +1,44 @@
...
+    warning('xQ.readoutDim parameter missing');
 end
 
 % Check for illegal sequence definitions
-if isfield(x.Q,'Sequence')
-    if isempty(regexpi(x.Q.Sequence, '(2d_epi|3d_grase|3d_spiral)'))
-        warning(['Unknown ASL sequence: ' xASL_num2str(x.Q.Sequence)]);
+if isfield(xQ,'Sequence')
+    if isempty(regexpi(xQ.Sequence, '(2d_epi|3d_grase|3d_spiral)'))
+        warning(['Unknown ASL sequence: ' xASL_num2str(xQ.Sequence)]);
```
HenkMutsaerts, line 39: You may get this warning a lot since we translate PulseSequenceType -> Sequence inside bids_parms2BIDS (somewhere in the general renaming)

Modules/Module_Processing/Module_Population/xASL_im_CreateGroupAnalysisMask.m
```diff
@@ -241,25 +232,6 @@
 x.S.CorSlices = [];
```
HenkMutsaerts, line 233: We have a separate visualization init function where these stuff should live. Not important for now, just that you know: `xASL_init_VisualizationSettings`

Modules/Module_Processing/xASL_init_DefaultEffectiveResolution.m
```diff
@@ -30,15 +30,23 @@
...
 %% ----------------------------------------------------------------------------------------
 %% Admin
-tIM = xASL_io_ReadNifti(PathASL);
+[tIM, tJSON] = xASL_io_ReadNifti(PathASL);
 NativeResolution = tIM.hdr.pixdim(2:4);
-if ~isfield(x.Q,'Sequence')
-    warning(['Setting x.Q.Sequence missing, skipping']);
+
+% Obtain the x.Q.Sequence field from the json
+if ~isempty(tJson) && isfield(tJson, 'Q') && isfield(tJson.Q, 'Sequence')
```
HenkMutsaerts, line 41: We still need to run xASL_bids_parms2BIDS here right, in case we're loading a BIDS json?

Functions/BIDS/xASL_bids_parms2BIDS.m
HenkMutsaerts: Also, does xASL_bids_parms2BIDS now move quantification fields like Sequence, PLD, etc to JSON.Q.fields or to JSON.fields? Because BIDS is JSON.fields...

So perhaps we should add either x.Q.Sequence or JSON.Sequence. But not JSON.Q.Sequence because we do this for none of the parameters right?

HenkMutsaerts (changes requested):
xASL_bids_parms2BIDS still has on top the translation from PulseSequenceType -> Sequence, that probably needs tweaking as well?

Modules/Module_Processing/xASL_init_DefaultEffectiveResolution.m
jan-petr: Haha. We don't. Cause BIDS2legacy conversion is an integral part of io_readNifti :) So bids_parms2BIDS is called inside the JSON loading. There was only a small bug in the code.

Modules/Module_Processing/Module_Population/xASL_im_CreateGroupAnalysisMask.m
jan-petr: OK. Good to know. I wasn't really checking the old code except for removing comments that are there for ages :)

Functions/Administration/xASL_adm_DefineASLSequence.m
jan-petr: True. And it can become annoying NEvertheless. If x.Q.Sequence is provided, it should be correct. And it shouldn't really happen that we don't have vendor+readoutDim.

Functions/BIDS/xASL_bids_parms2BIDS.m
jan-petr: 1. I've changed that to outParms.Q.Sequence. While it will end up in x.Q.Sequence, in theory, we just provide output as outParms and it's up to the user to add it to the x-struct or not. 
2. parms2BIDS does both BIDS->Legacy and Legacy->BIDS. So when doing Legacy->BIDS, it is all in JSON.fields. When doing BIDS->Legacy, some fields are in JSON.fields and some in JSON.Q.fields, depending on the Legacy format. Note that we don't use x or JSON, we call it outParms for simplicity.

I hope this makes sense. I've improved the comments.

Modules/xASL_module_ASL.m
```diff
@@ -874,9 +874,4 @@
...
 end
 
-
-%% 7. Define sequence (educated guess based on the Q field)
-x = xASL_adm_DefineASLSequence(x, bOutput);
-
-
-end
\ No newline at end of file
+end
```
HenkMutsaerts, line 877: This is not backward compatible, I'm fine with this, just that we notice this and make a full new version 2.0.0 and remove the whole _parms.mat stuff. Also, in the past, we added such sequence details to dataPar.json

HenkMutsaerts: Also see my main comment, we need to either load x.Q.Sequence at 6 locations throughout the xASL_module_ASL or keep this. So probably best to keep this here (replace by the [~, ~, json] = xASL_io_ReadNifti ...) and remove at the end

HenkMutsaerts (changes requested):
Just tested it, it works well, two residual points:

MAJOR
We use x.Q.Sequence on several points in ASL module:

xASL_wrp_RegisterASL
xASL_im_ASLSubtractionAveraging
xASL_im_CreatePseudoCBF
xASL_im_PVC_HiRes
xASL_wrp_CreateIndividualMask
xASL_wrp_ProcessM0
xASL_wrp_RegisterASL

So we either load it from the JSON every time, or we load it once at the beginning of the module and delete it at the end. I think the latter is probably easier, because there shouldn't be conflicts in x.Q.Sequence (unless merging different sequences but that probably doesn't happen).

Advantage of loading it every time: is that we can use these functions separately without xASL_module_ASL, but this probably doesn't happen now.

MINOR
* The loaded JSON has some quantification stuff in the Q field (but also non-quantification stuff like SoftwareVersions), and other quantification stuff in json. Would be good to harmonize this, but we probably want to do that when we fully go to BIDS (and design ASL-BIDS derivatives)

Modules/xASL_module_ASL.m
jan-petr: Don't worry. I have checked that to make sure it *is* backwards compatible.

In module_ASL at line 133, we call xASL_module_ASL_ParseParameters and this function makes sure that x-struct is loaded and parms.mat is loaded and everything from ASL4D.json is loaded. So I've only removed ASLDefineSequence because I knew that at the start of the ASL module, we will load the x.Q.Sequence from the JSON through parms2BIDS.

I agree that the whole stuff is convoluted etc. and we will get rid of _parms.mat in the future. But for now, we are remaining perfectly backwards compatible.

So this answers your questions and is in line with your suggestion, but we already have all that.

jan-petr:
> MAJOR We use x.Q.Sequence on several points in ASL module:
> xASL_wrp_RegisterASL xASL_im_ASLSubtractionAveraging xASL_im_CreatePseudoCBF xASL_im_PVC_HiRes xASL_wrp_CreateIndividualMask xASL_wrp_ProcessM0 xASL_wrp_RegisterASL

I know about this. That's why we initialize x.Q.Sequence at the start of the ASL module.

> So we either load it from the JSON every time, or we load it once at the beginning of the module and delete it at the end. I think the latter is probably easier, because there shouldn't be conflicts in x.Q.Sequence (unless merging different sequences but that probably doesn't happen).

Yes. the latter is the best and we already have that.
 
> Advantage of loading it every time: is that we can use these functions separately without xASL_module_ASL, but this probably doesn't happen now.

Fully agree. We load it at the start rather than in the middle as indeed this option you describe is not likely. Wrappers are not really for separate use as they depend on having a correctly initialized x-struct.
 
> MINOR
> 
> * The loaded JSON has some quantification stuff in the Q field (but also non-quantification stuff like SoftwareVersions), and other quantification stuff in json. Would be good to harmonize this, but we probably want to do that when we fully go to BIDS (and design ASL-BIDS derivatives)

Fully agree. This is a mess, but that's historically a mess. And we have several issues for simplifying this in the future on the way to full BIDS compatibility. So we are slowly moving in that direction.

Modules/xASL_module_ASL.m
HenkMutsaerts: I see, then we should only (now or in the future) make this more transparent, this what you're talking about is hidden in `xASL_module_ASL_ParseParameters` -> LoadParms -> BIDS2Parms -> DefineASLSequence.

Probably when we remove the parms.mat it is easier to simplify this. Should we move this to that issue? You can resolve this:)

HenkMutsaerts (changes requested):
Looks all nice, only remaining (potential) bug is that `xASL_bids_parms2BIDS` line 74 we need to remove the translation from PulseSequenceType -> Sequence as this will give problems.

So I would decide:
1. BIDS2parms -> defineASLSequence which takes readoutdim, PulseSequenceType, etc to create the legacy field "Sequence"

OPTIONS:
1. we keep PulseSequenceType (called PulseSequenceType)
2. we remove PulseSequenceType

BUT we don't literally translate PulseSequenceType -> Sequence because this is bug-prone/confusing, etc, even if it is temporarily.

(Note that because there is no BIDS derivatives for ASL yet, we're not really illegal by having non-BIDS json files, though this could also be confusing to users)

Modules/xASL_module_ASL.m
jan-petr: FUlly agree and added that to #1637

jan-petr:
> Looks all nice, only remaining (potential) bug is that `xASL_bids_parms2BIDS` line 74 we need to remove the translation from PulseSequenceType -> Sequence as this will give problems.
> 
> So I would decide:
> 
> 1. BIDS2parms -> defineASLSequence which takes readoutdim, PulseSequenceType, etc to create the legacy field "Sequence"
> 
> OPTIONS:
> 
> 1. we keep PulseSequenceType (called PulseSequenceType)
> 2. we remove PulseSequenceType
> 
> BUT we don't literally translate PulseSequenceType -> Sequence because this is bug-prone/confusing, etc, even if it is temporarily.
> 
> (Note that because there is no BIDS derivatives for ASL yet, we're not really illegal by having non-BIDS json files, though this could also be confusing to users)

I agree that this is very complex and error prone - but things are there for certain reason:
1. x.Q.Sequence is an old legacy thing. Since previously there wasn't a decision to remove it, I was keeping them in. I made sure in all the conversions that x.Q.Sequence is properly identified and initialized.
2. PulseSequenceType is actually a general BIDS field. This is read from DICOM and in case it is missing, I am correctly filling this based on other DICOM information.
3. The correct BIDS approach is to have EPI/GRASE saved in PulseSequenceType and 2D/3D in a BIDS field MRAcquisitionType.

I propose to make a new issue where we do the following:
1. Make sure that MRAcquisitionType and PulseSequenceType contain the correct information as specified above and that also these are always initialized. This will be done in import.
2. parms2BIDS makes sure that Legacy to BIDS correctly transforms x.Q.Sequence to MRAcquisitionType+PulseSequenceType.
3. parms2BIDS of BIDS to Legacy doesn't do anything to MRAcquisitionType+PulseSequence type and keep them as they are.
4. x.Q.Sequence field is removed everywhere and is replaced by the correct use of MRAcquisitionType+PulseSequenceType.

I'll make a new issue out of this.

HenkMutsaerts:
Sounds good, we could add that for backwards compatibility we take x.Q.Sequence and convert it into those two BIDS fields.

jan-petr:
> Sounds good, we could add that for backwards compatibility we take x.Q.Sequence and convert it into those two BIDS fields.

New issue #1674 was created. The backwards compatibility is of course there already.

HenkMutsaerts (approved):

PR #1668, opened 2024-03-30
Closes #1666

### LINKED ISSUE #1666: Population group mask goes wrong
### Description

Population group mask goes wrong, it removes the back of the brain and creates a far too stringent mask.

### Tasks

- [x] Susceptibility group masking should mask Threshold*max(intensities(:))) instead of Threshold*sortedIntensities (percentile). The latter doesn't work with the dummy susceptibility mask for 3D spiral (which is a brainmask)
- [x] Which masks are used for ROI values (e.g., also for GM or WM values?)

### Release notes

Nothing. Report with #870. Fixes a bug in mask creation.

HenkMutsaerts:
### Linked issue

Closes #1666


jan-petr (approved):
OK.

PR #1679, opened 2024-04-09
Feature #1635 any visit name

### LINKED ISSUE #1635: Visit management by date
### Description

Working on importing the Excersion-VCI study, I noticed two desires:

image

image

A single subject can be scanned twice (2 visits), where the scanning date denotes the visit. This is often the case in MRI studies. The nice thing is that BIDS accommodates this, e.g., you would have ses-20160101 ses-20180431 etc. Furthermore, the sorting automatically goes correct because the year is first. We only need to manage the alias: when we have visits as dates, and we leave out visit aliases (or we have special aliases saying 'yyyymmdd') that it will just take the visits as they are. 

### Detailed description - ticked off when a corresponding specific task is described
- [x] dicom2BIDS has two options: We only use a detection token in `folderHierarchy`, and do not convert the visit name by default, unless the user specifically wants this by supplying a `tokenVisitAliases` with the second index for the name the visit should get.
- [x] OPTION 1 (no visit translation)
dicom2BIDS: we allow all visits names and simply don't rename them.
E.g., subject(001|002)_(20\d{2}\d{6}) etc
Visit token in folder hierarchy = ['\d{8}']
tokenVisitAliases is empty or not provided
becomes sub-001/ses-20240406
- [x] OPTION 2 (visit translation)
E.g., subject(001|002)_month(12|24|36) etc
tokenVisitAliases = ['12', 1, '24', 2, '36', 3]
becomes sub-001/ses-1
- [x] The only restriction we need, is that either all visits get renamed, or none. Anything else gives an error.
It would be tricky to have some visits renamed and others stay the same, let's not do this granularity because it probably practically is not required. 
JAN: This is very simple. Either tokenVisitAliases is empty and nothing is renamed. Or tokenVisitAliases is provided and things are renamed accordingly. We don't need to have a single-index `tokenVisitAliases` because we already do visit token regexp in `folderHierarchy`
- [x] Then BIDS2Legacy: We just keep the session name as is.
- [x] This makes our life easy for sorting in case of dates
- [x] Some studies have multiple scans from different visits, in BIDS we then want the full messy dataset and the user can pick and choose (e.g. T1w from 20240224 was good and can be combined with the ASL from 20240324; this works nicely in BIDS).
- [x] BIDS rawdata is initially a full non-selective copy of DICOM sourcedata, the user can do the sorting
- [x] If we want automatic sorting, this is much easier and readable in BIDS rawdata than in DICOM sourcedata
- [x] If age at scanning date is specified exactly, DOB can be calculated from scanning date so not GDPR-proof. But this is not the responsibility of the code.

#### Specific tasks
- [x] Revamp rawdata for Siemens_PCASL_multiSequence_VE11C_1
- [x] Revamp derivatives for Siemens_PCASL_multiSequence_VE11C_1
- [x] dicom 2 BIDS import is allowed to have any visit format. If tokenVisitAliases does not exist in sourceStructure.json, it will not rename the visit.  520e3a32
- [x] DICOM 2 BIDS remove all non-alphanumericals from the visit name. 520e3a32
- [x] xASL_imp_DetermineStructureFromSourcedata/xASL_imp_AddVisitNames and xASL_imp_PrintSubject - make sure the visit name and not number is printed. 830cbb2b
- [x] Revamp DCM2NII. 830cbb2b
- [x] Adapt xASL_imp_DetermineStructureFromTempdata 86304aea
- [x] Revamp NII2BIDS 86304aea
- [x] Lock-files deal with visit names, not visit index in Import 86304aea
- [x] Adapt xASL_imp_DetermineStructureFromRawdata and check deface. JAN: It was working already.
- [x] Lock-files deal with visit names, not visit index in BIDS2Legacy 377c929f
- [x] Load Legacy from derivatives without rawdata 377c929f
- [x] Test BIDS2Legacy with rawdata without sourcedata 377c929f
- [x] BIDS 2 Legacy will keep visits as is, but removing everything that is not alfanumerical (already done in DICOM2BIDS, but we want to do it again in case of weird external data. dcec094d
- [x] Test DICOM2BIDS when different subjects have different visits (sub1 - visit 1 and 2; sub2 - visit 2 and 3). Check what happens if we import visit 3 and then run import again for visit 1. dcec094d
- [x] Does `xASL_module_Structural`
- [x] Does `xASL_module_LongReg` need updating?
- [x] Does `xASL_module_DARTEL` need updating?
- [x] Does `xASL_module_LongReg` need updating?
- [x] Does `xASL_module_ASL` need updating?
- [x] Does `xASL_module_Population` need updating?
- [x] Test all processing in Siemens_PCASL_multiSequence_VE11C_1
- [x] TokenVisitAliases explanation needs to be explain in the documentation.

### Open questions
- [x] JAN: Longitudinal registration might be tricky as the baseline session can change when adding a new visit.
HENK that hasn't changed from the previous implementation. E.g., in the previous implementation you could still have had _2 _3 and later added _1. It's a problem, but not a new problem (afaik) and it is a problem for the user.
We don't user longitudinal registration very often anyway. Though we could print a subtle warning (something like fprintf('WARNING: ....') (we don't need a formal warning for this because it is not wrong behavior)
- [x]  JAN: xASL_init_Parallelization splits along subject_visit, not along `subject` only. Is this a problem? Not for BIDS2Legacy and processing. And import is not parallelized. So should be ok?
I think so. But I thought we split by subjects, not by subject_visit? In fact if I remember correctly you were the one pushing this :) So I implemented this, do you mean this is now broken? That would be weird.
- [x] HENK: We will provide a warning if tokenVisitAliases~='STRING' or something there are <9 visits to warn the user if didn't forget tokenVisitAliases.
JAN: In import, we gather a list of all visits. I am just not too sure what kind of warning we want to do. If the visitNames are strings, we can expect anything. If the visits are numbers, we can also have the numbers of sessions, dates, or some extra pseudonyms. So I'm not really sure what kind of criteria to have to produce this warning... We can of course simply always print that no Aliases are provided and we will not rename visits...
HENK: I meant that — as an extreme example — if you have 1000 subjects and only 3 unique visits that are not indicated as tokens, this is probably going wrong (unless everyone has been scanned at exactly the same dates :)
So I more or less thought we would control it a bit more, but I'm fine only controlling if tokens are provided as you suggest. So you can ignore this point, I just wrote this here to explain what I meant.

### Release notes

Allow any visit (BIDS session) name.

jan-petr:
### Linked issue

Close #1635


Modules/Module_Import/Import/xASL_imp_DetermineStructureFromSourcedata.m
```diff
@@ -38,6 +38,14 @@
         x.modules.import.settings.bUseVisits = true;
         % vVisitIDs: cell vector with extracted session IDs (for all subjects, sessions and scans)
         x.modules.import.listsIDs.vVisitIDs = x.modules.import.tokens(:, x.modules.import.imPar.tokenOrdering(2));
+		% Check if tokenVisitAliases were defined
+		if isempty(x.modules.import.imPar.tokenVisitAliases)
+			% In case the aliases are not defined, we take the sessions as they are without any renaming
```
HenkMutsaerts: visits not sessions. Within import we keep legacy terminology right? Until we fully convert everything to BIDS terminology?

Modules/Module_Import/Import/xASL_imp_DetermineStructureFromSourcedata.m
```diff
@@ -38,6 +38,14 @@
...
+			x.modules.import.imPar.tokenVisitAliases = unique(x.modules.import.listsIDs.vVisitIDs);
+
+			% However, we do remove all non-alphanumerical entries
+			x.modules.import.imPar.tokenVisitAliases(:,2) = cellfun(@(y) xASL_adm_CorrectName(y, 2), x.modules.import.imPar.tokenVisitAliases(:,1), 'UniformOutput', false );
```
HenkMutsaerts: space after false

Modules/Module_Import/Import/xASL_imp_DetermineStructureFromSourcedata.m
```diff
@@ -83,19 +91,19 @@
...
+% -----------------------------------------------------------------
 function xASL_imp_PrintOverview(x)
 
     if isfield(x,'importOverview')
         % Get all main fields
         overviewFields = fieldnames(x.importOverview);
         for iField=1:numel(overviewFields)
             thisSubject = overviewFields{iField};
-            if ~isempty(regexpi(thisSubject, 'subject_'))
+            if ~isempty(regexpi(thisSubject, 'subject_', 'once'))
```
HenkMutsaerts, line 104: This 'once' annoys me. Theoretically it is faster but it won't matter on many occasions and simpler code can make it more readable. In general. I don't mind it now of course. Just because you keep updating it throughout, and this is just noise to me, takes time to check, has no effect (unless you have a 1000-loop)

Modules/Module_Import/Import/xASL_imp_DetermineStructureFromSourcedata.m
```diff
@@ -155,20 +175,20 @@ function xASL_imp_PrintRun(x,thisSubject,thisVisit,thisRun)
...
     
     for iSubject=1:numel(x.modules.import.listsIDs.subjectIDs)
         thisSubject = x.modules.import.listsIDs.subjectIDs{iSubject};
         x = xASL_imp_AddSubject(x, thisSubject, iSubject);
     end
-
 end
 
-
+% -----------------------------------------------------------------
```
HenkMutsaerts, line 187: I would add a bit more space between functions, the way you did it is also fine though.

Modules/Module_Import/Import/xASL_imp_DetermineStructureFromSourcedata.m
```diff
@@ -216,16 +234,16 @@ function xASL_imp_PrintRun(x,thisSubject,thisVisit,thisRun)
     
     % Sort by output
     if length(x.importOverview.(sFieldName).visitIDs)>1
-        for iV=1:numel(x.importOverview.(sFieldName).visitIDs)
+		IDrow = zeros(1, numel(x.importOverview.(sFieldName).visitIDs));
```
HenkMutsaerts: Same here. Initializing this has no true effect on code speed, it just creates more text. I understand that this is good practice and required for lower-level languages like C but just annoying when you need to check code.

And can potentially generate new bugs, e.g., now you have to check that nowhere in the function we check for the existence of IDrow as a variable. I would vote for not implementing accelerations that will create negligible improvements but make the code more complicated.

Modules/Module_Import/Import/xASL_imp_DetermineStructureFromSourcedata.m
```diff
@@ -240,42 +258,38 @@ function xASL_imp_PrintRun(x,thisSubject,thisVisit,thisRun)
...
 
-
+% -----------------------------------------------------------------
 %% Add visit names
+% -----------------------------------------------------------------
 function x = xASL_imp_AddVisitNames(x, sFieldName)
 
     if isempty(x.modules.import.imPar.visitNames)
         if isempty(x.importOverview.(sFieldName).visitIDs)
+			% In case that visit names are not filled, we name the visit by their number
```
HenkMutsaerts: What happens here if some fields are empty and others are not? Then you get some visits having an index number and other visits having the original alphanumeric string. I know that this should not happen because of the unique function above but always better to have a check (I don't mind if the check is here or above).

Modules/Module_Import/Import/xASL_imp_DetermineStructureFromSourcedata.m
```diff
@@ -289,24 +303,15 @@ function xASL_imp_PrintRun(x,thisSubject,thisVisit,thisRun)
...
-				
-				% Extract the number of the visit
-				numberVisit = regexp(nameVisit, '\d+');
-				if ~isempty(numberVisit)
-					numberVisit = nameVisit(numberVisit:end);
-				else
-					numberVisit = sprintf('%g', idVisit);
-				end
-				x.modules.import.imPar.visitNames{iVisit} = sprintf('ASL_%s', numberVisit);
+				x.modules.import.imPar.visitNames{iVisit} = x.modules.import.imPar.tokenVisitAliases{idVisit, 2};
```
HenkMutsaerts: What happened here? What you write looks good but I don't know if there is a reason (== use case) for the `sprintf('ASL_%s', numberVisit` and other stuff you just deleted

Modules/Module_Import/Import/xASL_imp_DetermineStructureFromSourcedata.m
```diff
@@ -437,28 +443,28 @@ function xASL_imp_PrintRun(x,thisSubject,thisVisit,thisRun)
...
-	thisScanID = scanSessionTable{iID,2};
-	% Iterate over the sessions of this run
-	for iSession=1:numel(x.importOverview.(sFieldName).(vFieldName).sessions)
-		currentID = x.importOverview.(sFieldName).(vFieldName).sessions{iSession};
-		if ~isempty(regexpi(currentID, thisRegExp)) && strcmp(currentID, thisToken)
-			x.importOverview.(sFieldName).(vFieldName).(vSessionName).scanIDs{iSessionNew} = thisScanID;
-			x.importOverview.(sFieldName).(vFieldName).(vSessionName).ids{iSessionNew} = currentID;
-			iSessionNew = iSessionNew + 1;
-			continue;
+    % Iterate over scan IDs
```
HenkMutsaerts, line 401: This is annoying because I cannot see if you changed the content or just added a tab. I know that many people like an extra indent after function but I think it is just a waste of horizontal space (and extra useless commits in this case :)

Modules/Module_Import/Import/xASL_imp_DetermineSubjectStructure.m
```diff
@@ -112,37 +104,40 @@
...
-                    warning('Multiple underscores in subject/session name...');
-                end
+				listSessions{iSubSes} = '';
+			elseif numel(indexSeparator) == 1
+                % Multi-session notation
+                x.SUBJECTS{iSubSes} = curSubSes(1:indexSeparator-1);
+				listSessions{iSubSes} = curSubSes(indexSeparator+1:end);
+			else
+				% Multiple separators
+                warning('It was not possible to determine the subject name from the temp data because there are multiple underscores in subject/session name');
```
HenkMutsaerts: Not specific enough, I would make it:

warning('It was not possible to determine the subject (or session) name from temporary data, note that underscores in subject or session values are illegal. Please check your data with the BIDS validator.'

Modules/Module_Import/Import/xASL_imp_DetermineSubjectStructure.m
```diff
@@ -153,10 +148,19 @@
         warning('x.SUBJECTS is undefined, data loading is going to fail...');
     end
     
-
+	% If tokenVisitAliases is not provided, we assume visits are not renamed and create a dummy tokenVisitAliases list
+	if ~isfield(x.modules.import.imPar, 'tokenVisitAliases') || isempty(x.modules.import.imPar.tokenVisitAliases)
+		% Create a unique list of sessions
+		listSessions = unique(listSessions);
```
HenkMutsaerts: Calling it sessions here is tricky as we call runs sessions in these import functions (per legacy nomenclature). We need to keep legacy nomenclature throughout to avoid confusion and bugs.

So should become listVisits

Modules/Module_Import/Import/xASL_imp_DetermineSubjectStructure.m
```diff
@@ -153,10 +148,19 @@
...
+
+		% Exclude empty session names
+		listSessions = listSessions(~ismember(listSessions,''));
```
HenkMutsaerts: What will happen to sub-001/ses-<'illegalsymbols'> ? The symbols will be removed above and this sub-visit will crash somewhere down the line as it's session cannot be found?

Modules/Module_Import/Import/xASL_imp_DetermineSubjectStructure.m
```diff
@@ -153,10 +148,19 @@
...
+		
+		% Create a list of unique session names
+		if ~isempty(listSessions)
+			x.modules.import.imPar.tokenVisitAliases = listSessions(:);
```
HenkMutsaerts: Why first (:) and second not?

BTW it seems you want

x.modules.import.imPar.tokenVisitAliases = [listSessions listSessions];

(unless you're afraid that somehow listSessions got transposed

Modules/Module_Import/Import/xASL_imp_NII2BIDS_Run.m
```diff
@@ -70,29 +69,29 @@
...
 
+%% -------------------------------------------------------------
 %% Issue failed run as a warning
-function xASL_imp_NII2BIDS_RunIssueWarning(x,loggingEntry, Scantype, subjectSessionLabel, iRun)
+%% -------------------------------------------------------------
+function xASL_imp_NII2BIDS_RunIssueWarning(loggingEntry, Scantype, subjectSessionLabel, iRun)
 
     xASL_adm_BreakString('');
-    fprintf(2,'NII2BIDS failed for %s image of %s_run-%s\n',Scantype,subjectSessionLabel,xASL_num2str(iRun));
+    fprintf(2,'NII2BIDS failed for %s image of %s_run-%s\n', Scantype, subjectSessionLabel, xASL_num2str(iRun));
```
HenkMutsaerts: If you're revamping this, then also add a space after the comma before 'NII2BIDS ;) same next line

Modules/Module_Import/xASL_wrp_DCM2NII_Subject.m
```diff
@@ -43,21 +43,14 @@
     % Check subjectID
     thisSubject.subjectExport = xASL_imp_SubjectName(subjectID);
     
+	x.modules.import.imPar.visitNames = thisSubject.visitIDs;
```
HenkMutsaerts, line 46: if Names is the same as IDs, why do we have these two variables with the same content? Can't we replace them by a single variable? We should have names and indices. In BIDS nomenclature this would be "values" (key = session, value = 20240218)

Modules/Module_Import/xASL_wrp_DCM2NII_Subject.m
```diff
@@ -43,21 +43,14 @@
...
+
     %% 2. Iterate over visits
     for iVisit=1:thisSubject.nVisits
         
         % Get the current visit and the visit ID
         vFieldName = ['visit_' num2str(iVisit,'%03.f')];
         thisVisit = thisSubject.(vFieldName);
-        visitID = thisSubject.visitIDs{iVisit};
-
```
HenkMutsaerts, line 53: Why are you removing all this code? I would still keep the option (defaulting to false) to convert visits to indices, unless this would get messy.

Not sure what functionality you are removing now, it could of course be that you see that you can simplify things but haven't accounted for all use cases?

Modules/Module_Import/xASL_wrp_NII2BIDS_Subject.m
```diff
@@ -90,7 +88,7 @@
 					checkExpression = regexp(nameSubjectSession, [separator visitAliases{iAlias,1} '$'], 'once');
 				end
 				if ~isempty(checkExpression) % nameSubject should end in the visit alias
-					visitName = nameSubjectSession(checkExpression:end);
+					visitName = nameSubjectSession(checkExpression+1:end);
```
HenkMutsaerts: Are you sure this doesn't break old code?

Modules/Module_Processing/xASL_init_SubjectList.m
```diff
@@ -74,7 +74,7 @@
         elseif ~strcmp(visitName(1:4), 'ses-')
             warning(['Illegal BIDS session definition for ' subjectName '_' visitName]);
         else
-            visitName = visitName(5:end); % if we find a BIDS session, we add it as _n visit suffix in ExploreASL legacy format
+            visitName = xASL_adm_CorrectName(visitName(5:end), 2); % if we find a BIDS session, we add it as visit suffix in ExploreASL legacy format
```
HenkMutsaerts: No the BIDS session should already be corrected. I would not alter the visit/session names as they should in ExploreASL be identical to the names in BIDS.

So I would rather issue an error if there are illegal symbols instead of removing them without notifying the user.

Image use case 1:

sourcedata/participant1_2024+02+01

-> this is easy, it will just remove the illegal symbols

but use case 2, where we are not the ones that created rawdata:

rawdata/sub-001/ses-2024\02\01

Perhaps we should still correct this automatically, but then we should have at least a detailed warning

Modules/xASL_module_BIDS2Legacy.m
```diff
@@ -83,18 +80,20 @@
         % x.modules.bids2legacy.BIDS.subjects(1).perf(1) -> filename: 'sub-10015124_ses-1_run-1_asl.nii.gz'
         % x.modules.bids2legacy.BIDS.subjects(1).perf(2) -> filename: 'sub-10015124_ses-1_run-2_asl.nii.gz'
         % x.modules.bids2legacy.BIDS.subjects(1).perf(3) -> filename: 'sub-10015124_ses-1_run-1_m0scan.nii.gz'
-        % x.modules.bids2legacy.BIDS.subjects(1).perf(4) -> filename: 'sub-10015124_ses-1_run-2_m0scan.nii.gz'            
+        % x.modules.bids2legacy.BIDS.subjects(1).perf(4) -> filename: 'sub-10015124_ses-1_run-2_m0scan.nii.gz'  
+		%
+		% Finally note that visits can now be any name, number, data, or a string and are always treated as a string
```
HenkMutsaerts: Remove the word "Finally" as this is grammatically weird

-> is this a final note in a list of notes? Or should the user always wait with noting this until the "final" moment ;)

Or did you mean "Note that visits at this point can be any ..."

-> which is a moot statement. This code is completely independent of dicom2BIDS so this is the beginning, not final of something?

Modules/xASL_module_BIDS2Legacy.m
```diff
@@ -83,18 +80,20 @@
...
 
         % Subject ID
         SubjectID = x.modules.bids2legacy.BIDS.subjects(iSubjSess).name;
         SessionID = x.modules.bids2legacy.BIDS.subjects(iSubjSess).session;
         if isempty(SessionID)
-            iVisit = 1; % we default to a single visit
+            SessionID = '1';
         else
-            iVisit = str2num(SessionID(5:end));
+            SessionID = xASL_adm_CorrectName(SessionID(5:end), 2);
```
HenkMutsaerts: Have to issue a warning if something is corrected

HenkMutsaerts (changes requested):
Looks good! Few suggestions, and annoying comments about when not to write extra code IMHO ;)

Modules/Module_Import/Import/xASL_imp_DetermineStructureFromSourcedata.m
jan-petr: Agree.Done.

Modules/Module_Import/Import/xASL_imp_DetermineStructureFromSourcedata.m
jan-petr: Done

Modules/Module_Import/Import/xASL_imp_DetermineStructureFromSourcedata.m
jan-petr: Sorry. Matlab shows this for me as a potential issue in the editor. So it's nice to have a clean list without issue. But I agree that this can be annoying, so I'll stop doing it from now on ;)

Modules/Module_Import/Import/xASL_imp_DetermineStructureFromSourcedata.m
jan-petr: There's one other reason to do it. If you have such a vector that can have potentially variable length, and the vector is reused many time, then it's always good to clean it beforehand to avoid having residual values from the previous loop. So that's why I always try to clean reused variables to avoid these potential bugs. 

I agree that here it wasn't necessary, but I really do this as a pre-caution to avoid bugs rather than for speeding up which would make no difference here.

Modules/Module_Import/Import/xASL_imp_DetermineStructureFromSourcedata.m
jan-petr: I agree. And I have thought of this option. That's why we copy the original alphanumeric string, and in case it's empty we provide a number, but as a string. So in the end, all values are strings and thus have a consistent type.

So a check shouldn't be necessary.

Modules/Module_Import/Import/xASL_imp_DetermineStructureFromSourcedata.m
jan-petr: Very simple - previously, we only used numbers for visit names and we extracted them as numerical. Now, we save everything as string, so if we have a number in a string, we just save it without doing the number->string conversion.

Modules/Module_Import/Import/xASL_imp_DetermineStructureFromSourcedata.m
jan-petr: Not me. I hate extra indent for a new function and I'm usually removing them :) Not sure why this happened - probably a mistake... Cause nothing changed here

Modules/Module_Import/Import/xASL_imp_DetermineSubjectStructure.m
jan-petr: I changed the warning. But removed the last sentence. This is about DICOM data and DCM2NII - we can't run that through the BIDS validator.

Modules/Module_Import/Import/xASL_imp_DetermineSubjectStructure.m
jan-petr: You know that this is your code that I just modified? ;)

So the term "SubjectSession" is something you have mostly everywhere in the code. And you even have that in x-structure. I'm fine changing it here, but we have x.SUBJECTS and x.SESSIONS everywhere in the entire xASL for as long as xASL exists.

I agree that this is confusing, but that's a larger problem than this single function.

Modules/Module_Import/Import/xASL_imp_DetermineSubjectStructure.m
jan-petr: Actually, it shouldn't crash. This here creates a dummy tokenVisitAlias list for the case we don't rename. Session with all illegal characters will be skipped from this list here. So later on, we will not be able to identify them and it will just skip them. 

We can try to create a warning for this: In case that we go for non-renaming Visits, we will specifically check for sessions that turn out empty after removing special characters and we will report this to the user. 

Is that fine? To me, a session without any alphanumerical characters is so exotic I would just ignore it, but I'm happy to add this warning if you think this might occur.

Modules/Module_Import/Import/xASL_imp_DetermineSubjectStructure.m
jan-petr: nono. this is correct and on purpose.

we want tokenVisitAliases to be a cell array nVisits x 2.
listSessions comes as 1 x nVisits.

So by using listSessions(:) we turn it into nVisits x 1 irrespective if listSessions is nVisit x 1 or 1xnVisits.

So x.modules.import.imPar.tokenVisitAliases = listSessions(:);
gives an array nVisits x 1 - always.

This command then gives 
x.modules.import.imPar.tokenVisitAliases(:,2) = listSessions;
the second column that is identical to the first column and we receive an array nVisits x 2 as we want.

When the matrix already exists and we add a column, it makes no difference if we do
x.modules.import.imPar.tokenVisitAliases(:,2) = listSessions;
or
x.modules.import.imPar.tokenVisitAliases(:,2) = listSessions(:)
or 
x.modules.import.imPar.tokenVisitAliases(:,2) = listSessions'
when listSessions is a vector.

Just put a breakpoint there to see for yourself...

Modules/Module_Import/Import/xASL_imp_NII2BIDS_Run.m
jan-petr: Oops. I've overlooked that. Done.

Modules/Module_Import/xASL_wrp_DCM2NII_Subject.m
jan-petr: Michael!!!! That kept me puzzled as well. I've tried to remove some of the useless code, but I didn't want to do larger changes to avoid extra bugs. Shall we make a new separate issue where we revisit how this whole thing works and remove things that don't make sense?

Modules/Module_Import/xASL_wrp_DCM2NII_Subject.m
jan-petr: Related to the comment above. This was a pure Michael nonsense that had no function and was just cluttering the code. I'm simply cleaning.

This specifically - he took the visit token. Renamed this to a visit name/number, saved this into his huge structure. And then a few moments later he does the same action again :) 

So I'm just removing this stupid duplicity. I know it's hard to check, but don't worry, I was very carefully reviewing and identifying this nonsense and it made no sense to keep it.

Modules/Module_Import/xASL_wrp_NII2BIDS_Subject.m
jan-petr: Yes. VERY sure. This is Michael nonsense of saving visit names as "_1" instead of "1". 

So I cleaned all code and save really only the visit name and always add the underscore when necessary to connect into a filename. But having an underscore in the visit name was sooo dirty...

Now you see why I complained so much :)

Modules/Module_Processing/xASL_init_SubjectList.m
jan-petr: See the relevant discussion inside the issue.

1. We remove illegal characters in visit names when doing DICOM to BIDS. 
2. We keep visit names when doing BIDS to Legacy as illegal names were already removed in DICOM to BIDS
3. But we need to check visit names when doing BIDS to Legacy when we have external data.

So I would keep automatic correct - but a very good point that we should give a warning for every such case. Agree?

Modules/xASL_module_BIDS2Legacy.m
jan-petr: Replaced by "also"

Modules/xASL_module_BIDS2Legacy.m
jan-petr: I agree. Note that we are correcting the names while loading the list of session and while doing the conversion. We should only report the warning on a single location, so I would do it here, during the actual renaming, rather than when creating the list of subjects. Do you agree?

Modules/Module_Import/Import/xASL_imp_DetermineStructureFromSourcedata.m
HenkMutsaerts: Ah that's actually a good reason, to initialize a clean variable. I was thinking only about the speed increase.
Thanks, good point! Please resolve once you read this.

Modules/Module_Import/Import/xASL_imp_DetermineStructureFromSourcedata.m
HenkMutsaerts: You answer about an empty token, I meant if you have an empty token, and some dates are specified in the sourcedata folder format, and some or not.

So consider the following (with empty visittoken/no visit token specified):

/sourcedata/Subject001_20230101/
/sourcedata/Subject001_20240201/
/sourcedata/Subject001_/

What will happen? Do we give an error above once one of the unique visits is empty? Or a warning and we assign 'missingSessionValue'?

Probably the second is nicer for the user, because rawdata is more user-readable than our huge log files, and the huge sourcedata folder. In this case, it would become:

/rawdata/sub-001/20230101/
/rawdata/sub-001/20240101/
/rawdata/sub-001/missingSessionValue/

So I would go for the warning and replacement of empty alphanumerics by "missingSessionValue". In the case that no visit token is specified. In fact, we could do the same when visit tokens are specified, e.g. if we have specified this as visit tokens it would have been something like:

visitTokens = {'^20230101$', '1', '^20240201$', '2', '^20240401$', '3'}

and one would still get the above situation. So the replacement of empty visit values could go here or quite down in the code.

But now I actually think that Paul's findRegExp function would not detect visits with empty values, unless we change the regular expression to also except empty (which is easy, it would just be (| or |). ).
The same goes for runs. It would be good to add such functionality if it doesn't take too much to program (it remains of course the user's responsibility).

Modules/Module_Import/Import/xASL_imp_DetermineStructureFromSourcedata.m
HenkMutsaerts: 1.No need. xASL_num2str already does this for us.

xASL_num2str('20240203') == '20240203'
xASL_num2str(20240203) == '20240203'

So can we simplify our code with this?

2.Something else, why did we have 'ASL_%s' if this was about visits? You're sure your not removing an 'ASL_' prefix that should have stayed here?

Modules/Module_Import/Import/xASL_imp_DetermineSubjectStructure.m
HenkMutsaerts: Yes and SubjectSession x.SESSIONS all through the old ExploreASL code, session was run.

In this code, you are tweaking visits. So if you tweak something that was called session, you might change the session/run behavior rather than the visit/session behavior.

This is actually the whole reason of course why we should use one nomenclature throughout the whole import.

So would be good to check if this is indeed listSession (about BIDS runs) or should be listVisit (about BIDS sessions).

And I start then doubting what we have above, where you removed 'ASL_', did you check that you're sure that this is not about runs?

Modules/Module_Import/Import/xASL_imp_DetermineSubjectStructure.m
HenkMutsaerts: Yes if it's not too much work. Can we integrate this with my above comment with empty BIDS session values?

(in BIDS key = session/visit, value = the visit name)

Modules/Module_Import/Import/xASL_imp_DetermineSubjectStructure.m
HenkMutsaerts: I know this, read my message, in my last sentence I said "unless you're afraid that somehow listSessions got transposed" which is exactly what you explain to me here, that in the case of a transposed listSessions, you would transpose back. I'm fine either way, your code is a bit longer but more secure.

If the reason for transposed listSessions is a bug, then your code would not give an error but we would get an error later on.

No clue what is better or worse. You can resolve this.

Modules/xASL_module_BIDS2Legacy.m
HenkMutsaerts: Sounds good. Didn't xASL_adm_CorrectName already have an input boolean for issuing a warning or not when something is changed? Otherwise we could add this, because this is of course something that happens a lot; there are probably many cases where you want the user to know that a name is changed.

(meaning that this is nicer than having to program something like:

"if diff between pre-correctname and post-correctname is not empty, then issue warning"

Modules/xASL_module_BIDS2Legacy.m
```diff
@@ -83,18 +80,20 @@
         % x.modules.bids2legacy.BIDS.subjects(1).perf(1) -> filename: 'sub-10015124_ses-1_run-1_asl.nii.gz'
         % x.modules.bids2legacy.BIDS.subjects(1).perf(2) -> filename: 'sub-10015124_ses-1_run-2_asl.nii.gz'
         % x.modules.bids2legacy.BIDS.subjects(1).perf(3) -> filename: 'sub-10015124_ses-1_run-1_m0scan.nii.gz'
-        % x.modules.bids2legacy.BIDS.subjects(1).perf(4) -> filename: 'sub-10015124_ses-1_run-2_m0scan.nii.gz'            
+        % x.modules.bids2legacy.BIDS.subjects(1).perf(4) -> filename: 'sub-10015124_ses-1_run-2_m0scan.nii.gz'
```
HenkMutsaerts, line 83: I don't know what you changed here at line 86/83, it seems (and should) stay identical.

Modules/xASL_module_BIDS2Legacy.m
```diff
@@ -83,18 +80,20 @@
         % x.modules.bids2legacy.BIDS.subjects(1).perf(1) -> filename: 'sub-10015124_ses-1_run-1_asl.nii.gz'
         % x.modules.bids2legacy.BIDS.subjects(1).perf(2) -> filename: 'sub-10015124_ses-1_run-2_asl.nii.gz'
         % x.modules.bids2legacy.BIDS.subjects(1).perf(3) -> filename: 'sub-10015124_ses-1_run-1_m0scan.nii.gz'
-        % x.modules.bids2legacy.BIDS.subjects(1).perf(4) -> filename: 'sub-10015124_ses-1_run-2_m0scan.nii.gz'            
+        % x.modules.bids2legacy.BIDS.subjects(1).perf(4) -> filename: 'sub-10015124_ses-1_run-2_m0scan.nii.gz'  
+		%
+		% Also note that visits can now be any name, number, data, or a string and are always treated as a string
```
HenkMutsaerts: visits here should become session! (we're now in BIDS)

Modules/xASL_module_BIDS2Legacy.m
```diff
@@ -68,8 +67,6 @@
         %% 1. Get subjectID & sessionID from x.modules.bids2legacy.BIDS
 
         %   !!!!! NOTE THAT SESSION (BIDS) IS A VISIT (LEGACY) HERE, NOT A RUN (BIDS) !!!!!
```
HenkMutsaerts, line 69: In this function we use BIDS and legacy nomenclature interchanged, which is understandable because we do the conversion. But I would use BIDS nomenclature throughout in this function. We probably don't need the legacy nomenclature (sub-001_20240201 we can still call subjectSession).

Modules/xASL_module_BIDS2Legacy.m
```diff
@@ -83,18 +80,20 @@
...
         if isempty(SessionID)
-            iVisit = 1; % we default to a single visit
+            SessionID = '1';
         else
-            iVisit = str2num(SessionID(5:end));
+            SessionID = xASL_adm_CorrectName(SessionID(5:end), 2);
         end
 
-        SubjectVisit = [SubjectID '_' xASL_num2str(iVisit)]; % this is the legacy name
+        SubjectVisit = [SubjectID '_' SessionID]; % this is the legacy name
```
HenkMutsaerts: SubjectVisit -> SubjectSession (we are in BIDS nomenclature)

HenkMutsaerts (changes requested):
Few minor things

Modules/Module_Import/Import/xASL_imp_DetermineStructureFromSourcedata.m
jan-petr: 1. If folderhierarchy token regexp is (Subject\d(3))_(\d(8)) then it will simply not even find it. 

2. So I guess you are refering to the case of having (Subject\d(3))_(.*) which will find it and return an empty token. In case we have defined visitTokenAliases in sourcestructure.json, we will have to explicitly define an alias for this empty token - this already works. I would ignore this in case this alias doesn't fall within known aliases because what we will do in case we have several unknown token - I would skip all unknown tokens.

3. So the only relevant option that we should address is token renaming and an empty token, for which we can indeed create a "missingSessionValue". Done in the last commit.

Modules/Module_Import/Import/xASL_imp_DetermineStructureFromSourcedata.m
jan-petr: 1. True, I was simplifying, but didn't realize I could simplify more. Done in the last commit.

2. No clue why it was there - ask Michael. I'm pretty sure this was a nonsense and the whole Flavor import works, so that confirms my suspicion.

Modules/Module_Import/Import/xASL_imp_DetermineSubjectStructure.m
jan-petr: 1.Yes, I'm sure the 'ASL_' should be removed.

2. I didn't tweak anything here - this function was already using session as we are using legacy code to import into BIDS - there's not a perfect unique solution. I changed everything to visits here, but it's playing with unimportant details with no impact.

Modules/Module_Import/Import/xASL_imp_DetermineSubjectStructure.m
jan-petr: we now name empty visits as missingSessionValue.

All user output uses BIDS terminology.

Modules/Module_Import/Import/xASL_imp_DetermineSubjectStructure.m
jan-petr: Yes - upon creation, it's transposed. So it's a normal behavior that should be fixed. A normal transpose should be fine. But I'd rather use this secure thing to avoid future bugs when somebody changes this up in the code.

Modules/xASL_module_BIDS2Legacy.m
jan-petr: Done

Modules/xASL_module_BIDS2Legacy.m
jan-petr: Done

Modules/xASL_module_BIDS2Legacy.m
jan-petr: Done

Modules/xASL_module_BIDS2Legacy.m
jan-petr: Done

Modules/Module_Import/Import/xASL_imp_DetermineStructureFromSourcedata.m
HenkMutsaerts: Let me try to explain again.

By "no visit token specified" I indeed meant something like 
folderHierarchy : (Subject\d(3))_(.*)
and no row that says visitTokenAliases

the sourcedata is:

/sourcedata/Subject001_20230101/
/sourcedata/Subject001_20240201/
/sourcedata/Subject001_/

So we need to fix two things:

1.  If you have no visitTokenAliases, you create a dummy visitTokenAlias for each unique value right?
So you would create this for '20230101' and '20240201' but not for ''. So you need to add '' if this is not too much to reprogram (like I describe above but some reason you didn't read? 

example dummy visitTokenAliases created internally if visitTokenAliases was not specified in sourceStructure.json in this example you probably create:

visitTokenAliases = {'^20230101$', '1', '^20240201$', '2'}

but it should create:

visitTokenAliases = {'^20230101$', '1', '^20240201$', '2','^$','missingSessionValue}

So it should always add this (in addition to the fix that you did in the end for NII2BIDS, which is also nice of course).

BTW: why do I have to explain this 3 times? You know that it takes me each time 10 minutes, so I don't hope you just read this too quickly so it takes 3 times 10 min from me, instead of you reading my message a bit more carefully and taking only 1 time 10 min ;) Because I really try to put effort in it to make it clear to you. Then if you only respond on part of what I wrote this is a bit annoying if you understand what I mean. No offense of course, just trying to preserve efficiency.

Modules/Module_Import/xASL_wrp_NII2BIDS_Subject.m
```diff
@@ -95,6 +95,13 @@
         end
     end
     
-    bidsLabel.subject = xASL_adm_CorrectName(subjectName,2);
-    bidsLabel.visit = xASL_adm_CorrectName(visitName,2);
+    bidsLabel.subject = xASL_adm_CorrectName(subjectName, 2);
+	if isempty(visitName)
+		bidsLabel.visit = 'missingSessionValue';
```
HenkMutsaerts, line 106: Don't we need a warning here as well?

Modules/Module_Processing/xASL_init_SubjectList.m
HenkMutsaerts: 1. Agreed, but we add a boolean to xASL_adm_CorrectName for issuing a warning if we changed something, that defaults to false. And in this case it is set to true, so if correcting the visit name the user will be warned.
2. Agreed
3. Not agreed. A pipeline that has BIDS input, assumes BIDS. Something that is nearly BIDS is not BIDS, so will break the pipeline. So A) do you mean that BIDS supports illegal characters (I would not think so, but I don't know?) B) if you want to add failsafes, that is fine with me, but it should be done at the point of BIDS2Legacy. You inserted this at the list loading. We cannot change things at the point of list loading, it has to be done at the point of creating legacy derivatives.

E.g.,

1. bids.layout reads rawdata in xASL_init_SubjectList

rawdata/sub-001/ses-2024\02\01 

becomes visit 2024\02\01 which is fine as string

2. then when we convert BIDS2Legacy, we do the conversion, both internally (visit name) and on the disk (newly saved derivatives/ExploreASL/sub-001_20240201 )

Modules/Module_Import/xASL_wrp_DCM2NII_Subject.m
HenkMutsaerts: Yes good point! (you can resolve)

Modules/Module_Import/Import/xASL_imp_DetermineSubjectStructure.m
HenkMutsaerts: Ah sorry I misunderstood. My points were also:

1. User doesn't know what temp data is. It doesn't exist to the user (and shouldn't, it is just a dirty in between-stage to make our life easy).
2. The problem probably doesn't lie in the temp data stage but in the initial naming of the folders.
3. The initial naming of the folders is often taken from standard scanner output.

So I would suggest then something like:

warning('In BIDS no values (==names) can have hyphens (-) or underscores (_) because these are used as separators. Removing them in the final BIDS values/names')

Modules/Module_Import/Import/xASL_imp_DetermineStructureFromSourcedata.m
```diff
@@ -240,42 +258,38 @@ function xASL_imp_PrintRun(x,thisSubject,thisVisit,thisRun)
...
+% -----------------------------------------------------------------
 function x = xASL_imp_AddVisitNames(x, sFieldName)
 
     if isempty(x.modules.import.imPar.visitNames)
         if isempty(x.importOverview.(sFieldName).visitIDs)
+			% In case that visit names are not filled, we name the visit by their number
             x.modules.import.imPar.visitNames = cell(x.importOverview.(sFieldName).nVisits,1);
             for iVisit=1:x.importOverview.(sFieldName).nVisits
-                x.modules.import.imPar.visitNames{iVisit} = sprintf('ASL_%g', iVisit);
+                x.modules.import.imPar.visitNames{iVisit} = xASL_num2str(iVisit);
```
HenkMutsaerts: I coded xASL_num2str such that it can function for entire cell arrays at once without having to loop over all elements, but this may need testing. So you can ignore and resolve this if you want as well :)

Modules/Module_Import/Import/xASL_imp_DetermineSubjectStructure.m
```diff
@@ -153,10 +148,19 @@
         warning('x.SUBJECTS is undefined, data loading is going to fail...');
     end
     
-
+	% If tokenVisitAliases is not provided, we assume visits are not renamed and create a dummy tokenVisitAliases list
+	if ~isfield(x.modules.import.imPar, 'tokenVisitAliases') || isempty(x.modules.import.imPar.tokenVisitAliases)
+		% Create a unique list of sessions
+		listVisits = unique(listVisits);
+
+		% Exclude empty session names
```
HenkMutsaerts: session -> visit also in this function

Plus, per discussion above, we should not exclude empty visit names right? Instead, we convert them to "missingSessionName"?

Modules/Module_Import/xASL_wrp_NII2BIDS_Subject.m
```diff
@@ -58,13 +57,12 @@
...
-    
 end
 
-
-
+%% -------------------------------------------------------------
 %% Check if there is a visit alias within the subject/visit name
-function bidsLabel = xASL_imp_CheckForAliasInVisit(imPar,nameSubjectSession)
+%% -------------------------------------------------------------
+function bidsLabel = xASL_imp_CheckForAliasInVisit(imPar, nameSubjectSession)
```
HenkMutsaerts: Again confusing. 

So at line 58 above, it is clear SubjVisitSess (which means SubjSessRun).
But here the subfunction name is in legacy and the input variable name is in BIDS? or legacy?

I don't care what we choose, as long as we choose 1 nomenclature per function :)
Or we fix that in a follow-up issue, if you have tested thoroughly that we're not mixing up visits and sessions. I would fix the easy things now already, we don't want to mix things up.

Modules/xASL_module_BIDS2Legacy.m
HenkMutsaerts: ???

I blindly accepted this because you said done, but I don't see where you have done this?

It seems you have now only done the first thing mentioned here (adding the xASL_adm_CorrectName), but didn't read my full message???

I provided a similar message in another function in this review about adding a warning boolean to xASL_adm_CorrectName, we may want to activate this here as well.

Are you in a rush when reading my comments? Doesn't really help except frustration ;)

HenkMutsaerts (changes requested):
Few minor things

Modules/Module_Import/Import/xASL_imp_DetermineStructureFromSourcedata.m
jan-petr: I understood it already the first time and did exactly what you asked for and wrote that I did that - unspecified/missing visit tokens were renamed to "missingSessionToken" sessions. I thought this will work without adding a dummy alias, which tests showed it won't, but there was no need to explain the whole issue again.

Anyway - it's done now.

Modules/Module_Processing/xASL_init_SubjectList.m
jan-petr: 1. xASL_adm_CorrectName doesn't accept a boolean for reporting. We report changes ourselves after applying xASL_adm_CorrectName.
2. OK
3. We accept BIDS. But A) BIDS doesn't forbid special characters by default, so we need to do it ourselves. B) We of course do the failsafe on both places. We have to do it also already when loading here, because we save the output to x-struct, so we can't have the non-corrected name there.

Modules/Module_Import/Import/xASL_imp_DetermineStructureFromSourcedata.m
jan-petr: OK. But better to loop over cells ourselves.

Modules/Module_Import/Import/xASL_imp_DetermineSubjectStructure.m
jan-petr: That's a subfunction for temp-data. The visit-names were renamed to missionSessionValue already. This is a failsafe.

Modules/Module_Import/xASL_wrp_NII2BIDS_Subject.m
jan-petr: All renamed to session.

But tokenVisitAlias and things inside imPar just have to stay visits...

Modules/Module_Import/Import/xASL_imp_DetermineSubjectStructure.m
HenkMutsaerts: 1. you didn't react to 
"session -> visit also in this function"
please read again (it means that the comments also need to be visits, or visit/session)

2. I'm confused. Your comment is not sufficiently explainable, so here are again 20 minutes of work :) At the first moment we process tokenVisitAliases in import, notice that it is missing, we recreate it from all existing visit values by saying that they are the same right?

Either this is 
situation A: we work with tempdata here.
In this case, the tokenVisitAlias stuff has been defined already above and this piece of code is unnecessary

situation B: we don't work with tempdata here (I.e. this is before the DICOM-conversion) and we should properly define the dummy tokenVisitAliases.

If you do want a failsafe, I would i) add a warning ii) convert the empty session name to something like 'sessionNameGoneWrong'. Again, in all cases we want everything converted to BIDS. If it is messy, the user can then read this in the BIDS stage. We don't want half of it converted because the other half had i/o issues and the user either a) not noticing that half was converted and reporting incomplete data OR b) has to go through the whole sourcedata folders to know if everything was converted properly or not.

So one important principle: we want to have the user know that everything was converted correctly. So errors or warnings are more important than code speed/failsafes that remove the potential for code crashing because of incorrect data. Of course, this would be in one of the import instances (assuming the import is a module where each subject has a log file).

Modules/Module_Processing/xASL_init_SubjectList.m
HenkMutsaerts: 1. xASL_adm_CorrectName should get a boolean for reporting, is what I say. This is much easier than having to code each instance "if changed, then report". And it serves as a reminder to provide feedback to the user for changing the name.
3. We have to have the non-corrected name here. I don't know what you mean by x-struct, but this is the point where a population-wide subject-list is read. Has nothing to do with the subject-specific x-struct.

A. When reading subject-list to memory: nothing changes in any of the subject/session/run name
B. When converting BIDS to legacy per subject, we correct the name in memory and for what we write to the disk.

BTW: assuming that rawdata and derivatives/ExploreASL always are on the same disk/filesystem etc, how can something be OK for the rawdata file/folder but not when saving as derivatives? It is exactly the same disk/filesystem with the same limitations/illegal characters. In Matlab strings, we can have whatever we want, so we're not restricted by our internal memory. So is this part even needed?

Modules/Module_Import/Import/xASL_imp_DetermineStructureFromSourcedata.m
```diff
@@ -240,73 +272,60 @@ function xASL_imp_PrintRun(x,thisSubject,thisVisit,thisRun)
...
             end
         else
             for iVisit=1:numel(x.importOverview.(sFieldName).visitIDs)
 				% Find the name id of the visit according to its number
-				idVisit = cellfun(@(y) regexp(y, x.importOverview.(sFieldName).visitIDs{iVisit}), x.modules.import.imPar.tokenVisitAliases(:,1), 'UniformOutput', false);
-				idVisit = find(cellfun(@(y) ~isempty(y), idVisit));
+				idVisit = cellfun(@(y) strcmp(y, x.importOverview.(sFieldName).visitIDs{iVisit}), x.modules.import.imPar.tokenVisitAliases(:,1), 'UniformOutput', false);
```
HenkMutsaerts: This was programmed in the way that tokenVisitAliases are regular expressions, are you sure this still works for that case? They should be regular expressions, like all aliases are.

Modules/Module_Import/Import/xASL_imp_DetermineStructureFromSourcedata.m
HenkMutsaerts: I'm not saying you understood things, I'm saying that if you don't communicate clearly to me what was done why, how am I supposed to check if I agree within reasonable time investment (i.e. without reading all the code? ;)

So: we have 2 situations, A) one with visitTokenAliases defined, B) one without being defined. We want to support not defining visitTokenAliases.

A) So you argue that if visitTokenAliases are explicitly defined, it would skip those that slightly differ or are empty. That's fine perhaps, because the user defining these tokens will probably have more control and check better than the user not defining visitTokenAliases, so we don't need to "help" the user by also converting the odd ones.
B) If you created a dummy visitTokenAliases from all visits, the empty ones need replacing by "missingSessionValue"


We might want to explain clearly that either A) all visit visitTokenAliases are explicitly defined, or B) visitTokenAliases are not defined at all. But that there is no in between.


Note also that the comment above is confusing, it says that for missing visit names, we will replace them by their number. Is this comment now incorrect, or is this part still replacing missing visit names?

i.e., does this code does:

A) convert
visitTokenAliases = {'^visit_1', 1, '^visit_2'}
to
visitTokenAliases = {'^visit_1', 1, '^visit_2', 2}

OR does this code does
B) create a dummy list of visitTokenAliases when visitTokenAliases was not provided?

HenkMutsaerts (changes requested):
Still a bit unclear to me. Perhaps because the import is very convoluted.

And I assume that in the case of missing tokenVisitAliases, you create a dummy tokenVisitAliases (including the alias '^$', 'missingSessionValue' for the empty visit)

, so the rest of the code doesn't need changing.

Modules/Module_Import/Import/xASL_imp_DetermineStructureFromSourcedata.m
jan-petr: > So: we have 2 situations, A) one with visitTokenAliases defined, B) one without being defined. We want to support not defining visitTokenAliases.

Yes. This was the original goal defined in the issue.

> A) So you argue that if visitTokenAliases are explicitly defined, it would skip those that slightly differ or are empty. That's fine perhaps, because the user defining these tokens will probably have more control and check better than the user not defining visitTokenAliases, so we don't need to "help" the user by also converting the odd ones. B) If you created a dummy visitTokenAliases from all visits, the empty ones need replacing by "missingSessionValue"

Following your previous comments, I have already implemented the following:
1) If tokens are defined, it only takes those that are defined and ignore those that are undefined - and report a warning for unknown ones.
2) If tokens are not defined, it copies them exactly.
3) Following your first comment, I have implemented that for non-defined tokens, it additionally converts a session with an empty token to "missingSessionValue" - e.g. case 2 above. Following your second comment, I have broaden the replacing the empty session by "missingSessionValue" for those cases where tokenVisitAliases are defined. Following your last comment, I have removed the renaming of empty tokens to missing when tokenVisitAliases are defined.

Because of this simplication (empty tokens are considered only when no tokenVisitAliases are defined), we can created the dummy aliases based on all tokens without removing the empty ones. The renaming of empty token to missingSessionValues can now happen only in nii2bids where all other alias renaming happens.

> We might want to explain clearly that either A) all visit visitTokenAliases are explicitly defined, or B) visitTokenAliases are not defined at all. But that there is no in between.

This will be added to the documentation before merging the issue. We have a task for this in the issue.

> Note also that the comment above is confusing, it says that for missing visit names, we will replace them by their number. Is this comment now incorrect, or is this part still replacing missing visit names?

This is not really confusing. We have three cases.
1) TokenvisitAliases defined -> use them.
2) TokenVisitAlises undefined, but tokenVisit defined. create aliases that equal the tokens
3) TokenVisit undefined.
For case 3, where we don't really even attempt to extract tokens for visit (token order = [1 0 2 3]), we always add visit _1 to each subject. That's something we added in previous issues on your request.

> i.e., does this code does:
> 
> A) convert visitTokenAliases = {'^visit_1', 1, '^visit_2'} to visitTokenAliases = {'^visit_1', 1, '^visit_2', 2}
> OR does this code does B) create a dummy list of visitTokenAliases when visitTokenAliases was not provided?

The code does neither - it doesn't work with tokenAliases, because no token was extracted.

Modules/Module_Processing/xASL_init_SubjectList.m
jan-petr: 1. This doesn't make sense as each adm_Correct name is applied to a very different task. So one fixes subject names, one visit names, one path name. So having reporting inside would report a name change, but it would be very unclear what kind of change that was.
2. It doesn't load it to memory. It saves it to x.dataset.TotalSubjects, which I assume is used later, so it needs to corrected name. Anyway - that's a useless comment, I keep both the corrected (saved in x-struct and used for writing to legacy) and uncorrected name (to read rawdata) and things work.
3. We have agreed upon removing non-alphanumeric characters from BIDS to Legacy. It's all written in the issue. I can of course remove it if you really want, but it's a very very minor thing that almost never will happen.

Modules/Module_Import/Import/xASL_imp_DetermineSubjectStructure.m
jan-petr: > 1. you didn't react to
>    "session -> visit also in this function"
>    please read again (it means that the comments also need to be visits, or visit/session)

I haven't reacted because all was changed already as far as I can see.

> 2. I'm confused. Your comment is not sufficiently explainable, so here are again 20 minutes of work :) At the first moment we process tokenVisitAliases in import, notice that it is missing, we recreate it from all existing visit values by saying that they are the same right?

Yes

> Either this is situation A: we work with tempdata here. In this case, the tokenVisitAlias stuff has been defined already above and this piece of code is unnecessary

We work with tempdata here. Since the import is modular, we have to make sure that tokenVisitAlias is again correctly initiated as the renaming from  visit to visit alias happens in NII2BIDS and not in DCM2NII.
 
> situation B: we don't work with tempdata here (I.e. this is before the DICOM-conversion) and we should properly define the dummy tokenVisitAliases.

Case A is true, not case B.
> 
> If you do want a failsafe, I would i) add a warning ii) convert the empty session name to something like 'sessionNameGoneWrong'. Again, in all cases we want everything converted to BIDS. If it is messy, the user can then read this in the BIDS stage. We don't want half of it converted because the other half had i/o issues and the user either a) not noticing that half was converted and reporting incomplete data OR b) has to go through the whole sourcedata folders to know if everything was converted properly or not.
> 
> So one important principle: we want to have the user know that everything was converted correctly. So errors or warnings are more important than code speed/failsafes that remove the potential for code crashing because of incorrect data. Of course, this would be in one of the import instances (assuming the import is a module where each subject has a log file).

As above - all is converted to BIDS, missing sessions renamed to missingSessionValue -> this is done in dcm2nii. So all warnings are there. All previous checking is only constructing the list of subjects. Previously, I tried to have the '','missingSessionValue' token, but that's not necessary as the visit renaming anyway happens in nii2BIDS - so all is there now. No failsafes are used.

Modules/Module_Import/Import/xASL_imp_DetermineStructureFromSourcedata.m
jan-petr: Oh. That was a useful comment. I have to change this. I know use regexp (which is unable to handle empty strings) and if regexp fails, I try strcmp. This is done and can be resolved

HenkMutsaerts: I would just use a single regexp. Regexp is able to handle empty strings if you always add `|)` or `(|` by default right? You can resolve.

Modules/Module_Import/Import/xASL_imp_DetermineStructureFromSourcedata.m
HenkMutsaerts: Understood and agreed with all. Except for:

1. `Note also that the comment above is confusing` this concerned 
line 303: `% In case that visit names are not filled, we name the visit by their number`
This comment needs to be updated.

2.
`i.e., does this code does:

A) convert
visitTokenAliases = {'^visit_1', 1, '^visit_2'}
to
visitTokenAliases = {'^visit_1', 1, '^visit_2', 2}

OR does this code does
B) create a dummy list of visitTokenAliases when visitTokenAliases was not provided?`

Sorry by "this code" I meant the import code in general.
IF we have tokenVisit defined, but visitTokenAliases is {'^visit_1', 1, '^visit_2'}.
THEN what alias does the second visit token become?

Modules/Module_Import/Import/xASL_imp_DetermineSubjectStructure.m
HenkMutsaerts: 1. Line 153:
`	% Create a unique list of sessions`
`		listVisits = unique(listVisits);`
`		% Exclude empty session names`

so "all was not changed already" (or am I seeing something wrong?)

-> the temp step is an arbitrary in-between step, should not be "modular" necessarily.
We should not need to re-initiate tokenVisitAlias.
Either we manage the tokenVisitAlias fully at the start, or at the temp stage. Not partly at the start and partly at the temp stage.
But I guess this is something we need to do properly in the next issue.

It still shows me that it excludes empty session names, whereas you say that these are replaced by missingSessionValues (or here it should be 'sessionNameGoneWrong' -> if you read that part ;)

Do we miscommunicate because I see old code?

Modules/Module_Processing/xASL_init_SubjectList.m
HenkMutsaerts: 1. Not really, if you warn that a subject or visit or path name is changed, this is human readable
2. x.dataset.TotalSubjects == memory. I would not save both corrected and uncorrected, this is very messy in terms of provenance; i.e. harder to understand what was done if somebody reads later. I would only save corrected once the correction is performed on the disk. Until correction is actually done, we save uncorrected.
3. I know, and sometimes consequences of things that are planned only become clear when executing the planning. Fine to keep removing the non-alphanumeric characters, but it should at least issue a warning.

HenkMutsaerts (changes requested):

Modules/Module_Import/Import/xASL_imp_DetermineStructureFromSourcedata.m
jan-petr: This whole section was removed as completely useless. You can resolve.

Modules/Module_Processing/xASL_init_SubjectList.m
jan-petr: We added an output option to xASL_adm_CorrectName called bCorrected that is set to TRUE when some modifs were done. It actually makes no sense to return the original as the original was the input. And the new name is the output.

1. We use the output boolean of adm_CorrectName to decide on reporting the name changes and this is correctly used in DICOM2BIDS
2. I have removed the functionality of correcting names that are already in BIDS.

Modules/Module_Import/Import/xASL_imp_DetermineSubjectStructure.m
jan-petr: > 1. Line 153:
>    `	% Create a unique list of sessions`
>    `		listVisits = unique(listVisits);`
>    `		% Exclude empty session names`
> 
> so "all was not changed already" (or am I seeing something wrong?)

Yes - you are looking at the original comment - which of course displays the original commit (see the "outdated" label). This has been modified in between, but this is only displayed in the list file changes, and won't be altered here.

> -> the temp step is an arbitrary in-between step, should not be "modular" necessarily. We should not need to re-initiate tokenVisitAlias. Either we manage the tokenVisitAlias fully at the start, or at the temp stage. Not partly at the start and partly at the temp stage. But I guess this is something we need to do properly in the next issue.

You probably don't remember this. But this is your decision. We wanted to have a joint/semi-separate DCM2NII and NII2BIDS. But you strongly insisted that these two things have to be completely separated and fully modular. That took as a lot of work to separate them completely, but it was you who strongly insisted that we should do all the initializations again and from scratch. So you can't now complain about this ;)

> It still shows me that it excludes empty session names, whereas you say that these are replaced by missingSessionValues (or here it should be 'sessionNameGoneWrong' -> if you read that part ;)
> 
> Do we miscommunicate because I see old code?

Yes - here you of course only see the old code. See the label "outdated", it won't reinitialize this code excerpt.

HenkMutsaerts: I see your point. Then tokenVisitAlias should neither be initiated/managed in dcm2nii OR in NII2BIDS, but in the general import initialization right?

-> How do I see the new code with the old comment then? If I want to check if code is changed according to a comment (that is the whole point of seeing comments "inline")

HenkMutsaerts (approved):
Looks all good. Just to verify: we initiate tokenVisitAliases once, keep only a single list etc nicely simple? Or else we can fix that later :)

PR #1681, opened 2024-04-10
Bug #1674 q sequence revamp

### LINKED ISSUE #1674: x.Q.Sequence revamp
### Description

Use of x.Q.Sequence is an error-prone legacy thing. This is used in all processing although `Sequence` is complementary to the BIDS fields `PulseSequenceType` and `MRAcquisitionType`.

PulseSequenceType is actually a general BIDS field. This is read from DICOM and in case it is missing, I am correctly filling this based on other DICOM information. The correct BIDS approach is to have EPI/GRASE saved in PulseSequenceType and 2D/3D in a BIDS field MRAcquisitionType. 

### Tasks

I propose to make a new issue where we do the following:

- [x] Make sure that MRAcquisitionType and PulseSequenceType contain the correct information as specified above and that also these are always initialized. This will be done in import. 587b1973
- [x] parms2BIDS of BIDS2Legacy and Legacy2BIDS doesn't do anything to MRAcquisitionType+PulseSequence type and keep them as they are. Sequence and readoutDim are ignored. 4ced567c
- [x] parms2BIDS makes sure that Legacy to BIDS correctly transforms x.Q.Sequence to MRAcquisitionType+PulseSequenceType for old Legacy. And x.Q.readoutDim goes to MRAcquisitionType. This is a one way conversion for backwards compatibility. 4ced567c
- [x] Fix xASL_adm_DefineASLSequence 4ced567c
- [x] Make sure that all possible loading of dataPar/studyPar/ASL.json considers this
- [x] x.Q.Sequence field is removed everywhere and is replaced by the correct use of MRAcquisitionType+PulseSequenceType.
- [x] readoutDim field is removed from everywhere and replaced by MRAcquisitionType. 207ba233
- [x] Update the documentation.

### Testing
- [x] GE TestDataSet not finished
- [x] Unit-Test Check deprecated fields failed
- [x] Flavor Siemens_PCASL_multiSequence_VE11C_1 crashed
- [x] Minor flavor differences
- [x] Rerun flavor test
- [x] Rerun unit test

### Release notes

Stop using Legacy fields Sequence and readoutDim and replace them with BIDS fields MRAcquisitionType and PulseSequenceType

jan-petr:
### Linked issue

Close #1674


Functions/Administration/xASL_adm_DefineASLSequence.m
```diff
@@ -5,87 +5,86 @@
...
+    warning('xQ.MRAcquisitionType parameter missing');
 end
 
 % Check for illegal sequence definitions
-if isfield(xQ,'Sequence')
-    if isempty(regexpi(xQ.Sequence, '(2d_epi|3d_grase|3d_spiral)'))
-        warning(['Unknown ASL sequence: ' xASL_num2str(xQ.Sequence)]);
+if isfield(xQ, 'PulseSequenceType')
+    if isempty(regexpi(xQ.PulseSequenceType, '(epi|grase|spiral)', 'once'))
+        warning(['Unknown ASL sequence: ' xASL_num2str(xQ.PulseSequenceType)]);
```
HenkMutsaerts: should be

warning(['Unknown ASL readout PulseSequenceType: ' xASL_num2str(xQ.PulseSequenceType)]);

Functions/Administration/xASL_adm_DefineASLSequence.m
```diff
@@ -5,87 +5,86 @@
...
     bVerbose = true;
 end
 
-if ~isfield(xQ, 'readoutDim') || isempty(xQ.readoutDim)
-    warning('xQ.readoutDim parameter missing');
+if ~isfield(xQ, 'MRAcquisitionType') || isempty(xQ.MRAcquisitionType)
+    warning('xQ.MRAcquisitionType parameter missing');
 end
 
 % Check for illegal sequence definitions
```
HenkMutsaerts: This was done for Sequence which is MRAcquisitionType _ PulseSequenceType. So now we only do this for PulseSequenceType, let's also do this for MRAcquisitionType.

Functions/Administration/xASL_adm_DefineASLSequence.m
```diff
@@ -5,87 +5,86 @@
 % FORMAT: [xQ] = xASL_adm_DefineASLSequence(xQ, bVerbose)
```
HenkMutsaerts: Now we improve the nomenclature, shouldn't we call this function:

xASL_adm_DefineASLReadout

instead?

Functions/Administration/xASL_adm_DefineASLSequence.m
```diff
@@ -5,87 +5,86 @@
...
-elseif  isempty(regexpi(xQ.Vendor, 'Gold Standard Phantoms|GE|Philips|Siemens'))
+elseif  isempty(regexpi(xQ.Vendor, 'Gold Standard Phantoms|GE|Philips|Siemens', 'once'))
     warning('Unknown Vendor specified in xQ.Vendor');
-elseif ~isempty(regexpi(xQ.Vendor, 'Gold Standard Phantoms'))
+elseif ~isempty(regexpi(xQ.Vendor, 'Gold Standard Phantoms', 'once'))
     fprintf('%s\n', 'Digital Reference Object ASL-DRO detected');
 end
 
 
 %% Try to work out which ASL sequence we have
```
HenkMutsaerts: then this would become

%% Try to work out which ASL readout we have

(and similar perhaps throughout this function)

Functions/Administration/xASL_adm_DefineASLSequence.m
```diff
@@ -5,87 +5,86 @@
...
 % First assume that 2D is 2D EPI, irrespective of Vendor
-if ~isfield(xQ, 'Sequence') && isfield(xQ, 'readoutDim') 
-	if strcmpi(xQ.readoutDim,'2D')
-		xQ.Sequence = '2D_EPI';
+if ~isfield(xQ, 'PulseSequenceType') && isfield(xQ, 'MRAcquisitionType') 
+	if strcmpi(xQ.MRAcquisitionType, '2D')
+		xQ.PulseSequenceType = 'EPI';
```
HenkMutsaerts: shouldn't this be lowercase per BIDS?

I know we used uppercase but are increasingly using strcmpi and regexpi so this is a good check and easy BIDS compatibility.

Functions/Administration/xASL_adm_DefineASLSequence.m
```diff
@@ -5,87 +5,86 @@
...
         if bVerbose
-            if ~isempty(regexpi(xQ.Vendor,'Gold Standard Phantoms'))
+            if ~isempty(regexpi(xQ.Vendor, 'Gold Standard Phantoms', 'once'))
```
HenkMutsaerts: We could change xQ.Vendor to the BIDS keyname (Manufacturer) but I would not do that now.
We could though change x.Q.Sequence now to x.Q.Readout or something, which is more correct and also nice for ourselves.

Functions/Administration/xASL_adm_DefineASLSequence.m
```diff
@@ -5,87 +5,86 @@
...
 		    else
 			    fprintf('%s\n', '2D readout detected, assuming 2D EPI');
             end
         end
-	elseif isfield(xQ, 'Vendor') && strcmpi(xQ.readoutDim,'3D')
-		if  ~isempty(regexpi(xQ.Vendor, 'Philips')) || ~isempty(regexpi(xQ.Vendor, 'Siemens'))
-			xQ.Sequence = '3D_GRASE'; % assume that 3D Philips or Siemens is 3D GRASE
+	elseif isfield(xQ, 'Vendor') && strcmpi(xQ.MRAcquisitionType, '3D')
+		if  ~isempty(regexpi(xQ.Vendor, 'Philips', 'once')) || ~isempty(regexpi(xQ.Vendor, 'Siemens', 'once'))
+			xQ.PulseSequenceType = 'GRASE'; % assume that 3D Philips or Siemens is 3D GRASE
```
HenkMutsaerts: Again, I would make all lowercase, is more consistent and perhaps also BIDS compliant.

Functions/Administration/xASL_adm_DefineASLSequence.m
```diff
@@ -99,17 +98,17 @@
 % We default to 3D spiral because this sequence has whole-brain coverage
 % and negligible geometric distortion, so we don't do any special
 % readout-specific image processing for this sequence
-if ~isfield(xQ,'Sequence') || isempty(xQ.Sequence)
-    warning('No xQ.Sequence defined');
+if ~isfield(xQ, 'PulseSequenceType') || isempty(xQ.PulseSequenceType)
```
HenkMutsaerts: I should have changed this when making the population module sequence-independent.
Also note that this is not the same as checking Sequence (2d_epi vs epi).

We can/should remove all lines from 98 until 109.
If we cannot detect the PulseSequenceType and/or MRAcquisitionType, we simply give an error (we would never have this because we create ASL BIDS in our import)

1. In the ASL module, we always need x.Q.Readout. If it is missing, we can at the location where we need it (e.g., susceptibility mask creation) just not mask susceptibility -> the processing stays the same as what we do now, but without adding a potentially incorrect PulseSequenceType & MRAcquisitionType to the provenance.
2. In the population module, we assume that PulseSequenceType & MRAcquisitionType are in the json sidecar. If they are not (which should be a rare case), then we still not want incorrect PulseSequenceType & MRAcquisitionType but the effective resolution calculation part will give a warning and assume nominal resolution == effective resolution.

Modules/Module_Processing/Module_ASL/xASL_wrp_ProcessM0.m
```diff
@@ -146,7 +146,7 @@ function xASL_wrp_ProcessM0(x)
 
 if isfield(x.modules.asl, 'bRegisterM02ASL') && ~x.modules.asl.bRegisterM02ASL
     fprintf('M0 registration (to ASL or T1w) is skipped upon request\n');
-elseif ~strcmpi(x.Q.M0, 'UseControlAsM0') && isempty(regexpi(x.Q.Sequence, 'spiral'))
+elseif ~strcmpi(x.Q.M0, 'UseControlAsM0') && isempty(regexpi(x.Q.PulseSequenceType, 'spiral', 'once'))
```
HenkMutsaerts, line 149: So this has to be x.Q.Readout -> '3d_spiral' 

(note that we had this, assuming that x.Q.Sequence could only contain '2d_epi, 3d_grase, 3d_spiral, which we created ourselves. Now it can have 2d_spiral)

HenkMutsaerts (changes requested):
Note that you cannot do a simple find/replace, because the value/meaning of PulseSequenceType (e.g. epi) is not identical to Sequence (e.g., 2d_epi).

We still need Sequence (e.g., 2d_epi), because many of our processing is dependent on it. You could simplify to epi grase spiral but you can have 2d spiral or 3d epi, even though it doesn't occur frequently, we need to catch this.

So I suggest the following:

1. readoutDim -> MRAcquisitionType so it is BIDS compliant (both json and internal)
2. x.Q.Sequence -> x.Q.Readout
3. we add PulseSequenceType

jan-petr:
> Note that you cannot do a simple find/replace, because the value/meaning of PulseSequenceType (e.g. epi) is not identical to Sequence (e.g., 2d_epi).
> 
> We still need Sequence (e.g., 2d_epi), because many of our processing is dependent on it. You could simplify to epi grase spiral but you can have 2d spiral or 3d epi, even though it doesn't occur frequently, we need to catch this.
> 
> So I suggest the following:
> 
> 1. readoutDim -> MRAcquisitionType so it is BIDS compliant (both json and internal)
> 2. x.Q.Sequence -> x.Q.Readout
> 3. we add PulseSequenceType

This is explained in the issue. MRAcquisitionType and PulseSequenceType are both BIDS fields. x.Q.Readout isn't. So no point introducing Readout when MRAcquisitionType and PulseSequenceType provide all the necessary information.

Functions/Administration/xASL_adm_DefineASLSequence.m
jan-petr: Done

Functions/Administration/xASL_adm_DefineASLSequence.m
jan-petr: No. EPI is according to BIDS - I've checked that before implementing.

Functions/Administration/xASL_adm_DefineASLSequence.m
jan-petr: MRAcquisitionType_PulseSequenceType is really used in function xASL_adm_DefineASLResolution - I have restored the original functionality and I am checking for both readout and sequence now.

jan-petr: For all files where x.Q.Sequence was used, we are now checking for both PulseSequenceType and MRAcquisitionType. 

You can resolve.

Functions/Administration/xASL_adm_DefineASLSequence.m
jan-petr: It is still defining MRAcquisitionType and PulseSequenceType, so Sequence is equally valid as before. Readout isn't incorrect either, but I don't see why a change would be necessary.

Functions/Administration/xASL_adm_DefineASLSequence.m
jan-petr: Same as above - nothing has changed about the functionality.

Functions/Administration/xASL_adm_DefineASLSequence.m
jan-petr: Vendor -> Manufacturer. I agree, I made an issue #1683.

No point in changing x.Q.Sequence to x.Q.Readout as this is not a BIDS fields and we want to stay BIDS compatible -> we use MRAcquisitionType and PulseSequenceType which are BIDS compatible.

Functions/Administration/xASL_adm_DefineASLSequence.m
jan-petr: 3D and GRASE are BIDS compliant.

Functions/Administration/xASL_adm_DefineASLSequence.m
jan-petr: Removed and we give an error.

Modules/Module_Processing/Module_ASL/xASL_wrp_ProcessM0.m
jan-petr: Yes. I have changed this everywhere. But note that this old version was already looking for any 'spiral', e.g. 2D_spiral or 3D_spiral. So I can add 3D_spiral, but that would be changing the functionality compared to the older version.

HenkMutsaerts: The point is that in the old version, we used a parameter created by ourselves (x.Q.Sequence), which could exclusively be 2d_epi, 3d_grase, 3d_spiral. Now, we have replaced this by a BIDS parameter that has multiple options. So we changed the functionality (in general, not in the most common cases).

So we can only replace checking x.Q.Sequence by checking both MRAcquisition & PulseSequenceType

Functions/Administration/xASL_adm_DefineASLSequence.m
HenkMutsaerts: Because ASL sequence = labeling + readout. All this concerns the readout only. It's not a big deal, but if we 're changing the nomenclature anyway to more correct/clear, then readout is more clear than "sequence".

What sequence is it? A PCASL. Is said as more often than "What type of ASL sequence is it?" "2D_EPI"
So not (in)correct, just more clear.

I don't feel strong about this though :)

Functions/Administration/xASL_adm_DefineASLSequence.m
HenkMutsaerts: Not what I meant. I meant that we are checking the readout, not the sequence (sequence being labeling + readout)

HenkMutsaerts (changes requested):
1. The point is that in the old version, we used a parameter created by ourselves (x.Q.Sequence), which could exclusively be 2d_epi, 3d_grase, 3d_spiral. Now, we have replaced this by a BIDS parameter that has multiple options. So we changed the functionality (in general, not in the most common cases).

So we can only replace checking x.Q.Sequence by checking both MRAcquisition & PulseSequenceType

So again, check throughout all cases where you changed x.Q.Sequence to PulseSequenceType, and see if this should not be MRAcqsuition & PulseSequenceType

2. This is all about the readout, not the sequence (sequence == labeling+readout). So calling it readout instead of sequence can be more clear.

I feel strong about point 1, not about point 2

Modules/Module_Processing/Module_ASL/xASL_wrp_ProcessM0.m
jan-petr: Yes, I understand. And I can add the 3D check here. But my point was that it's not really necessary here. As said - we check for dimension everywhere else. I thought adding here is not necessary, but I can do it if you want.

jan-petr:
> 1. The point is that in the old version, we used a parameter created by ourselves (x.Q.Sequence), which could exclusively be 2d_epi, 3d_grase, 3d_spiral. Now, we have replaced this by a BIDS parameter that has multiple options. So we changed the functionality (in general, not in the most common cases).
> 
> So we can only replace checking x.Q.Sequence by checking both MRAcquisition & PulseSequenceType
> 
> So again, check throughout all cases where you changed x.Q.Sequence to PulseSequenceType, and see if this should not be MRAcqsuition & PulseSequenceType
> 
> 2. This is all about the readout, not the sequence (sequence == labeling+readout). So calling it readout instead of sequence can be more clear.
> 
> I feel strong about point 1, not about point 2

Point 1 -> this has been changed everywhere. We are checking both for readoutType and readoutDim everywhere. Except in one single case, where I think that checking for 3D is not necessary - see the comment above. But even there, I can add it if you want.

Point 2 -> It's a small change -> done

Functions/Administration/xASL_adm_DefineASLSequence.m
jan-petr: I updated to "readout".

Functions/Administration/xASL_adm_DefineASLSequence.m
jan-petr: Updated to readout.

Modules/Module_Processing/Module_ASL/xASL_wrp_ProcessM0.m
HenkMutsaerts: So everywhere (not only here) where you renamed x.Q.Sequence to *PulseSequenceType, you potentially need to add the MRAcquisitionType check (even though you think it won't matter). That is my only wish ;)

HenkMutsaerts (approved):

PR #1685, opened 2024-04-19
Closes #1680

### LINKED ISSUE #1680: Reading hematocrit from participants.tsv
### Description

When running ExploreASL('path', 0, 0) – after adding 'hematocrit' column to participants.tsv inside derivatives folder – a warning appears saying that it cannot read participants.tsv

We realized that is was due to this double quoting marks "" in the subject names in the 'participants_id' column

ANOTHER BUG [17-4-2024]:
image

### Tasks

- [x] Ensure that the participants.tsv file in derivatives does not add double quotation marks in the participants_id column, when created.
- [x] Check if there are double quotation marks, and remove them if found.
- [x] read x.Q.Hematocrit(iSubjectSession) from x.Hematocrit (from SetsIDs from participants.tsv)

### Release notes

Fix bug in loading and using Hematocrit values.

HenkMutsaerts:
### Linked issue

Closes #1680

### DIFF
--- a/Modules/Module_Processing/Module_ASL/xASL_wrp_Quantify.m
+++ b/Modules/Module_Processing/Module_ASL/xASL_wrp_Quantify.m
@@ -202,7 +202,8 @@ end
 
 % c. We convert x.Hematocrit -> x.Q.BloodT1
 if isfield(x,'Hematocrit')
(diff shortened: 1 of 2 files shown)


jan-petr (approved):
OK

PR #1686, opened 2024-04-19
Closes #1667

### LINKED ISSUE #1667: ConvertDicomFolderStructure_CarefulSlow doesn't detect weird DICOM filenames
### Description

`ConvertDicomFolderStructure_CarefulSlow` doesn't move files in cases of weird DICOM filenames.
In the case of a period (.) in the filename, the extension may be detected incorrectly and we need to append .dcm.

@jan-petr you may know a better solution for my initial commit

JAN: I've checked the older solution and I don't see a reason this wouldn't work with '.' anywhere in the filename. I've tried different scenarios and they all worked with both versions.

Try filenames like:

00245087q346097q46.szlghzsrlguhq345097.dcm.

[~, name] = fileparts (...)

will give

name 00245087346097q46 (only the part before the first period).

So it will try to copy 00245087346097q46.dcm, which doesn't exist.

JAN: Yes, I have tried exactly this and it wasn't a problem
[~,Fname] = fileparts('/home/test/00245087q346097q46.szlghzsrlguhq345097.dcm')
Fname =
    '00245087q346097q46.szlghzsrlguhq345097'

Sorry, the problem occurs when the DICOM files have no extension:

path =

    'path/00245087q346097q46.szlghzsrlguhq345097'

>> [~, name] = fileparts(path)

name =

    '00245087q346097q46'

So the old code would take fullfile(path, [name '.dcm]) and copy this, but this file doesn't exist because the file had no extension and a dot inside the filename.

### Release notes

Better filename handling when renaming DICOM folder-structure.

HenkMutsaerts:
### Linked issue

Closes #1667

### DIFF
--- a/WorkInProgress/Import/ConvertDicomFolderStructure_CarefulSlow.m
+++ b/WorkInProgress/Import/ConvertDicomFolderStructure_CarefulSlow.m
@@ -32,8 +32,14 @@ end
             NewDir = fullfile(ROOT,Fname);
             xASL_adm_CreateDir(NewDir);
 
-            [~, Pname] = fileparts(Flist{iL});
-            NewFile = fullfile(NewDir,[Pname '.dcm']);
+            [~, Pname, Pext] = fileparts(Flist{iL});
+            if strcmpi(Pext, '.ima') || strcmpi(Pext, '.dcm')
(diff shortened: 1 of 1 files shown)


jan-petr (approved):
OK

PR #1688, opened 2024-04-21
#1687 Fix empty visit token in import

### LINKED ISSUE #1687: Bug in Import with missing and non-defined visit
### Description

Minor bug in import

### Tasks

- [x] String correction of a missing string is incorrectly reported as a change.
- [x] when visit aliases nor token are defined, the import crashes

HENK: 
Section 1: No VisitToken defined -> default is to take what is recognized and call this all visit 1
(We could create a dummy visitToken in this case that is like ‘’, that will recognize all as visit 1, if you want to reduce the number of if-statements)
-> the remainder of the code should not need to know if there is a VisitToken or not (there is always one, or a dummy one). So this is probably more tricky to get more “modular/independent” but not impossible
JAN: I though about this solution, but it proved to be practically very difficult. Plus it would mean that we randomly start renaming visits to "1" which isn't really correct. The easiest solution is to keep VisitToken empty and visitTokenAliases empty. It doesn't produce extra if-statements.

HENK: I meant that if no VisitToken is defined, and also no VisitTokenAlias is defined (i.e., there are NO visits), then we default to empty vistToken () and  VisitTokenAlias = {'', '_1'}. We already do this effectively, but at another location, with an if-statement (if no visitToken was defined, then set ... to 1). My suggestion would ensure that after the first reading/managing of visit tokens, throughout the whole import you always have both a visitToken and visitTokenAlias.
-> But I don't feel strong about this, you can remove this.

+ visitTokenAlias always gets an empty visit token, irregardless of what happened before in the code. Because in all scenarios, we want to have an empty visit token, translate it to visit 1, AND give a warning.
JAN: For user-defined visitTokenAlias we agreed that no empty token is added and empty session are reported as an error. 
HENK: Are you sure? I think I proposed to add an empty token and have empty sessions as warning, with the name "VisitTokenMissing" or something. The reason is that inside rawdata, the user can easily curate all this, whereas this is challenging for most users in the sourcedata. But perhaps we discussed this already and I forgot that we decided what you said, in that case you can ignore this.
JAN: Yes, I implemented "empty" visit handling for both visitTokenAliases defined and non-defined. And renaming to "missingSessionValue" with a warning. But then we decided that we only issue an error without renaming when this doesn't fit to the user-provided tokenAliases, so I removed that functionality.

### Release notes

Nothing. Related to #1635

jan-petr:
### Linked issue

Close #1687

### DIFF
--- a/Functions/Administration/xASL_adm_CorrectName.m
+++ b/Functions/Administration/xASL_adm_CorrectName.m
@@ -109,12 +109,10 @@ function [strOut, bCorrected] = xASL_adm_CorrectName(strIn, bOption, strExclude)
         strOut = strrep(strOut,'_','-');
     end
     
-	% Report when the string was modified
-    if ~strcmp(strIn, strOut) 
-		% strcmp doesn't work properly when both strings have zero length, so we have to rule out this option before reporting a change
-		if numel(strIn) || numel(strOut)
-			bCorrected = true;
-		end
+	% Report when the string was modified (difference between strIn and strOut). 
+	% Though note that strcmp doesn't work properly when both strings have zero length, so we have to rule out this option before reporting a change
+	if ~strcmp(strIn, strOut) && (numel(strIn) || numel(strOut))
+		bCorrected = true;
 	end
 end
 
--- a/Functions/Administration/xASL_adm_CorrectName.m
+++ b/Functions/Administration/xASL_adm_CorrectName.m
@@ -110,8 +110,11 @@ function [strOut, bCorrected] = xASL_adm_CorrectName(strIn, bOption, strExclude)
     end
     
 	% Report when the string was modified
-    if ~strcmp(strIn, strOut)
-		bCorrected = true;
+    if ~strcmp(strIn, strOut) 
+		% strcmp doesn't work properly when both strings have zero length, so we have to rule out this option before reporting a change
+		if numel(strIn) || numel(strOut)
+			bCorrected = true;
+		end
 	end
 end
 
--- a/Modules/Module_Import/Import/xASL_imp_DetermineSubjectStructure.m
+++ b/Modules/Module_Import/Import/xASL_imp_DetermineSubjectStructure.m
@@ -113,6 +113,7 @@ function [x] = xASL_imp_DetermineStructureFromTempdata(x)
     
 	% Initialize an empty list of visits
 	listVisits = {};
+	bVisitsDetected = false; % Set to true once at least a single visit is detected
 
     % Get subjects from list
     if numel(listSubjectsVisits)>0
@@ -129,6 +130,7 @@ function [x] = xASL_imp_DetermineStructureFromTempdata(x)
                 % Multi-visit notation
                 x.SUBJECTS{iSubVis} = curSubVis(1:indexSeparator-1);
 				listVisits{iSubVis} = curSubVis(indexSeparator+1:end);
+				bVisitsDetected = true; % a true visit (with an empty or non-empty name was detected)
 			else
 				% Multiple separators
                 warning('In BIDS, subject and session names shouold not contain hyphens (-) or underscores (_) as these are used as separators. Removing them in the final BIDS values/names');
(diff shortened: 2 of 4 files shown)


Functions/Administration/xASL_adm_CorrectName.m
```diff
@@ -110,8 +110,11 @@
     end
     
 	% Report when the string was modified
-    if ~strcmp(strIn, strOut)
-		bCorrected = true;
+    if ~strcmp(strIn, strOut) 
+		% strcmp doesn't work properly when both strings have zero length, so we have to rule out this option before reporting a change
```
HenkMutsaerts: Nicely catched. Next time I would add this as && instead of creating an extra condition, this is a smaller commit/change.

Modules/Module_Import/xASL_wrp_NII2BIDS_Subject.m
```diff
@@ -109,7 +109,7 @@
 		% The session name is OK. We correct it for special characters
 		[bidsLabel.visit, bCorrected] = xASL_adm_CorrectName(sessionName, 2);
 		if bCorrected
-			warning(['Subject ' subjectName ', changed visit name from: ' sessionName ' into ' bidsLabel.visit]);
+			warning(['Subject ' subjectName ', changed visit name from: ' sessionName ', into: ' bidsLabel.visit]);
```
HenkMutsaerts: I would shorten:

warning([subjectName ' changed visit name: ' sessionName ' -> ' bidsLabel.visit]);

Modules/Module_Import/Import/xASL_imp_DetermineSubjectStructure.m
```diff
@@ -113,6 +113,7 @@
     
 	% Initialize an empty list of visits
 	listVisits = {};
+	bVisitsDetected = false; % Set to true once at least a single visit is detected
```
HenkMutsaerts: We need to be sparingly with adding more booleans, I would only add new booleans if really needed, because the more variables the more difficult code becomes to read.

Probably a nicer way to do this, is replace bVisitsDetected throughout by numel(listSubjectsVisits)>0 [or ~isempty(listSubjectsVisits)]

-> smaller change in github history
-> less booleans -> easier to read (you don't have to search back in the code what the boolean means)

ALSO: note that now `bVisitsDetected` has no meaning for the use case where tokenVisitAliases was defined. This makes it confusing to read, because what happens when `bVisitsDetected` is false but tokenVisitAliases was defined? This should give a warning or whatsoever. You would answer me that this doesn't matter, is taken into account in other parts of the code. And then the answer of the first-time reader is "but then the general `bVisitsDetected` is a confusing boolean. -> Hence, should be replaced by a non-boolean (or call the boolean bVisitsDetected_UseOnlyWhenNoTokenVisitAliasesAreDefined which is senseless ;)

Modules/Module_Import/Import/xASL_imp_DetermineSubjectStructure.m
```diff
@@ -150,13 +152,19 @@
...
+		if bVisitsDetected
+			% Create a unique list of visits
+			listVisits = unique(listVisits);
+
+			% Create a list of unique visits names
+			if ~isempty(listVisits)
+				x.modules.import.imPar.tokenVisitAliases = listVisits(:);
+				x.modules.import.imPar.tokenVisitAliases(:,2) = x.modules.import.imPar.tokenVisitAliases(:,1);
+			end
+		else
```
HenkMutsaerts: I would in general suggest that we start with the default/fallback case for an if-statement (which is usually a small part -> we default/fallback to empty something) but then the true case can give a page of code. So easier to read the default/fallback version first and then read what happens in the true case (long code).

HenkMutsaerts (changes requested):
Some code improvement suggestions, I agree with the bugfix of course.
We just need to keep our code easy to understand for non-Jan and non-Henk ;) (and increase our efficiency for debugging!)

Functions/Administration/xASL_adm_CorrectName.m
jan-petr: I wanted to make it easier to understand. But fine with single line.

Done - you can resolve.

Modules/Module_Import/xASL_wrp_NII2BIDS_Subject.m
jan-petr: Agree. Done. You can resolve.

Modules/Module_Import/Import/xASL_imp_DetermineSubjectStructure.m
jan-petr: I made more comments in the code. But we can't replace the boolean in the way you suggested. Because that would simply not work at all.

Your suggestion "numel(listSubjectsVisits)>0" wouldn't work as this list contains a mixture of subject-visits. We are not interested in subject-visits but in what visits we have. We are primarily interested if the subjects are "sub1_" - where visit name is missing. Or "sub1" where visit name wasn't defined. It is really important to know if visit wasn't defined so we ignore missing visits. Or in case where visit names are simply missing and we need to replace this with "missingSession value".

However, we are constructing listVisits only to recreate tokenVisitAliases if needed. So we can skip nonexisting visits and only list empty visits. 

So i removed the boolean and changed the way the listVisit is indexed.
Done. You can resolve this.

Modules/Module_Import/Import/xASL_imp_DetermineSubjectStructure.m
jan-petr: OK. No prob. I have switched the statements. 

Done. You can resolve.

Modules/Module_Import/Import/xASL_imp_DetermineSubjectStructure.m
HenkMutsaerts: Thanks.
Sorry I now saw only the changes, with the residual code compressed in between. So it seemed as if 

bVisitsDetected == numel(listSubjectsVisits)>0

but I now see this is not the case.

So you can ignore this comment, and resolve once you read this.

HenkMutsaerts (approved):

Modules/Module_Import/Import/xASL_imp_DetermineSubjectStructure.m
jan-petr: Yes. I know. It looked simple and the initial implementation was like this, until I started testing properly :)

PR #1691, opened 2024-04-21
#1690 Fix Saving of PWI4D.json

### LINKED ISSUE #1690: MRAcquisition not defined for PWI4D.json
### Description

PWI4D.json only contains PLD, LabDur, and other basic parameters. We avoid saving all parameters, cause not all JSON parameters for ASL4D are still valid for PWI4D. But the trouble is that when loading all JSON, we runs BIDS2Parms function which runs ASLDefineReadout. And this last function is then missing MRAcquisitionType, PulseSequenceType, and Vendor files. It therefore crashes. We have several options how to fix this.

1.⁠ ⁠Put defineASLReadout where it was initially -> at the start of the ASL module only and not inside BIDS2Legacy JSON conversion.
2.⁠ ⁠⁠Suppress all warnings of missing PulseSequenceType and MRAcquisitionType as these will be bothering us all the time.
3. For PWI4D etc, we will additionally save Vendor, MRAcquisitionType, PulseSequenceType as these are fields that should in theory be there and valid also for PWI4D.

Solution 3 is probably the most optimal as easy and without compromise. So probably my choice. Solution 1 would also work, but you didn't want that. Solution 2 - I don't really like.

HENK:Yes, solution 3 is the most future-proof and won't go wrong if merging different sequences. Ideally, all sequence-dependent parameters are copied from ASL4D.json to PWI4D.json etc. I would just implement this consistently for all relevant parameters that we use in the ASL module, it becomes messy when some parameters are read from ASL4D.json, some from memory, and some from PWI4D. Of course authors, site-name etc don't need copying.

JAN: Yes, lets consistently copy this from a single ASL4D.json location. Just check if in all cases, we provide a reference ASL4D.json to copy from or if we are sometimes saving just PLD/LabDur/TE from memory - then we have to add the reference.
Manufacturer
MagneticFieldStrength
PulseSequenceType
MRAcquisitionType
EchoTime
SliceTiming
FlipAngle
RepetitionTimePreparation
ArterialSpinLabelingType
PostLabelingDelay
BackgroundSuppression
LabelingDuration
### Tasks

- [x] In all cases of ASL subtraction - make sure that those fields above are saved (Manufacturer, MagneticFieldStrength, PulseSequenceType, MRAcquisitionType, ArterialSpinLabelingType)
- [x] Those were already saved (EchoTime, PostLabelingDelay, LabelingDuration)
- [x] These are tricky as they are given per slice and we can't give them per volume if they would differ between two merged sequences (SliceTiming)
- [x] These can in theory be given as vectors, but that would require a new issue (FlipAngle, RepetitionTimePreparation, BackgroundSuppression)

### Release notes

Nothing. Report with #1674

jan-petr:
### Linked issue

#1690

### DIFF
--- a/Functions/InputOutput/xASL_io_ASLSubtractionAveraging.m
+++ b/Functions/InputOutput/xASL_io_ASLSubtractionAveraging.m
@@ -217,7 +217,21 @@ function xASL_io_ASLSubtractionAveraging_sub_SavePWI_JSON(x, path2save, image2sa
     
         if isfield(x.Q, fieldNameTE)
             jsonFields.EchoTime = x.Q.(fieldNameTE);
-        end
+		end
+
+		% Even if we don't copy all orig-json, we save certain basic parameters
+		listFields2Save = {'Vendor' 'MagneticFieldStrength' 'PulseSequenceType' 'MRAcquisitionType' 'LabelingType'};
+
+		% We check the existence of all these fields in both x or x.Q
+		for iField = 1:length(listFields2Save)
+			if isfield(x.Q, listFields2Save{iField})
+				jsonFields.(listFields2Save{iField}) = x.Q.(listFields2Save{iField});
+			end
+
+			if isfield(x, listFields2Save{iField})
+				jsonFields.(listFields2Save{iField}) = x.(listFields2Save{iField});
+			end
+		end
     
         xASL_io_SaveNifti(pathReference, path2save, image2save, 32, 0, [], bCopyOrigJson, jsonFields, true);
     end


Functions/InputOutput/xASL_io_ASLSubtractionAveraging.m
```diff
@@ -217,7 +217,21 @@ function xASL_io_ASLSubtractionAveraging_sub_SavePWI_JSON(x, path2save, image2sa
     
         if isfield(x.Q, fieldNameTE)
             jsonFields.EchoTime = x.Q.(fieldNameTE);
-        end
+		end
+
+		% Even if we don't copy all orig-json, we save certain basic parameters
+		listFields2Save = {'Vendor' 'MagneticFieldStrength' 'PulseSequenceType' 'MRAcquisitionType' 'LabelingType'};
+
+		% We check the existence of all these fields in both x or x.Q
```
HenkMutsaerts, line 225: Nice, but I don't get it why you have EchoTime separately then above? Shouldn't that be in this list?

And what about the other parameters that I suggested to you? I would just copy all sequence-related BIDS parameters, those that you need for quantification. E.g., FA could also differ between sequences that you merge.

HenkMutsaerts (changes requested):

Functions/InputOutput/xASL_io_ASLSubtractionAveraging.m
jan-petr: All have a very good reason:
1. TE,LD,PLD are all separate because you are saving to JSON as TE, but inside x.Q, you have TE_PWI4D, TE_PWI3D etc - so you need to pick the correct one based on what you are saving (control3D, PWI4D etc). Please don't criticize this, you have introduced this :)

2. So Vendor, PulseSequence etc are things that are always the same as in ASL4D and don't differ between PWI3D or PWI4D. So we can save them directly without checking. And that's why we previously thought it's not worth saving them :)

3. The other parameters you suggested - that's all written in the issue. For your convenience - see a copy here:
- We now save these fields as well (Manufacturer, MagneticFieldStrength, PulseSequenceType, MRAcquisitionType, ArterialSpinLabelingType)
- Those were already saved previously (EchoTime, PostLabelingDelay, LabelingDuration)
- These are tricky as they are given per slice and we can't give them per volume if they would differ between two merged sequences (SliceTiming). We would have to create a special matrix, discuss it's format etc. So all really tricky. We might need to deal with that in the future. So far it wasn't a problem as all merged sequences were 3D. We can create a new issue for this.
- These can in theory be given as vectors, but we would have to properly deal with them and do exactly the same handling as we do with TE and PLD and LabDur. It wouldn't be enough to add them as this wouldn't work nicely for merged sequences. So we can do that, but that would require a new issue (FlipAngle, RepetitionTimePreparation, BackgroundSuppression).

HenkMutsaerts: 1. Then I would do exactly the same for the other parameters, for two reasons:
* merging (you can merge from different sequences/scanners)
* consistency. Easier to understand for others if everything works the same way than some parameters in this way and other parameters that way

I'm not saying we have to implement this now but I would encourage it because it makes it easier to read:
A. all parameters are loaded the same way, depending on the filename they are suffixed 4D/3D/none.
B. merging etc all does the same for all parameters
c. etc

2. You probably thought. I suggested from the start to do everything the same way afaik :)
3. SliceTiming: Good point, and especially here I agree that it would take too much time for something that wouldn't occur often
4. As per (1): I would add all parameters as vectors.

If you want to keep it as you have it now, I would suggest only:
A. Adding (FlipAngle, RepetitionTimePreparation, BackgroundSuppression) to the second list
B. adding big comments to the sections where parameters are loaded differently, saying that this is pending on revamping

HenkMutsaerts (changes requested):

Functions/InputOutput/xASL_io_ASLSubtractionAveraging.m
jan-petr: > 1. Then I would do exactly the same for the other parameters, for two reasons:
> 
> * merging (you can merge from different sequences/scanners)
> * consistency. Easier to understand for others if everything works the same way than some parameters in this way and other parameters that way
> 
> I'm not saying we have to implement this now but I would encourage it because it makes it easier to read: A. all parameters are loaded the same way, depending on the filename they are suffixed 4D/3D/none. B. merging etc all does the same for all parameters c. etc
> 
> 2. You probably thought. I suggested from the start to do everything the same way afaik :)
> 3. SliceTiming: Good point, and especially here I agree that it would take too much time for something that wouldn't occur often
> 4. As per (1): I would add all parameters as vectors.
> 
> If you want to keep it as you have it now, I would suggest only: A. Adding (FlipAngle, RepetitionTimePreparation, BackgroundSuppression) to the second list B. adding big comments to the sections where parameters are loaded differently, saying that this is pending on revamping

We *can't* do most of what is suggested above for many reasons.
For (Manufacturer, MagneticFieldStrength, PulseSequenceType, MRAcquisitionType, ArterialSpinLabelingType):
1. We can't add them as a vector as BIDS strictly forbids that.
2. It makes no sense to have those fields as vectors for PWI4D - it is just useless.
3. It makes no sense to have those fields as vectors for merged sequences because we are not merging sequences across vendors/MFstrengths/LabelingTypes, it doesn't make any sense to do that, ExploreASL wouldn't be able to handle that, even BASIL is not able to handle that, there's not even a theoretical framework how to deal with that.
4. It doesn't make sense to do this now or even later. The best we can do is to produce a warning when merging two incompatible sequences.

For (FlipAngle, RepetitionTimePreparation, BackgroundSuppression)
1. We are fixing an error now, not extending functionality. For extending functionality, we define a new issue.
2. We can't add them to the second list, as this would create an inconsistency. We either have to add as vectors (but we need a new issue), or we don't add them at all.
3. If we add flipangle and repetitionTime as vectors, then we need to update a lot, we need to update BASIL etc. That's not a simple change to be done with this small bugfix.
4. BackgroundSuppression - the same conditions apply as for the manufacturer. This simply can't be done easily. We probably have a theoretical framework how to quantify, but not even BASIL supports that.

HenkMutsaerts (approved):
I see your point. I would indeed in a new issue deal with this, either as vectors or single numbers. Just to keep things consistent, now for parameters we go to:
A) memory
B) ASL4D
C) PWI4D

You would just in the new extension (BIDS derivatives) have a good criterion for this, like e.g., "parameters that can be changed easily for a sequence are saved as vectors, examples PLD, slice thickness, etc." And "parameters that cannot easily be changed are saved as scalar, examples FieldStrength, Manufacturer, etc."

PR #1697, opened 2024-04-23
Feature #1696 dicom sorting

### LINKED ISSUE #1696: Improve DICOM sorting script
### Description

The script for sorting DICOMs to directories needs cleaning and improving (ConvertDICOMStructure_carefulSlow)

### Tasks

- [x] Add an option to dcmtk_Read to skip DCMTK. And use xASL_io_DcmtkRead directly. ad496dc
- [x] Rename the sorting function to a proper name. 3e2441c8
- [x] Clean DCMTK reading and let be errors handled inside the dcmtk_reading function. 3e2441c8
- [x] Add Acquisition Number. 49bdf8a
- [x] Make a recursive function for deleting directories. 49bdf8a
- [x] Allow converting several directories at once. 49bdf8a
- [x] Don't check for values in TE and TR - these are missing in report files. ImageType is present always though. 49bdf8a

### Release notes

Function for sorting DICOMs to directories based on series names.

jan-petr:
### Linked issue

Close #1696


Functions/Administration/xASL_adm_SortDicomToFolders.m
```diff
@@ -0,0 +1,117 @@
+function xASL_adm_SortDicomToFolders(pathDICOM, nDirLayers, bUseDCMTK, bVerbose)
+%xASL_adm_SortDicomToFolders Sorts DICOMs in the pathDICOM folder to directories based on the sequence name
```
HenkMutsaerts: Do you mean based on SeriesDescription (or ProtocolName if SeriesDescription doesn't exist)?

Functions/Administration/xASL_adm_SortDicomToFolders.m
```diff
@@ -0,0 +1,117 @@
+function xASL_adm_SortDicomToFolders(pathDICOM, nDirLayers, bUseDCMTK, bVerbose)
```
HenkMutsaerts, line 1: What is the difference between this function and ConvertDicomFolderStructure?

And why is this one in Adm and the other in WIP?

Functions/InputOutput/xASL_io_DcmtkRead.m
```diff
@@ -16,21 +17,32 @@
...
+if nargin < 3 || isempty(bTryDCMTK)
+	bTryDCMTK = true;
+end
+
+if bTryDCMTK
+	try
+		% Read using DCMTK
+		header = xASL_mex_DcmtkRead(filepath, bPixel);
+	catch ME
+        fprintf('%s%s\n','Warning, xASL_io_DcmtkRead failed with following warning: ', ME.message);
```
HenkMutsaerts: Better to issue a Matlab warning and put ME.message on the next line right?

Functions/InputOutput/xASL_io_DcmtkRead.m
```diff
@@ -185,49 +197,65 @@
...
 	if isfield(timingItem,'RepetitionTime')
 		header.RepetitionTime = timingItem.RepetitionTime;
 	end
 end
 
 %% Philips fields
-if isfield(header,'PerFrameFunctionalGroupsSequence')
-	echoItem = header.PerFrameFunctionalGroupsSequence.Item_1.MREchoSequence.Item_1;
+if isfield(header, 'PerFrameFunctionalGroupsSequence')
+	privatItem = [];
```
HenkMutsaerts: e missing privateItem

privat is German ;)

HenkMutsaerts (changes requested):
Looks good, I just need some minor explanation. Just so I am uptodate :)

Functions/Administration/xASL_adm_SortDicomToFolders.m
jan-petr: I wanted to avoid using the DICOM tag names and have a short version to describe what it does. And then give the DICOM tag names in the full description. 

Happy to change this, but would prefer not to name all the DICOM tags here specifically.

Functions/Administration/xASL_adm_SortDicomToFolders.m
jan-petr: this is the same function :) I just wanted to get rid of "CarefulSlow" and wanted a better description - we are sorting files rather than converting them. And also, I wanted to clean this and move to the main functions instead of keeping it in WIP.

Feel free to suggest a better name. Though I don't thing "Convert" was a good one.

Functions/InputOutput/xASL_io_DcmtkRead.m
jan-petr: Fixed

Functions/InputOutput/xASL_io_DcmtkRead.m
jan-petr: Actually - I put a fprintf on purpose. This is about DCMTK failing and a fallback to SPM reading of DICOM. So things work fine in the end. That's why I think we can print a minimal warning rather than using 'warning' as this would print 10 lines per each file for a problem that is moderately severe.

I have modified the warning to make this clear.

Functions/Administration/xASL_adm_SortDicomToFolders.m
HenkMutsaerts: OK, I have extended the explanation here: you can also move this to the longer explanation/definition.
And you can resolve :)

%xASL_adm_SortDicomToFolders Sorts DICOMs in the pathDICOM folder to directories based on the sequence name as defined in DICOM tags "SeriesDescription" or "ProtocolName". Note that these are most likely the names set in the protocol on the scanner, or changed by the radiographics.

HenkMutsaerts (approved):
Read my final conversation which you can resolve and you can merge

Functions/Administration/xASL_adm_SortDicomToFolders.m
jan-petr: Done

PR #1704, opened 2024-04-29
Closes #1689 CommasInFoldername

### LINKED ISSUE #1689: Several bugs when processing
### Description

In some functions we check for a comma ',' as a comma in the extension should be removed: this occurs with spm paths in the format: 

dir/filename.nii(.gz),1
dir/filename.nii(.gz),2
dir/filename.nii(.gz),n

for different volumes in a 4D NIfTI file.

When there is a comma in a path (e.g. when mounting), it is useful if we don't have to move data to avoid having to resync it.

Example for me now:

'/Users/hjmutsaerts/surfdrive - Mutsaerts, H.J.M.M. (Henk-Jan)@surfdrive.surf.nl/HolidayPics/studyName'

Matlab otherwise works fine with this path, so spaces, commas, @ don't seem to be a problem in the foldername.

When converting csv to tsv, the comma is converted to a tab, creating another column. The number of columns will be inconsistent and csvtotsv will crash.
We can do two things (or both):
1. Not put the directory name in a stats/tsv file
2. Recognize when something is a path and ignore any delimiters in a path (this is perhaps more tricky)

Note that this is only valid for directory names, not for filenames. BIDS subject & visit names should not include illegal chars.

### Tasks

- [x] Extend xASL_fileparts that it detects ,1 as well as .nii(.gz)
- [x] Check for other potential path problems
- [x] Minor bug fix in xASL_adm_DeleteManyTempFiles, for xASL_TrackProgress, and speeding up by removing recursive file searching where not needed
- [x] xASL_qc_SortBySpatialCoV: Print warnings only once. We still repeat the fprintf for the scans that may need checking, but avoid repeating warnings for huge verbosity.
- [x] xASL_stat_PrintStats: bugfixes empty end of list. When scans are missing, they are not included but the rest should still be reported, filling empties with 'n/a'. Did didn't happen with empties at the end of the list, because the list was shorter than expected, this function crashed. Now fixed.
- [x] ps_LST_tlv: avoid listing the path, which avoids seeing a comma in a path as delimiter
- [x] xASL_qc_CollectQC_Structural & xASL_stat_GetVolumeStatistics: correct indices for not having a path in the tsv-files
- [x] xASL_wrp_GetVolumetrics: revamp (accordingly)
- [x] xASL_adm_GzipAllFiles: UNIX bugfix

### Release notes

Deals with commas in filename.

HenkMutsaerts:
### Linked issue

Closes #1689

### DIFF
--- a/External/SPMmodified/xASL/xASL_fileparts.m
+++ b/External/SPMmodified/xASL/xASL_fileparts.m
@@ -14,9 +14,11 @@ function [Fpath, Ffile, Fext, SuffixSPM] = xASL_fileparts(InputPath)
 %              If a file ending at nii.gz is given, then the whole nii.gz is returned as the extension.
 %              Does not verify the existence of the file, or existence of .nii or .nii.gz.
 %              This function uses the following steps:
-%              1. Catch SPM suffixes
-%              2. Manage .gz (double) extensions  
-%              3. Manage folders differently than files
+%
+%              1. Skip folders
+%              2. Catch SPM volume suffixes
+%              3. Manage .nii.gz (double) extensions  
+%
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
 % EXAMPLES: ['/path' 'file' '.nii']         = xASL_fileparts('/path/file.nii');
 %          ['/path' 'file' '.nii.gz']      = xASL_fileparts('/path/file.nii.gz');
@@ -34,20 +36,31 @@ function [Fpath, Ffile, Fext, SuffixSPM] = xASL_fileparts(InputPath)
         InputPath = InputPath{1};
     end
 
-    %% 1. Catch SPM suffixes
     [Fpath, Ffile, Fext] = fileparts(InputPath);
 
+    % Default output volume Suffix for SPM
+    SuffixSPM = '';
+
+    %% 1. Skip folders
+    if exist(InputPath, 'dir')
+        % when running this for a folder, there is no extension. So any 
+        % extension should be ignored and put back to the Ffile
+        Ffile = [Ffile Fext];
+        Fext = '';
+        return;
+    end
+
+
+    %% 2. Catch SPM volume suffixes
     Index1 = regexp(Fext, ',\d+$');
 
-	if isempty(Index1)
-		% No SPM suffix in a format like filename.nii,1 detected
-		SuffixSPM = '';
-	else      
+	if ~isempty(Index1)
+		% If a SPM volume suffix in a format like filename.nii,1 is detected
 		SuffixSPM = Fext(Index1:end);
 		InputPath = fullfile(Fpath, [Ffile Fext(1:Index1-1)]);
 	end
 
-    %% 2. Manage .gz (double) extensions
+    %% 3. Manage the double extension .nii.gz
     [Fpath, Ffile, Fext] = fileparts(InputPath);
     [~, Ffile2, Fext2] = fileparts(Ffile);
     
@@ -56,12 +69,5 @@ function [Fpath, Ffile, Fext, SuffixSPM] = xASL_fileparts(InputPath)
         Fext = [Fext2 Fext];
     end
     
-    %% 3. Manage folders differently than files
-    if exist(InputPath, 'dir')
-        % when running this for a folder, there is no extension. So any 
-        % extension should be ignored and put back to the Ffile
-        Ffile = [Ffile Fext];
-        Fext = '';
-    end
 
 end
--- a/External/SPMmodified/xASL/xASL_fileparts.m
+++ b/External/SPMmodified/xASL/xASL_fileparts.m
@@ -39,15 +39,13 @@ function [Fpath, Ffile, Fext, SuffixSPM] = xASL_fileparts(InputPath)
 
     Index1 = regexp(Fext, ',\d+$');
 
-    if isempty(Index1)
-        error('Something wrong with this file extension');
-        else
-            SuffixSPM = Fext(Index1:end);
-            InputPath = fullfile(Fpath, [Ffile Fext(1:Index1-1)]);
-        end
-    else
-        SuffixSPM = '';
-    end
+	if isempty(Index1)
+		% No SPM suffix in a format like filename.nii,1 detected
+		SuffixSPM = '';
+	else      
+		SuffixSPM = Fext(Index1:end);
+		InputPath = fullfile(Fpath, [Ffile Fext(1:Index1-1)]);
+	end
 
     %% 2. Manage .gz (double) extensions
     [Fpath, Ffile, Fext] = fileparts(InputPath);
--- a/External/SPMmodified/xASL/xASL_fileparts.m
+++ b/External/SPMmodified/xASL/xASL_fileparts.m
@@ -1,28 +1,31 @@
-function [Fpath, Ffile, Fext] = xASL_fileparts(InputPath)
-% Wrapper around the fileparts.m that treats nii.gz as an extension
-% (this also works with other extensions)
+function [Fpath, Ffile, Fext, SuffixSPM] = xASL_fileparts(InputPath)
+%xASL_fileparts Wrapper around fileparts for special extensions
 % FORMAT: [Fpath, Ffile, Fext] = xASL_fileparts(InputPath)
 %
 % INPUT:
-%   InputPath - input file name with path, ending either .ext or .ext.gz
+%   InputPath - input file name with path, ending either .ext(.gz) or .ext(.gz),n
 % OUTPUT:
 %   Fpath     - path or empty string
 %   Ffile     - file name
 %   Fext      - file extension or empty string
+%   SuffixSPM - volume extension for NIfTI files (per spm usage)
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
 % DESCRIPTION: Returns the path, file name, and file extension for InputPath using the fileparts.m function.
 %              If a file ending at nii.gz is given, then the whole nii.gz is returned as the extension.
-%              Does not verify the existence of the file, or existence of .nii or .nii.gz
-% EXAMPLE: xASL_fileparts('/path/file.nii');
-%          xASL_fileparts('/path/file.nii.gz','file');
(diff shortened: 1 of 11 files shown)


External/SPMmodified/xASL/xASL_fileparts.m
```diff
@@ -1,63 +1,69 @@
...
         InputPath = InputPath{1};
     end
-        
-    % Now catch SPM suffixes
-    [Ind1, Ind2] = regexp(InputPath,',\d');
+
+    %% 1. Catch SPM suffixes
+    [Fpath, Ffile, Fext] = fileparts(InputPath);
+
+    [Ind1, Ind2] = regexp(Fext, ',\d+');
```
jan-petr: To avoid bugs, best to look at endings only
',\d+$'
otherwise, it could find this in the middle of the path for weird cases like this... "/tmp/test,1Henk/file.nii"

if doing this, you can skip Ind2

`Ind1 = regexp(Fext, ',\d+$');`

External/SPMmodified/xASL/xASL_fileparts.m
```diff
@@ -1,63 +1,69 @@
...
     if ~isempty(Ind1)
-        SuffixSPM = InputPath(Ind1:Ind2);
-        InputPath = InputPath(1:Ind1-1);
+        if Ind2~=length(Fext)
```
jan-petr: See above, we look for file endings only, so skip Ind2 and you don't need this condition again as this will never occur.

External/SPMmodified/xASL/xASL_fileparts.m
```diff
@@ -1,63 +1,69 @@
...
+            error('Something wrong with this file extension');
+        else
+            SuffixSPM = Fext(Ind1:Ind2);
```
jan-petr: And you replace Ind2 with end

Modules/Module_Processing/Module_Population/xASL_adm_DeleteManyTempFiles.m
```diff
@@ -46,8 +46,8 @@ function xASL_adm_DeleteManyTempFiles(x)
     %% 2. Native space — ASL module files
     fprintf('Deleting temporary ASL files in native space folders:    ');
 
-    for iP=1:length(Files2DelNativeASL)
-        xASL_TrackProgress(iP, length(Files2DelNativeASL));
+    for iPath=1:length(Files2DelNativeASL)
+        xASL_TrackProgress(iPath, length(Files2DelNativeASL));
         
         for iSession=1:length(listASLDone)
             sessionIndex = regexp(listASLDone{iSession}, 'ASL_\d*$');
```
jan-petr, line 53: iP -> iPath on line 58!!!

Modules/Module_Processing/Module_Structural/xASL_wrp_GetVolumetrics.m
```diff
@@ -23,53 +23,65 @@ function xASL_wrp_GetVolumetrics(x)
...
-        GMvol = catVol.S.subjectmeasures.vol_abs_CGW(2)/1000; % volume was in cm^3, /1000 = dm^3 or Liter
-        WMvol = catVol.S.subjectmeasures.vol_abs_CGW(3)/1000;
-        CSFvol = catVol.S.subjectmeasures.vol_abs_CGW(1)/1000;
+fileCAT12Volume = ['cat_' x.P.STRUCT '_' x.P.SubjectID '.mat']; % CAT12 results
+pathCAT12Volume = fullfile(x.D.TissueVolumeDir, fileCAT12Volume);
+
+fileSPM12Volume = [x.P.STRUCT '_seg8.mat']; % SPM12 results
+pathSPM12Volume = fullfile(x.dir.SUBJECTDIR, fileSPM12Volume);
+
+SaveFile = fullfile(x.D.TissueVolumeDir, ['TissueVolume_' x.P.SubjectID '.tsv']); % ExploreASL results
```
jan-petr: to avoid doubt - should we rename SaveFile to SaveFileTSV?

jan-petr (changes requested):
A few minor things

Modules/Module_Processing/Module_Structural/xASL_wrp_GetVolumetrics.m
HenkMutsaerts: Done in comment "review suggestions Jan"

Modules/Module_Processing/Module_Population/xASL_adm_DeleteManyTempFiles.m
HenkMutsaerts: Thanks! Done in comment "review suggestions Jan"

External/SPMmodified/xASL/xASL_fileparts.m
HenkMutsaerts: >> [fPath, fFile, fExt] = fileparts('/tmp/test,1Henk/file.nii')

fPath =

    '/tmp/test,1Henk'


fFile =

    'file'


fExt =

    '.nii'

I don't mind replacing this, but the Fext already has the end. It could still be any weird thing right?
E.g., we would catch fExt = '.nii,1.gz'
Anyway, I don't mind much here :)

jan-petr: You are right. That example above will work. But I meant a weird case, but that is not likely to happen but in theory it could.

Like
[Fpath, Ffile, Fext] = fileparts('perfectlynormalifile/savedbyHenk.M.M.Mutsaerts,1April-withoutextension')
For that, we would loose a large part of the file as a supposedly SPM extension.

So my correction would be able to handle this without affecting the other functionality.

HenkMutsaerts: [Fpath, Ffile, Fext] = fileparts('perfectlynormalifile/savedbyHenk.M.M.Mutsaerts,1April-withoutextension')

Fpath =

    'perfectlynormalifile'


Ffile =

    'savedbyHenk.M.M'


Fext =

    '.Mutsaerts,1April-withoutextension'

my solution:


[Ind1, Ind2] = regexp(Fext, ',\d+')

[Ind1, Ind2] = regexp(Fext, ',\d+')

Ind1 =

    11


Ind2 =

    12

Your solution:

Ind1 = regexp(Fext, ',\d+$')

Ind1 =

     []


But the whole point here is to fix an extension. So if the extension is incorrect, the whole code is moot. This code only happens for NIfTIs, where they obligatory end with .nii(.gz). But I'm still happy to change to your code, it will still function the same :)

Probably your code is better, issuing an error when the catched SPM extension is empty. So I implemented this in d6d9e196dcefd6856168a35fbd52eb08ca906ca5.

But I now spotted that this is actually worse, because:

1. the Fext was .nii. Then we want an empty regexp, and say SuffixSPM = ' ';
2. the Fext was .nii,\d Then we want a SuffixSPM
3. something went wrong (your example) and we want to throw an error. So I should probably revert to my solution, because with your solution; it is more coding to differentiate between [1] and [3]

jan-petr: Yes, you are of course right in your observation with the current code. Because I haven't described the proposed change fully. 
To make my change fully working, you of course remove the condition
``if isempty(Index1)
        error('Something wrong with this file extension');``
Because having commas in the filename, but not having the "comma+number" in the end is a situation that can easily occur and it shouldn't be an error, but it should only mean that there's no SPM extension and there are weird commas inside.

If Index1 is empty, then SuffixSPM = ''.  If not empty, we assign the suffix.
This way, we make it work with nii.gz, we make it work with weird filenames "fileblabla,1April, blabla.nii.gz" etc.

I fixed this in e0675af

HenkMutsaerts: No, I disagree, because there should not be a comma in the filename. That is why I wrote we should revert to what I had. So your solution is technically better, but content-wise not:

This function does:

1. A filename with a comma is illegal (but this is for the BIDS validity check, not for this function)
    So it either ignores the comma, or throws an error (which is what we had)
2. A filename with .nii.gz extension needs to be dealt with as .nii extension
3. A filename with .nii(.gz),1 extension needs to be dealt with as .nii extension
4. Anything that doesn't fit should lead to an error (you have removed this functionality)
5. Foldername with comma should be fine and passed through
-> for this we potentially need to change xASL_io_ReadNifti as well:
I created a branch on my computer to fix this:
tempBranch_xASL_io_ReadNifti

jan-petr: I'm confused - you are saying that a filename with comma is illegal. And that's true for BIDS, but not for any other file. And fileparts is a function that should work universally.

So your solution will work nicely for BIDS, but will potentially crash for any other file format and it will crash for BIDS as well if there's a comma in the folder name.

HenkMutsaerts: Good point, xASL_fileparts is designed to fix NIfTIs only. Initially for .nii vs .nii.gz, later to also detect ,1 ,2 ,3 volume additions by SPM.

So it should first run
1. [folder, filename, extension] = fileparts(pathInput);

2. folder can contain commas or whatever.

3. If we detect a NIfTI file (extension contains .nii) then we continue (otherwise we skip our addition)

4. filename cannot contain a comma (but it doesn't matter for this code, it will be catched elsewhere, and we don't change the filename with this code)

5. Now we're left with the extension, if we detect .nii that is followed by a comma, we remove the comma

Do you agree with this proposition?

jan-petr: Henk, please read the discussion above. 

We have a working solution. This works perfectly with commas in foldername. It works perfectly with commas in filename (and yes, these are allowed and filenames do include commas sometimes).

So I don't understand why can't we accept this solution and have to change it.

jan-petr (approved):
All solved

jan-petr (changes requested):
Edits in printStats directly conflict with previously approved and merged #1723.

The combination creates code that has bug and won't run...

jan-petr (approved):
Conflicts resolved. Ready to merge.

PR #1706, opened 2024-05-16
Bug #1705 tumor population

### LINKED ISSUE #1705: Tumor population analysis
### Description

Making sure that the population analysis runs correctly for Ivar with Lesion_T1_1.nii

### Tasks

- [x] Incorrect removal of temporary data.
- [x] Make sure the standard space analysis runs.
- [x] Native space analysis has issues with MRAcquisitionType not found.
- [x] Make sure the native space analysis runs.
- [x] Check that Lesion files are created in native space
- [x] Run stats also on Lesion files -> or create a new issue for that. Will be done in #58 

### Release notes

Transform Lesion and ROI - T1 files into ASL space.

jan-petr:
### Linked issue

Close #1705

### DIFF
--- a/Modules/Module_Processing/Module_ASL/xASL_wrp_PreparePV.m
+++ b/Modules/Module_Processing/Module_ASL/xASL_wrp_PreparePV.m
@@ -34,10 +34,7 @@ function x = xASL_wrp_PreparePV(x, bStandardSpace)
 %
 % EXAMPLE: xASL_wrp_PreparePV(x);
 % __________________________________
-% Copyright (C) 2015-2020 ExploreASL
-
-
-
+% Copyright (C) 2015-2024 ExploreASL
 
 %% ------------------------------------------------------------------------------------------
 %% 0)   Admin
@@ -181,7 +178,7 @@ end
 % The maps have then have to be binarized again with a given threshold before using them again as masks.
 
 % Create a list of lesions
-LesionROIList = xASL_adm_GetFileList(fullfile(x.D.ROOT, x.SUBJECT), '(?i)^(Lesion|ROI)_(T1|FLAIR|T2)_\d*\.nii)', 'FPList', [0 Inf]);
+LesionROIList = xASL_adm_GetFileList(fullfile(x.D.ROOT, x.SUBJECT), '(?i)^(Lesion|ROI)_(T1|FLAIR|T2)_\d*\.nii', 'FPList', [0 Inf]);
 
 if ~isempty(LesionROIList)
     fprintf('\n');
--- a/Modules/Module_Processing/Module_ASL/xASL_wrp_PreparePV.m
+++ b/Modules/Module_Processing/Module_ASL/xASL_wrp_PreparePV.m
@@ -143,22 +143,54 @@ else
 end
 
 %% ------------------------------------------------------------------------------------------
-% 6     Prepare pGM and pWM in the ASL native space
-% by presmoothing before downsampling to ASL space
+% 6     Prepare pGM, pWM, pCSF, WMH_Segm, and also Lesion and ROI maps in the ASL native space
+
+% We need to take into account the resolution difference between T1 and ASL.
+% The correct downsampling is done by first presmoothing the image and then reslicing using normal SPM routines.
+
+% Transform GM and WM segmentation to the ASL space
 xASL_im_PreSmooth(x.P.Path_PWI,x.P.Path_c1T1, x.P.Path_PVgm,x.S.optimFWHM_Res_mm,[],x.P.Path_mean_PWI_Clipped_sn_mat, 1);
 xASL_im_PreSmooth(x.P.Path_PWI,x.P.Path_c2T1, x.P.Path_PVwm,x.S.optimFWHM_Res_mm,[],x.P.Path_mean_PWI_Clipped_sn_mat, 1);
 xASL_spm_reslice(x.P.Path_PWI, x.P.Path_PVgm, x.P.Path_mean_PWI_Clipped_sn_mat, 1, x.settings.Quality, x.P.Path_PVgm);
 xASL_spm_reslice(x.P.Path_PWI, x.P.Path_PVwm, x.P.Path_mean_PWI_Clipped_sn_mat, 1, x.settings.Quality, x.P.Path_PVwm);
 
-if xASL_exist(x.P.Path_c3T1) % for backward compatibility, where c3T1 wasnt always available
+% Transform CSF segmentation to the ASL space
+if xASL_exist(x.P.Path_c3T1) % Check file existence as for older versions c3T1 was not always available
 	xASL_im_PreSmooth(x.P.Path_PWI,x.P.Path_c3T1,x.P.Path_PVcsf,x.S.optimFWHM_Res_mm,[],x.P.Path_mean_PWI_Clipped_sn_mat, 1);
 	xASL_spm_reslice(x.P.Path_PWI, x.P.Path_PVcsf, x.P.Path_mean_PWI_Clipped_sn_mat, 1, x.settings.Quality, x.P.Path_PVcsf);
 end
 
+% Transform WMH segmentation maps to the ASL space
 if xASL_exist(x.P.Path_WMH_SEGM) 
 	xASL_im_PreSmooth(x.P.Path_PWI,x.P.Path_WMH_SEGM, x.P.Path_PVwmh,x.S.optimFWHM_Res_mm,[],x.P.Path_mean_PWI_Clipped_sn_mat, 1);
 	xASL_spm_reslice(x.P.Path_PWI, x.P.Path_PVwmh, x.P.Path_mean_PWI_Clipped_sn_mat, 1, x.settings.Quality, x.P.Path_PVwmh);
 end
 
+% Transform Lesion and ROI masks to the ASL space. Note that we transform binary masks to maps with values between 0 and 1 so that the PV factor is recorded.
+% The maps have then have to be binarized again with a given threshold before using them again as masks
+
+% Create a list of lesions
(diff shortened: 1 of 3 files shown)


Modules/Module_Processing/Module_ASL/xASL_wrp_PreparePV.m
```diff
@@ -143,22 +143,54 @@
...
+
+% Create a list of lesions
+LesionROIList = xASL_adm_GetFileList(fullfile(x.D.ROOT, x.SUBJECT), '(?i)^(Lesion|ROI)_(T1|FLAIR|T2)_\d*\.(nii|nii\.gz)', 'List', [0 Inf]);
+
+% Go through all the lesions
+for iROI = 1:length(LesionROIList)
+	LesionROIName = LesionROIList{iROI};
+
+	% Unzip and remove .gz extension if necessary
+	if ~isempty(regexpi(LesionROIName, '\.gz$', 'once'))
```
HenkMutsaerts: This is redundant code, you can just run xASL_adm_UnzipNifti which will automatically only unzip if needed

Modules/Module_Processing/Module_ASL/xASL_wrp_PreparePV.m
```diff
@@ -143,22 +143,54 @@
...
 end
 
+% Transform WMH segmentation maps to the ASL space
 if xASL_exist(x.P.Path_WMH_SEGM) 
 	xASL_im_PreSmooth(x.P.Path_PWI,x.P.Path_WMH_SEGM, x.P.Path_PVwmh,x.S.optimFWHM_Res_mm,[],x.P.Path_mean_PWI_Clipped_sn_mat, 1);
 	xASL_spm_reslice(x.P.Path_PWI, x.P.Path_PVwmh, x.P.Path_mean_PWI_Clipped_sn_mat, 1, x.settings.Quality, x.P.Path_PVwmh);
 end
 
+% Transform Lesion and ROI masks to the ASL space. Note that we transform binary masks to maps with values between 0 and 1 so that the PV factor is recorded.
+% The maps have then have to be binarized again with a given threshold before using them again as masks
```
HenkMutsaerts: Where do we binarize again?
The threshold to binarize is unknown to the user? So depending on resolution differences, the mask will change. 
I would binarize again in the new space (which also saves space, and then we control this just using p>0.5). One other thing we could do, is adapting the threshold such that the volume stays the same in old and new space. But I don't know how significant this bias could be/how important this is. I would in any case not leave this to the user:

We have either probabilistic or binary masks. This should always be due to the segmentation, not to our resampling. Because it becomes then difficult to judge by the user if the probabilities are "real" probabilities (method/PV) or "induced by smoothing/resampling".

Modules/Module_Processing/Module_ASL/xASL_wrp_ResampleASL.m
```diff
@@ -319,8 +319,6 @@ function xASL_wrp_ResampleASL(x)
     end
     xASL_adm_DeleteFilePair(x.P.Path_rtemp_despiked_ASL4D, 'mat', 'json');
     xASL_adm_DeleteFilePair(x.P.Path_temp_despiked_ASL4D, 'mat', 'json');
-else
```
HenkMutsaerts, line 322: Why is this removed? I cannot judge this from this issue?

Probably we want to keep the temporary ASL4D when DELETETEMP is off, because you use it to create new derivatives in standard space, for CBF4D or whatsoever. There must be a reason I put it there ;)

Modules/Module_Processing/xASL_init_DefaultEffectiveResolution.m
```diff
@@ -48,6 +48,17 @@
     return;
 end
 
+% Obtain the x.Q.MRAcquisitionType field from the json
```
HenkMutsaerts: Are you sure we also have PulseSequence here? Because remember we concluded they should always be together (one is 2D/3D, the other is EPI/GRASE/spiral).

Modules/Module_Processing/xASL_init_DefaultEffectiveResolution.m
```diff
@@ -48,6 +48,17 @@
...
+if ~isempty(tJson) && isfield(tJson, 'Q') && isfield(tJson.Q, 'MRAcquisitionType')
+	x.Q.MRAcquisitionType = tJson.Q.MRAcquisitionType;
+end
+
+% If sequence is still missing we skip this function
```
HenkMutsaerts, line 49: Will it not be defaulted to something that is potentially wrong? Not sure if we should always avoid crashing in case of "illegal" data.

HenkMutsaerts (changes requested):

Modules/Module_Processing/Module_ASL/xASL_wrp_PreparePV.m
jan-petr: I know. xASL_adm_UnzipNifti can be run without it. But I need to adapt the name of the lesion for input/output and also for calling SPM.

Modules/Module_Processing/Module_ASL/xASL_wrp_PreparePV.m
jan-petr: Currently nowhere, because we are not yet doing native-space analysis with ROIs and Lesions -> this is currently missing and has to be implemented. But rather than doing the thresholding here, it might be wiser to do it directly in the population moduel as we can nicely adapt this and change this later or even make it somehow variable. 
Haha - of course not leaving that to the user. I'm just leaving the question of correct threshold for the population module. This comment is not for the user but for our record.

Fully agree about probabilistic/binary. But since I call this PVLesion_T1_1, it makes it clear in the name that this is probabilistic rather than binary. 

Yes - in ideal world, we don't change probabilistic to binary and vice versa. But tumor are rather tricky and I don't want to make a clear decision on how to threshold it. It would be wrong to cut it at 50% as we can miss a large part of the tumor as these tend to be very thin at certain times. So this decision should be made in the population module and we either consider the PV or we threshold.

Modules/Module_Processing/Module_ASL/xASL_wrp_ResampleASL.m
jan-petr: That was a bug. tmp_despiked was already saved in the code above. 
So this was a forgotten saving that was useless as no "ASL" variable exists anymore at this line. 

So we were revamping the code, moved the saving above and forgot to delete this line in previous issues.

Modules/Module_Processing/xASL_init_DefaultEffectiveResolution.m
jan-petr: Yes. I'm sure. We have always a combination PulseSequenceType and MRAcquisitionType (like you say). And we always load them together and evaluate together. But we somehow forgot to do this also here.

Modules/Module_Processing/xASL_init_DefaultEffectiveResolution.m
jan-petr: Previously, we were defaulting to 3DGrase or something. And you decided to remove this defaulting. 

I'm happy to insert "crash" here if you want...

Modules/Module_Processing/Module_ASL/xASL_wrp_PreparePV.m
HenkMutsaerts: I would argue why the ASL native space lesion mask needs to be implemented. "Because Ivar asks for it" is not a good argument. Otherwise, I see your point.

Modules/Module_Processing/Module_ASL/xASL_wrp_PreparePV.m
HenkMutsaerts: Doesn't xASL_adm_UnzipNifti give the bare .nii name as output? I thought that was the whole reason of this function, to avoid 5 lines of code.

Modules/Module_Processing/xASL_init_DefaultEffectiveResolution.m
HenkMutsaerts: I would say that we need a single check that either MRAcquisitionType or PulseSequenceType is missing, and then we issue an error. Because this should already be covered by BIDS.

We can also add this to our generic parameter check at the beginning of the ASL module, then you can remove the check here (because submodule you start the ASL module with, it always needs to pass this generic parameter check).

Modules/Module_Processing/Module_ASL/xASL_wrp_PreparePV.m
jan-petr: no. it gives the full path. i need the name only...

Modules/Module_Processing/Module_ASL/xASL_wrp_PreparePV.m
jan-petr: we wanted to have native space analysis and that has to be complete. It's only that I forgot about this option when doing native space analysis. And only Ivar actually discovered this shortcoming. So it's not adding that because he wanted it. But it's somethign that should have been there from the get go, but it wasn't and he discovered that.

Modules/Module_Processing/xASL_init_DefaultEffectiveResolution.m
jan-petr: We are already checking the parameter presence at the start of ASL modules.

I've changed the code to an error. So in case they are missing, I'm issuing a full error. But that's a safety catch cause that situation wouldn't happen here.

Modules/Module_Processing/Module_ASL/xASL_wrp_PreparePV.m
HenkMutsaerts: You know my opinion ;)

I have now fixed this by adding many warnings and making the code clear that this needs fixing when we fix the whole native space mess.

Modules/Module_Processing/Module_ASL/xASL_wrp_PreparePV.m
HenkMutsaerts: See my fix

HenkMutsaerts (approved):
OK, if you agree with my minor revamp

PR #1708, opened 2024-05-16
#1707 Fix mirrored Siemens MOSAIC import

### LINKED ISSUE #1707: Siemens MOSAIC incorrectly imported
### Description

dcm2niix is struggling with import of Siemens MOSAIC images. The worst issue is when the Phoenix field is missing. When at the same time is the ImageOrientationPatient expecting superior to inferior orientation, the output is incorrectly rendered, mirroring the brain above the top of the head.

This special case needs to be detected and mirrored back manually after dcm2niix.

Example is:
Siemens_PCASL_3DGRASE_VE11C_5

### Tasks

- [x] Identify the special case
- [x] Flip the orientation and matrix manually
- [x] Finalize and add the flavor
- [x] Run flavor testing

### Release notes

Deal with mirrored Siemens MOSAIC DICOMs.

jan-petr:
### Linked issue

Close #1707

### DIFF
--- a/External/SPMmodified/xASL/xASL_io_SaveNifti.m
+++ b/External/SPMmodified/xASL/xASL_io_SaveNifti.m
@@ -226,7 +226,8 @@ end
 %% ====================================================================================
 %% 3. Create new NIfTI
 if ~isempty(changeMat)
-	newNifti.mat = changeMat;
+	newNifti.mat  = changeMat;
+	newNifti.mat0 = changeMat;
 end
 
 newNifti.dat.scl_slope = 1;
--- a/External/SPMmodified/xASL/xASL_io_SaveNifti.m
+++ b/External/SPMmodified/xASL/xASL_io_SaveNifti.m
@@ -240,6 +240,7 @@ if ~isempty(changeMat)
 end
 
 if ~isempty(changeMat0)
+	warning('Note that mat0 (the original NIfTI orientation) is changed upon request. This can jeopardize re-running ExploreASL when done in the processing module!');
 	newNifti.mat0 = changeMat0;
 end
 
--- a/External/SPMmodified/xASL/xASL_io_SaveNifti.m
+++ b/External/SPMmodified/xASL/xASL_io_SaveNifti.m
@@ -1,7 +1,7 @@
-function xASL_io_SaveNifti(pathOrigNifti, pathNewNifti, imNew, nBits, bGZip, changeMat, bCopyOrigJson, JsonFields, bLegacy2BIDS, bOverwrite)
+function xASL_io_SaveNifti(pathOrigNifti, pathNewNifti, imNew, nBits, bGZip, changeMat, bCopyOrigJson, JsonFields, bLegacy2BIDS, bOverwrite, changeMat0)
 % Save a file to a Nifti format, while taking the parameters from another file
 %
-% FORMAT: xASL_io_SaveNifti(pathOrigNifti, pathNewNifti, imNew[, nBits, bGZip, changeMat, bCopyOrigJson, JsonFields, bLegacy2BIDS, bOverwrite])
+% FORMAT: xASL_io_SaveNifti(pathOrigNifti, pathNewNifti, imNew[, nBits, bGZip, changeMat, bCopyOrigJson, JsonFields, bLegacy2BIDS, bOverwrite, changeMat0])
 %
 % INPUT:
 %   pathOrigNifti  Path to the original Nifti file to take parameters from (REQUIRED)
@@ -26,6 +26,8 @@ function xASL_io_SaveNifti(pathOrigNifti, pathNewNifti, imNew, nBits, bGZip, cha
 %   bOverwrite     Vector with 3 Booleans for overwriting the following pre-existing destination files:
 %                  1) NIfTI file, 2) JSON file, 3) mat-orientation motion file (for 4D NIfTIs) (OPTIONAL, DEFAULT = [1 1 1]; 
 %                  1 -> [1 1 1] and 0 -> [0 0 0] and creates a warning; [1 0] - error
+%   changeMat0     New orientation matrix mat0 4x4. Note that mat0 should always stay as it was originally from the scanner, 
+%                  but we can exceptionally change it if mat0 was completely wrong (OPTIONAL, DEFAULT same as previous)
 %                  
 %
 % JSON saving options are:
@@ -53,7 +55,7 @@ function xASL_io_SaveNifti(pathOrigNifti, pathNewNifti, imNew, nBits, bGZip, cha
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
 %
 % __________________________________
-% Copyright 2015-2023 ExploreASL
+% Copyright 2015-2024 ExploreASL
 
 %% ----------------------------------------------------------------------------
 %% Admin
@@ -100,6 +102,14 @@ elseif length(bOverwrite) == 1
 	warning('bOverwrite should have a length of 3');
 elseif length(bOverwrite) == 2
 	error('bOverwrite should have a length of 3');
+end
+
+if nargin < 11 || isempty(changeMat0)
+	changeMat0 = [];
+else
+	if ~isequal(size(changeMat0),[4 4])
+		error('changeMat0 has to be 4x4');
+	end
 end
 
 % JSON saving, see detailed explanation above.
@@ -227,7 +237,10 @@ end
 %% 3. Create new NIfTI
 if ~isempty(changeMat)
 	newNifti.mat  = changeMat;
-	newNifti.mat0 = changeMat;
+end
+
+if ~isempty(changeMat0)
+	newNifti.mat0 = changeMat0;
 end
 
 newNifti.dat.scl_slope = 1;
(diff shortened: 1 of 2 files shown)


External/SPMmodified/xASL/xASL_io_SaveNifti.m
```diff
@@ -226,7 +226,8 @@ function xASL_io_SaveNifti(pathOrigNifti, pathNewNifti, imNew, nBits, bGZip, cha
 %% ====================================================================================
 %% 3. Create new NIfTI
 if ~isempty(changeMat)
-	newNifti.mat = changeMat;
+	newNifti.mat  = changeMat;
```
HenkMutsaerts, line 239: mat0 should be the original orientation, as it comes from the scanner. This is useful for provenance checks in the orientation changes, e.g. for checking if there is a left-right flip, so theoretically we should never change mat0.

mat is whatever we want it to be, so can be changed

Modules/Module_Import/Import/xASL_imp_DCM2NII_Subject_SortASLVolumes.m
```diff
@@ -146,6 +146,21 @@
         elseif  ~isempty(strfind(scan_name,'M0'))
             [nii_files,ASLContext] = xASL_bids_MergeNifti(nii_files, 'M0', niiTable);
         end
+	else
+		% In case of a single file, we need to check issues with orientation
+		% Check for a special case of Siemens MOSAIC files without Phoenix field
+		if isfield(tmpJSON, 'Manufacturer') && strcmpi(tmpJSON.Manufacturer, 'Siemens') && isfield(tmpJSON, 'ImageType') && ~isempty(regexpi(tmpJSON.ImageType, 'MOSAIC')) && (~isfield(tmpJSON, 'PhoenixProtocol') || isempty(tmpJSON.PhoenixProtocol))
```
HenkMutsaerts: For readability, I would put the if cases on separate rows:

if isfield(tmpJSON, 'Manufacturer') && strcmpi(tmpJSON.Manufacturer, 'Siemens')
% this happens only for Siemens scans

if isfield(tmpJSON, 'ImageType') && ~isempty(regexpi(tmpJSON.ImageType, 'MOSAIC'))
% Siemens MOSAIC scans specifically

if ~isfield(tmpJSON, 'PhoenixProtocol') || isempty(tmpJSON.PhoenixProtocol)
% Siemens MOSAIC scans that have no Phoenix protocol stored in the DICOM header

Modules/Module_Import/Import/xASL_imp_DCM2NII_Subject_SortASLVolumes.m
```diff
@@ -146,6 +146,21 @@
...
+			niftiData = xASL_io_ReadNifti(nii_files{1});
+			% We check for a special case of image center in the Z-direction not being crossing 0
+			if (sum(niftiData.mat(:,3)) > 0) && (niftiData.mat(3,4) > 0)
+				% The image is mirrored in the Z-direction, we need to fix the image matrix and the orientation matrix
+				im = xASL_io_Nifti2Im(nii_files{1});
+				im = im(:,:,end:-1:1,:,:,:);
+				mat = niftiData.mat;
```
HenkMutsaerts: You can move this line up so we have:

if (sum(mat(:,3)) > 0) && (mat(3,4) > 0)

Modules/Module_Import/Import/xASL_imp_DCM2NII_Subject_SortASLVolumes.m
```diff
@@ -146,6 +146,21 @@
         elseif  ~isempty(strfind(scan_name,'M0'))
             [nii_files,ASLContext] = xASL_bids_MergeNifti(nii_files, 'M0', niiTable);
         end
+	else
+		% In case of a single file, we need to check issues with orientation
+		% Check for a special case of Siemens MOSAIC files without Phoenix field
+		if isfield(tmpJSON, 'Manufacturer') && strcmpi(tmpJSON.Manufacturer, 'Siemens') && isfield(tmpJSON, 'ImageType') && ~isempty(regexpi(tmpJSON.ImageType, 'MOSAIC')) && (~isfield(tmpJSON, 'PhoenixProtocol') || isempty(tmpJSON.PhoenixProtocol))
+			niftiData = xASL_io_ReadNifti(nii_files{1});
```
HenkMutsaerts, line 156: Should we have a warning if there are more than 1 nii_files?

Modules/Module_Import/Import/xASL_imp_DCM2NII_Subject_SortASLVolumes.m
```diff
@@ -146,6 +146,21 @@
...
+			% We check for a special case of image center in the Z-direction not being crossing 0
+			if (sum(niftiData.mat(:,3)) > 0) && (niftiData.mat(3,4) > 0)
+				% The image is mirrored in the Z-direction, we need to fix the image matrix and the orientation matrix
+				im = xASL_io_Nifti2Im(nii_files{1});
+				im = im(:,:,end:-1:1,:,:,:);
```
HenkMutsaerts: For some reason I remember that we should have 7 dimensions (not that we ever have this but this is SPM default I thought). And it's a lucky Christianity number ;)

HenkMutsaerts (changes requested):
Minor^2 comments

Modules/Module_Import/Import/xASL_imp_DCM2NII_Subject_SortASLVolumes.m
jan-petr: Now we have religion also in coding?! :) 

Sure - adding 7 :)

External/SPMmodified/xASL/xASL_io_SaveNifti.m
jan-petr: Good point. We keep mat0 as it is... Except in this case. dcm2niix really messes up import and creates a completely incorrect mat and mat0 that have nothing to do with the scanner orientation. So we need to change both in this case. So I'm adding mat0 as an extra parameter to xASL_io_SaveNifti and comment that this is exceptional.

Modules/Module_Import/Import/xASL_imp_DCM2NII_Subject_SortASLVolumes.m
jan-petr: Done

Modules/Module_Import/Import/xASL_imp_DCM2NII_Subject_SortASLVolumes.m
jan-petr: I've separated it to four lines. But I avoided doing several IFs cause this creates too much ifs, ends, and too many line indents...

Modules/Module_Import/Import/xASL_imp_DCM2NII_Subject_SortASLVolumes.m
jan-petr: Not needed - we're in an IF branch that only has 1 nii_files. So we only need {1} to move from cell to character array, but it cannot happen here that we have more than 1 nii_files at this line.

External/SPMmodified/xASL/xASL_io_SaveNifti.m
HenkMutsaerts: Annoying... I would be in favor not clouding our code for a very special occasion that will be soon fixed by dcm2nii (at least it should). Nobody else will change the mat0

External/SPMmodified/xASL/xASL_io_SaveNifti.m
```diff
@@ -226,7 +236,11 @@ function xASL_io_SaveNifti(pathOrigNifti, pathNewNifti, imNew, nBits, bGZip, cha
 %% ====================================================================================
 %% 3. Create new NIfTI
 if ~isempty(changeMat)
-	newNifti.mat = changeMat;
+	newNifti.mat  = changeMat;
+end
+
+if ~isempty(changeMat0)
+	newNifti.mat0 = changeMat0;
```
HenkMutsaerts, line 244: Please add a warning here that you change mat0

warning('Note that mat0 (the original NIfTI orientation) is changed upon request, this can jeopardize re-re-running ExploreASL!');

Because when we rerun ExploreASL, several orientations are defaulted back to mat0. So in this case, you need this warning. If you don't need it for import, then we need a bVerbose boolean to suppress such warnings.

But if this is a very exceptional case, then we can just always issue the warning.

HenkMutsaerts (changes requested):
Nice, small comment

External/SPMmodified/xASL/xASL_io_SaveNifti.m
jan-petr: Yes. Annoying, but dcm2niix has a lot of bugs and shortcomings that are not going to be fixed in the near future... So we have to accept that fact.

External/SPMmodified/xASL/xASL_io_SaveNifti.m
jan-petr: It's a very exceptional case and the user can be aware of that. So I'd indeed always issue this warning without a possibility to suppress. 

Warning added.

External/SPMmodified/xASL/xASL_io_SaveNifti.m
```diff
@@ -26,6 +26,8 @@ function xASL_io_SaveNifti(pathOrigNifti, pathNewNifti, imNew, nBits, bGZip, cha
 %   bOverwrite     Vector with 3 Booleans for overwriting the following pre-existing destination files:
 %                  1) NIfTI file, 2) JSON file, 3) mat-orientation motion file (for 4D NIfTIs) (OPTIONAL, DEFAULT = [1 1 1]; 
 %                  1 -> [1 1 1] and 0 -> [0 0 0] and creates a warning; [1 0] - error
+%   changeMat0     New orientation matrix mat0 4x4. Note that mat0 should always stay as it was originally from the scanner, 
+%                  but we can exceptionally change it if mat0 was completely wrong (OPTIONAL, DEFAULT same as previous)
```
HenkMutsaerts: instead of "exceptionally change it" (which is not grammatically correct I believe :) I would write:

"but in rare cases we change it if the orientation matrix was not correctly converted from the DICOM header."

HenkMutsaerts (approved):
minor comment left

External/SPMmodified/xASL/xASL_io_SaveNifti.m
jan-petr: Fixing

PR #1709, opened 2024-05-19
#1694 Check merging consistency in JSON

### LINKED ISSUE #1694: Check consistency when merging sequences
### Description

We now save extra tags in PWI4D.json - Manufacturer, MagneticFieldStrength, PulseSequenceType, MRAcquisitionType, ArterialSpinLabelingType, Background suppression. 
When merging two PWI4D.json, we have to check that these fields are consistent and then report an error if they are not.

### Release notes

Nothing. Report with #997

jan-petr:
### Linked issue

Close #1694

### DIFF
--- a/Functions/ImageProcessing/xASL_im_MergePWI4D.m
+++ b/Functions/ImageProcessing/xASL_im_MergePWI4D.m
@@ -97,7 +97,7 @@ for iSpace = 1:2
  			if isfield(jsonPWI4Dcurrent, listFields2Merge{iField2Merge}) && ~isempty(jsonPWI4Dcurrent.(listFields2Merge{iField2Merge}))
 				if iSession == 1 
 					% For the first session, just add the field
-					jsonPWI4DConcatenated.(listFields2Merge{iField2Merge}) = jsonPWI4Dcurrent.(listFields2Merge{iField2Merge});
+					jsonPWI4DConcatenated.(listFields2Merge{iField2Merge}) = jsonPWI4Dcurrent.(listFields2Merge{iField2Merge})(:)';
 				elseif ~isfield(jsonPWI4DConcatenated, listFields2Merge{iField2Merge})
 					% Session > 1, but the field was not added for session 1, we report a conflict
 					error('foo:bar',['Cannot merge sessions for subject ' x.SUBJECT '\n JSON field ' listFields2Merge{iField2Merge} ' is missing for session ' x.modules.asl.sessionsToMerge{1} ' but not for session ' x.modules.asl.sessionsToMerge{iSession}]);
@@ -105,10 +105,10 @@ for iSpace = 1:2
 					% The field seems to be consistently present/missing/empty on both sessions. The field is either check for consistency across sessions or concatenated
 					if bFields2Concatenate(iField2Merge)
 						% Concatenation
-						jsonPWI4DConcatenated.(listFields2Merge{iField2Merge}) = [jsonPWI4DConcatenated.(listFields2Merge{iField2Merge}); jsonPWI4Dcurrent.(listFields2Merge{iField2Merge})];
+						jsonPWI4DConcatenated.(listFields2Merge{iField2Merge}) = [jsonPWI4DConcatenated.(listFields2Merge{iField2Merge}), jsonPWI4Dcurrent.(listFields2Merge{iField2Merge})(:)'];
 					else
 						% Just checking consistency
-						if ~isequal(jsonPWI4DConcatenated.(listFields2Merge{iField2Merge}), jsonPWI4Dcurrent.(listFields2Merge{iField2Merge}))
+						if ~isequal(jsonPWI4DConcatenated.(listFields2Merge{iField2Merge}), jsonPWI4Dcurrent.(listFields2Merge{iField2Merge})(:)')
 							% Otherwise (field added for first session and we are now adding more) check for consistency and report an error if non-consistent
 							error('foo:bar',['Cannot merge sessions for subject ' x.SUBJECT '\n JSON field ' listFields2Merge{iField2Merge} ' differs between sessions ' x.modules.asl.sessionsToMerge{iSession} ' and ' x.modules.asl.sessionsToMerge{1}]);
 						end
--- a/Functions/ImageProcessing/xASL_im_MergePWI4D.m
+++ b/Functions/ImageProcessing/xASL_im_MergePWI4D.m
@@ -17,8 +17,7 @@ function [path_PWI4D, path_PWI4D_Pop] = xASL_im_MergePWI4D(x)
 %              merges them to a single NII with a predefined name. It returns this new name for both native and standard space and it also creates the 
 %              appropriate JSON sidecars.
 %              
-%              We merge NIfTIs and JSONs as are, without interpretation (the parameter interpretation, conversion BIDS->Legacy etc is done later)
-%              We only merge required (per BIDS/ExploreASL) JSON fields, so if any field is missing we issue an error
+%              We merge NIfTIs and JSONs as they are, without interpretation (the parameter interpretation, conversion BIDS->Legacy etc is done later)
 %
 % EXAMPLE: 
 %   [path_PWI4D, path_PWI4D_Pop] = xASL_im_MergePWI4D(x)
@@ -44,10 +43,8 @@ for iSession = 1:numel(x.modules.asl.sessionsToMerge)
 	mergedSessionsName = [mergedSessionsName '_' x.modules.asl.sessionsToMerge{iSession}];
 end
 
-
 for iSpace = 1:2
 	for iSession = 1:numel(x.modules.asl.sessionsToMerge)
-
         %% ---------------------------------------------------------------------
         %% 1. Load NIfTI and JSONs
 		% Here we get the path of the session, by replacing the session names 
@@ -85,7 +82,7 @@ for iSpace = 1:2
 				% Here we concatenate (cat) over the 4rd dimension (4), PWI (total concatenated image matrix) with the new current PWI
                 imPWI4DConcatenated = cat(4, imPWI4DConcatenated, imPWI4Dcurrent);
 			else
-				error(['Cannot concatenate sessions as session '  x.modules.asl.sessionsToMerge{iSession} ' has a different matrix size']);
+				error(['Cannot concatenate sessions: session '  x.modules.asl.sessionsToMerge{iSession} ' has a different matrix size']);
 			end
 		end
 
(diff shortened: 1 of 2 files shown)


HenkMutsaerts (changes requested):
This is suboptimal. I would generate as follows, though I understand that we might want to do this later.

*** Merging
1. We always merge [session1 ... session 2 ... sessionN] irregardless of it being a scalar or vector, if needed we copy for each volume
2. Only if fields are the same for all volumes, we "collapse" the vector back to a scalar
3. We do this for all fields in the original JSONs

*** Quantification
1. We issue an error for each incompatible quantification (which will decrease over time, as we will be more and more able to correct)

So this would be:
A) we merge all parameters
B) everything is concatenated as vector unless it can be summarized by a scalar

Whereas now:
A) Some things are merged, others are not (== assumed to be the same...)
B) Some things are concatenated, others are not
C) Some things have to be there, others you say that "if non-existing, just create" ...
    (whereas e.g. MagneticFieldStrength is not necessarily less important than 
    e.g. LabelingDuration)

So I would vouch for more standardization if possible.

E.g., if session1 has no MagneticFieldStrength but session2 has it, you just use session2, that's very messy. You cannot assume them to be the same (it makes sense, but you never know what the user might do, the user might want to work with between-scanner comparisons or longitudinal data).

BTW: after the license change we fortunately can stop changing the copyright years everytime ;)

jan-petr:
We have actually discussed all these previously and came to the conclusion that vectorizing everything is too much of a trouble and decided to vectorize only those allowed by BIDS and do warnings for the rest. Reasons being:

1. Some fields cannot be vectorized according to BIDS
2. Our quantification model is fully unable to work with merged data from two vendors, PCASL+PASL etc.
3. It would be very difficult to implement and there is no literature or models for that.

Regarding the issue with combining missing+non-missing values. I agree - this should create a warning as well! My reasoning was that this will not really happen as Manufacturer and FieldStrength are required fields and won't be inconsistent in DICOMs for two sessions. But I agree to change the code.

> *** Merging
> 
> 1. We always merge [session1 ... session 2 ... sessionN] irregardless of it being a scalar or vector, if needed we copy for each volume
> 2. Only if fields are the same for all volumes, we "collapse" the vector back to a scalar
> 3. We do this for all fields in the original JSONs

> *** Quantification
> 
> 1. We issue an error for each incompatible quantification (which will decrease over time, as we will be more and more able to correct)
> 
> So this would be: A) we merge all parameters B) everything is concatenated as vector unless it can be summarized by a scalar
> 
> Whereas now: A) Some things are merged, others are not (== assumed to be the same...) B) Some things are concatenated, others are not C) Some things have to be there, others you say that "if non-existing, just create" ... (whereas e.g. MagneticFieldStrength is not necessarily less important than e.g. LabelingDuration)
> 
> So I would vouch for more standardization if possible.
> 
> E.g., if session1 has no MagneticFieldStrength but session2 has it, you just use session2, that's very messy. You cannot assume them to be the same (it makes sense, but you never know what the user might do, the user might want to work with between-scanner comparisons or longitudinal data).
> 
> BTW: after the license change we fortunately can stop changing the copyright years everytime ;)

Functions/ImageProcessing/xASL_im_MergePWI4D.m
```diff
@@ -121,15 +121,20 @@
...
-					if ~isequal(jsonPWI4DConcatenated.(listFields2Merge{iField2Merge}), jsonPWI4Dcurrent.(listFields2Merge{iField2Merge}))
-						error(['Cannot merge JSON field ' listFields2Merge{iField2Merge} ' as it differs between sessions']);
-					end
+				elseif ~isfield(jsonPWI4DConcatenated, listFields2Merge{iField2Merge})
+					% Session > 1, but the field was not added for session 1
+					% We report a conflict
+					error(['Cannot merge JSON field ' listFields2Merge{iField2Merge} ' as it was missing for the first session']);
+				elseif ~isequal(jsonPWI4DConcatenated.(listFields2Merge{iField2Merge}), jsonPWI4Dcurrent.(listFields2Merge{iField2Merge}))
+					% Otherwise (field added for first session and we are now adding more) check for consistency and report an error if non-consistent
+					error(['Cannot merge JSON field ' listFields2Merge{iField2Merge} ' as it differs between sessions']);
```
HenkMutsaerts: These warnings should be:
'cannot merge ' path ' with ' path ' because field ' jsonField ' differs between sessions.
(or replace path by session)

because in the case of ~bConcatenate, parameters need to be equal and you don't merge parameters, you merge the sessions.

HenkMutsaerts (changes requested):
Why don't you do the same for all parameters and have a boolean bConcatenate

I mean, for all parameters you need to do the same checks
1. Does the field exist
2. Is the field not empty
3. Does the field have the expected number of elements
Etc.
Then:
if bConcatenate
   out = [in1 in2];
else
   if ~isequal(in1, in2)
         error('Can not merge ....');
   end
end

The only difference between A vectors and B scalars is that we A concatenate and B check isequal.

Functions/ImageProcessing/xASL_im_MergePWI4D.m
jan-petr: Done.

jan-petr:
Yes, all parameters in a single loop, with bConcatenate parameter.

> Why don't you do the same for all parameters and have a boolean bConcatenate
> 
> I mean, for all parameters you need to do the same checks
> 
> 1. Does the field exist
> 2. Is the field not empty
> 3. Does the field have the expected number of elements
>    Etc.
>    Then:
>    if bConcatenate
>    out = [in1 in2];
>    else
>    if ~isequal(in1, in2)
>    error('Can not merge ....');
>    end
>    end
> 
> The only difference between A vectors and B scalars is that we A concatenate and B check isequal.

Functions/ImageProcessing/xASL_im_MergePWI4D.m
```diff
@@ -79,39 +79,48 @@
...
+		% Go through all the fields to merge
+		for iField2Merge = 1:length(listFields2Merge)
+			% The field exists and is not empty, we therefore need to check if we can merge it across sessions
+ 			if isfield(jsonPWI4Dcurrent, listFields2Merge{iField2Merge}) && ~isempty(jsonPWI4Dcurrent.(listFields2Merge{iField2Merge}))
+				if iSession == 1 
+					% For the first session, just add the field
+					jsonPWI4DConcatenated.(listFields2Merge{iField2Merge}) = jsonPWI4Dcurrent.(listFields2Merge{iField2Merge});
+				elseif ~isfield(jsonPWI4DConcatenated, listFields2Merge{iField2Merge})
+					% Session > 1, but the field was not added for session 1, we report a conflict
+					error(['Cannot merge sessions as the JSON field ' listFields2Merge{iField2Merge} ' is missing for session ' x.modules.asl.sessionsToMerge{1} ' but not for session ' x.modules.asl.sessionsToMerge{iSession}]);
```
HenkMutsaerts: Is this correct? Here, we only know that it was missing for iSession 1 right?
I would keep it simple and only throw the error that it was missing for session 1.
And can we add the path to the error?

Functions/ImageProcessing/xASL_im_MergePWI4D.m
```diff
@@ -79,39 +79,48 @@
...
+				else
+					% The field seems to be consistently present/missing/empty on both sessions. The field is either check for consistency across sessions or concatenated
+					if bFields2Concatenate(iField2Merge)
+						% Concatenation
+						jsonPWI4DConcatenated.(listFields2Merge{iField2Merge}) = [jsonPWI4DConcatenated.(listFields2Merge{iField2Merge}); jsonPWI4Dcurrent.(listFields2Merge{iField2Merge})];
```
HenkMutsaerts: This is incorrect right?
Session1 = [Volume1 Volume2]

Session2 = [Volume3 Volume4]

Concatenated = [Session1, Session2]

NOT

Concatenated = [Session1; Session2]

Functions/ImageProcessing/xASL_im_MergePWI4D.m
```diff
@@ -79,39 +79,48 @@
...
+					else
+						% Just checking consistency
+						if ~isequal(jsonPWI4DConcatenated.(listFields2Merge{iField2Merge}), jsonPWI4Dcurrent.(listFields2Merge{iField2Merge}))
+							% Otherwise (field added for first session and we are now adding more) check for consistency and report an error if non-consistent
+							error(['Cannot merge sessions as the JSON field ' listFields2Merge{iField2Merge} ' differs between sessions ' x.modules.asl.sessionsToMerge{iSession} ' and ' x.modules.asl.sessionsToMerge{1}]);
```
HenkMutsaerts: We need to issue the path here, and the session that cannot be merged.

Functions/ImageProcessing/xASL_im_MergePWI4D.m
```diff
@@ -79,39 +79,48 @@
...
+						end
+					end
+				end
+			elseif isfield(jsonPWI4DConcatenated, listFields2Merge{iField2Merge}) && ~isempty(jsonPWI4DConcatenated.(listFields2Merge{iField2Merge}))
+				% Field is missing for the current session, but it was added for the first session
+				error(['Cannot merge sessions as the JSON field ' listFields2Merge{iField2Merge} ' is missing for session ' x.modules.asl.sessionsToMerge{iSession} ' but not for session ' x.modules.asl.sessionsToMerge{1}]);
```
HenkMutsaerts: Again, we don't care for which session the field is NOT missing, but only which session the field IS missing. And the path! (Or subject ID)

HenkMutsaerts (changes requested):
minor minor

Functions/ImageProcessing/xASL_im_MergePWI4D.m
jan-petr: Correct. That's how things come out of JSON. I've checked that. I had [Session1, Session2] initially, but that wouldn't work.

Functions/ImageProcessing/xASL_im_MergePWI4D.m
jan-petr: Yes - this here means that we are missing the parameter for session 1. But that by itself is not a problem. We are checking here consistency across session. So if it is missing for session 1 and all other sessions as well, we won't report an error here. We only report an error in case of non-consistent values. We can mention only session 1, but that wouldn't be complete information.

Yes - adding subject name. In combination with session, the user will have the complete information.

Functions/ImageProcessing/xASL_im_MergePWI4D.m
jan-petr: Subject added - that's better than a full path as that's too long information and the user knows the studypath.

Sessions are already in.

Functions/ImageProcessing/xASL_im_MergePWI4D.m
jan-petr: SubjectID added.

We mention both missing and non-missing sessions. As said above -> we are checking consistency, so we need to report both.

Functions/ImageProcessing/xASL_im_MergePWI4D.m
HenkMutsaerts: Are you sure? Internally, throughout the ASL module, we use e.g.,

x.module.asl.etc.parameter.PLD = [500 1000 1500 2000];

So if it comes from JSON as 

x.module.asl.etc.parameter.PLD = 
[500
1000
1500
2000];

then it should be transposed right, and we keep everything internally the same (always horizontal, never vertical)

I mean, most important is that we are consistent throughout ExploreASL. And because we always used horizontal rather than vertical for these vectors (and BIDS as well within the JSON file), we should read a JSON to internal memory horizontal. Right?

jan-petr: See above. I'm sure. 

Internally, we don't have a convention, we deal with vectors as vectors and always make sure that the code can work with any orientation to avoid issues with vectors are entered by a different orientation by the user.

I can make the code here also work with any orientation if you want. But I wouldn't impose a strict orientation as this is not needed.

HenkMutsaerts (approved):

PR #1719, opened 2024-06-14
Feature #1552 asl output change

### LINKED ISSUE #1552: Update PDF report for new x.output format
### Description

PDF report has been revamped. But we have now also changed the way the QC is created. After merging #1225 , 
Basically, we don't now have info inside x.output.ASL, but there are subfields based on sessions.
Same for structural.
And we have to be careful that all from x,output.ASL is now deleted what is not inserted in these subfields - so that could apply to sub fields inserted by the PDF report.

Basically, we are synchronizing edits made in #1225 and #1514. Both were already merged. #1551 is also already merged.

### Tasks

- [x] Change the way x.output and x.output_im is used in xASL_qc_ParsePdfConfig.m
- [x] Manage default number of floating points (digits after floating point)
- [x] Change the way x.output and x.output_im is used in xASL_qc_ParsePdfConfig.m - line 90, we create the PDF - but do we create it for each ASL session? Or once per subject? Each has its disadvantages as we either have the structural module multiple times or we need to detect the last ASL session and call it for all previous ASL sessions. But maybe a re-creation for all ASL sessions will work with some extra overhead.
- [x] xASL_qc_ParsePdfConfig.m/xASL_sub_createModuleContent - For ASL module, we use an extra subfield session. x.Output.ASL.ASL_1 and x.Output_im.ASL.ASL_1.
- [x] xASL_qc_ParsePdfConfig.m - take the current ASL session into account. Add the session name to the PDF filename.
- [x] Update Docs to explain the changes

HENK: 31-12-2024
- [x] bugfix: for 1:size() -> for 1:length()
- [x] convert pageIdentifier to identifier in /WorkInProgress/QualityControl/PDFGenerator.md
- [x] recreate the config file for each subject, because each subject can have different modules
- [x] move the config file location to the subject rather than derivatives root folder
- [x] do the crawling for creating the config file not module-specific, but based on having number of sessions (dwi & fmri can also have multiple sessions, but these modules are not really used)

### How to test

- [x] Unit Test
- [x] Henk: tested with 1. Structural + ASL_1 modules (TestDataSet) & 2. ASL_1 + ASL_2 FlavorDatabase/Siemens_PCASL_2DEPI_E11_1)
- [x] Henk: Checked the PDF

### Release notes

Generate PDF reports also for multiple ASL runs

maartenhammer:
closes #1552

### DIFF
--- a/Functions/QualityControl/exampleConfigReportPDF.json
+++ b/Functions/QualityControl/exampleConfigReportPDF.json
@@ -7,7 +7,8 @@
         "fontWeight": "normal",
         "color": "k",
         "axesVisible": "off",
-        "HorizontalAlignment": "left"
+        "HorizontalAlignment": "left",
+        "numberFormat": "%.2f"
     },
     "pages": [
         {
@@ -137,7 +138,11 @@
                     "parameter": "T1w_GM_vol_mL",
                     "module": "Structural",
                     "range": "1000-1200",
-                    "unit": "mL"
+                    "unit": "mL",
+                    "textSettings": {
+                        "numberFormat": "%.0f"
+                    }
+
                 },
                 {
                     "category": "content",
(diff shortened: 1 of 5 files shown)


jan-petr (approved):
OK

PR #1725, opened 2024-06-28
Bug #1723 population multi session

### LINKED ISSUE #1723: Population stats multisession error
### Description

The entire reporting of Stats for multi-session setup is full of bugs. Several things are wrong:

1. In case we have 3 subjects and 2 ASLs. The file with GM volumes then contains 6 lines. But in the case that processing of ASL_1 goes wrong and we only have CBF for ASL_2, then the CBF-stats file contains only 3 lines (ASL_2 for all subjects). But when it adds GM volumes inside, it takes the first three (sub1+ASL1, sub1+ASL2, sub2+ASL1). 
2. The second error is when we have CBF files that are wrong (e.g. have 4th dimension 2 and not 1). That's of course an error, but the Population module should be able to handle that and print "error" on that specific CBF line. Issues 1 and 2 are relevant for TBI data from Ulm.
3. Once we fix that above, we have to check cases where we have for Sub1 CBF1 and 2, but for Sub2 only CBF 1.
4. Once we fix that above, we have to check what happens if we have Sub1 ASL1 and ASL2, but for sub2 only ASL1 (similar to above, but above affects the CBF part and this below affects also the GM volume part). At the same time, we can check scenario that for sub1, we have ASL2 and 3 and for sub 2 only ASL3.

### Tasks

- [x] Fix issue with ASL_1 not having CBF quantified. This is because in printing, we use two different SubjSession indexes. One that goes through all existing ASL sessions (for GM volume part) and one that goes through CBF images (for CBF values). Those two indices differ, but the code was using only the second index for both cases. 9839f7c
- [x] Fix issue with wrong CBF images. This actually crashes in the template creation. So I fixed that template creation is skipped for wrong CBF images and warning reported. And for calculation of CBF mean, we report warning as well and assign N/A to values. 7fb3b437
- [x] Fix issue with missing CBF images. This worked out of the box.
- [x] Fix issue with missing and skipped ASL sessions. This worked out of the box.

### Release notes

Fix population stats printing for missing sessions and CBF images.

jan-petr:
### Linked issue

Close #1723

### DIFF
--- a/Functions/ImageProcessing/xASL_im_IM2Column.m
+++ b/Functions/ImageProcessing/xASL_im_IM2Column.m
@@ -39,7 +39,7 @@ if length(size(ImageIn))>3
 		ImageIn = shiftdim(ImageIn,length(size(ImageIn))-3);
 	else
 		% Issue a warning and output all NaNs
-		warning('Cannot accept 4D image when ApplyShiftDim is FALSE');
+		warning('Cannot accept 4D image when ApplyShiftDim is false, check if your data were correctly processed');
 		ImageIn = nan(size(ImageIn));
 	end
 else
--- a/Functions/ImageProcessing/xASL_im_IM2Column.m
+++ b/Functions/ImageProcessing/xASL_im_IM2Column.m
@@ -25,7 +25,7 @@ function [ColumnOut] = xASL_im_IM2Column(ImageIn, BrainMask, ApplyShiftDim)
 %
 % EXAMPLE: ColumnOut = xASL_im_IM2Column(ImageIn, BrainMask);
 % __________________________________
-% Copyright (C) 2015-2021 ExploreASL
+% Copyright (C) 2015-2024 ExploreASL
 
 
 %%  ------------------------------------------------------------
@@ -34,8 +34,14 @@ if nargin<3 || isempty(ApplyShiftDim)
     ApplyShiftDim = true;
 end
 
-if ApplyShiftDim && length(size(ImageIn))>3
-    ImageIn = shiftdim(ImageIn,length(size(ImageIn))-3);
+if length(size(ImageIn))>3
+	if ApplyShiftDim
+		ImageIn = shiftdim(ImageIn,length(size(ImageIn))-3);
+	else
+		% Issue a warning and output all NaNs
+		warning('Cannot accept 4D image when ApplyShiftDim is FALSE');
+		ImageIn = nan(size(ImageIn));
+	end
 else
     ApplyShiftDim = false; % to avoid re-shifting below
 end
--- a/Functions/Statistics/xASL_stat_PrintStats.m
+++ b/Functions/Statistics/xASL_stat_PrintStats.m
@@ -46,7 +46,7 @@ function [x] = xASL_stat_PrintStats(x, bFollowSubjectSessions)
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
 % EXAMPLE: x = xASL_stat_PrintStats(x);
 % __________________________________
-% Copyright 2015-2020 ExploreASL
+% Copyright 2015-2024 ExploreASL
 
 
 %% Admin
@@ -132,13 +132,15 @@ else
 
     printedSessionN = 0;
     
-    % Initialize empty table
+    % Initialize empty table with dimensions based on what was created in xASL_stat_GetROIStatistics
     numElements = size(x.modules.population.(thisFile),2);
     numSubjectsSessions = length(x.S.SubjectSessionID);
     for iSubjSess=1:numSubjectsSessions
+		% Fill the table with NaNs
         x.modules.population.(thisFile)(2+iSubjSess,:) = repmat({nan(1, 1)}, 1, numElements);
     end
     
+	% Go through all sessions that have CBF detected in xASL_stat_GetROIStatistics
     for iSubjSess=1:numSubjectsSessions
         % x.S.SubjectSessionID == subject/session IDs created in xASL_stat_GetROIStatistics
 
@@ -153,7 +155,6 @@ else
             
             if isempty(startSessionIndex) || isempty(endSessionIndex)
 				% If the sessionID was not found, then report a warning and use the entire expression for subjectID
-				SubjectID = x.S.SubjectSessionID{iSubjSess}(startSubjectIndex:endSubjectIndex);
                 fprintf(2, ['Could not find session for ' x.S.SubjectSessionID{iSubjSess} '\n']);
 			else
 				% If session ID was identified, we have to double-check that session ID is not part of subject ID and exclude if necessary
@@ -175,7 +176,7 @@ else
                 x.dataset.currentSubjectID = SubjectID;
                 x.dataset.currentSessionID = SessionID;
                 
-                % Fill stat cell array
+                % Fill stat cell array - subject name and session
                 x.modules.population.(thisFile) = xASL_stat_PrintStats_AddSubjectSessionStatCellArray(x,x.modules.population.(thisFile),iSubjSess);
                 
 
@@ -191,20 +192,19 @@ else
                         SessionColumn = find(strcmpi(x.S.SetsName, 'session'));
                         if isempty(SessionColumn) || length(SessionColumn)>1
                             fprintf(2, 'Could not find session data\n');
-                        else
-                            SessionN = iSubjSess;
-                            if isempty(SessionN) || ~isnumeric(SessionN)
+						else
+                            if isempty(iSubjSess) || ~isnumeric(iSubjSess)
                                 fprintf(2, ['Something wrong with session ' SessionID '\n']);
                             elseif SessionIndex>x.dataset.nSessions
-                                if ~max(printedSessionN==SessionN)
+                                if ~max(printedSessionN==iSubjSess)
                                     fprintf(2, 'Could not find values for other covariates\n');
                                 end
-                                printedSessionN = [printedSessionN SessionN];
+                                printedSessionN = [printedSessionN iSubjSess];
                             else
                                 
                                 %% Print the covariates and data
-                                iSubjectSession_SetsID = x.dataset.nSessions*(SubjectIndex-1)+SessionN;
-                                iSubjectSession_DAT = iSubjSess;
+                                iSubjectSession_SetsID = x.dataset.nSessions*(SubjectIndex-1)+SessionIndex; % This is for the total set and is based on the real names
+                                iSubjectSession_DAT = iSubjSess; % This is for the DATA and is based on xASL_wrp_GetROIStatistics
                                 bPrintSessions = false;
                                 
                                 % Write it to the cell array instead
@@ -349,6 +349,10 @@ end
 %% Fill the stat cell array with all subjects and sessions xASL_tsvWrite later on
 function statCell = xASL_stat_PrintStats_FillStatCellArray(x,statCell, rowNum, iSubjectSession_SetsID, iSubjectSession_DAT, bPrintSessions)
 
+    % statCell - we write into this structure
+	% rowNum   - number of the row in this structure to write to
+	% iSubjectSession_SetsID - index in the general dataset of all ASL scans
+	% iSubjectSession_DAT - index in the specific extracted dataset by xASL_wrp_GetROIStatistics - these two indexes can differ
     % Skip the first two rows and columns (Labels & Legend, Subjects & Sessions)
     rowNum = rowNum+2;
     iCell = 3;
@@ -359,19 +363,22 @@ function statCell = xASL_stat_PrintStats_FillStatCellArray(x,statCell, rowNum, i
     if bPrintSessions
         printMatrix = x.S.SetsID;
         optionsMatrix = x.S.SetsOptions;
-        sampleMatrix = x.S.Sets1_2Sample;
+        %sampleMatrix = x.S.Sets1_2Sample;
     else
         SessionColumn = find(strcmpi(x.S.SetsName, 'session'));
-        vector2Print = [1:size(x.S.SetsName,2)];
+        vector2Print = 1:size(x.S.SetsName,2);
         vector2Print = vector2Print(vector2Print~=SessionColumn);
         
         printMatrix = x.S.SetsID(:, vector2Print);
         optionsMatrix = x.S.SetsOptions(:, vector2Print);
-        sampleMatrix = x.S.Sets1_2Sample(:, vector2Print);
+        %sampleMatrix = x.S.Sets1_2Sample(:, vector2Print);
     end
     
     for iPrint=1:size(printMatrix,2)
-        String2Print = printMatrix(iSubjectSession_DAT, iPrint); % Previously we used iSubjectSession_SetsID
+		% Previously we incorrectly used iSubjectSession_DAT, but SetsID and DAT can differ
+		% Previously, this was changes to DAT, because there was a bug in SetsID calculation
+		% Now we are back at the correct SetsID and but we have also corrected calculation of this index
+        String2Print = printMatrix(iSubjectSession_SetsID, iPrint);
 
         if length(optionsMatrix{iPrint})>1 % we need options
             if length(optionsMatrix{iPrint}) >= length(unique(printMatrix(:,iPrint)))-2 % allow for zeros & NaNs
@@ -390,6 +397,4 @@ function statCell = xASL_stat_PrintStats_FillStatCellArray(x,statCell, rowNum, i
         statCell{rowNum,iCell} = xASL_num2str(x.S.DAT(iSubjectSession_DAT, iPrint));
         iCell = iCell+1;
     end
-
-
 end
--- a/Functions/Statistics/xASL_stat_PrintStats.m
+++ b/Functions/Statistics/xASL_stat_PrintStats.m
@@ -55,6 +55,25 @@ if nargin<2 || isempty(bFollowSubjectSessions)
 end
 
 
+% Manage missing type of statistics
+% x.S.Sets1_2Sample indicates the type of data
+% x.S.Sets1_2Sample==1 means paired data, e.g. sessions
+% x.S.Sets1_2Sample==2 means a two-sample data, e.g. cohorts
+% x.S.Sets1_2Sample==3 means continues data, e.g. age
+
+if ~isfield(x, 'S')
+    error('Missing field x.S, something went wrong');
+end
+nSets = size(x.S.SetsID, 2);
+if ~isfield(x.S, 'Sets1_2Sample')
+    x.S.Sets1_2Sample(1:nSets) = 3; % we default to continuous data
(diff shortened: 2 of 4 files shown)


Functions/Statistics/xASL_stat_PrintStats.m
```diff
@@ -132,13 +132,15 @@
 
     printedSessionN = 0;
     
-    % Initialize empty table
+    % Initialize empty table with dimensions based on what was created in xASL_stat_GetROIStatistics
```
HenkMutsaerts: `xASL_stat_PrintStats` was a general function that would work with any input, not only from xASL_stat_GetROIStatistics but also 

`040_GetDICOMStatistics`
`050_GetVolumeStatistics`
`060_GetMotionStatistics`
`065_GetRegistrationStatistics`

so it allows you to use the whole stats stuff I build around it (automatically performing t-tests if e.g., you included sex or cohort as "covariates" in x.S.Sets*.

Now we don't do this anymore and simplified this, because it just got too complicated to also do all these stats (I just checked).

But one feature that I had (deactivated by default) if you can remember, was the voxel-wise GLM. This needs this function to work generically, because if you do a GLM, it generates significant clusters, and those clusters statistics (e.g. peak-t location, size, etc) are then printed with xASL_stat_PrintStats.

Because GLM is a feature that is requested by quite some of our students, I would not abandon it (though not sure when I will have time to fix its dependencies :)

I changed the comment so this is fine, but just a heads-up in general :)

Functions/Statistics/xASL_stat_PrintStats.m
```diff
@@ -132,13 +132,15 @@
...
     numElements = size(x.modules.population.(thisFile),2);
     numSubjectsSessions = length(x.S.SubjectSessionID);
     for iSubjSess=1:numSubjectsSessions
+		% Fill the table with NaNs
         x.modules.population.(thisFile)(2+iSubjSess,:) = repmat({nan(1, 1)}, 1, numElements);
     end
     
+	% Go through all sessions that have CBF detected in xASL_stat_GetROIStatistics
```
HenkMutsaerts: No, xASL_stat_GetROIStatstics is designed to work on any input file. Can be CBF, but can also be M0, or rc1T1, or whatever. It defaults to a single session. So here we don't go through all the "sessions that have CBF detected..." Rather, we just go through any subject/session combination.

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
```diff
@@ -277,7 +277,9 @@
             TotalRows = x.dataset.nSubjects;
 		else
 			x.S.SubjectSessionID{SubjSess,1} = [x.SUBJECTS{iSubject} '_' listSessions{iSess}];
-            TotalRows = x.dataset.nSubjects * x.dataset.nSessions;
+			% The number of sessions have to be consistent with the detected one. If we have 2 sessions, but only session 2 has quantified CBF,
```
HenkMutsaerts: I'm not too sure about this. We rather have "dirty" stats files that show the error rather than cleaning it such that it looks better but the user doesn't note the error.

E.g., if a study has only rawdata run-2, then this should be converted to only having ASL_2 and x.dataset.nSessions should be 1, where x.dataset.SESSIONS = {'ASL_2'};

E.g., if a study has rawdata run-1 and run-2, then this should be converted to ASL_1 and ASL_2, where it would be useful for the user to see for all participants that ASL_1 is n/a.

By the way, I rephrased this because at this point it is not about "printing lines" but it is about which we will attempt to perform stats on, and you rather have a 1000 warnings on missing sessions (which we could perhaps catch and only print once to the screen) rather than catching it here and not processing it.

The alternative is to do this with nSessions (what you propose) but issue a warning that nSessions~=x.dataset.nSessions. We could already issue that warning at the top where we obtain nSessions.

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
```diff
@@ -488,7 +490,11 @@
 			FilePath = fullfile(x.dir.SESSIONDIR, [x.S.InputDataStrNative '.nii']);
 			if xASL_exist(FilePath, 'file')
 				Data3D = xASL_io_Nifti2Im(FilePath);
-				DataIm = xASL_im_IM2Column(Data3D,x.S.masks.WBmask);
+				if size(Data3D,4) > 1
```
HenkMutsaerts: I have a problem here. Because I am bit by bit transforming to the feature of allowing statistics for any number of image volumes. I need this feature soon, and the population module should be oblivious/independent about what went wrong in earlier modules (issues would be caught there/warnings/errors produced, and that the population module then behaves weirdly is fine).

So we have the following use cases:

case 1: the same results for non-4D files, what we had
case 2: 4D results which I need soon (e.g. this summer) when I'm working on trying to do a temporary SD spatial CoV or vascular distality analysis with the French data
case 3: give weird results for studies where the processing went wrong, i.e. it should have created single volume files in the population folder but it created 4D files. This seems to be your case, and should provide errors or warnings elsewhere.

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
```diff
@@ -504,7 +510,11 @@
 			FilePath = fullfile(x.D.PopDir, [x.S.InputDataStr '_' x.S.SubjectSessionID{SubjSess,1} '.nii']);
 			if xASL_exist(FilePath, 'file')
 				Data3D = xASL_io_Nifti2Im(FilePath,[121 145 121]);
-				DataIm = xASL_im_IM2Column(Data3D,x.S.masks.WBmask);
+				if size(Data3D,4) > 1
```
HenkMutsaerts: Same

Functions/Statistics/xASL_stat_PrintStats.m
```diff
@@ -153,7 +155,6 @@
             
             if isempty(startSessionIndex) || isempty(endSessionIndex)
 				% If the sessionID was not found, then report a warning and use the entire expression for subjectID
-				SubjectID = x.S.SubjectSessionID{iSubjSess}(startSubjectIndex:endSubjectIndex);
```
HenkMutsaerts, line 156: See the comment the line before. If no session ID was found, e.g. for printing rc1T1 statistics, then we need the entire "expression" for SubjectID?

Did you consider this use case? I'm not sure, just asking.

Functions/Statistics/xASL_stat_PrintStats.m
```diff
@@ -46,7 +46,7 @@
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
 % EXAMPLE: x = xASL_stat_PrintStats(x);
 % __________________________________
-% Copyright 2015-2020 ExploreASL
+% Copyright 2015-2024 ExploreASL
```
HenkMutsaerts, line 49: Please change the header accordingly, for what you propose to change (including variables used inside this function)

Functions/Statistics/xASL_stat_PrintStats.m
```diff
@@ -359,19 +363,22 @@
     if bPrintSessions
         printMatrix = x.S.SetsID;
         optionsMatrix = x.S.SetsOptions;
-        sampleMatrix = x.S.Sets1_2Sample;
+        %sampleMatrix = x.S.Sets1_2Sample;
```
HenkMutsaerts: This breaks functionality (even though we don't use this often), I fixed this.

Functions/Statistics/xASL_stat_PrintStats.m
```diff
@@ -359,19 +363,22 @@
...
-        sampleMatrix = x.S.Sets1_2Sample(:, vector2Print);
+        %sampleMatrix = x.S.Sets1_2Sample(:, vector2Print);
     end
     
     for iPrint=1:size(printMatrix,2)
-        String2Print = printMatrix(iSubjectSession_DAT, iPrint); % Previously we used iSubjectSession_SetsID
+		% Previously we incorrectly used iSubjectSession_DAT, but SetsID and DAT can differ
+		% Previously, this was changes to DAT, because there was a bug in SetsID calculation
+		% Now we are back at the correct SetsID and but we have also corrected calculation of this index
+        String2Print = printMatrix(iSubjectSession_SetsID, iPrint);
```
HenkMutsaerts, line 407: Nice!

HenkMutsaerts (changes requested):
Few comments, and for a few I already improved your changes. Nice that you spotted these bugs (though some of them may be features ;)

Functions/Statistics/xASL_stat_PrintStats.m
jan-petr: Thanks. Yes, the comment was a bit limited.

Functions/Statistics/xASL_stat_PrintStats.m
```diff
@@ -362,7 +389,7 @@
         sampleMatrix = x.S.Sets1_2Sample;
```
jan-petr, line 392: Are you sure you don't want to comment out the line that is currently not used? For me, it's a bit better to keep it commented until we start really using it.

But if you say that it would be too difficult to comment it now and then uncomment it later, then we can keep it as it is.

jan-petr (commented):
I agree with all your edits. And I have read the corresponding comments - and when I agreed fully, I simply resolved them.

I only kept open the comments where you asked for an action or where I needed to explain something...

Functions/Statistics/xASL_stat_PrintStats.m
jan-petr: Understood. Agree.

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
jan-petr: If we have ASL_2 only and then obviously only CBF_2, then everything is fine - just a single session is detected and written both for volumes and CBF.

The trouble is when we have ASL_1 and ASL_2. But quantification of ASL_1 fails, so we only have CBF_2. Then the two indexing options do not match and it results in a messed up stats table. Note that the issue is not with run_1 -> ASL_1 conversion. But with quantification of Legacy ASL_1 to CBF_1. 

As discussed -> I only programmed this as a second line safety check. This error should be captured elsewhere and reported. But we can't count that the users will do everything right and check. So if a user ignores errors, then we still have to make sure that errors are not propagated to the final table.

I'm just clarifying the motivation. I agree with the edit you made to the comment and about the extra check of nSession and x.dataset.nSessions -> these are useful 2nd/3rd line checks ;) 

You can resolve after reading.

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
jan-petr: Yesyes, I know about this and I'm fine with your future changes. The trouble is that currently, CBF in MNI should be 3D. And if it's 4D, then xASL_im_IM2Column doesn't mask it correctly considering the 4th dimension, but rather selects something weird and the mean CBF after this masking is then completely off.

So before we make the 4D feature and before we make sure that xASL_im_IM2Columne works fine with 4D data, we have to report an error here.

To sum up - case 1 and case 2 - I agree but the xASL_im_IM2Column and all remaining functions should be first able to handle 4D data. They are not able to do this now.

Case 3 -> As explained above. Yes, error should be checked by the user. But a real experience says that users ignore this simply because they are not experienced. So the error here is to help them to realize something went wrong. 

I'd thus keep it as it is now and change in the future -> if you agree about temporary addressing this before the new processing is out.

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
jan-petr: Same ;)

Functions/Statistics/xASL_stat_PrintStats.m
jan-petr: I considered it, bu I removed this line because the variable created here (SubjectID) wasn't ever used anywhere :)

Functions/Statistics/xASL_stat_PrintStats.m
jan-petr: Are you sure you don't want to comment out the line that is currently not used? For me, it's a bit better to keep it commented until we start really using it.

But if you say that it would be too difficult to comment it now and then uncomment it later, then we can keep it as it is.

Functions/Statistics/xASL_stat_PrintStats.m
jan-petr: Sorry - I resolved this and rather copied this as an answer to your comment on this topic.

Functions/Statistics/xASL_stat_PrintStats.m
```diff
@@ -191,19 +211,22 @@
...
+                                printedSessionN = [printedSessionN iSubjSess];
                             else
                                 
                                 %% Print the covariates and data
-                                iSubjectSession_SetsID = x.dataset.nSessions*(SubjectIndex-1)+SessionN;
+                                % SetsID -> contains throughout ExploreASL the stats read from participants_id, e.g., subject, session, covariates, names, etc
+                                % iSubjectSession_SetsID here defines the names that will be printed
+                                % _DAT & _DATA is about the data (mostly numerical), the stats coming out of e.g., xASL_wrp_GetROIStatistics that will
+                                % be printed
+                                iSubjectSession_SetsID = x.dataset.nSessions*(SubjectIndex-1)+SessionIndex;
```
jan-petr, line 232: Check lines 132-138 - it's a bit worrisome that these differ from what is here. Also the comment on line 132 isn't very reassuring. I propose to adapt lines 132-138 to match what we have here and what I've tested.

Any better idea on this?

Functions/Statistics/xASL_stat_PrintStats.m
jan-petr: Actually, I haven't really modified the I/O. I only fixed an error from previously.

Any propositions what to change and how?

jan-petr (commented):
2 more comments.

Functions/Statistics/xASL_stat_PrintStats.m
HenkMutsaerts: Yes, e.g., you changed a variable name that was just copied to a new variable so was useless. But I agree that doesn't have to be in the header. If you didn't change the functionality you don't need to change the header of course, I thought that you changed the functionality but you probably didn't (e.g., if you disabled something it would be nice to put it to the header for us to remember)

I guess you can resolve

Functions/Statistics/xASL_stat_PrintStats.m
HenkMutsaerts: I thought I saw that it is now also created elsewhere. Are you really sure? If it doesn't hurt, I would just comment it out instead of removing it; as I cannot oversee now how this works when doing rc1T1 stats. Easier for me when I'm going to improve the voxel-wise stats

jan-petr: Not the exact same parameter. I full-text search the entire file to always make sure it's nowhere used, saved, or passed as an output. So it was really useless. I comment out things that can be potentially used in the future, but this didn't seem to be the case ;) Fine to keep it like this?

Functions/Statistics/xASL_stat_PrintStats.m
jan-petr: Agree in general!

Functions/Statistics/xASL_stat_PrintStats.m
HenkMutsaerts: Michael Stritt added that non-reassuring comment 3 years ago :)
I have now improved the comment at line 132. In an exceptional case where we want to enforce using the subjectSession definition from participants.TSV rather than x.S.DAT, we use this. So this is correct.

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
HenkMutsaerts: No, please don't because you make your life more difficult with the other branch that I'm working on for this. I'm saying this because your use case (3) is quite rare, and there are other places than deep down in the ROIstats thingy to check for being 4D. E.g, if you would throw an extra warning or error in the template creation, I would be fine. Or perhaps better for both of us: to put this warning or error in xASL_im_IM2Columns, with something like "4D image detected whereas the mask is not 4D, check your data, outputting NaN instead now'

jan-petr: Yes - I throwing an error during template creation cause it first crashes there.

I am putting a warning to xASL_im_IM2Columns. Note that the default here is to apply a shiftdim, which would make the 121x145x121x2 into 2x121x145x121, so we would still detect incorrect 4th dimension. But to make it a bit more transparent, I do check for the 4th dimension only when ApplyDimShift is FALSE. I can still remove this condition and only keep checking the 4th dimension, as this would capture my error, but it would be overall less elegant. See add277c2

HenkMutsaerts (changes requested):
See comments and my recent commit

Functions/ImageProcessing/xASL_im_IM2Column.m
```diff
@@ -34,8 +34,14 @@
...
 end
 
-if ApplyShiftDim && length(size(ImageIn))>3
-    ImageIn = shiftdim(ImageIn,length(size(ImageIn))-3);
+if length(size(ImageIn))>3
+	if ApplyShiftDim
+		ImageIn = shiftdim(ImageIn,length(size(ImageIn))-3);
+	else
+		% Issue a warning and output all NaNs
+		warning('Cannot accept 4D image when ApplyShiftDim is FALSE');
```
HenkMutsaerts: Nice, I would then only change this to:

warning('Cannot accept 4D image when ApplyShiftDim is false, check if your data were correctly processed');

Functions/Statistics/xASL_stat_PrintStats.m
HenkMutsaerts: I would comment it out rather than delete it here, just because I don't know how this affects the usage for images that don't have sessions like rc1T1. We can also test this use case but easier to just move it back in and comment out? Up to you.

HenkMutsaerts (approved):
Nice, see my new comment and addition to existing comment, but up to you, they are minor comments.

Functions/Statistics/xASL_stat_PrintStats.m
jan-petr: Sorry. it really makes no sense to put this back. Cause the indices are empty, so that's why we are not using them...

Functions/ImageProcessing/xASL_im_IM2Column.m
jan-petr: Done

PR #1728, opened 2024-07-01
#1726 Load and deal with multi-TR M0

### LINKED ISSUE #1726: Trivial version of mutli-TE multi-TI M0 quantification
### Description

DEBBIE (but also other sequences) could have M0 images scanned at multi-TE and multi-TI. This currently will either not work, or it will incorrectly average over those and mess up the quantification.

### Tasks

- [x] Make sure that multi-TI setup for an M0 is identified correctly during the import. d84b1285
- [x] Identify that this is happening in the M0 quantification (and later as well). d84b1285
- [x] Throw away all multi-TE but the first one d84b1285
- [x] Throw away all multi-TI but the last one with the highest TI. d84b1285
- [x] Do the simple single-TE single-TI M0 quantification and make sure the correct parameters are used. d84b1285
- [x] Add RepetitionTimeM0 somehwere in the documentation
- [x] Update all Flavor sequences to use the multi-TI version automatically
- [x] Run all testing

### Release notes

Select optimal TI for multi-TI M0scans.

jan-petr:
### Linked issue

Close #1726

### DIFF
--- a/External/SPMmodified/xASL/xASL_io_Nifti2Im.m
+++ b/External/SPMmodified/xASL/xASL_io_Nifti2Im.m
@@ -18,7 +18,7 @@ function [imOut, json] = xASL_io_Nifti2Im(niftiIn, ImageSize, bLoadAsSingle, bBI
 %
 % OUTPUT:
 %   imOut     - image matrix with single precision
-%   json      - json containing quantification parameters, if niftiIn is a path
+%   json      - json containing quantification parameters, if niftiIn is a path, otherwise empty (e.g., when the input is a matrix and not a path)
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
 % DESCRIPTION: This function loads a NIfTI image matrix with flexible input
 %              (as explained under INPUT: niftiIn). It does the following. It also loads the JSON sidecar and 
@@ -54,6 +54,11 @@ if nargin<4 || isempty(bBIDS2Legacy)
 end
 
 niiMat = false; % default
+
+% Initialize the json output
+if nargout >= 2
+	json = [];
+end
 
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
 % 1) Load the NIfTI (and JSON sidecar)
--- a/External/SPMmodified/xASL/xASL_io_Nifti2Im.m
+++ b/External/SPMmodified/xASL/xASL_io_Nifti2Im.m
@@ -56,9 +56,7 @@ end
 niiMat = false; % default
 
 % Initialize the json output
-if nargout >= 2
-	json = [];
-end
+json = [];
 
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
 % 1) Load the NIfTI (and JSON sidecar)
--- a/Functions/Quantification/xASL_quant_M0.m
+++ b/Functions/Quantification/xASL_quant_M0.m
@@ -58,7 +58,7 @@ elseif ~isfield(x,'M0_usesASLtiming')
 end
 
 % Allow inputting path instead of image
-M0IM = xASL_io_Nifti2Im(inputM0);
+[M0IM, jsonM0] = xASL_io_Nifti2Im(inputM0, [], [], true);
 
 %% ------------------------------------------------------------------------------------------------------
 % 1. Correct scale slopes, if Philips
@@ -114,9 +114,15 @@ else
     if ~isempty(regexpi(x.Q.Vendor,'GE')) && isempty(regexpi(x.Q.Vendor,'Siemens')) &&  isempty(regexpi(x.Q.Vendor,'Philips'))
         TR = 2000; % GE does an inversion recovery, which takes 2 s and hence signal has decayed 2 s
         fprintf('%s\n','GE M0 scan, so using 2 s as TR (GE inversion recovery M0)');
-    else
-        M0_parms = xASL_adm_LoadParms(M0ParmsMat, x);
-        TR = M0_parms.RepetitionTime; % This will be either the separate M0-scan value or the ASL scan value (if UseControlAsM0)
+	else
+		% Preferably load the TR from the rM0.json
+		if isempty(jsonM0) || ~isfield(jsonM0, 'RepetitionTime')
+			% Outdated and soon to be discontinued variant
+			M0_parms = xASL_adm_LoadParms(M0ParmsMat, x);
+			TR = M0_parms.RepetitionTime; % This will be either the separate M0-scan value or the ASL scan value (if UseControlAsM0)
+		else
+			TR = jsonM0.RepetitionTime;
+		end
     end
 
     %% ------------------------------------------------------------------------------------------------------
--- a/Modules/Module_Processing/Module_ASL/xASL_wrp_ProcessM0.m
+++ b/Modules/Module_Processing/Module_ASL/xASL_wrp_ProcessM0.m
@@ -118,52 +118,54 @@ end
 %% 0C. Use the M0 with the lowest TE (if we have multiple M0s)
 if ~isempty(jsonM0)
 	if isfield(jsonM0.Q, 'EchoTime')
-		% Select only non-zero TEs
-		uniqueNonzeroTE = unique(jsonM0.Q.EchoTime(jsonM0.Q.EchoTime > 0));
+		% Select unique TEs
+		uniqueTE = unique(jsonM0.Q.EchoTime);
 
 		% Check if we have multiple TE to decide to select only a single value
-		if length(uniqueNonzeroTE) > 1
+		% Only the smallest and single TE is relevant for M0 quantification. The entire TE vector is only
+		% useful for fitting T2 time and we do not need it for obtaining the equilibrium M0 for CBF quantification
+		if length(uniqueTE) > 1
             % First check if the number of TEs equals the number of M0 volumes
-            if length(uniqueNonzeroTE)>size(imM0,4)
+            if length(uniqueTE)>size(imM0,4)
                 warning('M0 JSON contains more unique echo times than number of M0.nii volumes');
                 fprintf('%s\n', 'Hint: check if your data are BIDS compatible');
             end
 
-			% Find the smallest of the previously selected positive nonzero TEs
-			minTE = min(uniqueNonzeroTE);
+			% Find the smallest of the TEs
+			minTE = min(uniqueTE);
 			% JSON values were converted from BIDS (seconds) to Legacy (millisecond) during JSON reading
 			fprintf('Multi-TE M0 present. Using the shortest TE = %.1f ms for M0 calibration.\n', minTE);
 
 			% Find the correct indices and update the jsonM0 parameter
-			indMinTE = jsonM0.Q.EchoTime == minTE;
-			jsonM0.Q.EchoTime = jsonM0.Q.EchoTime(indMinTE);
+			indexMinTE = jsonM0.Q.EchoTime == minTE;
+			jsonM0.Q.EchoTime = jsonM0.Q.EchoTime(indexMinTE);
 			if isfield(jsonM0, 'RepetitionTime') && length(jsonM0.RepetitionTime) > 1
-				jsonM0.RepetitionTime = jsonM0.RepetitionTime(indMinTE);
+				jsonM0.RepetitionTime = jsonM0.RepetitionTime(indexMinTE);
 			end
 
 			% Select the minimal TEs
-			imM0 = imM0(:, :, :, indMinTE);
+			imM0 = imM0(:, :, :, indexMinTE);
 		end
 	end
 
 %% 0D. Use the M0 with the TR closest to 2000 ms (if we have multiple M0s)
 	if isfield(jsonM0, 'RepetitionTime')
-		uniqueNonzeroTR = unique(jsonM0.RepetitionTime(jsonM0.RepetitionTime > 0));
+		uniqueTR = unique(jsonM0.RepetitionTime);
 
 		% Check if we have multiple TR to decide to select a single value only
-		if length(uniqueNonzeroTR) > 1
+		if length(uniqueTR) > 1
 			% We currently do not do inversion fitting in multi-TR M0, so we issue a warning and use the TR value closest to 2000ms
-			[~, optimalTR] = min(abs(uniqueNonzeroTR-2000));
-			optimalTR = uniqueNonzeroTR(optimalTR);
-			indOptimalTR = jsonM0.RepetitionTime == optimalTR;
+			[~, optimalTR] = min(abs(uniqueTR-2000));
+			optimalTR = uniqueTR(optimalTR);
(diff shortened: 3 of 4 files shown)


Functions/Quantification/xASL_quant_M0.m
```diff
@@ -58,7 +58,7 @@
 end
 
 % Allow inputting path instead of image
-M0IM = xASL_io_Nifti2Im(inputM0);
+[M0IM, jsonM0] = xASL_io_Nifti2Im(inputM0, [], [], true);
```
HenkMutsaerts, line 61: So this function allows inputting an image instead of a path. Will this still work, because jsonM0 is NaN by default? Or do we break this functionality (which might be fine).

To explain: when you feed an image to xASL_io_Nift2IM, it will just throughput it. So having this function at the start of a function, allows running the function on either a path (opening the image) or on an image directly (e.g. internally).

Functions/Quantification/xASL_quant_M0.m
```diff
@@ -114,9 +114,15 @@
     if ~isempty(regexpi(x.Q.Vendor,'GE')) && isempty(regexpi(x.Q.Vendor,'Siemens')) &&  isempty(regexpi(x.Q.Vendor,'Philips'))
         TR = 2000; % GE does an inversion recovery, which takes 2 s and hence signal has decayed 2 s
         fprintf('%s\n','GE M0 scan, so using 2 s as TR (GE inversion recovery M0)');
-    else
-        M0_parms = xASL_adm_LoadParms(M0ParmsMat, x);
-        TR = M0_parms.RepetitionTime; % This will be either the separate M0-scan value or the ASL scan value (if UseControlAsM0)
+	else
+		% Preferably load the TR from the rM0.json
+		if isempty(jsonM0) || ~isfield(jsonM0, 'RepetitionTime')
```
HenkMutsaerts, line 119: You checked that it is empty rather than e.g. NaN by default?

Also, BIDS has also second TR definition right (RepetitionTimePreparation), are these already converted to both being RepetitionTime here?

Modules/Module_Processing/Module_ASL/xASL_wrp_ProcessM0.m
```diff
@@ -71,11 +69,12 @@ function xASL_wrp_ProcessM0(x)
 % Use either original or motion estimated ASL4D
 % Use despiked ASL only if spikes were detected and new file has been created
 % Otherwise, despiked_raw_asl = same as original file
-if ~xASL_exist(x.P.Path_despiked_ASL4D,'file')
-    x.P.Path_despiked_ASL4D = x.P.Path_ASL4D;
-end
-tempnii = xASL_io_ReadNifti(x.P.Path_despiked_ASL4D);
-nVolumes = double(tempnii.hdr.dim(5));
+% PM: The code below is unused
```
HenkMutsaerts, line 86: I see the issue here, we used it but we used it below, I have now fixed this. In the first usage of Path_ASL this didn't matter, but in the second we resliced M0 to ASL for the "conventional" M0 image processing, which should be the despiked_ASL4D as after motion correction there could have been some reslicing.

Modules/Module_Processing/Module_ASL/xASL_wrp_ProcessM0.m
```diff
@@ -120,10 +119,36 @@ function xASL_wrp_ProcessM0(x)
 
 			% Find the smallest of the previously selected positive nonzero TEs
 			minTE = min(uniqueNonzeroTE);
-			fprintf('Multi-TE M0 present. Using the shortest TE=%.1f ms for M0 calibration.\n', minTE*1000);
+			% JSON values were converted from BIDS (seconds) to Legacy (millisecond) during JSON reading
```
HenkMutsaerts, line 136: Can we rewrite from line 112 above, such that we always select based on the TE and TR (which will be the single one in the case of a single M0)?

Effectively, this gives the same result, but the code will have fewer if-statements (except for printing to the screen), so easier to read.

I improved the header and the code "chapter" titles.

Modules/Module_Processing/Module_ASL/xASL_wrp_ProcessM0.m
```diff
@@ -120,10 +119,36 @@ function xASL_wrp_ProcessM0(x)
...
+		end
+	end
 
-			% Average M0 image across all these minimal TEs
-			imM0 = xASL_stat_MeanNan(imM0(:,:,:,jsonM0.Q.EchoTime == minTE), 4);
+	% Check if there are multiple different TR times and select the optimal one for M0 quantification
+	if isfield(jsonM0, 'RepetitionTime')
+		uniqueNonzeroTR = unique(jsonM0.RepetitionTime(jsonM0.RepetitionTime > 0));
+		if length(uniqueNonzeroTR) > 1
+			% We currently do not do inversion fitting in multi-TE M0, so we issue a warning and use the TR value closest to 2000ms
```
HenkMutsaerts: Same comment above for multi-TR M0. We don't do any fitting here.
And we don't issue a warning here.
I would issue a single fprintf('Warning: ...) see also comment above.

HenkMutsaerts (changes requested):
Some suggestions

Functions/Quantification/xASL_quant_M0.m
jan-petr: I understand. I need to check what happens and make sure that an empty JSON is outputted and that this is catched here...

Functions/Quantification/xASL_quant_M0.m
jan-petr: About NaN - You are right, this is fixed now. a80fba07

TR -> this is after BIDS to Legacy conversion, RepetitionTime is the only option in Legacy.

Modules/Module_Processing/Module_ASL/xASL_wrp_ProcessM0.m
jan-petr: That was something that somebody changed previously, but it looks good now...

Modules/Module_Processing/Module_ASL/xASL_wrp_ProcessM0.m
jan-petr: OK, I will try to reduce the number of IF-statements, but we will need to have the if statements to report that we were choosing single volume from multiple. I'll try to also simplify this, but just saying that we won't be able to remove all IFs.

Modules/Module_Processing/Module_ASL/xASL_wrp_ProcessM0.m
jan-petr: OK will do...

Functions/Quantification/xASL_quant_M0.m
jan-petr: Updated io_Nifti2IM. If input is a matrix, it returns an empty JSON. And the quant_M0 function knows how to deal with this. a80fba07

Modules/Module_Processing/Module_ASL/xASL_wrp_ProcessM0.m
jan-petr: I've checked the code, but it can't be really simplified to reduce the number of IF statements:
1. We need an IF to check if we have JSONM0
2. We need an IF to check presence of TE and TR field.
3. We need an IF again to check the presence of TR and TE fields to know if we update them.
4. So the only IF to get rid of potentially is the IF that checks if we have or haven't multiple values. But we need this IF anyway to report to the user that we found multiple values and are choosing a single one.

So we can't really simplify the code and I don't think this code is complicated. I've added extra comments.

Modules/Module_Processing/Module_ASL/xASL_wrp_ProcessM0.m
jan-petr: COrrected multi-TE to multi-TR. And print the message similar as for the multi-TE case, but with a warning.

Modules/Module_Processing/Module_ASL/xASL_wrp_ProcessM0.m
HenkMutsaerts: 1. We need this now, but not when we go for BIDS-requirement (which we almost have)
2. Same as (1)
3. Same as (1)
But I agree.

Modules/Module_Processing/Module_ASL/xASL_wrp_ProcessM0.m
```diff
@@ -120,10 +131,40 @@ function xASL_wrp_ProcessM0(x)
 
 			% Find the smallest of the previously selected positive nonzero TEs
 			minTE = min(uniqueNonzeroTE);
-			fprintf('Multi-TE M0 present. Using the shortest TE=%.1f ms for M0 calibration.\n', minTE*1000);
+			% JSON values were converted from BIDS (seconds) to Legacy (millisecond) during JSON reading
+			fprintf('Multi-TE M0 present. Using the shortest TE = %.1f ms for M0 calibration.\n', minTE);
```
HenkMutsaerts, line 137: Why is this not a warning and 0D has a warning?

Modules/Module_Processing/Module_ASL/xASL_wrp_ProcessM0.m
```diff
@@ -120,10 +131,40 @@ function xASL_wrp_ProcessM0(x)
...
+
+			% Find the correct indices and update the jsonM0 parameter
+			indMinTE = jsonM0.Q.EchoTime == minTE;
+			jsonM0.Q.EchoTime = jsonM0.Q.EchoTime(indMinTE);
+			if isfield(jsonM0, 'RepetitionTime') && length(jsonM0.RepetitionTime) > 1
+				jsonM0.RepetitionTime = jsonM0.RepetitionTime(indMinTE);
+			end
+
+			% Select the minimal TEs
+			imM0 = imM0(:, :, :, indMinTE);
```
HenkMutsaerts: I would rewrite ind to index for readability but not a big deal of course

Modules/Module_Processing/Module_ASL/xASL_wrp_ProcessM0.m
```diff
@@ -120,10 +131,40 @@ function xASL_wrp_ProcessM0(x)
...
+		end
+	end
 
-			% Average M0 image across all these minimal TEs
-			imM0 = xASL_stat_MeanNan(imM0(:,:,:,jsonM0.Q.EchoTime == minTE), 4);
+%% 0D. Use the M0 with the TR closest to 2000 ms (if we have multiple M0s)
+	if isfield(jsonM0, 'RepetitionTime')
+		uniqueNonzeroTR = unique(jsonM0.RepetitionTime(jsonM0.RepetitionTime > 0));
```
HenkMutsaerts: What do we do when we have additional M0s with zeroes? Perhaps it never happens, but now you added the '> 0' clause specifically, we probably need a specific warning as well if there are M0s with zeroes?
Or do we in case of a single M0 with a zero just ignore it.

Modules/Module_Processing/Module_ASL/xASL_wrp_ProcessM0.m
```diff
@@ -120,10 +131,40 @@ function xASL_wrp_ProcessM0(x)
...
+
+		% Check if we have multiple TR to decide to select a single value only
+		if length(uniqueNonzeroTR) > 1
+			% We currently do not do inversion fitting in multi-TR M0, so we issue a warning and use the TR value closest to 2000ms
```
HenkMutsaerts, line 157: This comment also belongs to us not fitting multi-TE M0 currently right? Or do we now try to fit it?

External/SPMmodified/xASL/xASL_io_Nifti2Im.m
```diff
@@ -55,6 +55,11 @@
 
 niiMat = false; % default
 
+% Initialize the json output
+if nargout >= 2
```
HenkMutsaerts: It's very minor of course, but I would from the perspective of prefering readability not do this clause.
There is no harm in always defaulting to json = [] right, irrespective of the nargout?
And code speed effect is negligible, it just makes for fewer if-statements which is desirable for bug-fixing in general.

So I would say that we in general default all variables to be empty when starting a function, and after the function checking for empty. And json = NaN means that there was a json but something went wrong in the loading (would also be nice to default that).

HenkMutsaerts (approved):
Left several potential improvements but up to you, I don't feel strong about them and know that you will still read them and take them serious so I approved :)

Modules/Module_Processing/Module_ASL/xASL_wrp_ProcessM0.m
jan-petr: Yes - we will run later a full BIDS validator before running a pipeline later...

Modules/Module_Processing/Module_ASL/xASL_wrp_ProcessM0.m
jan-petr: Cause with multi-TE, it's kind of assumed that you only take the first TE with best signal as it has the same TE as the volumes really used for CBF quantification. So there's no logical reason to take other TEs. So this comments only documents that we did that.

For TR, that's more tricky. We could pick another TR as more optimal (or some people might prefer to), or we could run an inversion recovery fit. So this solution is not necessarily the only and the best, so we have to warn the user we did a choice for him...

Modules/Module_Processing/Module_ASL/xASL_wrp_ProcessM0.m
jan-petr: Yes. agree. Done

Modules/Module_Processing/Module_ASL/xASL_wrp_ProcessM0.m
jan-petr: Sorry. that's a historical thing. sometimes we have TE vector with some zeros in it. They don't correspond to any volume and are in DICOM only and are supposed to be removed. But this happens already in the import. Plus PLD of M0 in BIDS is supposed to be 0. So when looking for a PLD vector, we skipped 0s. 

So I was just copy pasting code. Here, there's no need to have this check anymore. I'm removing it for clarity.

Modules/Module_Processing/Module_ASL/xASL_wrp_ProcessM0.m
jan-petr: As explained above. Makes no sense to fit multi-TE for M0 quantification. So that's why I don't even mention it. I've added a comment about this.

External/SPMmodified/xASL/xASL_io_Nifti2Im.m
jan-petr: Sure. no prob in always default to [].

I'd keep a simply output [] in case things go wrong. It's easier to check afterwards. And we of course print an error when this happens and this error is propagated, so the user notices that.

PR #1729, opened 2024-07-02
Bug #1721 fix merged multi pld

### LINKED ISSUE #1721: Single-PLD sequence becomes multi-PLD after merging
### Description

Currently, we are checking if a sequence is multi-PLD at the start of the ASL module for each session. So for single-PLD sequences that become multi-PLD only after merging, we are setting the multi-PLD aspect to false at the start of the ASL module. And so later on during the quantification, we don't run BASIL by default. This is for example the case of Siemens_PCASL_3DGRASE_VE11C_5.

### Tasks

- [x] We first need to split ASL to ASL and M0 and only afterwards check the ASL image parameters like multi-PLD. While splitting was added as the last section of the config in the ASL module, this really made no sense, because we had to make sure that splitting was done correctly before calculating any further options. Otherwise, we incorrectly thought that there are multi-PLD acquisitions because M0 had a different PLD. The order is now correct. df70ee37
- [x] After splitting ASL to M0 and ASL_source with 8 PLDs, we should move PLD 2:8 to ASL and PLD 1 to M0. But instead, we move PLD 1:7 to ASL. I couldn't reproduce this.
- [x] SaveCBF4D option needs to be rechecked after merging.  ed507f21
- [x] We have a bQuantifyMultiPLD option to set in the config file. And we now use it later as well and adjust based on data. But that means we can readjust this option when needed. So we need to separate the input from data and the real detected value - there's really no way around that cause the first single-session pass always sets it to OFF otherwise. So we need an config value and the really used value. We have discussed this with Henk and new solution is above.
- [x] Put multi-parameter identification to a sub-function. Call it at the start and also after merging. dfced3b6
- [x] Set bQuantifyParameters only later after merging. Make sure none of the parameters is used before. 9c9bf810
- [x] Verify that this works.

### Release notes

Make multi-PLD quantification work even when multi-PLD is created by session merging.

jan-petr:
### Linked issue

Close #1721

### DIFF
--- a/Modules/xASL_module_ASL.m
+++ b/Modules/xASL_module_ASL.m
@@ -423,11 +423,8 @@ if ~x.mutex.HasState(StateName{iState}) || ~x.mutex.HasState(StateName{iState+1}
 	end
 end
 
-if x.modules.asl.bMergingSessions
-	% Check again if we have multiple values for certain ASL parameters (PLD, LD, TE)
-	fprintf('%s\n','Sessions were concatenated. Checking again if LD, PLD, TE have multiple values...');
-	x = xASL_module_ASL_MultiParameterParsing(x, x.P.Path_PWI4D_used);
-end
+% Check again if we have multiple values for certain ASL parameters (PLD, LD, TE) directly on the PWI file used for processing
+x = xASL_module_ASL_MultiParameterParsing(x, x.P.Path_PWI4D_used);
 
 % Check multi-parametric quantification options
 x = xASL_module_ASL_MultiParameterQuantificationOptions(x);
--- a/Modules/xASL_module_ASL.m
+++ b/Modules/xASL_module_ASL.m
@@ -420,6 +420,15 @@ if ~x.mutex.HasState(StateName{iState}) || ~x.mutex.HasState(StateName{iState+1}
 		% Note: The paths x.P.(Pop_)Path_PWI4D_used can differ from a PWI4D obtained from a simple subtraction as we may have merged sessions,
 		% and the user or ExploreASL may have removed volumes from PWI4D. That's why we save and use x.P.(Pop_)Path_PWI4D_used in memory, such
 		% that the correct NIfTI will be used later in the pipeline (e.g., in xASL_wrp_VisualQC)
+
+		% Check again if we have multiple values for certain ASL parameters (PLD, LD, TE)
+		fprintf('%s\n','Sessions were concatenated. Checking again if LD, PLD, TE have multiple values...');
+		[tempPath, tempFile] = xASL_fileparts(x.P.Path_PWI4D_used);
+		tempASL = xASL_io_Nifti2Im(x.P.Path_PWI4D_used);
+		nVolumes = size(tempASL, 4);
+		x = xASL_module_ASL_MultiParameterParsing(x, fullfile(tempPath, [tempFile '.json'], nVolumes);
+
+		% Check again that we can save CBF4D for the given ASL4D
 		x = xASL_module_ASL_CheckSaveCBF4D(x);
 	end
 end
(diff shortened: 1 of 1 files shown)


Modules/xASL_module_ASL.m
```diff
@@ -156,14 +153,21 @@
...
-%% F. Split ASL and M0 within the ASL time series
+%% E. Split ASL and M0 within the ASL time series
 % Run this when the images hasn't been touched yet
 % The first three states are here, because the first two are run only conditionally
 
 % NOTE THAT THIS PART IS BEING PHASED OUT, AND NOW PROCESSED IN
 % `xASL_bids_parseM0` (BIDS->Legacy conversion). We keep this here only for
 % backward compatibility, for images imported a long time ago
 if ~x.mutex.HasState(StateName{1}) && ~x.mutex.HasState(StateName{2}) && ~x.mutex.HasState(StateName{3})
+	% Initialize the DummyScan and M0 position fields - by default empty
```
HenkMutsaerts, line 164: Why did you change the order of everything, instead of simply creating new subfunctions from the big subfunction that we already had below, that are called from the big subfunction, as well as from the new location?

Modules/xASL_module_ASL.m
```diff
@@ -843,22 +727,6 @@
     x.Q.BackgroundSuppressionNumberPulses = x.BackgroundSuppressionNumberPulses;
 end
 
-if ~isfield(x.modules.asl, 'bUseBasilQuantification') || isempty(x.modules.asl.bUseBasilQuantification)
```
HenkMutsaerts, line 846: Why is this part not called here to the new function?

HenkMutsaerts (changes requested):
Very dificult for me to check now if the order is still correct.
Why not simply moving only stuff from the ParseParameters part, part-by-part into a new subfunction, and just calling the new subfunction at multiple locations?

Modules/xASL_module_ASL.m
jan-petr: It's easy to check. I haven't changed anything but order. You can see the order change on lines 24-27.

For the mult-parameter check where I did more changes, I created a subfunction

Modules/xASL_module_ASL.m
jan-petr: Parsing of PLDs is in subfunction which is called from the original location.

Setting bUseBasil and bQuantify* is moved to a subfunction (xASL_module_ASL_MultiParameterQuantificationOptions) that is only called later before quantification. We have discussed why it can't be called twice and agreed to move it before quantification.

HenkMutsaerts: Sorry but where is xASL_module_ASL_MultiParameterQuantificationOptions?
Either you give me in full detail what you changed and why, or I cannot check it.
I don't know the order by heart, so if you would have just moved parts of the subfunction to new subfunctions and didn't change the order, then it would be easy to check.

Now, for me to really review if you didn't break anything, I need to check in-depth what each part was doing, where it was used, where dependencies are, etc. Because you changed the original functionality, which we didn't agree upon. We only agreed on moving a part of the subfunction to a new subfunction. Not on changing the order of subfunctions or something.

So I'm fine approving this, but then we risk that you moved stuff around that breaks other things and is not properly checked. So I would be more happy if you can do as we agreed: only moving parts of subfunctions.

Afaik the golden rule of commits: changing as little as possible to obtain the new functionality. A larger order change like you did now makes multiple branches etc working at the same time on this function very challenging.

Modules/xASL_module_ASL.m
HenkMutsaerts: OK I'll try to split my question then :)

Modules/xASL_module_ASL.m
```diff
@@ -21,10 +21,10 @@
 % A. Check existence and validity of ASL
 % B. Manage mutex state processing step
 % C. Cleanup before rerunning
-% D. ASL processing & quantification parameters
-% E. Backward and forward compatibility of filenames
-% F. Split ASL and M0 within the ASL time series
-% G. DeltaM and CBF parsing - check if all/some volumes are deltams or CBFs
+% D. Backward and forward compatibility of filenames
```
HenkMutsaerts, line 24: Why was this order switch necessary if all we wanted was the initial checks to stay identically but have a new subfunction that is called after merging?

Again not saying this is wrong, just would like to understand it.

1. xASL_module_ASL_ParseParameters was moved to later in the xASL_module_ASL initialization
2. DummyScanPosition was moved out of xASL_module_ASL_ParseParameters to the main initialization script
3. Managing BASIL quantification and managing CBF4D was done after managing TimeEncoded, now it is before
4. Nice that you split MultiParameterQuantificationOptions and MultiParameterParsing? But these are now called at different locations, whereas they seem related. They were called after one another and now at different locations.

Furthermore, they are now conditional to states. It seems easier to insert these functions in between the State checks but perhaps you wanted them after the State checks to speed up, but then it would be potentially nicer still to have them independent of the State checks (i.e. always doing the same) but at the start of xASL_module_ASL skipping xASL_module_ASL if status 999 is present.

Modules/xASL_module_ASL.m
```diff
@@ -415,11 +420,22 @@
 		% Note: The paths x.P.(Pop_)Path_PWI4D_used can differ from a PWI4D obtained from a simple subtraction as we may have merged sessions,
 		% and the user or ExploreASL may have removed volumes from PWI4D. That's why we save and use x.P.(Pop_)Path_PWI4D_used in memory, such
 		% that the correct NIfTI will be used later in the pipeline (e.g., in xASL_wrp_VisualQC)
+
+		% Check again if we have multiple values for certain ASL parameters (PLD, LD, TE)
+		fprintf('%s\n','Sessions were concatenated. Checking again if LD, PLD, TE have multiple values...');
```
HenkMutsaerts: This whole checking here could go to the subfunction xASL_module_ASL_MultiParameterParsing right, such that this subfunction only needs `x`, and can be potentially called from several positions without having to repeat the PWI4D that we take, or do we potentially take a different PWI4D each time we need to call this subfunction?

Modules/xASL_module_ASL.m
```diff
@@ -415,11 +420,22 @@
...
+		tempASL = xASL_io_Nifti2Im(x.P.Path_PWI4D_used);
+		nVolumes = size(tempASL, 4);
+		x = xASL_module_ASL_MultiParameterParsing(x, fullfile(tempPath, [tempFile '.json']), nVolumes);
 	end
 end
+
 % Quantification is performed here according to ASL consensus paper (Alsop, MRM 2016)
 % Including PVC
 if ~x.mutex.HasState(StateName{iState}) && x.mutex.HasState(StateName{iState-4})
+	% Check multi-parametric quantification options
```
HenkMutsaerts: Why do we have a separate MultiParameterQuantificationOptions and MultiParameterParsing? Are those not related? And just required in any case after merging (potentially just doing nothing in the default case)? It seems easier to insert these functions in between the State checks but perhaps you wanted them after the State checks to speed up, but then it would be potentially nicer still to have them independent of the State checks (i.e. always doing the same) but at the start of xASL_module_ASL skipping xASL_module_ASL if status 999 is present.

Modules/xASL_module_ASL.m
HenkMutsaerts, line 846: Please in cases like this, give them a new section title %% Manage BASIL quantification and %% Manage saving CBF4D, and then move these elsewhere, so it is easier for me to check

HenkMutsaerts (changes requested):
Tried to explain some more what wasn't clear from your order changes, and why they didn't seem necessary vs only splitting the current subfunction and calling it from a new location (so again I'm probably fine with it, but just want to understand what you did so I can review it)

Modules/xASL_module_ASL.m
jan-petr: See my answer to this above.
To repeat:
1. I moved them out of state-checks. 
2. multiParsing was run already at the start and should be only run again when merging was done.
3. QuantOptions should be run only once - cause it redefines options that are provided in dataPar.json and it would be difficult to run it twice and keep track of what was changed in ASL_module and what was provided at dataPar.json. So that's why QuantOptions is run separately from MultiParsing.

Modules/xASL_module_ASL.m
jan-petr: Of course, no prob explaining the motivation. I wrote it in the issue:
- ASL split to ASL+M0 happens in BIDS2Legacy, but we still have a failsafe in ASL_module to do this in case people provide data in Legacy themselves. So when ASL+M0 split happens in ASL_module, it needs to happen before ASL parameter check. Because ASL/M0 tend to have different TR/TE/PLD and this would then be falsely identified as a multi-parameter setup. We could have of course done a complicated check in the parameter check, but it's much easier to split before looking for multi-PLD.

Answers to your questions:
1. We didn't do anything else than moving Split ASL forward.
2. DummyScanPosition was not really moved out of ASL_ParseParameters. It was in ASLSplit and stayed there. We only moved the parameter initialization to empty.
3. BASIL quant has to be initialized after bQuantifyMultiPLD as it needs to know this. We've agreed to move bQuantifyMultiPLD to shortly before quantification, so we had to move BASILquant as well.
4. One thing is to calculate parameters, which can be done at ASL module start, but has to be recalculated after merging. Quantification options have to be done before quantification, but has to be run only once to be able to deal with missing parameters and values from dataPar -> so it has to be called only after merging and before quant. So that's why Parameters are called twice, and options once.
5. Parameter function is unconditional at the start. And the recalculation is conditional not for speed-up, but because I need it to run only when merging is done. But I should perhaps run this conditional to merging, but outside of state-check. Same for QuantificationOptions - that has to be run always. Fixed in the last commit.

Modules/xASL_module_ASL.m
jan-petr: Good point. Next time...

Modules/xASL_module_ASL.m
jan-petr: Yes. I didn't do this previously because we indeed use x.P.Path_PWI4D_used or x.P.Path_ASL4D. And we passed jsonPath to the function. But I can simplify and unify this. Indeed a good idea to put all things inside. Fixed in the last commit.

Modules/xASL_module_ASL.m
HenkMutsaerts: Why, you can still add those titles where you moved them right? That is 3 seconds work. I don't need to check it, it just helps increasing the readability anyways

Modules/xASL_module_ASL.m
HenkMutsaerts: 1. OK
2. OK
3. OK
4. OK
5. I thought that we agreed that these parameters were only used for quantification, hence that you would move them to just before quantification (not needing the state checking), and irrespective of merging. Can we remove the merging check? Because I would like the pipeline to do the same irregardless of merging. Merging is a separate function that does merging-related stuff. So I would suggest:
option A: we have to call it twice, then add it to the merging wrapper instead of xASL_module_ASL so the transparency/modularity philosophy that merging doesn't change xASL_module_ASL but is only an insert, if fullfilled
option B: we only have to call this before quantification, then you can remove it at the start and also remove the "if bMerging statement"

6. As per my last comment, it would be great if you can add to the header of the (new) ParseParms functions what is done. So create sections (M0 management, BASIL management, etc), numbering them, and then adding them to the header, like we have in all our functions/wrappers (like here in xASL_module_ASL).

HenkMutsaerts (changes requested):
Minor question, I can probably quickly approve.

Modules/xASL_module_ASL.m
jan-petr: Aha. OK :) Sure, added!

Modules/xASL_module_ASL.m
jan-petr: 5.1 Yes - they are now just before merging and without "state checking".
5.A I can't move it inside the merging function. Because on a re-run, the state will be finished and merging won't be re-run.
5.B I don't want to remove it at the start because we want to have access to UniqueParameter vectors potentially even before quantification. Not the bQUantifyMultiPLD, but the nUniquePLD etc.

The pipeline of course does the same irregardless of the merging. If we run the MultiParam function again without merging, it will have 100% same output and zero effect. It will have zero influence to run it the second time, it will just consume a few extra seconds to read it again the file etc. I'd say it is a bad programming practice to run it the second time because we are 100% sure the effect will be the same because it will read the same ASL4D.json, there's no way anything could have changed - unless we merged into a new JSON. But if you really really really insist, I can do that ;)

6. Yes. Added

HenkMutsaerts (changes requested):
Nice! But you didn't work on my previous comment 5 yet? Or should I check this somewhere in the main issue?

Modules/xASL_module_ASL.m
HenkMutsaerts: 5. Okay, compromise, you can add it as elseif to the if-statement of the merging state control right?

If State{notyetmerged}
     Merging function
    % we need to reinitialize the parameters from the merged PWI4D.json
     ParseParameters
elseif bMerge
    % we need to reinitialize the parameters from the merged PWI4D.json
    ParseParameters
end

or at least just keep it in the same code section, that it is clear that this is "an insert".

But I understand now why you need to do it like this.
I would still say that if it costs 0 time, it might be bad programming practice but clearer since we will always do the same, i.e. rechecking the quantification parameters separately for the quantification compared to the processing.

Thinking about this, isn't this more logical:

1. ASL4D.json provides the settings/data for how we should process our data
2. PWI4D.json provides settings/data for how we should quantify our data (can be from ASL4D.json but can also be either manually or automatically merged)
3. xASL_wrp_Quantify re-reads both PWI4D.nii AND PWI4D.json. Because we actually don't know if the user has manually merged two NIfTIs!

Even if you disagree that the user should be able to manually merge two NIfTIs (which is transparent and doable now), this is even more clear. And we don't need to say "if merging, we decide that the quantification parameters potentially differ from the processing parameters, but if no merging, we are certain that this is not the case".

So I would propose doing this for simplification and transparency. Also easier for bug-fixing and provenance.

HenkMutsaerts (approved):

PR #1734, opened 2024-07-14
Closes #1733 xASL_wrp_LinearReg_T1w2MNI: check ASL orientation matrices

### LINKED ISSUE #1733: Issue warning when ASL registration inconsistencies for ExploreASL rerun
### Description

A potential error can occur if for whatever reason, in a previous rerun some where realigned (e.g., to MNI) and others were not.
So we will now issue a warning if any of the orientation matrices were changed (e.g. in an earlier processing run).

Note that this can be normal for FLAIR and its derivative WMH_SEGM, as these are first resampled to T1w.nii
For T1w.nii, this can also be normal if the T1w was already realigned to MNI in a previous run.
So we check this for ASL only, by comparing the nii.mat to nii.mat0. Either all were realigned, or none. If only some were realigned, we issue a warning.

### Tasks

- [x] xASL_wrp_LinearReg_T1w2MNI: check ASL orientation matrices

### Release notes

Fix minor bug in inconsistent orientation matrices during realignment.

HenkMutsaerts:
### Linked issue

Closes #1733

### DIFF
--- a/Modules/Module_Processing/Module_Structural/xASL_wrp_LinearReg_T1w2MNI.m
+++ b/Modules/Module_Processing/Module_Structural/xASL_wrp_LinearReg_T1w2MNI.m
@@ -18,12 +18,15 @@ function xASL_wrp_LinearReg_T1w2MNI(x, bAutoACPC)
 % a center of mass-based ACPC alignment, and then several iterations of SPM coregistration.
 % Assuming that this submodule is run at the start of ExploreASL, all NIfTI orientation matrices are restored before running the registration.
 %
+% This function performs the following steps:
+% 1. Obtain path lists
+% 2. Issue warnings if any of the orientation matrices were changed (e.g. in an earlier processing run)
+% 3. Perform the registration
+%
 % EXAMPLE: xASL_wrp_LinearReg_T1w2MNI(x);
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
 % __________________________________
-% Copyright 2015-2019 ExploreASL
-%
-% 2019-05-02 HJM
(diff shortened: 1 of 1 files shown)


Modules/Module_Processing/Module_Structural/xASL_wrp_LinearReg_T1w2MNI.m
```diff
@@ -53,34 +52,66 @@ function xASL_wrp_LinearReg_T1w2MNI(x, bAutoACPC)
...
+    end
+end
+
+% Check orientation differences for all ASL NIfTIs
+% between nii.mat0 (original orientation from the scanner)
+% and nii.mat (potential new orientation after registration)
+% If no registration happened, these are equal
+for iCheck=1:length(checkList)
+    nii = xASL_io_ReadNifti(checkList{iCheck});
+    matEqual(iCheck) = min(min(nii.mat==nii.mat0));
```
jan-petr: This would also work. But replacing by "isequal(nii.mat, nii.mat0)" is a bit more transparent.

jan-petr (changes requested):
A small fix is needed.

Modules/Module_Processing/Module_Structural/xASL_wrp_LinearReg_T1w2MNI.m
HenkMutsaerts: Yes, fixed in e1119ede4c2447e1602b94c5d58c5c2035de455a

jan-petr (approved):
OK

PR #1736, opened 2024-07-19
Closes #1039 xASL_im_FillNaNs

### LINKED ISSUE #1039: Incorrect flow field edges y_ASL.nii after presmoothing
### Description

We smooth `y_T1.nii` with the ASL resolution to create `y_ASL.nii`. Sometimes, `y_ASL.nii` still has incorrect edges, perhaps due to too small kernel sizes in `xASL_im_PreSmooth`? This seems independent of `x.settings.Quality`.

image

On the left, you can see `y_T1.nii`, on the right `y_ASL.nii`, for `Philips_2DEPI_Lesion_NoFLAIR_NoM0_HiQ`. This results in erroneous lines at the bottom:

image

There seems to be something wrong with `xASL_im_PreSmooth`, I cannot find anything else. Alternatively, we keep it as it is, and always remove the outer e.g., 3 lines of `y_ASL.nii` and extrapolate with `xASL_im_ndnanfilter`, at the end of `xASL_im_CreateASLDeformationField`.

CORRECTION:
From the example of EPOD, I believe I found the culprit.
The NaNs outside the flowfield are locations in the new MNI space where we don't have information (because of limited FoV of e.g., the T1w). We used to set these NaNs to a value by extrapolating smoothing (or linearly extrapolating), which is incorrect. We should replace these NaNs by the "identity flowfield" (which is just the flow field values that set a voxel to its default location, which we have: `External/SPMmodified/MapsAdded/Identity_Deformation_y_T1.nii`. See my suggested fix of this bug. It seems that in most cases, this was not an issue (except for the weird edges of the population MNI images). But when the FoV of the T1w is too limited, this becomes an issue, because we interpolate values that translate to values within the brain. This is how a part of the brain was also shown in a weird deformed way in a corner of the MNI space for some EPOD subjects, see #1731.

I'm not sure if this fully fixes our bug described here, but it seems like a decent start! We should ask the students to do some testing.

Also: @jan-petr do you want to check if my fix is not too simple (e.g., we had quality and voxel-size options which I have now removed, if you know a mathematical way to do the same fix without needing the identity y_T1 NIfTI, that would be better).

### Requirements

- [x] xASL_im_FillNaNs: add option 4 , replacing NaNs with identity flowfield
- [x] Set default of xASL_im_FillNaNs to 1 instead of 3 (this was wrong)
- [x] Use new option 4 with xASL_im_FillNaNs
- [x] Avoid unzipping Identity_Deformation_y_T1.nii.gz

### Release notes

Fixed NaN filling using linear interpolation.

HenkMutsaerts:
### Linked issue

Closes #1039


jan-petr (changes requested):
I like the solution, and there's only one possible issue that I can think of:

The only issue with using the identity is that it might create a discontinuity at the edge of the NaNs. So I agree that the extrapolation was somehow wrong, but it had a smooth edge. We should combine the two properties - correct transformation, but smooth edges.

This is just a thought, I am not sure how much of a problem that really is. The current solution can give a small artifact on the places where the brain touches the FOV. Can you please verify in the data if that is a problem or it isn't?

If this is a problem, we can try the following:
1.⁠ ⁠Create NaN mask
2.⁠ ⁠⁠Replace y on Nanmask with identity (exactly as you have).
3.⁠ ⁠⁠Smooth the newly created mask with a small kernel
4.⁠ ⁠⁠Replace y on Nanmask with the smoothed mask from 3.

It will: look absolutely the same as your solution outside Nanmask
it will: solve the issue with y_T1 going crazy.
It will: solve the issue with discontinuity on the FoV border.

So please check this proposition, check the data, think about it and let me know if the current solution looks good -> then I only review the code and check it formally and merge.
If you think that the current solution doesn't look good, I can try to implement that small improvement.

HENK: Yes, I agree with you that this would be better.

HenkMutsaerts:
image
This is after replacing the NaNs and zeros by the identity field.
image
This is after smoothing the result, and replacing only the original NaNs and zeros (the mask) by the smoothed result.
You can see that the line is still there (the outer edge of the original FoV).
We can expand the mask by one line (or more) and apply the smoothing also there, to remove this line in the flowfield, but this is tricky if the brain would be at the edge of the flowfield, which happens seldomly. But then again, I don't know how important it is to avoid the line in the flowfield?

External/SPMmodified/MapsAdded/Dummy_Deformation_y_T1.nii.gz
jan-petr, line 1: I don't see this file ever used. Please remove.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
```diff
@@ -1,4 +1,4 @@
-function [IM] = xASL_im_FillNaNs(InputPath, UseMethod, bQuality, VoxelSize)
+function [IM] = xASL_im_FillNaNs(InputPath, UseMethod, bQuality, VoxelSize, bZeroes)
 %xASL_im_FillNaNs Fill NaNs in image
 % FORMAT: xASL_im_FillNaNs(InputPath[, UseMethod, bQuality, VoxelSize])
```
jan-petr: Please add bZeroes to the header.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
```diff
@@ -10,15 +10,21 @@
...
+%                 4: fill with flow field identity map (specific to MNI 1.5mm space used by CAT12/DARTEL)
 %                 (OPTIONAL, DEFAULT = 1)
 %   bQuality    - affects kernel size of the extrapolation performed in method 1
 %                 (extrapolating by smoothing goes much faster with smaller
 %                 kernel but this creates ringing artifacts)
 %                 (OPTIONAL, DEFAULT = 1)
 %   VoxelSize   - [X Y Z] vector for voxel size 
 %                 (OPTIONAL, DEFAULT = [1.5 1.5 1.5])
+%   bZeroes     - boolean for converting zeroes to NaNs and also changing them accordingly (OPTIONAL, DEFAULT = false)
+%   x           - x is only needed for the location of the deformation identity field, if UseMethod is 4 (default)
```
jan-petr: x and IdentityMNIEdges not defined anywhere in the function header nor comments.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
```diff
@@ -80,21 +86,33 @@
         else
                 VoxelSize = [1.5 1.5 1.5];
         end
-        
     end
 end
```
jan-petr, line 90: For the UseMethod ~=1 default of parameter 4 are not defined.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
```diff
@@ -80,21 +86,33 @@
...
 
+if nargin<5 || isempty(bZeroes)
+    if UseMethod==3
+        bZeroes = true;
+    else
+        bZeroes = false;
+    end
+end
+
+
```
jan-petr: Missing default definition for parameters 6 and 7.

Functions/ImageProcessing/xASL_im_CreateASLDeformationField.m
```diff
@@ -62,7 +62,8 @@ function xASL_im_CreateASLDeformationField(x, bOverwrite, EstimatedResolution, P
 sKernel(EstimatedResolution<TemplateResolution) = 0;
 
 %% 2) Fill NaNs at edges y_T1.nii flowfield to prevent interpolation artifact
-xASL_im_FillNaNs(x.P.Path_y_T1, 3, x.settings.Quality); % First fill NaNs, to prevent interpolation artifacts
+xASL_im_FillNaNs(x.P.Path_y_T1, 3, [], [], true); % First fill NaNs, to prevent interpolation artifacts
```
jan-petr, line 65: I'm missing the context why zeros are filled. This is a bit random. Are we expecting zeros around the edges or something? Then lets fill them as NaNs here before the call. Otherwise, we risk that normal zeros in the middle of the image are rewritten for no reason.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
```diff
@@ -158,29 +183,67 @@
...
+% nEdgeVoxels - number of voxels at the edges of the non-nan values that we check for weird interpolation values, 
+% to be set to NaN and ignored in the extrapolation
 
-% Zeros are also converted to NaNs - because this is sometimes unclear in the flow fields.
-IM(IM==0) = NaN;
+%% 0. Admin
+% Also catch Infs or imaginary numbers or whatever. Zeroes are already catched above in the main function.
 IM(~isfinite(IM)) = NaN;
 
+if nargin<2 || isempty(nEdgeVoxels)
```
jan-petr: I would remove the entire nEdgeVoxels parameter - it's never used, so it's only confusing. = you never call it in the function.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
```diff
@@ -158,29 +183,67 @@
...
+% together with the transformation difference from a smoothed transformation field
+% Note that this could also potentially fix extreme transformations
+
 % Get distance transform
-distMap = xASL_im_DistanceTransform(~isfinite(IM));
+distMap = xASL_im_DistanceTransform(isnan(IM));
+
+% Smooth the image outside of the NaNs
+IMsmooth = xASL_im_ndnanfilter(IM, 'gauss', [8 8 8], 1);
+% Get an absolute difference image between the original and the smoothed transformation field
```
jan-petr: I'm a bit confused what is happening here and why. I thought we said we'll replace NaNs by identity.
line 213-233 are very complex and I'm really not sure if we need anything like this.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
```diff
@@ -191,62 +254,162 @@
...
+end
 
 % Any remaining NaNs had to be inside, so we interpolate them (setting them
 % to zeros creates artifacts!)
 IM = xASL_im_ExtrapolateSmoothOverNaNs(IM, 0);
 
+end
+
+
+function [diffIs, minInd, maxInd, thresholdIs] = xASL_im_Extrapolate_FixEdges(imThisCol, bSide)
```
jan-petr: I miss the rationale for this statistics and checking. We should simply replace nans by linearly extrapolated values and not calculate stats etc. 

I understand that there were some errors, but this entire construction is extremely complicated.

jan-petr (changes requested):
Sorry, I don't understand the thing at all.

External/SPMmodified/MapsAdded/Dummy_Deformation_y_T1.nii.gz
HenkMutsaerts: It is useful for testing problems with y_T1.nii, and really small. I thought of moving it out of the repository elsewhere, but then it would be a lot searching next time. Though if we have a database of NIfTIs for "unit tests" I could put it there.

Functions/ImageProcessing/xASL_im_CreateASLDeformationField.m
HenkMutsaerts: Yes, sometimes NaNs at the flowfield edges have turned into zeroes. So we want to convert zeroes to NaNs. Then we run our filling, which only fills stuff at the edges of the flowfield. Any NaNs (or original zeroes) that remain are in the middle of the flowfield, would be interpolated by the ndnan_filter. I agree with you in theory that this could go wrong, but in reality I don't think there will ever be a 0.0000000000.

But I can change this here, that is probably fine.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
HenkMutsaerts: See whatsapp. I will add a rationale, but this part didn't work well, the distance-based outlier exclusion worked better.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
HenkMutsaerts: No, identity (which means not moving anything) was illegal according to you. Instead, we should take the borders of the non-values values and extrapolate them linearly.

The point is that the borders (single voxels) can have weird artifacts which would then be extrapolated. So we want to be a bit more robust. So we first want to check if there are significant outliers at the border. This algorithm here checks for those severe outliers and removes them, before doing the linear extrapolation.

I have now added explanations, is this better?

External/SPMmodified/xASL/xASL_im_FillNaNs.m
HenkMutsaerts: It is used, but disabled by default.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
HenkMutsaerts: No these are not used, I removed them.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
HenkMutsaerts: Done

External/SPMmodified/xASL/xASL_im_FillNaNs.m
HenkMutsaerts: That's what the comment says there, Method 1 is the only method that uses that parameter :)

External/SPMmodified/xASL/xASL_im_FillNaNs.m
HenkMutsaerts: These parameters don't exist

External/SPMmodified/MapsAdded/Dummy_Deformation_y_T1.nii.gz
jan-petr: For testing, this is of course really good. But lets move it to the "Testing" repository. Cause here it might be confusing.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: I know and agree. But for consistency, you should set the parameter to empty or something. So that the parameter exists.

Functions/ImageProcessing/xASL_im_CreateASLDeformationField.m
jan-petr: I agree. Don't close yet, I'll add details to comments and see if I can improve this a bit. But agree in general that this is needed.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: Yes - I used the "identity" parallel, but indeed meant the "extrapolated identity".

OK. I will check the code and see if we can get rid of those outliers before having them, or otherwise keep your solution and clean it.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: OK. I'll check in detail.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
HenkMutsaerts: Done

External/SPMmodified/xASL/xASL_im_FillNaNs.m
```diff
@@ -6,12 +6,9 @@
...
 %   UseMethod   - one of the following methods to fill the NaNs:
-%                 1: extrapolate by smoothing (useful for any image, e.g.
-%                    extrapolating M0 smooth biasfield) (DEFAULT)
-%                 2: set to 0 (useful for e.g. DARTEL displacement fields,
-%                    where 0 means no displacement)
-%                 3: linearly extrapolate 
-%                 4: fill with flow field identity map (specific to MNI 1.5mm space used by CAT12/DARTEL)
+%                 1: extrapolate by smoothing (useful for any image, e.g. extrapolating M0 smooth biasfield) (DEFAULT)
+%                 2: set to 0 (useful for e.g. DARTEL displacement fields, where 0 means no displacement)
+%                 3: extrapolate deformation field linearly - this method is adapted to SPM deformation fields XxYxZx3
```
HenkMutsaerts: should be XxYxXx1x3

External/SPMmodified/xASL/xASL_im_FillNaNs.m
```diff
@@ -73,18 +70,23 @@
     bQuality = 1; % default
 end
 
-if UseMethod==1 % only method that needs VoxelSize
-    if nargin<4 || isempty(VoxelSize)
+if UseMethod == 1 || UseMethod == 3 % only methods that needs VoxelSize
```
HenkMutsaerts: Why did you make method 3 voxel-size dependent? That's tricky because we also use it inside CAT12 where voxel-size can change dependent on our quality settings. I thought we decided to do this voxel-size ignorant, by just checking how the deformation field behaves.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
```diff
@@ -73,18 +70,23 @@
...
+    if nargin < 4 || isempty(VoxelSize)
         if ~isnumeric(InputPath) && xASL_exist(InputPath, 'file')
             nii = xASL_io_ReadNifti(InputPath);
             VoxelSize = nii.hdr.pixdim(2:4);
+		elseif UseMethod == 3
```
HenkMutsaerts: See above, you don't want the method to be dependent on inputting voxel-size or not. The previous method worked really nicely without having to input the voxel size.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
```diff
@@ -110,39 +111,45 @@
     IM(IM==0) = NaN;
 end
 
+% Also catch Infs or imaginary numbers or whatever. Zeroes are already catched above in the main function.
+IM(~isfinite(IM)) = NaN;
```
HenkMutsaerts: I considered this as well, but you don't want this. When we use this function to just remove NaNs in images, sometimes you may want imaginary numbers or Inf's to remain. What happens with Inf's or imaginary numbers that are not at the edges?

External/SPMmodified/xASL/xASL_im_FillNaNs.m
```diff
@@ -110,39 +111,45 @@
...
-
-                CurrentIt = CurrentIt+1;
-            end
-        end
-    end
+% Run only if there were actual NaNs present
+if sum(isnan(IM(:))) > 0
+	switch UseMethod
+		case 1
+			% Method run is run independently for each 3D volume
```
HenkMutsaerts: Method run is run independently?

External/SPMmodified/xASL/xASL_im_FillNaNs.m
```diff
@@ -110,39 +111,45 @@
...
+						end
+					end
+				end
+			end
+			fprintf('\n');
+		case 2
+			IM(isnan(IM)) = 0;
+		case 3
+			% Use this for cases where NaNs need to expand linearly (e.g., extrapolating deformation flowfields)
+			if ndims(IM) ~= 4 || size(IM, 4) ~= 3
```
HenkMutsaerts: Incorrect. Deformation fields are always [X Y Z 1 3]

External/SPMmodified/xASL/xASL_im_FillNaNs.m
```diff
@@ -185,52 +178,81 @@
...
 % to be set to NaN and ignored in the extrapolation
+% VoxelSize   - voxel size is needed for extrapolating correctly as the transformation is given in mm and not voxels
 
 %% 0. Admin
 if nargin<2 || isempty(nEdgeVoxels)
     nEdgeVoxels = 0;
     % We can try this later, but it seems to work less well than doing the same based on the distance transform
 end
 
+if nargin < 3 || isempty(VoxelSize)
```
HenkMutsaerts: You said you would leave my method intact and only work from 250 onwards ;)

I see now that you only improved my IMsmooth to 3D (my bad/bug). Then I would move this VoxelSize thingy to later, because it doesn't apply to the IMsmooth

External/SPMmodified/xASL/xASL_im_FillNaNs.m
```diff
@@ -185,52 +178,81 @@
...
+	% The voxel-size is not provided and needs to be estimated from the transformation maps
+	% We set it here to empty in admin and estimate it later
+	VoxelSize = [];
+end
+
+%% Create masks and estimate voxel size
+% Create NaN mask for all 3D volumes together
+maskNaN = sum(isnan(IM), 4) > 0;
```
HenkMutsaerts, line 218: So if we e.g. know for a border voxel where it came from in AP and LR, but not IS, then we now extrapolate it? Because then only extrapolating IS and using the information from the other directions could still be beneficial? This should never occur except for special situations where borders are smoothed so perhaps this is a good thing.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
```diff
@@ -185,52 +178,81 @@
...
+
+% The voxel-size needs to be estimated
+if isempty(VoxelSize)
+	% For each volume, we calculate the difference of neighboring voxels
+	% We calculate the mean over nonNaN values, which gives the average update including the correct direction
+	diffIM = IM(2:end, :, :, 1) - IM(1:end-1, :, :, 1);
+	diffIM = diffIM(~maskNaN(2:end, :, :));
+	VoxelSize(1) = mean(diffIM(~isnan(diffIM)));
```
HenkMutsaerts, line 219: Could replace this by xASL_stat_MeanNan.

Are you sure that the each volume only contains one dimension? Because that would be quite inefficient... that would mean that you could describe the flowfields in a single volume...

HenkMutsaerts: Could be replaced by xASL_stat_MeanNan

External/SPMmodified/xASL/xASL_im_FillNaNs.m
```diff
@@ -185,52 +178,81 @@
...
+end
+
 %% 0.5 Fix interpolation edges
 % The first layer(s) next to the NaNs are sometimes interpolated as well
 % So here we try to estimate which voxels those are, by looking at the distance from the NaNs
 % together with the transformation difference from a smoothed transformation field.
 
-% First we get the distance transform
-distMap = xASL_im_DistanceTransform(isnan(IM));
+% First we get the distance transform from NaNs towards the real values
```
HenkMutsaerts: I would add (from the mask inwards) to clarify this

External/SPMmodified/xASL/xASL_im_FillNaNs.m
```diff
@@ -185,52 +178,81 @@
...
+
 % We obtain an absolute difference image between the original transformation field, and the smoothed transformation field
 % This difference will show us how "extreme" the transformations are.
-deltaIm = abs(IM-IMsmooth);
+deltaIm = mean(abs(IM-IMsmooth), 4);
 % This absolute difference image, is now divided by the distance.
 % Because we want to detect extreme transformation values that are close to the border (low distance)
 % So this would result in a high relativeDelta
-relativeDelta = deltaIm./distMap;
+relativeDelta = deltaIm./distMapInward;
```
HenkMutsaerts: so deltaIm is 3D and distMapInward is 4D? (or 5D)?

External/SPMmodified/xASL/xASL_im_FillNaNs.m
```diff
@@ -242,7 +264,8 @@
 edgeDistance = round(0.1*averageSize);
 
 % Set the voxels at the border where extreme values were detected to NaN as well
-IM(relativeDelta>thresholdIs & distMap<edgeDistance) = NaN;
+% We update the NaN mask instead of modifying the image
+maskNaN(relativeDelta > thresholdIs & distMapInward < edgeDistance) = 1;
```
HenkMutsaerts: Tricky, because that means that you have to use the maskNaN throughout. So there is the possibility to take the values that should have been set to NaN. This possibility is not there anymore when they are set to NaNs instead of expanding the NaN mask.

External/SPMmodified/xASL/xASL_im_DistanceTransform.m
```diff
@@ -9,7 +9,7 @@
 %
 % OUTPUT:
 %  dist     - distance map, distance in voxels to the closest point on the mask
-%  x,y,z    - X, Y, and Z coordinate of the closest voxel on the mask
+%  x,y,z    - X, Y, and Z coordinate of the closest voxel on the mask. Note that these are relative coordinates, and Y,X,Z
```
HenkMutsaerts: What do you mean by Y,X,Z? Shouldn't it be X Y Z? The fact that some software like SPM or Matlab turns X & Y around doesn't mean that we should do that too :)

External/SPMmodified/xASL/xASL_im_FillNaNs.m
```diff
@@ -267,184 +267,31 @@
...
-end
-
-% Any remaining NaNs had to be inside, so we interpolate them (setting them
-% to zeros creates artifacts!)
-IM = xASL_im_ExtrapolateSmoothOverNaNs(IM, 0);
-
-end
+%% Extrapolation of values based on the edge values 
+% Calculate the distance map from the non-Nan towards NaN
+[~, distY, distX, distZ] = xASL_im_DistanceTransform(~maskNaN);
```
HenkMutsaerts: very weird that this is not X Y Z but Y X Z :)

External/SPMmodified/xASL/xASL_im_FillNaNs.m
```diff
@@ -267,184 +267,31 @@
...
 
-% Mean difference, from the middle index to the index of the last value
-% Calculated by their difference with their neighboring voxel            
-
-if ~bSide % start/first
-    diffIs = imThisCol(minInd:midInd) - imThisCol(minInd+1:midInd+1);
-else
-    diffIs = imThisCol(midInd:maxInd) - imThisCol(midInd-1:maxInd-1);
+for iDim = 1:3
+	% To each NaN voxel, we assign the border value in the original image
```
HenkMutsaerts, line 303: You forgot to use the NaNmask here right? Now you reset all voxels

External/SPMmodified/xASL/xASL_im_FillNaNs.m
```diff
@@ -267,184 +267,31 @@
...
-% Any remaining NaNs had to be inside, so we interpolate them (setting them
-% to zeros creates artifacts!)
-IM = xASL_im_ExtrapolateSmoothOverNaNs(IM, 0);
-
-end
+%% Extrapolation of values based on the edge values 
+% Calculate the distance map from the non-Nan towards NaN
+[~, distY, distX, distZ] = xASL_im_DistanceTransform(~maskNaN);
 
+% We calculate the meshgrid
```
HenkMutsaerts: Can you explain here in lay terms what a meshgrid is? ;) Shouldn't be too difficult.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
```diff
@@ -267,184 +267,31 @@
...
-% input
-% imThisCol - vector of current column (REQUIRED)
-% bSide     - boolean specifying if we check the start/first non-NaN voxel (0)
-% %           or the end/last non-NaN voxel (1). (REQUIRED)
-
-
-minInd = find(~isnan(imThisCol), 1, 'first'); % Get the index of the first non-value
-maxInd = find(~isnan(imThisCol), 1, 'last'); % Get the index of the last non-value
-midInd = floor(minInd+(maxInd-minInd)/2); % Get the index of the middle voxel
+% We calculate the coordinates of the border voxel
```
HenkMutsaerts: This removes the functionality of calculating a robust version of the border voxel?
What we had, is that we for each border voxel, we take the previous 4 voxels, create a robust difference of those and calculate what the border voxel would have been without too much non-linear behavior (effectively smoothing the last non-values) and then taking the mean between the actual border voxel and this smoothed/robust version of the border voxel. This doesn't replace the border voxel, but is the new value for the closest NaN.

So e.g., if the edge voxels are 

1) 30 31.5 33 34.5 39 NaN NaN NaN NaN

you don't want 

2) 30 31.5 33 34.5 39 40.5 42 43.5 45 46.5

but you want

3) 30 31.5 33 34.5 39 37.5 39 40.5 42

(or the average between (2) and (3)


Also, consider the case:

1) 30 21.5 43 24.5 25.5 NaN NaN NaN NaN

Then the last voxel doesn't seem to be too different from the previous one, but there is just too much non-linear going on on the border. So to now the "starting value" to start extrapolating from, we need to have a smoothed version of the flowfield, is what I'm saying. Maybe that is something that is easy to fix for you?

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: Good point, but only party true. Inside cat, it is XxYxZx3, so we need both options, I've fixed the code to work with this.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: That's just a misunderstanding:
- CAT12 can indeed have different voxel-sizes and we don't have the NII.
- So that's why I estimate the voxel-size automatically as we agreed by looking at the deformation field.
- But to make this a bit transparent, I save the estimated voxel-size to the VoxelSize field.
- But in theory, you can override the VoxelSize with your own value.

So it works like we agreed, with an added possibility to add your own number, which is practically not used in our code. I can of course create an extra parameter for the estimated voxel-size if you want, but I found this way more transparent and simple to read.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: See above - it will estimate the Voxel-size automatically from the deformation fields. So it will work as intended.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: I can remove this line if you want. But this is not my idea, this line is from **your code**, I just moved it here from another location, because in my version, I need to do this before running the subfunction.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: Typo fixed.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: See above - both options are possible. I have modified this and it now works for both.

jan-petr: Duplicate to the other comment.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: Yes. And I did :) I've later told you that cosmetic updates are needed to update your method from 3D to work with the three fields at once :)

I need to keep the voxel here, as this is input parameter handling. So that's why I do a basic check here, and the real estimation later. But I can remove it from here completely if you want.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: Exactly - in theory, it should never happen that we have a Nan for AP but not for IS - cause al 3 coordinates are needed. So this should always work, or the deformation field is invalid. 

So we can resolve this?

External/SPMmodified/xASL/xASL_im_FillNaNs.m
```diff
@@ -185,52 +178,81 @@
...
+	end
 end
 
-
 %% ========================================================================================
 %% ========================================================================================
 %% METHOD 3
-function [IM] = xASL_im_ExtrapolateLinearlyOverNaNs(IM, nEdgeVoxels)
+function [IM] = xASL_im_ExtrapolateLinearlyOverNaNs(IM, nEdgeVoxels, VoxelSize)
 %xASL_im_ExtrapolateLinearlyOverNaNs
```
jan-petr, line 207: nEdgeVoxels are now not used at all - can I remove this parameter?

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: Good point. Fixed.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: Done

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: Good point. As commented above, this subfunction now runs with 4D = XxYxZx3
maskNaN and distMapInward are 3D.
IM is 4D.
deltaIm is 3D. 

So all should be OK.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: Not sure why tricky. As said above - all NaNs from the 3 directions are excluded. So we keep your previous functionality without any issues.

External/SPMmodified/xASL/xASL_im_DistanceTransform.m
jan-petr: The Mex files was returning YxXxZ previously but I haven't noticed as this wasn'y really used anyway. But I have now fixed it in the distanceTransform wrapper (as these coordinates were not used anywwhere, I was able to do this). And I've also fixed it in FillNaNs function. So all should be good now.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: Fixed. See above.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: Done.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: I don't understand - I just kept what you had till line 250 to discard the weird voxel values. 

For 0 31.5 33 34.5 39 NaN NaN NaN NaN
Your code will discard 39 and we'll have
0 31.5 33 34.5 36 37.5 39 - so exactly what you want.

30 21.5 43 24.5 25.5 NaN NaN NaN NaN
Your code will discard a border of 4 and give
30 31.5 33 34.5 

So I'm still using your code and our agreed solution, so I don't understand where the issue is.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: Haha - no I didn't ;) That's the beauty of the solution. For NaNs, we will reassign the new value. For non-NaNs, distX,Y,Z is zero, so we reassign the map with the exactly same value.

So this is exactly what we functionally want and it's much faster and more elegant.
I added a comment in the code.

But a good comment, because I have noticed that the coordinate update below should be based on the relative coordinates and not on the absolute coordinate ;) This is now fixed.

jan-petr (changes requested):
See my answers.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
HenkMutsaerts: There is a maskNaN which says where new values should go. But we also need to know what the "edge or border value" is that we want to extrapolate from. This is not always the edge voxel as in the beginning, it could be that we want to take a few layers deeper, or that we smooth the final few voxels with values for a robust edge value.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
```diff
@@ -291,26 +288,27 @@
...
-[~, distY, distX, distZ] = xASL_im_DistanceTransform(~maskNaN);
+% Calculate the relative distance map from the non-Nan towards NaN. Note that for non-NaNs, these relative coordinates are zero
+[~, distX, distY, distZ] = xASL_im_DistanceTransform(~maskNaN);
 
-% We calculate the meshgrid
+% We calculate the meshgrid - three 3D matrices that give, respectively, for each voxel its X, Y, and Z coordinates.
 [gridX, gridY, gridZ] = ndgrid(1:size(IM, 1), 1:size(IM, 2), 1:size(IM, 3));
 
-% We calculate the coordinates of the border voxel
+% We calculate the coordinates of the border voxel, note that for non-NaNs this gives the original voxel location
```
HenkMutsaerts, line 297: You can slightly improve this for understanding, e.g. it is not "borderX" but "newValuesX" right? (the border is at the border of the non-nans but the non-nans could be halfway the image).

gridX == coordinateX?
distX = distanceX? % 0 for non-nans

External/SPMmodified/xASL/xASL_im_FillNaNs.m
HenkMutsaerts: OK but then I don't understand. gridX gives the original coordinate of the matrix, not the original values of the non-nans right? And we need those to stay the original values of the non-nans (but perhaps that doesn't matter because you only change the values inside the MaskNan below)

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: 1. That's what I'm saying - this edge voxels are updated of course based on your exclusion.
After running your code, we update the edge-voxel border on line 296, so we ignore those problematic voxels and extrapolate based on the new edge you have created a few layers deeper.

BTW: Can you indicate - in your example file that you've sent - which are the problematic voxels? I have excluded them based on your code, but I couldnt' really locate any problematic voxels.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
```diff
@@ -49,10 +49,9 @@
 % Copyright (C) 2015-2024 ExploreASL
 
 %% Admin
-IM = NaN; % default
-
 if nargin<1 || isempty(InputPath)
     warning('Illegal input path, skipping');
+	IMout = NaN;
```
HenkMutsaerts: You removed the default of IM = NaN, and moved it to the clause of nargin<1 only. So if there are any other errors in the code, this script not only gives a warning but also crashes because IMout is not assigned.

That was the reason to default it to NaN before anything else. Any (unforeseen) warning or issue that might occur, will issue a warning and the code will continue.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: 1. We first calculate the border coordinates (borderX)
2. Then we take the border value and set each voxel to the border value
3. Then we update the value by the distance from border

So borderX is correct, it's not newValuesX
gridX can be indeed renamed to coordinateX
distX can be distanceX % 0 for non-NaNs

External/SPMmodified/xASL/xASL_im_FillNaNs.m
```diff
@@ -228,22 +228,27 @@
...
-	diffIM = diffIM(~maskNaN(:, 2:end, :));
-	VoxelSize(2) = xASL_stat_MeanNan(diffIM);
+diffIM = IM(:, :, 2:end, 3)-IM(:, :, 1:end-1, 3);
+diffIM = diffIM(~maskNaN(:, :, 2:end));
+VoxelSizeEst(3) = xASL_stat_MeanNan(diffIM);
 
-	diffIM = IM(:, :, 2:end, 3)-IM(:, :, 1:end-1, 3);
-	diffIM = diffIM(~maskNaN(:, :, 2:end));
-	VoxelSize(3) = xASL_stat_MeanNan(diffIM);
+if isempty(VoxelSize)
```
HenkMutsaerts: No, we never use the VoxelSize. I would remove the VoxelSize option fully. Because we never used it in the pipeline, and to stay consistent, I would also not start using it.

So I would remove the VoxelSize input for METHOD 3 fully.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
```diff
@@ -305,14 +310,8 @@
 end
 
 % We now subtract the voxel-size adjusted relative coordinate difference from these newly filled voxels
-IM(:, :, :, 1) = IM(:, :, :, 1) + distX(:, :, :)*VoxelSize(1);
-IM(:, :, :, 2) = IM(:, :, :, 2) + distY(:, :, :)*VoxelSize(2);
-IM(:, :, :, 3) = IM(:, :, :, 3) + distZ(:, :, :)*VoxelSize(3);
-
-%% Smoothing of voxels
```
HenkMutsaerts: Why did you remove this? Any residual NaNs inside the flowfield will corrupt the processing. So we need to fill them. We could issue a warning if you feel this shouldn't be the case.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: GridX gives the coordinate for each voxel (in the matrix notation). Each voxel has it's own coordinate in gridX. It's the position within the matrix, not the world coordinates.
GridX + distanceX gives the coordinate (in the matrix space, not in world coordinates) of the closest border voxel for each voxel. So non-NaNs get their own coordinates. This is still all about position in the matrix without looking at the transformation values.

Only now I start working with the values in the IM which are the SPM-like transformation coordinates.
I update the entire IM.  But updating IM at location gridX with values of IM at position gridX+distanceX will give border value to NaNs and will do nothing to non-NaNs. This is perfectly safe, secure, does what it should exactly and it's computationally much faster. So no reason to update only MaskNan, it's better to update all.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: All unforseen errors give the correct output with my solution. Functionality is perfectly fine... The reason was that it wouldn't work with the dimension swap with an initialized IMout.

But if you really insist, I can change it back and think of another solution.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: But the option is there, so why we have option that we are not using - that's very inconsistent.

But I can of course always automatically remove it.

External/SPMmodified/xASL/xASL_im_DistanceTransform.m
HenkMutsaerts: OK but then it should be X,Y,Z now right? Because the comment says Y,X,Z

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: That was valid with our old code that left NaNs inside certain structures. With our new code, all NaNs are already removed, so it makes no sense to keep filling NaNs because there can't possibly be any left.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
HenkMutsaerts: OK

BTW: Then it should not exclude anything right, if there are no problematic voxels? The problematic voxels are there if we e.g. run ExploreASL on low quality and the flowfields are created at [3x3x3] and resampled to [1.5x1.5x1.5]. So it should be a failsafe only.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
HenkMutsaerts: How? It still says Y X Z here

External/SPMmodified/xASL/xASL_im_FillNaNs.m
HenkMutsaerts: No, in the second case, my code will not discard the border of 4 right (at least it should not)?

There are two challenges:
1. Which voxels should be replaced, in addition to the NaNs?
2. What should we use as new values for the NaN?

We fixed challenge 1, but not challenge 2. You now just take the edge value and keep adding 1.5 right?

I'm talking about the whole flowfields being correct, and at the edge there is some high non-linear behavior at the border of the skull. There, we don't want to use a more robust value to extrapolate.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
HenkMutsaerts: How I read it (please correct me where I'm wrong so I understand, I trust that you're doing it right, just want to understand :)

1. GridX gives the coordinate for each voxel (which is wrong for non-NaNs because they should not have their own coordinates but their original non-linearly created values).
-> this is probably something I don't understand correctly?

2. We don't want to give border-values to non-NaNs, these need to be the smoothed border-values (more robust, less non-linear)
-> this is something I think you should add

External/SPMmodified/xASL/xASL_im_FillNaNs.m
HenkMutsaerts: OK so:

CoordinateBorderX
CoordinateOriginalX
distanceX

? Nice if we can change this to make it more readable (not very important)

External/SPMmodified/xASL/xASL_im_FillNaNs.m
HenkMutsaerts: No the option is there only for METHOD 1, to calculate the smoothing kernel only. Not for METHOD 3, and I would not use it to stay consistent, to always do the same.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
HenkMutsaerts: Yes

External/SPMmodified/xASL/xASL_im_FillNaNs.m
HenkMutsaerts: I know. I mean that now, this is run for all METHODS. Whereas we only want it for METHOD 3. Otherwise I'm fine.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
HenkMutsaerts: Just remove VoxelSize as input parameter to METHOD 3 and keep it for METHOD 1, would be my preference.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
HenkMutsaerts: I would just remove the VoxelSize because it confuses, it was only the input to METHOD 1 (for a very different reason). We can always think of all options that we generate inside code and add it as an input parameter. Your argument is that it was there for METHOD 1 already. But that false, it was there for a very different reason (for keeping a smoothing kernel relevant to the voxel size). So to me it overcomplicates the fix.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
HenkMutsaerts: OK let's keep it like this then

External/SPMmodified/xASL/xASL_im_FillNaNs.m
HenkMutsaerts: Nice

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: Done

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: Done

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: I added it for Method 3 only. But it has to be on this line, before checking for presence of NaNs.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: Yes, it should not exclude anything, but it's you who wrote this function ;) With the current dataset you've sent, it excludes something like 1-voxel layer. But I don't see any wrong voxels. So that's why I'm asking if it's really necessary and if we have an example where this helps.

External/SPMmodified/xASL/xASL_im_DistanceTransform.m
jan-petr: This comment is of course shows the state from when you made the comment. It's not updating with the later commits.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: See above.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: I've updated my last comment. I thought that this is clear because of the way gridX is created...

GridX is using the coordinates within the transformation matrix. So 1,2,3,4
Values of IM are the coordinates of the transformation in real world space. So 0.5mm, 1mm, 1.5mm etc.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: I know you talk about highly-nonlinear behavior on the border and my answers always talked about this. I just don't see any of the non-linearities in the example you sent. I need to see in which slices they are.

Your code discards a border (something like thickness 1 voxel in the Nexample you sent for y_ASL.nii). I use you code and discard exactly the same voxels your code discards.

I of course provide new values in IM for all NaNs that were originally NaNs, and also for all border voxels discarded by you. And of course, we ignore all discarded border voxels and don't use them for extrapolation.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: Done

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: Done

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: Done

External/SPMmodified/xASL/xASL_im_FillNaNs.m
HenkMutsaerts: The reason for that was that any NaNs inside the flowfield (e.g. the center of the flowfield) potentially need a different treatment than NaNs at the border of the flowfield. So it was on purpose that the old code "left some NaNs".

Can you confirm what happens with NaNs inside (e.g. the center) of the flowfield in your code?

External/SPMmodified/xASL/xASL_im_FillNaNs.m
HenkMutsaerts: But that it what I didn't understand. IM should not be 0.5mm 1mm 1.5mm, it should only be that for the voxels in NaNmask. The voxels outside NaNmask should stay the same as they were (the non-linear flowfield), not being replaced by "identity". Perhaps I just didn't understand your code enough. I guess you can resolve and try to explain me in the Whatsapp conversation :)

External/SPMmodified/xASL/xASL_im_FillNaNs.m
HenkMutsaerts: I know I wrote this function, which I think should still work but for the exclusion threshold (median+X*MAD), the X might need tweaking? I will send you some other examples.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
HenkMutsaerts: Then just move it 1 line before checking for the presence of NaNs, either inside METHOD 3 or just outside it.

It's probably not very important, but we are not testing the use cases of methods 1 and 2 so they should not change, and now all methods get this change because you put it in the general image loading section.

External/SPMmodified/xASL/xASL_im_DistanceTransform.m
jan-petr: It's fixed everywhere. Distance function outputs it correctly now X,Y,Z. You can resolve.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: I wasn't talking about the exact values. I was talking about the general units in GridX (voxel coordinates) and IM (real world transformation in mm). But IM of course isn't really nicely linearly increasing as it contains the nonlinear transform...

Voxels in IM outside maskNaN of course stay exactly the same and are not modified at all. Except for you excluding your margin. We kind of "touch" all values in IM, but we do it in a very smart way so that outside maskNaN voxels in IM are touched, but their values remain. Because IM value at ~maskNaN is replaced by itself. That allows to keep it fast and elegant the code.

You can resolve.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: NaNs inside and NaNs outside receive the exact same treatment. We look at nonNaNs, remove unsafe border, extrapolate NaNs linearly with values from the closest safe border.

Previously, with the complicated code, we were scanning each line separately going from image edges from both sides to the border. So that way, we didn't get into the central NaNs. The new solution is much easier and faster and simply looks for the closest border and works with that.

So with a much faster and simpler code, we deal the same way with all NaNs. And the nice thing is that previously, we scanned everything linewise, so the closest border was identified across line only and could lead to weird artifacts. Now, we look for the closest 3D Euclidean distance, so this hsould work much much better.

This was the main reason why I wanted to discard the old line-wise code and replace it.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: I'm fine with the function in general. I was just saying that I can't see issues, so that's why I can't properly debug and tweak. So no problem with the general functionality.

I'll check your examples. Please test them as well, you might find out that weird values are not so big problem with the new NaN-replacement funciton...

External/SPMmodified/xASL/xASL_im_FillNaNs.m
jan-petr: That's the problem. Before checking if we have method 1/2/3, we need to check if we have NaNs or not at all. So the INF->NAN has to happen before that. Or else we first check there are no NaNs, so decide to skip the filling, but leave INFs. 

Please check the new code how it looks after all changes -> it will be really clear to you then.

HenkMutsaerts: OK but then there are two issues:

1. We change the behavior of method 1 & method 2 which are not tested (we test method 3 only). Chances are small that somebody wants to keep Inf's in, but this could happen. E.g., if you use the old M0 method, there are Inf's at the border which you now replace by NaNs. Could even be good. But formally we need to test it, because method 1 & 2 are used in ExploreASL at other locations as method 3.
2. We clearly describe in the function header that not only NaNs but all non-real values are replaced.
3. We can avoid [1] by adding a warning if non-real non-nan voxels were found, and will be removed. Also fine with me.

jan-petr: Henk - please check new code before answering. INF->NAN is applied beforehand, but only for METHOD 3 - 3 comments ago :)

1. Functionality of 1&2 is not affected. This is only applied selectively for METHOD 3
2. OK - I describe this for method in a header now.

External/SPMmodified/xASL/xASL_im_FillNaNs.m
```diff
@@ -233,43 +233,46 @@
...
-	IMsmooth(:, :, :, iDim) = xASL_im_ndnanfilter(IM(:, :, :, iDim), 'gauss', [8 8 8], 1);
+% We also need to remove the borders if we want a proper distance also from the image edges
+distMapInward = maskNaN;
+distMapInward(1,:,:) = 1; distMapInward(end,:,:) = 1;
+distMapInward(:,1,:) = 1; distMapInward(:,end,:) = 1;
+distMapInward(:,:,1) = 1; distMapInward(:,:,end) = 1;
+distMapInward = xASL_im_DistanceTransform(distMapInward);
+
+% We exclude several border voxels - calculated as around 7% of the image size
+maskRestricted = distMapInward > mean(size(IM, 1:3))*0.07;
```
HenkMutsaerts, line 244: Add a comment that this is 10 voxels for [121 145 121]

External/SPMmodified/xASL/xASL_im_FillNaNs.m
```diff
@@ -233,43 +233,46 @@
...
-medianRelativeDelta = xASL_stat_MedianNan(relativeDelta(:));
-MADRelativeDelta = xASL_stat_MadNan(relativeDelta(:));
-thresholdIs = medianRelativeDelta + 3.5*MADRelativeDelta;
-
-% We want to have a hard cutoff of distance from NaNs that we won't change
-% This is 10% of the transformation field size (e.g, for 121*145*121 transformation field, this is 13 voxels)
-sizeIm = size(IM);
-if numel(sizeIm)<3
-    warning('Something wrong with this flowfield');
+% Perorm the final interpolation
```
HenkMutsaerts, line 263: Perorm -> Perform

jan-petr (approved):
All fixed and tested.

PR #1739, opened 2024-08-08
#1738 Fix M0 import report error

### LINKED ISSUE #1738: M0 import reporting issue
### Description

Missing revM0-PA error is reported even when the file is missing. 

The reason is that we are defining the parameters in all cases as we need this for all the ASL-M0 dependencies. But we obviously don't need to report errors for files that are not there.

### Release notes

Fix errors in error-printing in M0 import.

jan-petr:
### Linked issue

Close #1738


HenkMutsaerts (approved):
Nice!

PR #1741, opened 2024-08-09
 #1740 Add the supratentorial mask for population analysis

### LINKED ISSUE #1740: Allow use of brainmask_supratentorial
### Description

Provide the possibility to evaluate ROI in the whole brain without the cerebellum using the brainmask_supratentorial.

### Tasks
- [x] Adapt brainmask_supratentorial such that it has the same definition as totalGM.nii
- [x] Create .nii.mat sidecar
- [x] Rename to Supratentorial_GM_WM
- [x] Create bSubjectTissueMask with [GM WM WB] (we could add other tissue types like CSF or WMH that may have certain use cases, or do this later). Fixed by #1753 as x.S.TissueMasking
- [x] Add the atlas as dataPar.json option

Edited the supratentorial atlas:

image
Added this to the README. The code can be found in CustomScripts.

When merging, check the atlas/ROI moving in #900

### Release notes

Duplicate, nothing to report.

jan-petr:
### Linked issue

Close #1740


jan-petr:
Sorry, I previously forgot to create the PR, so you must have missed many things. This is copy-pasted full comment from the Issue.

Q: Do you want to have the WholeBrain mask, but then for supratentorial only?
A: Yes

Q: So this is for the total GM? Then I would call it totalGM_supratentorial.
A: Yes. OK, shall I keep the file brainmask_supratentorial and just rename the atlas tag?

Q: And we need to have the same brainmask as total GM (the supratentorial mask is a bit narrower).
A: Yes - that's why I've been asking if we have that mask and thought that brainmask_supratentorial is fine. So can we keep this or we create a new mask?

Q: If this is then only for total GM ROIs supratentorial, this is fine. And for smaller ROIs or WM this is not needed by design.
A: Yes, exactly for this, so I would keep the branmask_supratentorial and only change the atlas tag.

Q: Note BTW that the cerebellum is already excluded for sequences that have no information there. E.g., for most 2D EPI sequences it is included for the top only, and the 3D sequences do include it. It is a good idea to exclude the cerebellum though for total GM.
A: Yes, aware of if  - so I had the exact motivation to get rid of it for TotalGM if possible and asked for -> so to have TotalGM_Supratentorial.

Q: One other thing that we need to do is add a .nii.mat sidecar, which ExploreASL can use without having to unzip (keeping the repository unchanged).
A: Yes, I've done this technically already -> see the PR.

HENK ANSWERED (I can do this):
OK, so to do:
1. Adapt brainmask_supratentorial such that it has the same definition as totalGM.nii
YES!

2. rename to totalGM_supratentorial.nii
YES. And we want also totalWM_supratentorial and WB_supratentorial.

3. Create .nii.mat sidecar (with the same code as Henk did previously, it is in the WorkInProgress folder/InputOutput)
YES. Obviously:)

4. Check with the atlas change issue for the license
YES, of course :)

5. E.g., we might want to automate in xASL_adm_FileSystem for adding all files in the atlas folder that have .nii.mat as atlas.
I would be careful. Adding new atlases is easy in the code, so it needs no automation. But if you want to do something nice for the user, then you can add a special directory for user atlases. This directory will then be checked upon running and atlases correctly imported and used. So then a user can add his custom atlases easily anytime. But we are not responsible for any issues as this will be a clear responsibility of the user to do this right. If you want, we can make a new issue about this.

HENK:
Ad 2. Yes I agree with you actually to just make a bSubjectTissueMask with [GM WM WB] (we could add other tissue types like CSF or WMH that may have certain use cases, or do this later). I would move this to our xASL_stat_GetROIstatistics revamp though, or I can do it now, it is not a big change.
JAN: Ideally now. We need this.

Ad 5: I edit this atlas management in the license issue anyway, so can add it there. It is already there, no need to make a new issue. For now we can just add it manually as you did indeed.
JAN: Agree.

PR #1748, opened 2024-08-20
Closes #1746_ExtraFolderLayers

### LINKED ISSUE #1746: DICOM to Nifti conversion can't handle difference in depth of folder structure
### Description

Datasets may contain a variety of layers where the DICOMS can be found, at the 'scan' layer level token of the sourcestructure.json.  Currently, ExploreASL is not able to handle this. A solution would be to identify the level as depicted by the `sourcestructure.json`, and if no dicoms or zip's are found to continue searching for in a deeper folder (with a maximum).

### How to test

A test subjects is available in `~/r-divi/RNG/Projects/ExploreASL/Testing/Dicomconversionfolderdepth` 

Here, the same subject but with the deeper folder layer (OneLayerDeeper_1), which causes the issue, and at a working layer (NormalDepth_1) can be found in the structure that fits other patients that gave the error. 

### Release notes
Allow dummy folders in path during import.

HenkMutsaerts:
### Linked issue

Closes #1746

### How to test

See issue, for Mathijs to test (I tested only the changed code and conversion) and for Jan to review the code 

### Comments

This fixes the issue. It could generate a bug if there are multiple subfolders containing DICOMs, but that is already a bug, and should be fixed by changing the sourceStructure.json. So essentially this seems fine for now.

### DIFF
--- a/Modules/Module_Import/DCM2NII/xASL_imp_DCM2NII_Subject_StartConversion.m
+++ b/Modules/Module_Import/DCM2NII/xASL_imp_DCM2NII_Subject_StartConversion.m
@@ -62,6 +62,16 @@ function [globalCounts, x, summary_line, destdir, scanpath, scan_name, dcm2niiCa
         else %% Convert DICOM files
             %% Start the conversion. Note that the dicom filter is only in effect when a directory is specified as input.
             try
+                % First we try to see if there are deeper layers
+                [fPath, ~, fExt] = xASL_fileparts(scanpath);
+                if isempty(fExt) % run this part only for folders
+                    filepathsNonRecursive = xASL_adm_GetFileList(fPath, x.modules.import.imPar.dcmExtFilter, 'FPList');
+                    filepathsRecursive = xASL_adm_GetFileList(fPath, x.modules.import.imPar.dcmExtFilter, 'FPListRec');
+                    if isempty(filepathsNonRecursive) && ~isempty(filepathsRecursive)
+                        scanpath = fileparts(filepathsRecursive{1});
+                    end
+                end
+                
                 [nii_files, scan_name, first_match, MsgDcm2nii] = xASL_io_dcm2nii(scanpath, destdir, scan_name, x.modules.import.imPar, x.opts.MyPath);
 
                 % If dcm2nii produced a warning or error, catch this & store it
--- a/Modules/Module_Import/DCM2NII/xASL_imp_DCM2NII_Subject_StartConversion.m
+++ b/Modules/Module_Import/DCM2NII/xASL_imp_DCM2NII_Subject_StartConversion.m
@@ -25,12 +25,11 @@ function [globalCounts, x, summary_line, destdir, scanpath, scan_name, dcm2niiCa
 % EXAMPLE:     n/a
 %
 % __________________________________
-% Copyright 2015-2023 ExploreASL
+% Copyright 2015-2024 ExploreASL
 % Licensed under Apache 2.0, see permissions and limitations at
 % https://github.com/ExploreASL/ExploreASL/blob/main/LICENSE
 % you may only use this file in compliance with the License.
 % __________________________________
-
 
     %% Start the conversion if this scan should not be skipped
     first_match = [];
@@ -69,19 +68,16 @@ function [globalCounts, x, summary_line, destdir, scanpath, scan_name, dcm2niiCa
                     if isempty(filepathsNonRecursive) && ~isempty(filepathsRecursive) 
 						% There are no files in the directory but there are files in the subdirectories
 						for iFile = 1:length(filepathsRecursive)
-							% Remove the filename and keep only the path
+							% Remove the filename and keep only the path to the folder
 							filepathsRecursive{iFile} = xASL_fileparts(filepathsRecursive{iFile});
 						end
-						% Find the unique paths
+						% Find the unique folders
 						filepathsRecursiveUnique = unique(filepathsRecursive);
-						if isempty(filepathsRecursiveUnique)
-							% No files found recursively -> error
-							error(['No files found in ' scanpath]);
-						elseif length(filepathsRecursiveUnique) > 1
+						if length(filepathsRecursiveUnique) > 1
 							% Files at multiple directory levels -> error
-							error(['Multiple directory levels with files were found in ' scanpath]);
+							error(['Multiple subdirectories with (DICOM) files were found in ' scanpath]);
 						else
-							% Only a single directory level with files found - use that one
+							% Only a single subdirectory with files found - use that one
 							scanpath = filepathsRecursiveUnique{1};
 						end
                     end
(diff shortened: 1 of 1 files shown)


jan-petr (changes requested):
Several things are need improvement:
1. You are checking if the path is a folder, but currently, you might mistake a file without extension for a folder (DICOM files often do not have an extension).
2. You removed the last dir (examples/sub1/ASL - fPath was only examples/sub1), so the recursive listing gave you all sequences (ASL, T1 etc) or rather randomly the first one. So I changed that to recursively look into the directory without removing the extra layer.
3. Last issue is that when files are at multiple levels (which they shouldn't) then the program won't notice this and will select a random directory. We either want to report this as an error (I did that now). Or worst case select the highest directory level and let dcm2nii read all layers.

I have fixed all these issues. So please only check my newest commit and I'll then approve everything and merge it. I've created a special example for this in the Flavors, so it gets tested everytime and will run all Flavors on it as well.

Modules/Module_Import/Import/xASL_imp_DCM2NII_Subject_StartConversion.m
```diff
@@ -58,12 +58,27 @@
...
             try
                 % First we try to see if there are deeper layers
-                [fPath, ~, fExt] = xASL_fileparts(scanpath);
-                if isempty(fExt) % run this part only for folders
-                    filepathsNonRecursive = xASL_adm_GetFileList(fPath, x.modules.import.imPar.dcmExtFilter, 'FPList');
-                    filepathsRecursive = xASL_adm_GetFileList(fPath, x.modules.import.imPar.dcmExtFilter, 'FPListRec');
-                    if isempty(filepathsNonRecursive) && ~isempty(filepathsRecursive)
-                        scanpath = fileparts(filepathsRecursive{1});
+                if xASL_exist(scanpath, 'dir') == 7 % run this part only for folders
+                    filepathsNonRecursive = xASL_adm_GetFileList(scanpath, x.modules.import.imPar.dcmExtFilter, 'FPList'); % List files directly in scanpath
```
HenkMutsaerts: Will this also find files without extensions? Same for below (I know I had this as well :) just asking, as DICOM files can be extension free.

Also, this way, xASL_adm_GetFileList defaults to NOT reporting subfolders, so that should be OK

Modules/Module_Import/Import/xASL_imp_DCM2NII_Subject_StartConversion.m
```diff
@@ -58,12 +58,27 @@
...
+                    filepathsRecursive = xASL_adm_GetFileList(scanpath, x.modules.import.imPar.dcmExtFilter, 'FPListRec'); % List all files including subdirectories
+                    if isempty(filepathsNonRecursive) && ~isempty(filepathsRecursive) 
+						% There are no files in the directory but there are files in the subdirectories
+						for iFile = 1:length(filepathsRecursive)
+							% Remove the filename and keep only the path
+							filepathsRecursive{iFile} = xASL_fileparts(filepathsRecursive{iFile});
+						end
+						% Find the unique paths
```
HenkMutsaerts: path can be folder+file, so you mean folder here instead of path right (in the comments)

Modules/Module_Import/Import/xASL_imp_DCM2NII_Subject_StartConversion.m
```diff
@@ -58,12 +58,27 @@
...
+						filepathsRecursiveUnique = unique(filepathsRecursive);
+						if isempty(filepathsRecursiveUnique)
```
HenkMutsaerts: This is not possible right? Following logically from above.
As you search for files only (xASL_adm_GetFileList), and then unique their folders, which should always be >0.

Also, your comment says "no files found" which should be "no folders found" as you before this command took their folders from the full path name.

Modules/Module_Import/Import/xASL_imp_DCM2NII_Subject_StartConversion.m
```diff
@@ -58,12 +58,27 @@
...
+							% No files found recursively -> error
+							error(['No files found in ' scanpath]);
+						elseif length(filepathsRecursiveUnique) > 1
```
HenkMutsaerts: What happens if a user has an ASL sequence, where the user has sorted on SeriesDescription, which in e.g. Siemens results in different folders for control label etc. And the user wants dicom2nii to concatenate them in a single NII, that is probably not something that will work right? Just asking

Modules/Module_Import/Import/xASL_imp_DCM2NII_Subject_StartConversion.m
```diff
@@ -58,12 +58,27 @@
...
+							% Files at multiple directory levels -> error
+							error(['Multiple directory levels with files were found in ' scanpath]);
+						else
+							% Only a single directory level with files found - use that one
```
HenkMutsaerts: You're saying "single directory level" but it can be that you find a "single directory level" but on this same "level" you have multiple directories. Just to have a correct error message and comment, for the user to understand.

E.g.:

/dicomFolderTokenSpecified/ExtraDirectoryLevel/ControlSiemens/dicoms.dcm
/dicomFolderTokenSpecified/ExtraDirectoryLevel/LabelSiemens/dicoms.dcm

so the error message should be:

error(['Multiple subdirectories with (DICOM) files were found in ' scanpath]);

and the comments should be like:
% Only a single subdirectory with files found - use that one

so call them "subdirectories" instead of "directory levels" because you don't discriminate "horizontal vs vertical" levels if you understand what I mean

HenkMutsaerts (commented):
Few minor points, generally I agree

Modules/Module_Import/Import/xASL_imp_DCM2NII_Subject_StartConversion.m
jan-petr: Yes - the way how DICOM extensions work isn't affected. So it should still keep finding DICOMs without an extension.

Modules/Module_Import/Import/xASL_imp_DCM2NII_Subject_StartConversion.m
jan-petr: Yes. Fixed

Modules/Module_Import/Import/xASL_imp_DCM2NII_Subject_StartConversion.m
jan-petr: True. Nice catch. If I check for emptiness before unique, it really shouldn't be empty a few lines below :)
Removed...

Modules/Module_Import/Import/xASL_imp_DCM2NII_Subject_StartConversion.m
jan-petr: Very good point. And no. This wasn't working previously and isn't working now either. Single folder with files. We would have to rework everything to make it work with multiple folders. Same for e.g. PLDs.

Modules/Module_Import/Import/xASL_imp_DCM2NII_Subject_StartConversion.m
jan-petr: Thanks. It was indeed tricky to nicely describe this.

Code changed accordingly.

jan-petr (approved):
OK

PR #1752, opened 2024-08-22
#1751 stat_GetDICOMStatistics Bug fix

### LINKED ISSUE #1751: Unit test fail
### Description

Unit tests fails at several places as a result of previous untested merges of #1717 

### Tasks

- [x] ExploreASL unittest fails in last 4 subtests. The error is in TSV writing in stat_GetDICOMStatistics. Sometimes, DICOM tag can be composed of multiple fields, which then turns into a cell and crashes TSV writing. So we need to turn cells into a composed string fe8e7eef
- [x] BIDS2Legacy unittest fails -> this was related to the error above. No further fix needed. 

### Release notes

Report together with #1717 as it fixes error in that issue.

jan-petr:
### Linked issue

Close #1751

### DIFF
--- a/Modules/Module_Processing/Module_Population/xASL_stat_GetDICOMStatistics.m
+++ b/Modules/Module_Processing/Module_Population/xASL_stat_GetDICOMStatistics.m
@@ -92,9 +92,13 @@ for iSubject=1:x.dataset.nSubjects
 end
 fprintf('\n');
 
-
+% Verify that all gathered cells include strings and not cells inside
+listCells = find(cellfun(@iscell, TSV(:)));
+for indexCells = 1:numel(listCells)
+	% Replace all cells by a string created by concatenating the strings inside
+	TSV{listCells(indexCells)} = strjoin(TSV{listCells(indexCells)}, '/');
+end
 %% 2. Write TSV file
 xASL_tsvWrite(TSV, PathTSV, bOverwrite);
-
 
 end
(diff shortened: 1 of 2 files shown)


Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
```diff
@@ -538,7 +538,7 @@
 			DataIm(:) = NaN;
 		end
 		%% 4.d Show ROIs projected on ASL image (%% PM %% : ALSO DONE BELOW)
-		LabelIM = xASL_vis_Convert4D_3D_atlas(xASL_im_Column2IM(SubjectSpecificMasks(:,[1:3:end]), x.S.masks.WBmask));
+		LabelIM = xASL_vis_Convert4D_3D_atlas(xASL_im_Column2IM(SubjectSpecificMasks(:, 1:3:end), x.S.masks.WBmask));
```
HenkMutsaerts, line 541: This will create a conflict with another recent issue, so because it doesn't seem necessary, I would revert this one. But your decision, you will probably be the one resolving the conflict :)

Modules/Module_Processing/Module_Population/xASL_stat_GetDICOMStatistics.m
```diff
@@ -92,9 +92,13 @@ function xASL_stat_GetDICOMStatistics(x, ScanType, HasSessions, bOverwrite)
 end
 fprintf('\n');
 
-
+% Verify that all gathered cells include strings and not cells inside
+listCells = find(cellfun(@iscell, TSV(:)));
```
HenkMutsaerts, line 96: I thought I fixed this with the xASL_num2str or something. But if this fix by you works and you tested it, I'm fine.

HenkMutsaerts (approved):
Approved with 2 questions, up to you

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
jan-petr: Haha - no worries - that's a conflict I can handle :)

Issue #1314, opened 2023-02-05
Reference atlases, Mindboggle, xASL_stat_GetROIstats


HenkMutsaerts:
### Description

Adds 3 population-based atlases, based on the ExploreASL paper.
Registers the MindBoggle atlas.
Adds bWMH boolean for (de-)activating WMH exclusion from WM mask
And a few minor fixes.

### Tasks

- [x] Add SD maps to ExploreASL for populations shown in ExploreASL paper (NOVICE, Sleep, EPAD) and add license (per ExploreASL paper). Start with EPAD only. The between-subject CoV doesn't appear smooth though, so it could be helpful to smooth individual images before creating the bs-CoV template; or to do this in a PVC-way.
- [x] Add license for these atlases
- [x] Register MindBoggle atlas to CAT12-IXI space
- [x] Fix warnings for xASL_spm_affine & xASL_vis_CreateVisualFig
- [x] Bugfixes xASL_stat_GetROIstatistics: ROI visualization
- [x] Improve printing which masks are used
- [x] Add bWMH boolean for (de-)activating WMH exclusion from WM mask
-> see #1745 for the explanation regarding the bWMH

HENK PM: if registering Mindboggle still needs improvement (I don't think so) we could try the following:
       Mindboggle is registered with ICBM152 (e.g., http://www.bic.mni.mcgill.ca/ServicesAtlases/ICBM152NLin2009)
       & we are registered with CAT12s slightly different MNI atlas.
       a. Solution is to affine transform Mindboggle to our CAT12 MNI atlas, using the templates from ICBM152 to 
       CAT12:
       mni_icbm152_nlin_sym_09c_nifti/mni_icbm152_nlin_sym_09c/mni_icbm152_(gm|wm|csf)_tal_nlin_sym_09c.nii
        /ExploreASL/External/SPMmodified/MapsAdded/rc1T1.nii (OR /SPMmodified/toolbox/CAT12/templates/last iteration

### Release notes

#1314 Adds 3 population-based atlases, based on the ExploreASL paper.
#1314 Registers the MindBoggle atlas.
#1314 Adds bWMH boolean for (de-)activating WMH exclusion from WM mask

PR #1754, opened 2024-08-23
Closes #1753_xASL_stat_GetROIstatistics_tissuetype

### LINKED ISSUE #1753: xASL_stat_GetROIstatistics: create tissue-type choice
### Description

Create bSubjectTissueMask with [GM WM WB] (we could add other tissue types like CSF or WMH that may have certain use cases, or do this later).

### Tasks

- [x] In the code, default x.S.TissueMasking at the same location where x.S.Atlases is defaulted, these should match. If TissueMasking is missing or non-matching (because a user uses an old dataPar.json) the user will be correctly notified and hinted on what should be the input.
HENK: Ah, reading your point below I understand what you mean by "backward compatibility", that the input settings are also still compatible. To me, "backward compatibility" is mostly "can you use partly processed data correctly with the new pipeline". Of course, new settings that are unspecified will lead to errors but those will be unknown to the user. I agree that it would be nice to provide an error that x.S.TissueMasking needs to be provided.
JAN: Yes :) The compatibility of partly processed data will be there as well as atlases are only dealt with in the Population module. So that's not an issue. I've updated the description above.
HENK: x.S.Atlases was defaulted at the start of xASL_module_Population, and I have now moved this together with defaulting x.S.TissueMasking to a subfunction of xASL_module_Population, including the user feedback that you propose. Done in c2a0eed625816dd344b8513734221a9985a6cda2.
- [x] x.S.TissueMasking, x.S.bTissueMasking is used inside the GetROI function.
- [x] Replace automatic GM-WM-WB recognition from atlas name to x.S.bTissueMasking
- [x] Add this to the Documentation. Jan can do this, Henk drafted an explanation:
ExploreASL's population module produces ROI values that are obtained by a combination of an MNI ROI atlas with subject-specific GM and WM segmentations. 
In 1.11.0 and earlier versions, xASL_module_Population->xASL_wrp_GetROIstatistics automatically decided if GM, WM, or wholebrain (meaning GM+WM) values should be extracted from an ROI, based on the name of the atlas. It would default to GM values, unless an atlas contained the name WM or WB (wholebrain).
From 1.12.0 onwards, the subject-specific tissue masking is specified by the user, using the parameter x.S.TissueMasking, with option 1=GM, 2=WM, 3=WB (GM+WM).
x.S.Atlases has not changed, but the parameter x.S.TissueMasking should be the same length as x.S.TissueMasking.
When not specified, these default to:
x.S.Atlases = {'TotalGM' 'DeepWM'};
x.S.TissueMasking = [1 2];
JAN: Isn't it better to use string instead of numbers for TissueMasking? Maybe the answer for backwards compatibility issue will be that an error will be reported if Atlases are defined and TissueMasking undefined. So any older version users, will be notified to add the TissueMasking vector. This might be the easiest solution. It might be better than trying to guess TissueMasking from Atlases when TissueMasking would be missing. And the error could sort of provide some nice hints on what to fill and where.
HENK: Changed to strings in 6ca435024689dde2b6495a806334d1fae5bd3be9.
PS: may change 1.12.0 to 2.0.0 with the new license.

### How to test

I tested this in xASL_module_Population and it works. Easiest to visualize test this is by enabling x.S.SubjectWiseVisualization and looking in the population stats Check... folder.

### Release notes

Separate the option for subject-wise tissue masking (x.S.TissueMasking) and ROI atlas selection (x.S.Atlases)

HenkMutsaerts:
### Linked issue

Closes #1753.

### Comments

Note that this is based on #1314, which needs merging first. CAREFULLY check the commit ids and potentially cherry-pick into a different branch!!!!!

### DIFF
--- a/Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
+++ b/Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
@@ -31,6 +31,19 @@ function [x] = xASL_stat_GetROIstatistics(x)
 %                       0 = [0 0 0 0]
 %                       Can be useful for e.g. loading lesion masks outside the GM
 %                       (OPTIONAL, DEFAULT=1)
+%   x.S.bWMH           - boolean specifying if we should mask a ROI with pvWMH (by subtracting pvWMH in ASL resolution)
+%                        true = include WMH (no masking), false = exclude WMH (OPTIONAL, DEFAULT=false)
+%   x.S.bTissueMasking - value specifying which subject-wise tissue type we will mask with. Cannot be a vector
+%                        (loop outside this function for multiple tissue types)
+%                        1 = gray matter (GM)
+%                        2 = white matter (WM)
+%                        3 = whole brain (WB = GM+WM)
+%                        4 = ...
+%                        (REQUIRED) ->>>>>>>>>>>>>>>>>>>>>>>>>>>, set DEFAULT outside this code as [1 2])
+%                                   ->>>>>>>>>>>>>>>>>>>>>>>>>>>
+%                        Note that a ROI needs to have a minimal volume. So a deepWM ROI for which pvGM is requested will be skipped.
+%
+%
 % OUTPUT:
 %   x                            - same as input
 %   x.S.SubjectSessionID 		 - %   x.S.SubjectSessionID       - names of the Subject/Sessions in the order of x.S.SetsID/x.S.DAT
@@ -596,22 +609,34 @@ for iSubject=1:x.dataset.nSubjects
 
 			% Skip PVC for certain ROI and mask combination as the result might not be meaningful
 			bSkipPVC = 0;
-			
+
+            % pGM_here is the main tissue type that is investigated
+            % pWM_here is the other tissue type that is used with PVC
+
+            %   x.S.bTissueMasking - value specifying which subject-wise tissue type we will mask with. Cannot be a vector
+            %                        (loop outside this function for multiple tissue types)
+            %                        1 = gray matter (GM)
+            %                        2 = white matter (WM)
+            %                        3 = whole brain (WB = GM+WM)
+
 			if ~x.S.IsASL
 				pGM_here = ones(size(DataIm)); % no masking
 				bSkipPVC = 1;
-			elseif ~isempty(findstr(lower(namesROIuse{iROI}),'whole brain')) || ~isempty(findstr(lower(namesROIuse{iROI}),'wb')) || ~isempty(findstr(lower(namesROIuse{iROI}),'wholebrain'))
+            elseif x.S.bTissueMasking==1
+                % we want GM, so keep tissue masks as they are
+				pGM_here = pGM;
+				pWM_here = pWM;
+			elseif x.S.bTissueMasking==2
+                % we want WM, so we flip the ROIs, using GM as "WM" in PVC
+                pGM_here = pWM;
+				pWM_here = pGM;
+			elseif x.S.bTissueMasking==3
+                % we want WB, i.e. GM+WM, using CSF as "WM" in PVC
 				pGM_here = pGM+pWM;
 				pWM_here = ones(size(pGM)) - pGM - pWM; % CSF
 				bSkipPVC = 1;
-			elseif  ~isempty(findstr(lower(namesROIuse{iROI}),'white matter')) || ~isempty(findstr(lower(namesROIuse{iROI}),'wm'))  || ~isempty(findstr(lower(namesROIuse{iROI}),'whitematter'))
-				% flip the ROIs:
-				% we are interested in WM instead of pGM
-				pGM_here = pWM;
-				pWM_here = pGM;
-            else % we want GM, so keep tissue masks as they are
-				pGM_here = pGM;
-				pWM_here = pWM;
+            else
+                error('Unknown tissue type chosen, should be one out of 1:GM, 2:WM, 3:WB');
             end
 
             % Defaulting outputs for ROI visualization
@@ -654,6 +679,9 @@ for iSubject=1:x.dataset.nSubjects
                     % Check for empty CBF map first
                     fprintf('%s\n', ['Warning: Empty image for ' x.SUBJECTS{iSubject} '_ASL_' xASL_num2str(iSess)]);
 				end
+
+                % pGM_here is the main tissue type that is investigated
+                % pWM_here is the other tissue type that is used with PVC
 
 				if x.S.bMasking(3)==0 % no tissue-masking
                     pGM_here = ones(size(DataIm));
--- a/Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
+++ b/Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
@@ -39,8 +39,8 @@ function [x] = xASL_stat_GetROIstatistics(x)
 %                        2 = white matter (WM)
 %                        3 = whole brain (WB = GM+WM)
 %                        4 = ...
-%                        (REQUIRED) ->>>>>>>>>>>>>>>>>>>>>>>>>>>, set DEFAULT outside this code as [1 2])
-%                                   ->>>>>>>>>>>>>>>>>>>>>>>>>>>
+%                        (REQUIRED, single value)
+%                        Cannot be a vector. Iterate outside this function over multiple x.S.Atlases-x.S.TissueMasking combinations.
 %                        Note that a ROI needs to have a minimal volume. So a deepWM ROI for which pvGM is requested will be skipped.
 %
 %
--- a/Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
+++ b/Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
@@ -33,13 +33,13 @@ function [x] = xASL_stat_GetROIstatistics(x)
 %                       (OPTIONAL, DEFAULT=1)
 %   x.S.bWMH           - boolean specifying if we should mask a ROI with pvWMH (by subtracting pvWMH in ASL resolution)
 %                        true = include WMH (no masking), false = exclude WMH (OPTIONAL, DEFAULT=false)
-%   x.S.bTissueMasking - value specifying which subject-wise tissue type we will mask with. Cannot be a vector
+%   x.S.TissueMaskingLocal - value specifying which subject-wise tissue type we will mask with. Cannot be a vector
 %                        (loop outside this function for multiple tissue types)
 %                        'GM' = gray matter (GM)
 %                        'WM' = white matter (WM)
 %                        'WB' = whole brain (WB = GM+WM)
 %                        (REQUIRED, single value)
-%                        Cannot be a vector. Iterate outside this function over multiple x.S.Atlases-x.S.bTissueMasking combinations.
+%                        Cannot be a vector. Iterate outside this function over multiple x.S.Atlases-x.S.TissueMaskingLocal combinations.
 %                        Note that a ROI needs to have a minimal volume. So a deepWM ROI for which pvGM is requested will be skipped.
 %
 %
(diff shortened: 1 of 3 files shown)


Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
```diff
@@ -31,6 +31,18 @@
 %                       0 = [0 0 0 0]
 %                       Can be useful for e.g. loading lesion masks outside the GM
 %                       (OPTIONAL, DEFAULT=1)
+%   x.S.bWMH           - boolean specifying if we should mask a ROI with pvWMH (by subtracting pvWMH in ASL resolution)
```
jan-petr, line 34: This was added previously, right?

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
```diff
@@ -31,6 +31,18 @@
 %                       0 = [0 0 0 0]
 %                       Can be useful for e.g. loading lesion masks outside the GM
 %                       (OPTIONAL, DEFAULT=1)
+%   x.S.bWMH           - boolean specifying if we should mask a ROI with pvWMH (by subtracting pvWMH in ASL resolution)
+%                        true = include WMH (no masking), false = exclude WMH (OPTIONAL, DEFAULT=false)
+%   x.S.bTissueMasking - value specifying which subject-wise tissue type we will mask with. Cannot be a vector
```
jan-petr: Is this definition correct? Boolean, but you talk about GM, WM...

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
```diff
@@ -31,6 +31,18 @@
...
 %                       (OPTIONAL, DEFAULT=1)
+%   x.S.bWMH           - boolean specifying if we should mask a ROI with pvWMH (by subtracting pvWMH in ASL resolution)
+%                        true = include WMH (no masking), false = exclude WMH (OPTIONAL, DEFAULT=false)
+%   x.S.bTissueMasking - value specifying which subject-wise tissue type we will mask with. Cannot be a vector
+%                        (loop outside this function for multiple tissue types)
+%                        'GM' = gray matter (GM)
+%                        'WM' = white matter (WM)
+%                        'WB' = whole brain (WB = GM+WM)
+%                        (REQUIRED, single value)
+%                        Cannot be a vector. Iterate outside this function over multiple x.S.Atlases-x.S.bTissueMasking combinations.
```
jan-petr: A general question
can we give Atlases as {'TotalGM' 'Supratentorial' 'Supratentorial'} - with two names repeated as 
TissueMasking will be {'GM' 'WM' 'GM'}.

I understand that this is possible and it is good. Just asking if the code will work with that.

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
```diff
@@ -596,22 +608,37 @@
...
-			elseif  ~isempty(findstr(lower(namesROIuse{iROI}),'white matter')) || ~isempty(findstr(lower(namesROIuse{iROI}),'wm'))  || ~isempty(findstr(lower(namesROIuse{iROI}),'whitematter'))
-				% flip the ROIs:
-				% we are interested in WM instead of pGM
-				pGM_here = pWM;
-				pWM_here = pGM;
-            else % we want GM, so keep tissue masks as they are
-				pGM_here = pGM;
-				pWM_here = pWM;
+            else
+                switch x.S.bTissueMasking
```
jan-petr: Didn't we say that bTissueMasking is a boolean? And TissueMasking the vector with strings?

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
```diff
@@ -655,6 +682,9 @@
                     fprintf('%s\n', ['Warning: Empty image for ' x.SUBJECTS{iSubject} '_ASL_' xASL_num2str(iSess)]);
 				end
 
+                % pGM_here is the main tissue type that is investigated
```
jan-petr: We might actually later change the names pGM_here as pMain and pWM_here as pSecond or something like that to avoid a possible confusion with swapping GM and WM.

I'm of course fine with that personally and know the code is correct. It just might be a bit confusing.

Modules/xASL_module_Population.m
```diff
@@ -247,7 +244,17 @@
...
-        for iAtlas=1:length(x.S.Atlases)        
+        for iAtlas=1:length(x.S.Atlases)
+            % Note that the number of ROI atlases here should be the same the number of tissue masking chosen
+            % If needed, an atlas or tissue type can be provided multiple times in different combinations
+
+            % We use the specified tissue type
+            % 'GM' = gray matter
+            % 'WM' = white matter
+            % 'WB' = whole brain (= GM+WM)
+            x.S.bTissueMasking = x.S.TissueMasking(iAtlas);
```
jan-petr: Now I get it. bTissueMasking is the single instance. Then we should still change the name,

Modules/xASL_module_Population.m
```diff
@@ -356,4 +364,54 @@
...
+
+
+%% -----------------------------------------------------------------------------
+%% -----------------------------------------------------------------------------
+function [x] = xASL_wrp_Population_PrepareAtlas4ROI(x)
+%xASL_wrp_Population_PrepareAtlas4ROI Parse x.S.Atlases & x.S.TissueMasking
+
+bAtlasTissueMatch = true;
+if isfield(x.S, 'Atlases') && ~isfield(x.S, 'TissueMasking')
+    fprintf('\n%s\n', 'Warning: custom ROI atlas-selection detected without the tissue-types for these ROIs');
```
jan-petr: Give the variable name x.S.Atlases.

Modules/xASL_module_Population.m
```diff
@@ -356,4 +364,54 @@
...
+    bAtlasTissueMatch = false;
+elseif ~isfield(x.S, 'Atlases') && isfield(x.S, 'TissueMasking')
+    fprintf('\n%s\n', 'Warning: custom tissue-types specified without ROI atlas-selection');
```
jan-petr: Also clarify that you talk about x.S.TissueMasking

Modules/xASL_module_Population.m
```diff
@@ -356,4 +364,54 @@
...
+    bAtlasTissueMatch = false;
+elseif isfield(x.S, 'Atlases') && isfield(x.S, 'TissueMasking') && length(x.S.Atlases)~=length(x.S.TissueMasking)
+    fprintf('\n%s\n', 'Warning: not the same number of ROI atlases as subject-wise tissue-types provided');
```
jan-petr, line 386: Same. And you can print the content to help the user...

Modules/xASL_module_Population.m
```diff
@@ -356,4 +364,54 @@
...
+    bAtlasTissueMatch = false;
+end
+if ~bAtlasTissueMatch
+    fprintf('\n%s\n', 'ExploreASL''s population module produces ROI values that are obtained by a ');
+    fprintf('%s\n', 'combination of an MNI ROI atlas with subject-specific GM and WM segmentations.');
+    fprintf('%s\n', 'When custom ROI atlases are provided in x.S.Atlases, their tissue types ');
+    fprintf('%s\n', 'need to be provided as well, with either option ''GM'', ''WM'', ''WB'' (GM+WM).');
+    fprintf('%s\n', 'E.g., when not provided, these default to:');
```
jan-petr, line 396: This is nice!

jan-petr (changes requested):
Functionality good. Some questions about parameter names and comments.

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
HenkMutsaerts: No I added this two weeks ago (or I changed it two weeks ago)

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
HenkMutsaerts: Yes, xASL_stat_GetROIstatistics loops over x.S.Atlases and will take the same index from x.S.TissueMasking. Previously, we check that those have equal length. And if they don't make sense (e.g. TotalGM combined with 'WM') then it will return n/a to the tsv because it is empty.

Thinking about this, I need to check the variable name in the tsv. It should be TotalGM_GM Supratentorial_GM Supratentorial_WM in this example.

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
HenkMutsaerts: Fixed in a7370150eb9b7101ef9e47766dabb1ef1b82e7ad by changing:
xS.bTissueMasking->xS.TissueMaskingLocal

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
HenkMutsaerts: Yes I thought about this as well. Done in f1cd5484f028a2ca0c198750bcc69d7539e7cfe8

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
HenkMutsaerts: This was another issue, that we either recently merged or not

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
HenkMutsaerts: Fixed in a7370150eb9b7101ef9e47766dabb1ef1b82e7ad by changing:
xS.bTissueMasking->xS.TissueMaskingLocal

Modules/xASL_module_Population.m
HenkMutsaerts: Fixed in a7370150eb9b7101ef9e47766dabb1ef1b82e7ad by changing:
xS.bTissueMasking->xS.TissueMaskingLocal

Modules/xASL_module_Population.m
HenkMutsaerts: Done in fd5b7c0cabc7ffc54dce588ca69ee2bbbfbe5e80

Modules/xASL_module_Population.m
HenkMutsaerts: Done in fd5b7c0cabc7ffc54dce588ca69ee2bbbfbe5e80

Modules/xASL_module_Population.m
HenkMutsaerts: Done in fd5b7c0cabc7ffc54dce588ca69ee2bbbfbe5e80

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
HenkMutsaerts: I have now fixed this in 5b9132ee1eb5bbab7416f8ac3fe43b3ea2da27d7 such that the resulting filename is:

/Population/Stats/mean_qCBF_GM_StandardSpace_TotalGM_n=1_13-Sep-2024_PVC2.tsv

When we merge this and the other issue, this should become e.g.,:

/Population/Stats/mean_qCBF_GM_StandardSpace_WholeBrain_n=1_13-Sep-2024_PVC2.tsv

OR e.g.,

/Population/Stats/mean_qCBF_GM_StandardSpace_Supratentorial_n=1_13-Sep-2024_PVC2.tsv

jan-petr: Can you check and confirm before merging?

jan-petr (approved):
All answered

PR #1755, opened 2024-08-25
Closes #900 license

### LINKED ISSUE #900: License
### Description

Preparations for ExploreASL to fullfil the Apache 2.0 license.

### To do
- [x] Change license file

** External folder **
A note to Vladimir: Everything in the folder External is not a part of ExploreASL, but considered an external package which is packaged together with ExploreASL to ensure that the correct versions are provided. The idea is that this helps us to avoid some of the issues with e.g. SPM license. We have some (1%) of the code that is a change of SPM, which is thus licensed inside SPM as GPL. The ExploreASL code is separate from everything in the External folder, hence the different license.

- [x] Change the folder names in xASL_dev_MakeStandalone according to the changes done in this issue
- [x] Move atlas path definition from xASL_init_FileSystem to xASL_init_MapsAndAtlases, because they are not module or subject-specific (which the remainder of xASL_init_FileSystem is)
- [x] Move atlas paths from x.P.Atlas.* to x.D.Atlas* (from subject-specific to general)
- [x] Move maps to the following structure:
New distribution maps:
1. /Maps: ExploreASL maps for image processing, ExploreASL license: 
2. /External/Atlases4ROIs: all atlases used for ROIs
    /External/Atlases4ROIs/LicenseLimited
    /External/Atlases4ROIs/LicensePermissive
3. /External/SPMmodified/MapsAdded: maps used for image processing, SPM/CAT12, GPL license
-> this way, the licenses are nicely separated. And the Atlases can be added by a quick recursive filesearch.
- [x] Copyright addition will be done in #1778

**Atlases**
- [x] Move MindBoggle to the free atlases folder
- [x] xASL_stat_PrintCluster_ROI_Stats -> added warning that this is a commercial atlas. When using this code, we should replace HO(cort|sub) by Mindboggle
- [x] xasL_io_ReadDataPar: 76 -> here we should remove Hammers, HO(cort|sub)
- [x] Add a citation part to the license text files
- [x] Split atlases to free and non-free
- [x] When running the population module, add that atlases should be cited and licensed respected
**BIDS-matlab** : GNU GPL 3 (2007)
update to newest version which has the MIT license
-> Not required, because of our discussion within AUMC, that we are fine being separate from GNU GPL
**SPMmodified**
The Apache License 2.0 is not directly compatible to GNU GPL 2 (everything inside SPM, so CAT12, DARTEL, GeodesicShooting, etc) is all GPL.
-> Not required, because of our discussion within AUMC, that we are fine being separate from GNU GPL
**SPMmodified**-**xASL** 
This is the ExploreASL change of SPM that glues with the GPL2 part, and ExploreASL itself is separate from SPMmodified. We keep this as is to avoid Apache-GPL incompatibility.

### Removed or resolved
**isnear**
Used in a few functions only and the license should be fine, but still
- [x] Replace by our own simple function xASL_isnear. Done in #1655 
**fsl config file**
- [x] Moved to /Functions/FSL/
**ExploreQC** @luislorenzini
Property of Amsterdam UMC & not-used
- [x] Has been removed (moved to the CustomScripts private repository)
**MPdenoising**
- [x] Same as for ExploreQC

### Already open-source compatible
**DCMTK**
BSD License, compatible to Apache 2.0
**MRIcron**
Custom license that has similarities with permissive open-source licenses
**SuperGzip**: MIT License, compatible. Check if we use this, otherwise remove
Michael used this for faster Windows zipping. We can keep this for now.
**vtkwrite**: check if we use this, otherwise remove.
Michael used this to write some surfaces or volumes in a vector format. So that is useful for some visualization, but no clue if we practically use it anywhere. We don't currently use it, but it has MIT so we can keep it.

### Release notes
Update ExploreASL License to Apache 2.0

HenkMutsaerts:
### Linked issue

Closes #900

### How to test

Henk tested this for the ExploreASL-included TestDataSet only

### Comments

Still need to change LICENSE-EXPLOREASL & append the Apache statement to each .m header (except for the External folders). Also, we might want to reversion the this beta release to 2.0.0


jan-petr (approved):
OK

PR #1756, opened 2024-08-26
#1711 xASL_module_ASL: for 3D spiral 0->NaN

### LINKED ISSUE #1711: Set zeroes outside 3D spiral FoV to NaN
### Description

Torstein Torgan (ASL in CO intoxication, Oslo) had a single GE 3D PCASL scan that was acquired misoriented, see below slice. While most of the cortex is still in the FoV, he noticed that there were some negatives included from interpolation artifacts at the edges, which when divided by in M0 division became extreme large values. These originated from artificial zeroes from the GE 3D spiral FoV which is round, not Cartesian, as the FoV stopped there. Usually this is not a concern because those zeroes are well outside the brain but when they are close to the brain, it is essential that they are set to NaN to be masked out. Like SPM ensures that everything outside the original FoV is set to NaN when moving to a different space.

Henk requested the T1w for this example as well, to check if this influences #1039. It did not.

For Henk to check:
/Users/hjmutsaerts/ExploreASL/TestDataSets/GE_3Dspiral_misaligned_LoQ

image

Now this nicely ends up at the correct location, including the missing FoV:

image

### Tasks

- [x] Very early in the ASL module, set "true" zeroes to NaN (not isnear), for 3D spiral specifically (because with Cartesian the FoV == the reconstruction matrix more or less)
- [x] Add this example to TestDataSets
- [x] Once merged, notify Torstein Torgan

### How to test

- [x] Test this with the new example in TestDataSets

### Release notes

Fix issue with wrongly planned GE ASL scan.

HenkMutsaerts:
### Linked issue

Closes #1711

### How to test

I tested this with https://github.com/ExploreASL/TestDataSets/tree/main/GE_3Dspiral_misaligned_LoQ/
see issue.

### Comments

See issue, please check if you like the current location where I do this. I don't think we should change this for the BIDS conversion, this is an image processing issue.

### DIFF
--- a/Modules/xASL_module_ASL.m
+++ b/Modules/xASL_module_ASL.m
@@ -213,6 +213,29 @@ end
 %% G. ASL processing & quantification parameters
 x = xASL_module_ASL_ParseParameters(x, bOutput);
 
+%% H. For GE 3D spiral ASL and M0, check for parts outside of the FoV
+% With a non-Cartesian FoV, there can be voxels in the image matrix that are outside the FoV.
+% If these are not explicitly to NaNs, the original FoV can be wrongly interpreted.
+% This led to a misalignment in the case of registering an ASL where the brain was too close to the edge of the FoV.
+% By setting these voxels to NaNs, the registration algorithms know to skip these voxels outside the FoV.
+
+if strcmpi(x.Q.Vendor, 'GE') && strcmpi(x.Q.MRAcquisitionType, '3D') && strcmpi(x.Q.PulseSequenceType, 'spiral')
+    % Once we have a GE 3D spiral sequence
+    
+    niftiList = {x.P.Path_ASL4D, x.P.Path_M0};
+    for iNifti=1:length(niftiList)
+        if xASL_exist(niftiList{iNifti})
+    
+            IM = xASL_io_Nifti2Im(niftiList{iNifti});
+            IM(IM==0) = NaN;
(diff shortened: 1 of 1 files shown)


Modules/xASL_module_ASL.m
```diff
@@ -213,6 +213,29 @@
 %% G. ASL processing & quantification parameters
 x = xASL_module_ASL_ParseParameters(x, bOutput);
 
+%% H. For GE 3D spiral ASL and M0, check for parts outside of the FoV
+% With a non-Cartesian FoV, there can be voxels in the image matrix that are outside the FoV.
+% If these are not explicitly to NaNs, the original FoV can be wrongly interpreted.
+% This led to a misalignment in the case of registering an ASL where the brain was too close to the edge of the FoV.
+% By setting these voxels to NaNs, the registration algorithms know to skip these voxels outside the FoV.
+
+if strcmpi(x.Q.Vendor, 'GE') && strcmpi(x.Q.MRAcquisitionType, '3D') && strcmpi(x.Q.PulseSequenceType, 'spiral')
```
jan-petr: GE should be regexpi - cause we might have GE_wip.

jan-petr (changes requested):
Minor comment.

Modules/xASL_module_ASL.m
HenkMutsaerts: changed to "contains" in b8c3e11495fd6dc8805c5e4aeed3f7d6ed4c1c28

jan-petr (approved):
OK.

PR #1758, opened 2024-08-26
Closes #1103 xASL_module_BIDS2Legacy: add backward compatibility

### LINKED ISSUE #1103: <SUBJECTNAME> vs <SUBJECTNAME>_1

HenkMutsaerts:
### Linked issue

Closes #1103

### DIFF
--- a/Modules/xASL_module_BIDS2Legacy.m
+++ b/Modules/xASL_module_BIDS2Legacy.m
@@ -102,7 +102,26 @@ function [result, x] = xASL_module_BIDS2Legacy(x, bOverwrite, bVerbose)
             %% 2. Parse modality
             
             pathLegacy_SubjectSession = fullfile(x.dir.xASLDerivatives, SubjectSession);
-            
+            bFolderExisted = exist(pathLegacy_SubjectSession, 'dir')==7;
+
+            % Backwards compatibility: rename sub-*** to sub-***_1
+            if strcmp(SessionID, '1')
+                pathLegacy_SubjectSession_OLD = fullfile(x.dir.xASLDerivatives, SubjectID);
+                if ~bFolderExisted && exist(pathLegacy_SubjectSession_OLD, 'dir')==7
+                    warning(['Renaming ' pathLegacy_SubjectSession_OLD '->' SubjectSession]);
+                    xASL_Move(pathLegacy_SubjectSession_OLD, pathLegacy_SubjectSession, 1, 0);
+                    bFolderExisted = true;
+                end
+            end
+
+            % Provide warning that we are going to overwrite the folder
+            if bFolderExisted
+                warning(['Overwriting rawdata copied to: ' pathLegacy_SubjectSession]);
+                fprintf('%s\n', 'Note that we do not overwrite any derivatives');
+                fprintf('%s\n', 'Rather, we merge them with the freshly copied rawdata');
+                fprintf('%s\n', 'It is always cleanest to remove all derivatives before rerunning');
+            end
+
             % Create subject/session directory (to enable reruns for pre-imported or crashed datasets, we need a subject level here/above!)
             xASL_adm_CreateDir(pathLegacy_SubjectSession);
 
--- a/Modules/xASL_module_BIDS2Legacy.m
+++ b/Modules/xASL_module_BIDS2Legacy.m
@@ -106,10 +106,34 @@ function [result, x] = xASL_module_BIDS2Legacy(x, bOverwrite, bVerbose)
 
             % Backwards compatibility: rename sub-*** to sub-***_1
             if strcmp(SessionID, '1')
-                pathLegacy_SubjectSession_OLD = fullfile(x.dir.xASLDerivatives, SubjectID);
+                pathLegacy_SubjectSession_OLD = fullfile(x.dir.xASLDerivatives, SubjectID); % it should be [SubjectID '_' SessionID] now
                 if ~bFolderExisted && exist(pathLegacy_SubjectSession_OLD, 'dir')==7
                     warning(['Renaming ' pathLegacy_SubjectSession_OLD '->' SubjectSession]);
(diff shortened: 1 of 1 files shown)


Modules/xASL_module_BIDS2Legacy.m
```diff
@@ -102,7 +102,26 @@
...
             pathLegacy_SubjectSession = fullfile(x.dir.xASLDerivatives, SubjectSession);
-            
+            bFolderExisted = exist(pathLegacy_SubjectSession, 'dir')==7;
+
+            % Backwards compatibility: rename sub-*** to sub-***_1
+            if strcmp(SessionID, '1')
+                pathLegacy_SubjectSession_OLD = fullfile(x.dir.xASLDerivatives, SubjectID);
+                if ~bFolderExisted && exist(pathLegacy_SubjectSession_OLD, 'dir')==7
+                    warning(['Renaming ' pathLegacy_SubjectSession_OLD '->' SubjectSession]);
+                    xASL_Move(pathLegacy_SubjectSession_OLD, pathLegacy_SubjectSession, 1, 0);
```
jan-petr: A minor issue - we rename the folder, but don't rename the lock-files. So the old lock-files stay there. 

Shall we ignore that or keep that? Removing might be nice to avoid potential future issues. On the other hand, it's only relevant to old data. So just think about it and decide what you want ;)

Modules/xASL_module_BIDS2Legacy.m
```diff
@@ -102,7 +102,26 @@
...
+                    bFolderExisted = true;
+                end
+            end
+
+            % Provide warning that we are going to overwrite the folder
+            if bFolderExisted
+                warning(['Overwriting rawdata copied to: ' pathLegacy_SubjectSession]);
```
jan-petr: A bit confusing so the user might think you are overwriting raw data.

Modules/xASL_module_BIDS2Legacy.m
```diff
@@ -102,7 +102,26 @@
...
+                fprintf('%s\n', 'Note that we do not overwrite any derivatives');
```
jan-petr: You mean, we don't overwrite processed files. Cause the origin files derivatives/ExploreASL/sub1/ASL_1/ASL.nii is kind of a derivative file.

In other words, we are not overwriting derived files, but this message can be confusing and the user might think we talk about the entire "derivatives" folder.

Modules/xASL_module_BIDS2Legacy.m
```diff
@@ -102,7 +102,26 @@
...
+                fprintf('%s\n', 'Rather, we merge them with the freshly copied rawdata');
```
jan-petr: We merge previously processed data with ...

Modules/xASL_module_BIDS2Legacy.m
```diff
@@ -102,7 +102,26 @@
...
+                fprintf('%s\n', 'It is always cleanest to remove all derivatives before rerunning');
```
jan-petr: Nevertheless, ...

jan-petr (changes requested):
Minor comments.

Modules/xASL_module_BIDS2Legacy.m
HenkMutsaerts: Good point, but then we would need to rename all derivatives as well. Done in 16d0ee0c9a8ea0b26985b41e69d75091b4c951e9

Modules/xASL_module_BIDS2Legacy.m
HenkMutsaerts: a750e39ae434b87ce24d127bfc2a24d87652ab95

Modules/xASL_module_BIDS2Legacy.m
HenkMutsaerts: a750e39ae434b87ce24d127bfc2a24d87652ab95

Modules/xASL_module_BIDS2Legacy.m
HenkMutsaerts: a750e39ae434b87ce24d127bfc2a24d87652ab95

Modules/xASL_module_BIDS2Legacy.m
HenkMutsaerts: a750e39ae434b87ce24d127bfc2a24d87652ab95

jan-petr (approved):
OK

PR #1767, opened 2024-09-09
Closes #1766 xASL_imp_CreateSummaryFile: Fix missing scans index

### LINKED ISSUE #1766: xASL_imp_CreateSummaryFile crashes with too many missing scans per single Subject/Visit/Session
### Description

xASL_imp_CreateSummaryFile crashes with too many missing scans per single Subject/Visit/Session. In some cases we obtain multiple scans per single session/run, which might need concatenation or are erroneous, but in any case it gives a crash.
I fixed it with the changes below, but @jan-petr may want to check if we shouldn't also add a warning that something goes wrong in the session/scan definition?

Screenshot 2024-09-09 at 13 50 10

JAN: Problem is not indexing in summary files. The problem is that we have multiple versions of a single scan discovered. That cannot be tolerated as ExploreASL cannot deal with that. It then randomly converts a single scan and reports missing scans that are actually there.
So as a new solution, we report an error that the user needs to specify more clearly which scans to take to avoid ambiguous entries.

### Release notes

Warning when duplicate scans are detected.

HenkMutsaerts:
### Linked issue

Closes #1766

### DIFF
--- a/Modules/Module_Import/Import/xASL_imp_CreateSummaryFile.m
+++ b/Modules/Module_Import/Import/xASL_imp_CreateSummaryFile.m
@@ -131,13 +131,13 @@ function xASL_imp_CreateSummaryFile(thisSubject, PrintDICOMFields, x)
 			
 			for iSession=1:thisVisit.nSessions
 				fprintf(fid_summary,',"');
-				fprintf(fid_summary,'%s ',thisVisit.scanNames{logical(thisSubject.globalCounts.missing_scans(iSubject, iVisit, iSession,:))});
+				fprintf(fid_summary,'%s ',thisVisit.scanNames{logical(sum(thisSubject.globalCounts.missing_scans(iSubject, iVisit, iSession,:)))});
 				fprintf(fid_summary,'"');
 			end
 			
 			for iSession=1:thisVisit.nSessions
 				fprintf(fid_summary,',"');
-				fprintf(fid_summary,'%s ',thisVisit.scanNames{logical(thisSubject.globalCounts.skipped_scans(iSubject, iVisit, iSession,:))});
+				fprintf(fid_summary,'%s ',thisVisit.scanNames{logical(sum(thisSubject.globalCounts.skipped_scans(iSubject, iVisit, iSession,:)))});
 				fprintf(fid_summary,'"');
 			end
 			


jan-petr (changes requested):
It's a bit weird to me. Can you please show me a printout of what's in 
thisVisit.scanNames and in thisSubject.globalCounts.missing_scans(iSubject, iVisit, iSession,:)

It's possible that the issue are incompatible sizes, so using 'squeeze' instead of 'sum' would solve it. Even in the other cases, I would not use sum.

But please just show me what's in those variables when the crash happens.

### HENK
I would have to check again, but with the data of Udunna (the DEBBIE example pig Paulien you got from me) it would convert correctly, but these missing scans would have multiple values:

This is the error we get:

```
Message: Non-dicom: /Users/hjmutsaerts/ExploreASL/Pig_Paulien/sourcedata/Pig004/ENCODEDIMAGES_HAD04_03TIS_08TES_0028/Icon
   
Warning: A numerical field provided as a string (2).  It was correctly converted, but should be avoided next time.
Evaluating multi-parameter studyPar for subject/visit/session: Pig004//2
 For RegExp subject/visit/session //2, overwriting fields:  LabelingDuration, PostLabelingDelay, TimeEncodedMatrixSize, ASLContext,
Warning: concatenating multiple NIfTIs & jsons as output from dcm2niiX
Merging the echo numbers of the Hadamard encoded sequence...
Warning: echo times do not increase, resorting will be applied...
Corrected dcm2niiX output for following files:
/Users/hjmutsaerts/ExploreASL/Pig_Paulien/derivatives/ExploreASL/temp/Pig004/ASL_2/ASL4D.nii
Warning: Skipped scans had a different size than converted scans, fixing this but summary file could be incorrect 
> In xASL_imp_CreateSummaryFile (line 56)
In xASL_wrp_DCM2NII (line 57)
In xASL_module_Import (line 153)
In xASL_init_Iteration>runIteration (line 415)
In xASL_init_Iteration (line 152)
In ExploreASL_Import (line 50)
In ExploreASL (line 91) 
Warning: Missing scans had a different size than converted scans, fixing this but summary file could be incorrect 
> In xASL_imp_CreateSummaryFile (line 62)
In xASL_wrp_DCM2NII (line 57)
In xASL_module_Import (line 153)
In xASL_init_Iteration>runIteration (line 415)
In xASL_init_Iteration (line 152)
In ExploreASL_Import (line 50)
In ExploreASL (line 91) 

ERROR: Import module terminated for subject 1: Pig004

ans =

    'The logical indices contain a true value outside of the array bounds.
     
     Error in xASL_imp_CreateSummaryFile (line 134)
     				fprintf(fid_summary,'%s ',thisVisit.scanNames{logical(thisSubject.globalCounts.missing_scans(iSubject, iVisit, iSession,:))});
     
```

If I check the parameter value by debugging:
```
thisSubject.globalCounts.missing_scans(iSubject, iVisit, iSession,:)

  1×1×1×7 uint8 array

ans(:,:,1,1) =

   0


ans(:,:,1,2) =

   0


ans(:,:,1,3) =

   0


ans(:,:,1,4) =

   0


ans(:,:,1,5) =

   1


ans(:,:,1,6) =

   1


ans(:,:,1,7) =

   1
```

squeezing gives:

```
squeeze(thisSubject.globalCounts.missing_scans(iSubject, iVisit, iSession,:))

ans =

  7×1 uint8 column vector

   0
   0
   0
   0
   1
   1
   1
```
And it needs to be an index to `thisVisit.scanNames`

Original line:
`fprintf(fid_summary,'%s ',thisVisit.scanNames{logical(thisSubject.globalCounts.missing_scans(iSubject, iVisit, iSession,:))});`

It makes a logical as you see, so apparently the index should always be one if any of the converted scans was missing. So I figured `sum` should be fine. Do you agree? Otherwise, you need to suggest a viable alternative ;)
NOTE that this is just for the SummaryFile

PR #1771, opened 2024-09-11
#1760 im_CreateASLDeformationField Fix y_ASL PreSmooth

### LINKED ISSUE #1760: Flow-field error on the edge of y_ASL.nii
### Description

PS: this issue should not be confused with #1039.

We smooth `y_T1.nii` with the ASL resolution to create `y_ASL.nii`. Sometimes, `y_ASL.nii` still has incorrect edges, perhaps due to too small kernel sizes in `xASL_im_PreSmooth`? This seems independent of `x.settings.Quality`.

image

On the left, you can see `y_T1.nii`, on the right `y_ASL.nii`, for `Philips_2DEPI_Lesion_NoFLAIR_NoM0_HiQ`. This results in erroneous lines at the bottom:

image

There seems to be something wrong with `xASL_im_PreSmooth`, I cannot find anything else. Alternatively, we keep it as it is, and always remove the outer e.g., 3 lines of `y_ASL.nii` and extrapolate with `xASL_im_ndnanfilter`, at the end of `xASL_im_CreateASLDeformationField`.

JAN:
This is caused by Matlab's convn that leaves a weird border. So we need to remove the border and fill it using xASL_im_FillNaNs.

### Release notes

Nothing. Report together with #1039.

jan-petr:
### Linked issue

Close #1760

### DIFF
--- a/Functions/ImageProcessing/xASL_im_CreateASLDeformationField.m
+++ b/Functions/ImageProcessing/xASL_im_CreateASLDeformationField.m
@@ -34,7 +34,7 @@ function xASL_im_CreateASLDeformationField(x, bOverwrite, EstimatedResolution, P
 % EXAMPLE:      xASL_im_CreateASLDeformationField(x);
 %               xASL_im_CreateASLDeformationField(x, 1, [3 3 7]);
 % __________________________________
-% Copyright (C) 2015-2021 ExploreASL
+% Copyright (C) 2015-2024 ExploreASL
 
 
 %% Admin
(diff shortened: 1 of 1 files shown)


HenkMutsaerts (changes requested):
I would change it to 3 layers, that should be fine and never cause any issues, because no voxels should end up at the edges of MNI (in fact 7 layers should still be fine)

jan-petr:
No real need to do that. Because we remove a single layer, but then run fillNan - which checks another thick layer for outliers. So any further issues would have been detected. The only reason why I really need to remove the first layer is that FillNaN needs to have NaNs to start checking. If there wouldn't be any NaNs, not even 1 layer, it would have kept it how it was.

HenkMutsaerts (approved):

PR #1774, opened 2024-09-11
#1750 module_ASL Skip CBF quantification by default

### LINKED ISSUE #1750: Skip CBF quantification
### Description

If ASLContext is CBF, then we skip the quantification and take what is there. This is useful when we have both CBF ASL image, or when we have DSC image that we just want to run through as a dummy ASL.

### Tasks

- [x] Detect CBF in ASL_Context
- [x] ApplyQuantification defaults to 0 when ASL_Context contains CBF

### Release notes

Skip quantification for ASL-Context containing a CBF image.

jan-petr:
### Linked issue

Close #1750

### DIFF
--- a/Modules/xASL_module_ASL.m
+++ b/Modules/xASL_module_ASL.m
@@ -185,7 +185,8 @@ end
 %% F. DeltaM and CBF parsing - check if all/some volumes are deltams or CBFs
 % If TSV file exist
 % We don't have a subtraction image by default
-x.modules.asl.bContainsSubtracted = false;
+x.modules.asl.bContainsSubtracted = false; % Contains any kind of subtracted image like deltaM or CBF
+x.modules.asl.bContainsCBF = false; % Contains quantified CBF image
 % Load TSV file
 if xASL_exist(x.P.Path_ASL4Dcontext, 'file')
 	aslContext = xASL_tsvRead(x.P.Path_ASL4Dcontext);
@@ -198,6 +199,7 @@ if xASL_exist(x.P.Path_ASL4Dcontext, 'file')
 	% Check for presence of CBF volumes
 	if ~isempty(regexpi(strjoin(aslContext(2:end)),bidsPar.stringCbf, 'once'))
 		x.modules.asl.bContainsSubtracted = true;
+		x.modules.asl.bContainsCBF = true;
 	end
 else
 	% In case the ASL-context file is missing we set containsDeltaM to true for all NIfTIs with a single volume only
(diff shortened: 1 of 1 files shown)


HenkMutsaerts (changes requested):
Nice, though I'm not sure what happens if we would have a combination of "rawer" ASL volumes and a CBF, which can easily be the case e.g., for GE 3D spiral, if the user has added both the raw series and the scanner-quantified volumes in the same BIDS volume. 

Not sure how we deal with this when this happens in sourceData, but of course the user could mess this up in rawData.

Your change means that in this case we skip quantification. Though the same logic applies to bContainsSubtracted, where we potentially already have a "bug".

So do we first need a check that if we have "more/less raw" volumes in ASL4D (as specified by ASLcontext), that we remove the less raw volumes and process the more raw images, and provide a warning?

Because this is not a bug that is introduced in #1750 (we already have this potential bug for bContainsSubtracted), we should perhaps address this as a new issue.

If you agree, do you want to make a new issue, and I can approve #1750?

Another minor point of improvement, is that this behavior leads to extra booleans, whereas we could also just let x.bidsPar travel through the code as global parameter, from which ASLcontext is checked. If you would vote against this, we should at least add a comment to both the initialization of these boolean where/how they are used downstream, and at the locations where we use these booleans, how are they are defined from ASLcontext upstream. We could already do this in this issue right?

jan-petr:
Definitively a good point and I already thought about that.
Lets make a new issue as you say. The current behavior is that M0 gets separated automatically. So if you have "cbf,M0scan" or "deltaM,M0scan" then things will work correctly and automatically with the code we have.
If the user messses it up and provides "cbf,control,label" or "deltaM,control,label" or "cbf,deltaM" - the code won't work correctly. As it will not keep all in and not quantify it.

In the new issue, we can issue some special warnings for this:
1. CBF+control+label detected - removing CBF
2. CBF+deltaM detected - removing CBF
3. deltaM,control,label detected - removing deltaM
Or something like this. Feel free to adapt this in the issue.

Regarding you last question -> lets also add this to this new issue. I'm undecided if we should keep this in x.bidsPar or in x.module.asl - it feels kind of correct in both variables, so I'd stick to x.module.asl as we have read it from JSON and interpreted the JSON, so it's no longer pure BIDS, but our interpretation of it. But I definitively agree that these parameters need to be very well documented, commented, and defined what happens with them!

HenkMutsaerts (approved):
New issue created: #1775

PR #1777, opened 2024-09-13
Closes #1776 sticky diary

### LINKED ISSUE #1776: Disable sticky diary file
### Description

If a diary file "sticks" in Matlab's memory from a previous ExploreASL run (or another software for that matter) it can cause issues down the road as we check with xASL_TrackProgress if a diary is on, turn it off and turn it on again. If that diary file is in the meantime removed, we will get an error. So out of precaution, we disable any diary at the initialization of ExploreASL

### Tasks

- [x] ExploreASL_Initialize: disable diary by default

### Release notes

Fix bug with sticky/unclosed Matlab diary

HenkMutsaerts:
### Linked issue

Closes #1776

### DIFF
--- a/ExploreASL_Initialize.m
+++ b/ExploreASL_Initialize.m
@@ -34,6 +34,16 @@ function [x] = ExploreASL_Initialize(varargin)
 
 
     %% 1. Admin
+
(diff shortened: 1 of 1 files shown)


jan-petr (approved):
OK

PR #1804, opened 2024-10-01
Bug #1788 test data set bugs

### LINKED ISSUE #1788: TestDataSet
### Description

Has several errors that need fixing

### Tasks
- [x] Fix Siemens_PCASL_3DGRASE_VE11C_DEBBIE_2 import on Luna - done in issue #1800 

image

[BEA: I can't move past the import module because of this bug, but if I just want to re-run the quantification of and LCBC subject with a different T2 value, I added a printscreen below of what is happening. If you have any idea let me know Jan]

- [x] GE_3Dspiral_misaligned_LoQ: It ignores that bUseMNIasDummyStructural is provided. Fixed previously in TestDataSets.
- [x] xASL_stat_GetDICOMStatistics fails in some datasets as there's a wrong struct for printing at the end. 
7a79ac1
- [x] Fix Philips_2DEPI_FLAIR_Segm_PWI4D_Quant_Parameters_LoQ - fixed dataPar in TestDataSets
- [x] Fix Philips_2DEPI_FLAIR_Segm_PWI4D_Quant_Parameters_LoQ - error in TC calculation - dataPar was in second and not miliseconds. Fix in TestDataSet repository.
- [x] Update the Table with new reference values (GE_3Dspiral_misalig)
- [x] Check values against older reference (Siemens3DGRASE_Stru)
- [x] Comparison returns all nans
- [x] Population module tissue masking and atlases done in #1801 
- [x] Run a final test.

### Release notes

Added extra GE TestDataSet.

jan-petr:
### Linked issue

Close #1788

### DIFF
--- a/Modules/Module_Processing/Module_Population/xASL_stat_GetDICOMStatistics.m
+++ b/Modules/Module_Processing/Module_Population/xASL_stat_GetDICOMStatistics.m
@@ -89,6 +89,7 @@ for iSubject=1:x.dataset.nSubjects
 					newField = strjoin(newField, '/');
 				elseif isstruct(newField)
 					% Structs are also not allowed
+					fprintf('Warning, field %s contains a struct, so we cannot print it in the DICOM tag overview\n', jsonFields{iField});
 					newField = 'n/a';
 				end
                 % Check if field already exists as column
--- a/Modules/Module_Processing/Module_Population/xASL_stat_GetDICOMStatistics.m
+++ b/Modules/Module_Processing/Module_Population/xASL_stat_GetDICOMStatistics.m
@@ -81,35 +81,31 @@ for iSubject=1:x.dataset.nSubjects
         nFields = length(jsonFields);
         for iField=1:nFields
             if ~isempty(Parms.(jsonFields{iField}))
+				% Saves the new field for further testing and formatting
+				newField = Parms.(jsonFields{iField});
+
+				if iscell(newField)
+					% Verify that all gathered cells include strings and not cells inside
+					newField = strjoin(newField, '/');
+				elseif isstruct(newField)
+					% Structs are also not allowed
+					newField = 'n/a';
+				end
                 % Check if field already exists as column
                 iColumn = find(strcmp(TSV(1,:), jsonFields{iField}));
                 if isempty(iColumn)
                     TSV{1,end+1} = jsonFields{iField}; % create new column for this parameter
-                    TSV{1+iSubjSess,end} = xASL_num2str(Parms.(jsonFields{iField}), [], [], [], true); % add the value to the new column, and run unique
+                    TSV{1+iSubjSess,end} = xASL_num2str(newField, [], [], [], true); % add the value to the new column, and run unique
                 elseif numel(iColumn)>1
                     error('When making the DICOMparameters.TSV, columns with the same name');
                 else
-                    TSV{1+iSubjSess,iColumn} = xASL_num2str(Parms.(jsonFields{iField}), [], [], [], true); % add the value to the existing column, and run unique
+                    TSV{1+iSubjSess,iColumn} = xASL_num2str(newField, [], [], [], true); % add the value to the existing column, and run unique
                 end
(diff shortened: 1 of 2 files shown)


Modules/Module_Processing/Module_Population/xASL_stat_GetDICOMStatistics.m
```diff
@@ -98,6 +98,14 @@ function xASL_stat_GetDICOMStatistics(x, ScanType, HasSessions, bOverwrite)
 	% Replace all cells by a string created by concatenating the strings inside
 	TSV{listCells(indexCells)} = strjoin(TSV{listCells(indexCells)}, '/');
 end
+% Find all columns with struct in it
```
HenkMutsaerts: Would be nicer if we can disentangle the structs to columns, if they are numel(struct)<20 . Because if this includes e.g. multiple PLDs, we want to see them here. This function generates a TSV overview that users can use to compare different subjects to verify that there were no sequence changes along the acquisition.

Ideally this struct unwrapping to string columns happens inside the loop above — i.e. on the single subject level — because the struct can contain different fields for different subjects?

Thinking about it, the input for this function is NOT dicom fields anymore but the BIDS ASL4D.JSONs. So the fields that there are, are well controlled. So it would be helpful to check which JSON field caused the problem for Bea, and ensure that this ends up in this TSV table.

Especially since this is a main QC point, that avoids the user (or us for clinical trials!) going through all the JSONs one-by-one, I do think we should fix this now, even though this may take some time.

HenkMutsaerts (changes requested):
Single issue

Modules/Module_Processing/Module_Population/xASL_stat_GetDICOMStatistics.m
jan-petr: Fully agree. And I rewrote how we check for cells and structs.

In this special case - we simply had config structure from x-struct like structural/asl/segmentCAT12 etc. No clue how did it propagate here. But that's something we can safely ignore. 

I agree that we can unpack some BIDS structs, but in BIDS, everything is in a single layer, so this should not happen.

I thus propose to keep this slightly better handling of cells and struct and get back to the problem when we have an example that is relevant for unpacking. Because all we have now is being dealt with fine (cells are unpacked to strings, struct is x-struct and should be ignored cause that is a lot of fields).

HenkMutsaerts: I see, then issuing a warning is probably best for now? As you say something wrong happened, and it would be good if we can check how this happened (although not too important) and we should not print the structs.

Modules/Module_Processing/Module_Population/xASL_stat_GetDICOMStatistics.m
```diff
@@ -76,36 +76,32 @@ function xASL_stat_GetDICOMStatistics(x, ScanType, HasSessions, bOverwrite)
         nFields = length(jsonFields);
         for iField=1:nFields
             if ~isempty(Parms.(jsonFields{iField}))
+				% Saves the new field for further testing and formatting
+				newField = Parms.(jsonFields{iField});
+
+				if iscell(newField)
+					% Verify that all gathered cells include strings and not cells inside
```
HenkMutsaerts, line 88: I think we do this already with xASL_num2str below, but this probably doesn't hurt?
Or is xASL_num2str better because it can contain numbers which would not correctly be converted to strings here?

HenkMutsaerts (approved):
minor check needed, let me know if you need my help with this but probably you can easily solve this and merge

Modules/Module_Processing/Module_Population/xASL_stat_GetDICOMStatistics.m
jan-petr: This is actually not about numbers but about strings. So a DICOM field having a couple of strings, which then gets converted to a cell array and we only need to concatenate to strings again :)

PR #1805, opened 2024-10-01
#1766 wrp_DCM2NII_Subject warning multiple scans

### LINKED ISSUE #1766: xASL_imp_CreateSummaryFile crashes with too many missing scans per single Subject/Visit/Session

jan-petr:
### Linked issue

Close #1766

### DIFF
--- a/Modules/Module_Import/xASL_wrp_DCM2NII_Subject.m
+++ b/Modules/Module_Import/xASL_wrp_DCM2NII_Subject.m
@@ -105,8 +105,8 @@ function [x, PrintDICOMFields, dcm2niiCatchedErrors] = xASL_wrp_DCM2NII_Subject(
             %% 4. Iterate over scans
 			% Safety-check, we can't allow multiple matches for the same sequence
 			if numel(thisRun.scanIDs) ~= numel(unique(thisRun.scanIDs))
-				fprintf('Detected duplicit scans: %s\n',strjoin(thisRun.scanIDs, ', '));
-				error('We do not allow multiple matches per scan for. Please delete some sequences or adjust sourcestructure.json');
+				fprintf('Detected duplicate scans: %s\n',strjoin(thisRun.scanIDs, ', '));
+				error('We do not allow multiple token matches per scantype. Please adjust sourcestructure.json');
 			end
 
             for iScan=1:numel(thisRun.scanIDs)
--- a/Modules/Module_Import/xASL_wrp_DCM2NII_Subject.m
+++ b/Modules/Module_Import/xASL_wrp_DCM2NII_Subject.m
@@ -103,6 +103,12 @@ function [x, PrintDICOMFields, dcm2niiCatchedErrors] = xASL_wrp_DCM2NII_Subject(
             end
             
             %% 4. Iterate over scans
+			% Safety-check, we can't allow multiple matches for the same sequence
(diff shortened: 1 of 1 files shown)


Modules/Module_Import/xASL_wrp_DCM2NII_Subject.m
```diff
@@ -103,6 +103,12 @@
             end
             
             %% 4. Iterate over scans
+			% Safety-check, we can't allow multiple matches for the same sequence
+			if numel(thisRun.scanIDs) ~= numel(unique(thisRun.scanIDs))
+				fprintf('Detected duplicit scans: %s\n',strjoin(thisRun.scanIDs, ', '));
+				error('We do not allow multiple matches per scan for. Please delete some sequences or adjust sourcestructure.json');
```
HenkMutsaerts: should be

error('We do not allow multiple token matches per scantype. Please adjust sourcestructure.json');

For clarity, but most importantly users should NOT delete sourcedata! Ideally this is always readonly, because from the philosophy of BIDS sourcedata is always exactly what you got, without restructuring etc. I know this is not what we do yet for practical reasons, but you get my point :)

Modules/Module_Import/xASL_wrp_DCM2NII_Subject.m
```diff
@@ -103,6 +103,12 @@
             end
             
             %% 4. Iterate over scans
+			% Safety-check, we can't allow multiple matches for the same sequence
+			if numel(thisRun.scanIDs) ~= numel(unique(thisRun.scanIDs))
+				fprintf('Detected duplicit scans: %s\n',strjoin(thisRun.scanIDs, ', '));
```
HenkMutsaerts: duplicit -> duplicate

No, “duplicit” is not a correct word in English. The correct word you’re thinking of is “duplicate”, which refers to something that is an exact copy or reproduction of another item.

There might be confusion with “duplicitous,” which is an adjective used to describe someone as being deceitful or two-faced. However, “duplicit” on its own is not recognized as an English word.

Thanks chatGPT :)

HenkMutsaerts (approved):
Two minor comments that you can easily fix without me having to re-approve

Modules/Module_Import/xASL_wrp_DCM2NII_Subject.m
jan-petr: nice. fixed.

Modules/Module_Import/xASL_wrp_DCM2NII_Subject.m
jan-petr: good point. fixed.

PR #1806, opened 2024-10-01
#1801 Fix tissue mask warning and defaults

### LINKED ISSUE #1801: Population mask and tissue input check
### Description

Population module tissue masking and atlases sometimes creates an error when mask and tissue input doesn't match. This should be catched and correctly reported. And not crashing the code.

image

### Tasks

- [x] Inform Mathijs

### Release notes
Nothing to report. Report along with #1753

jan-petr:
### Linked issue

Close #1801

### DIFF
--- a/Modules/Module_Processing/xASL_init_LoadDataPar.m
+++ b/Modules/Module_Processing/xASL_init_LoadDataPar.m
@@ -143,7 +143,7 @@ end
 function [x] = xASL_initLoadDataPar_PrepareAtlas4ROI(x)
 %Check for x.S.Atlases & x.S.TissueMasking and provide default values or print instructions
 
-bAtlasTissueMatch = true; % Necessary condition
+bAtlasTissueMatch = true; % x.S.Atlases & x.S.TissueMasking should match before we can continue (needed for the population module)
 
 if ~isfield(x, 'S') || (~isfield(x.S,'Atlases') && ~isfield(x.S, 'TissueMasking'))
 	% Default atlases/ROIs & tissue masks if nothing is provided
@@ -157,7 +157,7 @@ elseif ~isfield(x.S, 'Atlases') && isfield(x.S, 'TissueMasking')
     bAtlasTissueMatch = false;
 elseif isfield(x.S, 'Atlases') && isfield(x.S, 'TissueMasking') && length(x.S.Atlases)~=length(x.S.TissueMasking)
 	% Non matching lengths, cannot continue
-    warning('The same number of ROI atlases as subject-wise tissue-types provided does not match:');
+    warning('The number of ROI atlases x.S.Atlases as subject-wise tissue-types x.S.TissueMasking provided does not match:');
     fprintf('%s\n', ['S:{Atlases:["' strjoin(x.S.Atlases, '", "') '"]}']);
     fprintf('%s\n', ['S:{TissueMasking:["' strjoin(x.S.TissueMasking, '", "') '"]}']);
     bAtlasTissueMatch = false;
@@ -184,8 +184,8 @@ elseif 	isfield(x.S, 'Atlases') && ~isfield(x.S, 'TissueMasking')
 end
 if ~bAtlasTissueMatch
     fprintf('%s\n', 'When ROI atlases are provided in S.Atlases, their tissue types');
-    fprintf('%s\n', 'need to be provided as well in dataPar.json as S.TissueMasking, with either option ''GM'', ''WM'', ''WB'' (GM+WM).');
-    fprintf('%s\n', 'When not provided, these default to:');
+    fprintf('%s\n', 'need to be provided as well in dataPar.json as S.TissueMasking, with either option ''GM'', ''WM'', ''WB'' (==GM+WM).');
+    fprintf('%s\n', 'The default values when Atlases and Tissues masking are not provided are:');
     fprintf('%s\n', 'S:{Atlases:["TotalGM", "DeepWM"],');
     fprintf('%s\n\n', '  TissueMasking:["GM", "WM"]}');
 
--- a/Modules/Module_Processing/xASL_init_LoadDataPar.m
+++ b/Modules/Module_Processing/xASL_init_LoadDataPar.m
@@ -124,7 +124,9 @@ function [x] = xASL_init_LoadDataPar(x)
     if ~isfield(dataPar.x.settings,'DELETETEMP')
         dataPar.x.settings.DELETETEMP = true;
 	end
-	
+
+	% Check for Atlases and TissueMasking parameters
+	dataPar.x = xASL_initLoadDataPar_PrepareAtlas4ROI(dataPar.x);
 
     %% Write final dataPar
     x.dir.dataPar = fullfile(x.dir.xASLDerivatives, 'dataPar.json');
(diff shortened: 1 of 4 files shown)


Modules/Module_Processing/xASL_init_LoadDataPar.m
```diff
@@ -134,4 +136,61 @@
     %% Load dataPar
     x = xASL_adm_MergeStructs(dataPar.x, x);
     
-end
\ No newline at end of file
+end
+
+%% -----------------------------------------------------------------------------
+%% -----------------------------------------------------------------------------
+function [x] = xASL_initLoadDataPar_PrepareAtlas4ROI(x)
```
HenkMutsaerts, line 143: Shouldn't we keep/move this to the population module where it belongs (same as the check above)?

As it (should have) no effect on the rest of the processing.
But on the other hand, perhaps good to issue warnings here rather than at the end for users that specify this in dataPar.json

Modules/Module_Processing/xASL_init_LoadDataPar.m
```diff
@@ -134,4 +136,61 @@
...
+    x.S.TissueMasking = {'GM' 'WM'}; % GM WM, fits with the TotalGM & DeepWM above
+    % Note that this should be in the same order as the atlases/ROIs
+    % A mismatch (e.g. TissueMasking=GM for Atlases=deepWM) would result in an empty ROI, producing a NaN in the .tsv table
+elseif ~isfield(x.S, 'Atlases') && isfield(x.S, 'TissueMasking')
+	% Missing Atlases, but provided TissueMasking - cannot continue
+    warning('Custom tissue-types (x.S.TissueMasking) specified without ROI atlas-selection (x.S.Atlases). Atlases need to be provided. See instructions below:');
+    bAtlasTissueMatch = false;
+elseif isfield(x.S, 'Atlases') && isfield(x.S, 'TissueMasking') && length(x.S.Atlases)~=length(x.S.TissueMasking)
+	% Non matching lengths, cannot continue
+    warning('The same number of ROI atlases as subject-wise tissue-types provided does not match:');
```
HenkMutsaerts: Should be:

warning('The number of ROI atlases `x.S.Atlases` as subject-wise tissue-types `x.S.TissueMasking` provided does not match:');

Modules/Module_Processing/xASL_init_LoadDataPar.m
```diff
@@ -134,4 +136,61 @@
...
+	% TissueMasking not provided, so it has to be extracted from Atlases as previously.
+    warning('ROIs provided in S.Atlases without the tissue-types for these ROIs in S.TissueMasking.');
+    bAtlasTissueMatch = false;
+	textRecommendation = 'S:{TissueMasking:["';
+	% Fill in TissueMasking based on the atlas names and default to GM. The user will see a warning and automatic tissue masks to verify
+	for iAtlas = 1:numel(x.S.Atlases)
+		if iAtlas > 1
+			textRecommendation = [textRecommendation '", "'];
+		end
+		if ~isempty(regexpi(x.S.Atlases{iAtlas}, 'WM')) || ~isempty(regexpi(x.S.Atlases{iAtlas}, 'whitematter'))
```
HenkMutsaerts, line 174: Nice, though we should be careful as the ignorant user may just copy what we write here which may not be correct. 

Note that the only case where this can be inferred from our ROI atlases is "DeepWM" (now totalGM==a WholeBrain mask already and we should rename in this issue).

For other provided atlases it is risky to infer this. E.g., most people assume that we provide GM values, so if they provide a custom atlas named "HarvardOxfordWholeBrain", they would assume that we analyze GM. They would fill in "WB" as per your recommendation and still think ExploreASL provided them GM CBF.

So, I wonder if we shouldn't just remove this here, and link to the tutorial (which can be a specific part of the tutorial where we explain this). And then I would explain "if you want to analyze GM, write "GM", etc, not 

"as example for an atlas ending with WB, you should write WB"

Modules/Module_Processing/xASL_init_LoadDataPar.m
```diff
@@ -134,4 +136,61 @@
...
+		else
+			textRecommendation = [textRecommendation 'GM'];
+		end
+	end
+	textRecommendation = [textRecommendation '"]}'];
+	fprintf('Recommended addition to the dataPar.json to match provided Atlases: %s\n\n', textRecommendation);
+end
+if ~bAtlasTissueMatch
+    fprintf('%s\n', 'When ROI atlases are provided in S.Atlases, their tissue types');
+    fprintf('%s\n', 'need to be provided as well in dataPar.json as S.TissueMasking, with either option ''GM'', ''WM'', ''WB'' (GM+WM).');
```
HenkMutsaerts: end should be

''WB'' (==GM+WM).');

Modules/Module_Processing/xASL_init_LoadDataPar.m
```diff
@@ -134,4 +136,61 @@
...
+    fprintf('%s\n', 'When not provided, these default to:');
```
HenkMutsaerts: This should either be

fprintf('%s\n', 'When not provided, these would have defaulted to:');

or not print it at all when `~bAtlasTissueMatch`, because it may confuse the reader in thinking that we default to this (which we don't necessarily when you issue this message)

Modules/Module_Processing/xASL_init_LoadDataPar.m
```diff
@@ -134,4 +136,61 @@
...
-end
\ No newline at end of file
+end
+
+%% -----------------------------------------------------------------------------
+%% -----------------------------------------------------------------------------
+function [x] = xASL_initLoadDataPar_PrepareAtlas4ROI(x)
+%Check for x.S.Atlases & x.S.TissueMasking and provide default values or print instructions
+
+bAtlasTissueMatch = true; % Necessary condition
```
HenkMutsaerts: rename comment to 

% x.S.Atlases & x.S.TissueMasking should match before we can continue (needed for the population module)

"necessary condition" confused me at least :)

Modules/xASL_module_Population.m
```diff
@@ -30,11 +30,11 @@
 % EXAMPLE: [~, x] = xASL_module_Population(x);
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
 % Copyright 2015-2024 ExploreASL
-% Licensed under Apache 2.0, see permissions and limitations at
-% https://github.com/ExploreASL/ExploreASL/blob/main/LICENSE
-% you may only use this file in compliance with the License.
-% __________________________________
-
+% Licensed under Apache 2.0, see permissions and limitations at
```
HenkMutsaerts: I see you had the same problem. Probably Matlab defaulting to different line endings in the m-files than how I programmed it with the cell splitting and joining thingy

Modules/xASL_module_Population.m
```diff
@@ -48,7 +48,17 @@
     return;
 end
 
-x = xASL_wrp_Population_PrepareAtlas4ROI(x); % Parse x.S.Atlases & x.S.TissueMasking
+% Check again for Atlases (main checking is done when loading dataPar)
```
HenkMutsaerts, line 51: I still tricky because "advanced" users my just run without specifying in the dataPar, start debugging mode here, and provide the x.S.Atlases & x.S.TissueMasking here. 

Perhaps it is best practice to move the checking we did here to a separate function, and calling this upon dataPar loading as well as just before running xASL_wrp_GetRoi...?

Testing/UnitTests/xASL_ut_master_xASL_ExploreASL.m
```diff
@@ -530,6 +530,7 @@
 % Create dataPar.json
 dataParStruct.x.settings.Quality = 0;
 dataParStruct.x.S.Atlases = {'TotalGM','DeepWM','Hammers','HOcort_CONN','HOsub_CONN','Mindboggle_OASIS_DKT31_CMA'};
```
HenkMutsaerts, line 532: Should we also unit test WB and a bit more WM?

HenkMutsaerts (changes requested):
Very nice! Few comments, and I think we should update the names totalGM -> WholeBrain (like we have for Supratentorial)

Modules/Module_Processing/xASL_init_LoadDataPar.m
jan-petr: We had it in the population module. And you asked to move this to the dataPar loading ;)

So I'd stick with the original comment that we want to warn the user about a basic wrong input right at the start.

Modules/Module_Processing/xASL_init_LoadDataPar.m
jan-petr: It was you who wrote that :) I just copy-pasted it here :)

But sure, changed...

Modules/Module_Processing/xASL_init_LoadDataPar.m
jan-petr: Good catch. Fixed.

Modules/Module_Processing/xASL_init_LoadDataPar.m
jan-petr: Yes, we should write this carefully and explain them the caveats. But to make their life easier, I also want to prepare the exact line they should add. So I'd like to keep the example that is adjusted to the correct length. Otherwise, they keep struggling with putting wrong lengths etc. Please suggest a text that explains that they should check the tissue types. Note that this recommendation is printed addiotional to referring them to the manual and additional to printing the instructions. So without this suggestion, the ignorant user will keep copy-pasting the example from the manual...

Modules/Module_Processing/xASL_init_LoadDataPar.m
jan-petr: done

Modules/Module_Processing/xASL_init_LoadDataPar.m
jan-petr: your code :)

I can remove the the default and I can remove my recommended hint. But then the user will struggle with not having any examples. And we wanted to give the user some examples. 

So see how I rewrote it...

Modules/xASL_module_Population.m
jan-petr: could be. lets ignore this as a minor nuissance...

Modules/xASL_module_Population.m
jan-petr: If the user doesn't provide dataPar, then default values will be used. So if the same user puts a breakpoint here and adds incorrect values here, it will check it below and create an error (so exit correctly), but it won't print him the manual on how to do this correctly.

I'd say that if somebody does anything like that, he doesn't need to have the manual printed. I don't think the full check function is needed here again as we have the full check of input validity here, which should be enough....

Testing/UnitTests/xASL_ut_master_xASL_ExploreASL.m
jan-petr: WB added.

Modules/xASL_module_Population.m
HenkMutsaerts: Also fine, I think in general it doesn't hurt to have such checks twice (hardly any time inefficiency but more warnings can still help us avoid explaining stuff). But you can choose and resolve.

Modules/Module_Processing/xASL_init_LoadDataPar.m
HenkMutsaerts: Yes I think it is the combination of the location that you moved this to. My point was that if you have 

~bAtlasTissueMatch

we do NOT default to anything, so this confuses the user.

jan-petr: OK. so I removed it - fully up to you. As I say - you wrote that warning code in that exact location ;)

Modules/Module_Processing/xASL_init_LoadDataPar.m
HenkMutsaerts: warning('x.S.TissueMasking input did not match x.S.Atlases');
fprintf(%s\n, 'If you want GM CBF from an atlas ROI, add '''GM'''. If you want WM CBF from an atlas ROI, add '''WM'''');
fprintf(%s\n, 'If you want both GM CBF and WM CBF from an atlas ROI, add the atlas ROI twice and add both '''GM''' and '''WM'''');
fprintf(%s\n, 'E.g., if you want both GM CBF and WM CBF from the Harvard-Oxford atlas, you would provide the following:');
fprintf(%s\n, 'x.S.Atlases = {''''HOConn''' ''''HOConn'''});
fprintf(%s\n, 'x.S.TissueMasking = {''''GM''' ''''WM'''});

Or something similar.
My main point is that we cannot advice the user based on trying to detect what they want based on the atlas name, we should remove the regexpi(x.S.Atlases{iAtlas}) etc.

We rather want the user to puzzle more or ask us, then somebody reporting WM CBF as GM CBF because they just copy-pasted and didn't understand it.

HenkMutsaerts (changes requested):
Single comment

Modules/Module_Processing/xASL_init_LoadDataPar.m
jan-petr: Warning added as proposed.

HenkMutsaerts (approved):

PR #1809, opened 2024-10-03
Closes #1797 xASL_wrp_Quantify: add xASL_quant_AgeSex2Hct

### LINKED ISSUE #1797: Add AgeSex2Hct option & hct revamp
### Description

We realized that the function xASL_quant_AgeSex2Hct is not anywhere implemented in ExploreASL:
Screenshot 2024-09-27 at 15 21 19

So, we have agreed (27 September 2024) to add it to xASL_wrp_Quantify, in this part:
Screenshot 2024-09-27 at 12 31 54

### Tasks

- [x] Add xASL_quant_AgeSex2Hct to xASL_wrp_Quantify
- [x] Allow user to disable xASL_quant_AgeSex2Hct
- [x] Move hct parsing to xASL_quant_DefineQuantificationParameters
- [x] CleanUp x.Q at the start of xASL_module_ASL (check that this doesn't hurt stuff)
- [x] Jan adds to documentation: x.modules.asl.bHct2BloodT1. 0 = off, 1== use hct 2==use age & sex. DEFAULT = use hct if exist, use age&sex if exist, otherwise disabled (in this order)
- [x] Add bHct2BloodT1 into parameter loading
- [x] Backwards compatibility is removed, now only hematocrit from participants.tsv -> x.S.Sets* is used

### Testing

- [x] Henk tested with FlavorDatabase/Siemens_PCASL_2DEPI_E11_1

### Release
Improved hematocrit->bloodT1 management, including the ability to infer this from age & sex

HenkMutsaerts:
### Linked issue

Closes #1797 xASL_wrp_Quantify: add xASL_quant_AgeSex2Hct


BeatrizPadrela (approved):

Modules/Module_Processing/Module_ASL/xASL_wrp_Quantify.m
```diff
@@ -205,7 +207,40 @@ function xASL_wrp_Quantify(x, PWI4D_Path, pathOutputCBF, M0Path, SliceGradientPa
...
+    IndexSetsSex = find(strcmpi(x.S.SetsName, 'sex'));
+    if ~isempty(IndexSetsAge) || ~isempty(IndexSetsSex)
+        fprintf('%s\n', 'Trying to infer hematocrit from age & sex');
+        
+        age = x.S.SetsID(:, IndexSetsAge);
+        
+        % Convert sex correctly
+        sex = x.S.SetsID(:, IndexSetsSex);
+        sexOptions = x.S.SetsOptions{:, IndexSetsSex};
+        sexN = zeros(length(sex), 1)
```
jan-petr: We set everything to 0, then set some values to 1 or 2. And in the end set everything that isn't 1 or 2 to NaN.

It's much easier to set sexN to NaNs here at 223. And then remove line 231.

Modules/Module_Processing/Module_ASL/xASL_wrp_Quantify.m
```diff
@@ -205,7 +207,40 @@ function xASL_wrp_Quantify(x, PWI4D_Path, pathOutputCBF, M0Path, SliceGradientPa
...
+        for iOption=1:length(sexOptions)
+            if ~isempty(regexpi(sexOptions{iOption}, '^(male|m|man|men)$'))
+                sexN(sex==iOption) = 1;
```
jan-petr: I understand that this value is only used internally and that input in participants tsv is male/female. But it would be nicer to follow the statistical convention and define binary categorical variables like sex as 0/1 and not 1/2.

jan-petr (changes requested):
Minor comments.

Modules/Module_Processing/Module_ASL/xASL_wrp_Quantify.m
HenkMutsaerts: Shall we restart the same discussion? ;)

You cannot use 0 as an index so this will not work with the current setup. The way it works now, it saves x.S.SetsOptions, and the value 1 points to x.S.SetsOptions{1} etc. x.S.SetsOptions{1} etc can be 0 1 or male female or 1 2 whatever you want. So for categorical variables (binary or three values or whatever) are all stored as indices in x.S.SetsID.

Modules/Module_Processing/Module_ASL/xASL_wrp_Quantify.m
HenkMutsaerts: Fixed in b7f4265a627f48c78dae39af4356a335ff27a3f8

jan-petr (changes requested):
All is good. 
One last issue - if we provide Hematocrit, then we do automatic Hematocrit correction. Which is fine for me. 
But if the user provides sex+age, it also runs an automatic Hematocrit correction using two literature models "HCT->BloodT1" and "Sex+age->Hct". Perhaps that's not something the user will want! I'd say that there should be an option to turn this OFF/ON or some warnings.

I would probably turn this off by default, at least based on age+sex. But I don't have a strong opinion. Just think about it.

HENK: Good point, I now added a warning when age & sex are found, that it will by default change Hematocrit accordingly. And add a parameter that the user can disable this. And I moved this to the start of xASL_module_ASL such that these warnings are catched early. Can you add this to the documentation?

Modules/Module_Processing/Module_ASL/xASL_wrp_Quantify.m
```diff
@@ -187,70 +187,95 @@ function xASL_wrp_Quantify(x, PWI4D_Path, pathOutputCBF, M0Path, SliceGradientPa
...
-% We convert x.Hematocrit > x.hematocrit 
-% -> this accounts for the (backward compatibility) possibility
-% that the user provides hematocrit in dataPar.json
-if isfield(x, 'hematocrit') && isfield(x, 'Hematocrit')
-    warning('Two hematocrit fields, ignoring x.hematocrit');
-    x = rmfield(x, 'hematocrit');
-elseif isfield(x, 'hematocrit')
-    x.Hematocrit = x.hematocrit;
-	x = rmfield(x, 'hematocrit');
+if isfield(x, 'Hematocrit')
```
jan-petr: See other comments in the issue. I propose that x.Hematocrit is already ignored directly after reading dataPar.json. 

In participants.tsv, we can couple values and subject IDs. In dataPar.json not really, so lets forbid the option to add it in dataPar.json completely. I don't think anybody's using it.

Modules/Module_Processing/Module_ASL/xASL_wrp_Quantify.m
```diff
@@ -187,70 +187,95 @@ function xASL_wrp_Quantify(x, PWI4D_Path, pathOutputCBF, M0Path, SliceGradientPa
...
-                sexN(sex==iOption) = 2;
-            end
-        end
 
-        if sum(isnan(sexN))>0
-            warning('Unknown sex detected in participants.tsv, currently we can only infer hematocrit from male or female');
-            fprintf('%s\n', 'So in participants.tsv specify the words "male" and "female" only');
+% a2. Manage hematocrit usage parameter
+if isfield(x.modules.asl.bHct2BloodT1)
+    if x.modules.asl.bHct2BloodT1 == 2 && isempty(x.Q, 'Hematocrit')
```
jan-petr: You have "isempty" So hematocrit was found! Not that it wasn't found as you claim in the comment.

Modules/Module_Processing/Module_ASL/xASL_wrp_Quantify.m
```diff
@@ -187,70 +187,95 @@ function xASL_wrp_Quantify(x, PWI4D_Path, pathOutputCBF, M0Path, SliceGradientPa
...
+        warning('Parameter x.modules.asl.bHct2BloodT1 was set to 2: trying to infer hematocrit from age and sex, but hematocrit data were also found');
+        fprintf('%s\n', 'Consider setting x.modules.asl.bHct2BloodT1 to 1 to use the hematocrit data directly');
+    end
+    if x.modules.asl.bHct2BloodT1 == 2 && (isempty(IndexSetsAge) || isempty(IndexSetsSex))
+        warning('Parameter x.modules.asl.bHct2BloodT1 was set to 2: trying to infer hematocrit from age and sex, but age & sex data were incomplete');
+        fprintf('%s\n', 'Consider changing x.modules.asl.bHct2BloodT1 or ensure that age & sex data are present in participants.tsv');
+        x.modules.asl.bHct2BloodT1 = []; % Setting this to empty here, so it will be dealt with below
+    end
+
+if ~isfield(x.modules.asl.bHct2BloodT1) || isempty(x.modules.asl.bHct2BloodT1)
```
jan-petr: we are inside "if isfield(x.modules.asl.bHct2BloodT1)" is the field exist always here and you can remove the first condition.

Modules/Module_Processing/Module_ASL/xASL_wrp_Quantify.m
```diff
@@ -187,70 +187,95 @@ function xASL_wrp_Quantify(x, PWI4D_Path, pathOutputCBF, M0Path, SliceGradientPa
...
+        fprintf('%s\n', 'Setting x.modules.asl.bHct2BloodT1 to 1: converting hematocrit to blood T1 values');
+        fprintf('%s\n', 'Consider setting x.modules.asl.bHct2BloodT1 to avoid this warning');
+        x.modules.asl.bHct2BloodT1 = 1;
+    elseif ~isempty(IndexSetsAge) || ~isempty(IndexSetsSex)
+        warning('Parameter x.modules.asl.bHct2BloodT1 was not set but age & sex data were found');
+        fprintf('%s\n', 'Setting x.modules.asl.bHct2BloodT1 to option 2: trying to infer hematocrit from age & sex');
+        fprintf('%s\n', 'Consider setting x.modules.asl.bHct2BloodT1 to avoid this warning');
+        x.modules.asl.bHct2BloodT1 = 2;
+    else
+        fprintf('\n%s\n', 'No hematocrit data found, disabling x.modules.asl.bHct2BloodT1');
```
jan-petr: what if bHct2BloodT1 was set to 0 already, then this printf about disabling isn't necessary as things were already disabled.

Modules/Module_Processing/Module_ASL/xASL_wrp_Quantify.m
```diff
@@ -187,70 +187,95 @@ function xASL_wrp_Quantify(x, PWI4D_Path, pathOutputCBF, M0Path, SliceGradientPa
...
-            elseif ~isempty(regexpi(sexOptions{iOption}, '^(female|f|woman|women)$'))
-                sexN(sex==iOption) = 2;
-            end
-        end
 
-        if sum(isnan(sexN))>0
-            warning('Unknown sex detected in participants.tsv, currently we can only infer hematocrit from male or female');
-            fprintf('%s\n', 'So in participants.tsv specify the words "male" and "female" only');
+% a2. Manage hematocrit usage parameter
+if isfield(x.modules.asl.bHct2BloodT1)
```
jan-petr: You are checking existence of bHct2DBloodT1. But not checking existence of asl field.

This should be done already in the ASL module - setting it to [] if missing.

Modules/Module_Processing/Module_ASL/xASL_wrp_Quantify.m
```diff
@@ -187,70 +187,95 @@ function xASL_wrp_Quantify(x, PWI4D_Path, pathOutputCBF, M0Path, SliceGradientPa
...
+    age = x.S.SetsID(:, IndexSetsAge);
+    
+    % Convert sex correctly
+    sex = x.S.SetsID(:, IndexSetsSex);
+    sexOptions = x.S.SetsOptions{:, IndexSetsSex};
+    sexN = nan(length(sex), 1); % currently we can only infer hct from male/female
+    % So anything that is not detected, will remain NaNs
+    for iOption=1:length(sexOptions)
+        if ~isempty(regexpi(sexOptions{iOption}, '^(male|m|man|men)$'))
+            sexN(sex==iOption) = 1;
```
jan-petr: I understand that previously, we used it for indexing. But here, we can use sex only internally and therefore we can use the 0/1 form and not 1/2. Or am I missing something?

jan-petr (changes requested):
A few things need clarification.

Modules/xASL_module_ASL.m
```diff
@@ -577,6 +577,10 @@
 
 
 %% 1. Load ASL parameters (inheritance principle)
+
+% First ensure that we don't use quantification parameters from other sequences
+x.Q = struct;
```
jan-petr, line 607: From previously. We have a rather complicated system of saving things to x and then loading them. Which includes also structural data that are either added here after the structural module or reloaded. 

Completely deleting all that's in x.Q is likely going to have important negative consequences!

jan-petr (changes requested):
Check this and also all the older comments.

Modules/Module_Processing/Module_ASL/xASL_wrp_Quantify.m
HenkMutsaerts: In the new location (xASL_quant_DefineQuantificationparameters), we don't read from x.Hematocrit except for issuing a warning when x.Hematocrit was found. We used to do this in the past, by loading a separate hematocrit.mat within dataPar (that's why it used to be a .m-file). So this should be fine now? We don't need to have a list of parameters to exclude when reading dataPar.json?

Modules/Module_Processing/Module_ASL/xASL_wrp_Quantify.m
HenkMutsaerts: Thanks for the catch, fixed in c2e1f53882a9e4e48ef52e4417d86bbccf38b2e8

Modules/Module_Processing/Module_ASL/xASL_wrp_Quantify.m
HenkMutsaerts: Not applicable anymore in the new location?

Modules/Module_Processing/Module_ASL/xASL_wrp_Quantify.m
HenkMutsaerts: This line is only parsed if bHct2BloodT1 was not defined

Modules/Module_Processing/Module_ASL/xASL_wrp_Quantify.m
HenkMutsaerts: The ASL module uses x.modules.asl already at line 80, far before calling this quantification stuff, so it should already crash there. I believe we initialize x.module.asl in processing settings? Anyway, not something that needs changing now.

Modules/Module_Processing/Module_ASL/xASL_wrp_Quantify.m
HenkMutsaerts: This is only for checking if participants.tsv contains the words male and female only, as we currently do not parse anything else, to avoid confusion. And it has an indexing role (sort of). It would only confuse if at some locations we use 0-1 and others 1-2 so I would keep it like this.

Modules/xASL_module_ASL.m
HenkMutsaerts: I checked this, x.Q is only used for ASL quantification parameters, that we always reload after this line from ASL4d.json sidecars etc. And we had this already, just not at this earlier point. So we don't change things except from ensuring that settings from previous module runs are not propagated to new module runs (each module run should be independent of other module runs).

jan-petr (approved):
All fixed.

PR #1811, opened 2024-10-07
#1807 Minor fixes

### LINKED ISSUE #1807: DcmTk issues
### Description

I ran `xASL_adm_SortDicomToFolders` on macOS, which usually runs fine (perhaps a bit slower because of the SPM DICOM reading but otherwise fine).

When reading a dicom with DcmTk, if it fails, it will produce one message per DICOM, this needs catching and producing only a single warning.

>> xASL_adm_SortDicomToFolders('/Users/hjmutsaerts/ExploreASL/ASL/RogerMcIntosh/008/DICOM/23092619'00%Warning, reading a DICOM with DCMTK failed. Will continue with SPM. Warning message: Undefined function 'xASL_mex_DcmtkRead' for input arguments of type 'char

Then it gives a warning that it cannot print it fully (we can probably ignore this because solving the repetitive warning will solve this as well, this is due to the `xASL_adm_TrackProgress` trying to print a tracker on the same line, so warnings also go to a single line.

>> Text exceeds maximum line length for Command Window displa13... Output truncated. 

Then we get an error:

>> Unrecognized field name "PixelValueTransformationSequence".

Error in xASL_io_DcmtkRead>xASL_io_DcmtkRead_TrimSPM (line 237)
		pixelItem  = header.PerFrameFunctionalGroupsSequence{1}.PixelValueTransformationSequence{1};

Error in xASL_io_DcmtkRead (line 61)
		header = xASL_io_DcmtkRead_TrimSPM(header{1});

Error in xASL_adm_SortDicomToFolders (line 56)
		tDcm = xASL_io_DcmtkRead(Flist{iL}, false, bUseDCMTK, true);

I can share the dataset that caused this error if needed.

### Tasks

- [x] Avoid repeating the same warning for each DICOM (I believe we already solve this in another issue?) Yes, solved elsewhere...
- [x] Fix unrecognized field name "PixelValueTransformationSequence" by SPM (or fix DcmTk reading on macOS, Jan has macOS himself as well now ;)

### Release notes

Minor bugfix of DCMTK reading

jan-petr:
### Linked issue

Close #1807

### DIFF
--- a/Functions/Administration/xASL_adm_SortDicomToFolders.m
+++ b/Functions/Administration/xASL_adm_SortDicomToFolders.m
@@ -47,50 +47,53 @@ if nDirLayers > 0
 	end
 else
 	% Gather a list of all files in the directory
-	Flist   = xASL_adm_GetFileList(pathDICOM,'^.*.(?!(xlsx|ini|json))$','FPListRec',[0 Inf]);
+	Flist   = xASL_adm_GetFileList(pathDICOM,'.*','FPListRec',[0 Inf]);
 	for iL=1:length(Flist)
-		% Track progress if verbosity is on
-		if bVerbose; xASL_TrackProgress(iL, length(Flist)); end
+		[~, ~, Fext] = xASL_fileparts(Flist{iL});
+		if isempty(regexp(Fext, '(xlsx|ini|json|DS_Store)'))
+			% Track progress if verbosity is on
+			if bVerbose; xASL_TrackProgress(iL, length(Flist)); end
 
-		% Read the DICOM file - all error handling is inside this function
-		tDcm = xASL_io_DcmtkRead(Flist{iL}, false, bUseDCMTK, true);
+			% Read the DICOM file - all error handling is inside this function
+			tDcm = xASL_io_DcmtkRead(Flist{iL}, false, bUseDCMTK, true);
 
-		% Check that the header was read and contains the basic tags
-		if isempty(tDcm) || ~isfield(tDcm, 'EchoTime') || ~isfield(tDcm, 'RepetitionTime') ||...
-				~isfield(tDcm, 'ImageType') || isempty(tDcm.ImageType) || ~isfield(tDcm, 'ProtocolName') || isempty(tDcm.ProtocolName)
-			warning(['Incomplete DICOM header: ' Flist{iL}]);
-		else
-			% Always add the protocol name to the directory name
-			Fname = tDcm.ProtocolName;
-
-			% Add series description if available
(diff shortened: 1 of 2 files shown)


Functions/Administration/xASL_adm_SortDicomToFolders.m
```diff
@@ -47,50 +47,53 @@ function xASL_adm_SortDicomToFolders(pathDICOM, nDirLayers, bUseDCMTK, bVerbose)
 	end
 else
 	% Gather a list of all files in the directory
-	Flist   = xASL_adm_GetFileList(pathDICOM,'^.*.(?!(xlsx|ini|json))$','FPListRec',[0 Inf]);
+	Flist   = xASL_adm_GetFileList(pathDICOM,'.*','FPListRec',[0 Inf]);
```
HenkMutsaerts, line 50: No, you want to get all files but not xls, ini, or json, so this change should be reverted

Functions/Administration/xASL_adm_SortDicomToFolders.m
```diff
@@ -47,50 +47,53 @@ function xASL_adm_SortDicomToFolders(pathDICOM, nDirLayers, bUseDCMTK, bVerbose)
...
 	for iL=1:length(Flist)
-		% Track progress if verbosity is on
-		if bVerbose; xASL_TrackProgress(iL, length(Flist)); end
+		[~, ~, Fext] = xASL_fileparts(Flist{iL});
+		if isempty(regexp(Fext, '(xlsx|ini|json|DS_Store)'))
```
HenkMutsaerts, line 53: I don't see why you need to move it here.

Functions/Administration/xASL_adm_SortDicomToFolders.m
```diff
@@ -47,50 +47,53 @@ function xASL_adm_SortDicomToFolders(pathDICOM, nDirLayers, bUseDCMTK, bVerbose)
...
+			% Track progress if verbosity is on
+			if bVerbose; xASL_TrackProgress(iL, length(Flist)); end
```
HenkMutsaerts, line 55: The rest seems all the same. To me, this is a useless fix. Unless you explain me why this was needed ;)
The only thing I see is the index file DS_Store that macOS adds, but that could easily be added on the top list of the files that should not be listed.

HenkMutsaerts (changes requested):
The change of SortDicom2Folders makes no sense to me, needs reverting.
It probably needs to be

'^.*(?!(\.xlsx|\.ini|\.json|\.DS_Store))$'

The change to xASL_io_DcmtkRead is fine

HenkMutsaerts (approved):

PR #1812, opened 2024-10-07
Feature #1787 new dcm2nii

### LINKED ISSUE #1787: new DCMTK version
### Description

There's a new DCMTK version from February 2024, not only it should work better and have a better ASL support, but it should also work on the new MacOS

### Tasks

- [x] main question: doe you mean DCMTK or DCM2NIIX? afaik dcm2niix was never an issue with macOS, only Jan's dcmtk compilation. For Rik, this was somehow an issue, so doesn't hurt to update :)
- [x] JAN pick dataset that Henk tests DcmTK conversion with. JAN: Just try any dataset, any flavor and carefully check if DCMTK throws an error and rolls back to DCMTK or not.
- [x] JAN give address where Henk downloads dcmtk 2024. No need, this is donwloaded and integrated to ExploreASL already.
- [x] @HenkMutsaerts  tests dcm2niix by calling it on DICOMs from commandline (ideally by calling it through system from Matlab) and sees if it works (enough to test if it invokes help screen, but slightly better to test real conversion). -> HENK: I will test real conversion. Rik already tried.
- [x] @HenkMutsaerts Confirms that dcm2niix from 2022 that is included in ExploreASL works with Intel Mac and crashes with M1 Mac. Rik did that already
- [x] @HenkMutsaerts Confirms that dcm2nii from February 2024 (downloaded newly) work with both Intel Mac and M1 Mac
- [x] Once this is done, @jan-petr will initiate the update
- [x] Add new dcm2niix version to ExploreASL
- [x] Update code that 2024 version is taken by default
- [x] Inform Rik
- [x] Run all flavors and check for errors

### How to test

- [x] macARM dcm2niiX 2024 works, dcmtk doesn't work, not compiled yet
- [x] macINTEL dcm2niiX 2024 works, but dcmTk gives the following warning"
       DcmDataDictionary: Cannot open file: /opt/local/share/dcmtk/dicom.d100% -> is this path fixed? Because we 
       assume that users don't have dcmtk installed right? So we cannot point to a shared dcmtk library
- [x] Single-PLD doesn't have M0 PLD as 0 anymore 'GE_PCASL_3Dspiral_DV22.0_1' , 'GE_PCASL_3Dspiral_DV23.1_1'   ,'GE_PCASL_3Dspiral_DV24.0_1','GE_PCASL_3Dspiral_DV25.0_1.5T_1', 'GE_PCASL_3Dspiral_DV25.0_ADNI3_027_S_5079_1', 'GE_PCASL_3Dspiral_RX28.0_multiPLD_1'  
- [x] multi-PLD ignores studyPar values  ,'GE_PCASL_3Dspiral_DV25.0_eASLPLD3_1', ,'GE_PCASL_3Dspiral_DV25.0_eASLPLD7_1'
- [x] Philips - extra Tags
- [x] Philips Lock-Locker messed up
- [x] Siemens_PCASL_2DEPI_VB17A_BSupN_1 - wrong labeling duration
- [x] Siemens extra tags

### Release notes

jan-petr:
### Linked issue

Close #1787


Functions/Initialization/xASL_init_Import.m
```diff
@@ -65,4 +65,6 @@
     % Create logging directory if it does not exist already
     xASL_adm_CreateDir(fullfile(x.dir.DatasetRoot,'derivatives','ExploreASL','log'));
 
+	x = xASL_init_LoadDataPar(x);
```
HenkMutsaerts: This needs reverting: dataPar is for processing, not for import. We strictly separated this and probably even removed this init from here, remember?

Functions/InputOutput/xASL_io_dcm2nii.m
```diff
@@ -10,7 +10,7 @@
 %      destdir         target destination directory (REQUIRED)
 %	   series_name     target name of the NIfTI file (REQUIRED)
 %      imPar           imPar from x.modules.import.imPar (OPTIONAL, DEFAULT = [])
-%                      'imPar.dcm2nii_version' (DEFAULT 20220720)
+%                      'imPar.dcm2nii_version' (DEFAULT 20240202)
```
HenkMutsaerts, line 13: Did you test with all flavors that the new version works? We also have/had some exceptions where we defaulted to an older dcm2nii version, e.g., for PAR/REC, which was supported in an older version but not in the (back then) newer version.

Modules/Module_Import/NII2BIDS/xASL_bids_BIDSifyASLJSON.m
```diff
@@ -117,7 +117,7 @@
 					') & studyPar (' xASL_num2str(jsonOut.(fn{1})) '). '];
 
 				% Define the DICOM or studyPar priority
-				if strcmp(fn{1}, 'TotalAcquiredPairs') || strcmp(fn{1}, 'ArterialSpinLabelingType') || strcmp(fn{1}, 'EchoTime') || strcmp(fn{1}, 'VascularCrushing')
+				if strcmp(fn{1}, 'TotalAcquiredPairs') || strcmp(fn{1}, 'ArterialSpinLabelingType') || strcmp(fn{1}, 'EchoTime') || strcmp(fn{1}, 'VascularCrushing') || strcmp(fn{1}, 'PostLabelingDelay') || strcmp(fn{1}, 'LabelingDuration')
```
HenkMutsaerts, line 120: I don't get this. We made the following rule afaik:

1. studyPar always has priority over DICOM
2. Users should always start with an empty studyPar, check what is missing (i.e. not obtained from the DICOM) and then add this to the studyPar

Unless you can give me a use case where DICOM fields should have priority, and the user cannot simply remove the parameter from the studyPar.

So we should always compare all parameters between DICOM and studyPar.

I'm fine for now if we compare this only for a few and only issue a warning for a few, but we need to always prioritize studyPar, period :)
(for the simple reason that studyPar is only used exceptionally)

HenkMutsaerts (changes requested):
Few (probably minor) points.
Also note that one commit title says DcmTk instead of dcm2nii (but that's fine for me, just checking if it should really only be dcm2nii that is changed, not dcmtk)

Functions/Initialization/xASL_init_Import.m
jan-petr: Agree

Functions/InputOutput/xASL_io_dcm2nii.m
jan-petr: Yes, all flavors work. PAR/REC of Look-Locker needs to use the old version. All others work nicely.

Modules/Module_Import/NII2BIDS/xASL_bids_BIDSifyASLJSON.m
jan-petr: Yes. That's something old from Michael probably. Because studyPar has priority for all, doesn't need to be checked.

We should make a new issue and check that and fix that in agreement with what you say. 

For now, we just need to accommodate was has to be done for the new dcm2nii version.

HenkMutsaerts: OK can you create the new issue and resolve this comment?

HenkMutsaerts (changes requested):

HenkMutsaerts (approved):

Modules/Module_Import/NII2BIDS/xASL_bids_BIDSifyASLJSON.m
jan-petr: Agree. New issue created #1865

PR #1822, opened 2024-11-16
Feature #1821 vab yquantification

### LINKED ISSUE #1821: External VABY quantification
### Description

Adapt the call to VABY. We do it now by tweaking the FSL options to BASIL/FABBER/VABY
But future implementations need to do this properly

### Basic Tasks

- [x] Add path to VABY into FSL parameter in dataPar.json 7b24065f
- [x] Check FSL setup with VABY instead - do not search automatically 7b24065f
- [x] Add an option to quantify with VABY 7b24065f
- [x] Add VABY option to quant_FSL 66f5517f
- [x] Load and deal with inputs/outputs 66f5517f
- [x] Add VABY option to run_FSL 66f5517f
- [x] Adapt the options generation to work with multiple options 5b02d0f3
- [x] Adapt the VABY command and VABY options

### Extra tasks
- [x] Add FABBER and BASIL to ExternalQuantificationType
- [x] Add ExternalQuantificationType to dataPar defaults and make sure it is under modules.asl
- [x] FSLdir, RootFSLdir, VABY dir goes to external in code 
- [x] FSLdir, RootFSLdir, VABY dir goes to external in the documentation
- [x] Read and output ITT map. Add functionality in general, adapt filename for fabber/basil
- [x] Rename all commands and options after discussion with HJ

### Release notes

Added VABY external quantification. Rename the dataPar parameters for external quantification, but make them backwards compatible.

jan-petr:
### Linked issue

Close #1821


Functions/FSL/xASL_fsl_RunFSL.m
```diff
@@ -85,44 +85,48 @@
 %     ExistCUDA = false;
 % end
 
+% Setup VABY quantification
```
HenkMutsaerts: Well, I hope you understand that this is actually quick and dirty ;)

xASL_fsl_RunFSL is a script that prepares for running FSL, what are the settings that we should give FSL, what is the Bash initialization etc. So it is 90% for FSL specifically, so if you then use the 10% that is more generic for VABY that we will only use once that is asking for diluted code :)

I propose instead:
1. rename xASL_fsl_RunFSL -> xASL_run_FSL
2. create xASL_run_VABY

Functions/Administration/xASL_adm_GetDeprecatedFields.m
```diff
@@ -97,6 +97,7 @@
     'bRunLongReg'                   'modules'   'bRunLongReg'                        '';...
     'ForceInclusionList'            'dataset'   'ForceInclusionList'                 '';...
     'bAutomaticallyDetectFSL'       'external'  'bAutomaticallyDetectFSL'            '';...
+	'ExternalQuantificationType'    'external'  'ExternalQuantificationType'         '';...
```
HenkMutsaerts: ExternalQuantificationType should default to FSL?

But: don't we already have this parameter, and it defaults to ExploreASL quantification?
I propose:

Have a single parameter "QuantificationType" which can be ExploreASL/FSL/VABY.

(alternatively, we can have a boolean bExternalQuantification, but that is probably more work to keep track of)

Functions/FSL/xASL_fsl_SetFSLdir.m
```diff
@@ -55,7 +53,14 @@
     FSLdir = x.FSLdir;
     RootWSLdir = x.RootFSLdir;
     return;
-elseif isfield(x, 'FSLdir') && ~isfield(x, 'RootFSLdir') && ~ispc
+end
+
+% For VABY quantification, we cannot run automatic detection
+if isfield(x, 'external') && isfield(x.external, 'ExternalQuantificationType') && strcmp(x.external.ExternalQuantificationType, 'VABY')
```
HenkMutsaerts: This is very confusing to the user, using the FSL path for VABY.
What about cases where we use FSL TopUp + VABY quantification?

Functions/Quantification/xASL_quant_FSL.m
```diff
@@ -106,16 +106,30 @@
...
     %% 4. Create option_file that contains options which are passed to the FSL command
     % FSLOptions is a character array containing CLI args for the BASIL/FABBER command
-	% Define if BASIL or FABBER is used - multiTE needs FABBER 
-	if (isfield(x.modules.asl, 'bQuantifyMultiTE') && x.modules.asl.bQuantifyMultiTE) || bQuantifyMultiTE
-		bUseFabber = 1;
-        FSLfunctionName = 'fabber_asl';
+	
+	% Define if BASIL or FABBER is used - multiTE needs FABBER. VABY is an alternative
+	% Allow external input of quantification type
+	if isfield(x, 'external') && isfield(x.external, 'ExternalQuantificationType')
```
HenkMutsaerts: I don't see why we need to rename the string. If we need to specifically for VABY, we should do this in a subfunction.

We can rename this function to xASL_quant_External and have a VABY-specific subfunction, if VABY uses a lot of the settings similarly to FSL?

In any case, it would be good to split:
A. preparation of Tex/ATT/fitting parameters
B. specific parameters for the quantification modeling

HenkMutsaerts (changes requested):
I do actually find it “quick and dirty”, you predicted correctly, sorry 🙂 But you can of course use this branch for mediri. 

More importantly: what happens when you want to use FSL for non-quantification things like TopUp? Because that is not possible now in combination with VABY. We could integrate the SPM equivalent of TopUp though if needed.

Also: we call it quantification type now, but are these differences in quantification, or differences in quantification modeling, or differences in the program that does it?

E.g., we probably cannot say "the quantification is equal but the modeling is different", but we could say "the quantification/modeling is equal but the program that does it is different"?

Functions/FSL/xASL_fsl_RunFSL.m
jan-petr: I am aware of all these shortcomings. So alerted about that in the issue ;)

My main point was that I need to quickly set up the functionality. And then I'll distribute the functionality across the correct files as we agree together.

I agree with your proposition, but we need to adapt the code as we have an FSL directory in Functions.

I propose to create Functions/External
then have xASL_ext_runFSL, xASL_ext_runVABY, xASL_ext_SetFSLdir, xASL_ext_TopUpFSL

Agree like that, or other names?

Functions/Administration/xASL_adm_GetDeprecatedFields.m
jan-petr: We have bUseBasilQuantification from previously, and we had an internal parameter to distinguish between BASIL and FABBER as this could be decided automatically based on multi-TE data.

I now added ExternalQuantificationType because FABBER/VABY cannot be decided based on data.

We of course want a single parameter to make things easier, but I didn't want ot make the decision without you.

I propose bUseBasilQuantification -> bExternalQuantification. And then keep ExternalQuantificationType as is.

We can of course just have bQuantificationType as ExploreASL/FSL/VABY/BASIL/FABBER - I have no preference and can change that to make it backwards compatible...

Functions/FSL/xASL_fsl_SetFSLdir.m
jan-petr: Yes of course. This is related to all the comments above. Once we decide to split VABY as a separate thing and not as a variant of FSL, we can treat it as a fully separate function.

Functions/FSL/xASL_fsl_RunFSL.m
HenkMutsaerts: Yes that sounds good, though I would do xASL_fsl_TopUp xASL_vaby_Quant xASL_fsl_Setdir etc in the external folder
Or as you have it, but fsl before topup, not topupFSl

So our prefix then (ext if you want it) then the software name then the function name

HenkMutsaerts (changes requested):
See first comment (that was the only one you wanted my response on right? Checked it through my phone)

Functions/FSL/xASL_fsl_RunFSL.m
jan-petr: If you don't mind, I would move all to External and use xASL_ext_
The main reason is that we have FSL, VABY - that would be already two directories and could grow up a lot. So best to keep the number of directories in functions reasonable and have all in ext.
So the function names will be

xASL_ext_FSLrun
xasl_ext_VABYrun
xASL_ext_FSLsetDir
xASL_ext_FSLTopUp

OK? The other option is 
Functions/FSL/xASL_fsl_Run
Functions/VABY/xASL_vaby_Run etc...

Functions/Quantification/xASL_quant_FSL.m
jan-petr: previously, we had only FSL and then we renamed BASIL or FABBER based on input data... Now this changes.

VABY is very very similar, so we should keep it in a single function. 
ok xASL_quant_External.

We are of course splitting based on specific parameter...

Functions/FSL/xASL_fsl_RunFSL.m
HenkMutsaerts: Yes good idea, I only proposed to remove the ext prefix but you're right that it is better.

Functions/Quantification/xASL_quant_FSL.m
HenkMutsaerts: Fine to call it xASL_quant_External. But still, it seems that the parameter choices, model choices, and software (Vaby/ExploreASL later/FSL now) are a bit mixed.

So 1: in the near future I would like to separate these.

And 2:
My problem is that we are changing 99% of the code usage (FSL) for 1% of the code usage (Vaby, we won't use that many times). That doesn't seem like good practice.

So instead of:

if Vaby
     .....
else 
   ..... (what we had originally, which you have to move complete sections of code)
end

I would prefer:

[keep what we have]

at the end have a separate section

if Vaby
    ......
end

So it is very clear that Vaby is a single time exception, that we probably will remove in the near future. Right?

You can of course insert this at multiple locations, but trying to keep the original code unchanged.

If you do it this way, I'm fine not changing the name, keeping everything as is (except perhaps the FSL folder name arranging or something) and just inserting Vaby here and there as a "quick and dirty feature" which we will soon remove when we have our own tooling for this?

HenkMutsaerts (changes requested):
OK I thought a bit more about this. My guess is you agree with option A below, that is not too much work (just more explicitly having Vaby separated from FSL at several locations)

Option A: this is a temporary quick & dirty inclusion of Vaby as feature, which will be soon replaced. Then I'm fine keeping all the FSL names, I would only suggest that Vaby is in inserted code (not changing the original FSL code with if statements but new separate sections that override the previously set FSL parameters with the Vaby parameters, so it is clear that this is a "plug-in/insert". With some comments.

Option B: we feel that we will use Vaby the coming years and we want to organize this better, then we go with the full renaming etc.

Separately: we should keep BIDS (sequence) parameters/ modeling options/ software-specific input separated, but that can be a separate issue (or perhaps we have it already) for later.

Functions/FSL/xASL_fsl_RunFSL.m
jan-petr: OK. So the final list of changes is
- [x] Remove FSL directory in functions and create EXT  4380b354
- [x] xASL_fsl_RunFSL.m -> xASL_ext_FSLrun ff97b571
- [x] xASL_fsl_SetFSLdir.m -> xASL_ext_FSLsetDir ff97b571
- [x] xASL_fsl_TopUp.m -> xASL_ext_FSLtopUp ff97b571
- [x] xASL_quant_FSL -> xASL_quant_External ff97b571
- [x] Split code to a new function xASL_ext_VABYsetDir a71a30b3
- [x] Use VABYdir variable everywhere instead of FSLdir a71a30b3
- [x] Split code to a new function xASL_ext_VABYrun a71a30b3
- [x] Create VABYpath variables a71a30b3

Functions/Administration/xASL_adm_GetDeprecatedFields.m
jan-petr: A summary:
- [x] bUseBasilQuantification is renamed to bUseExternalQuantification bc9bdb0c
- [x] bUseBasilQuantification as an outdated variable gets renamed to bUseExternalQuantification bc9bdb0c
- [x] ExternalQuantificationType is "BASIL" by default, but can be "FABBER" or "VABY" - we add a default and complain if another value is used bc9bdb0c

Functions/Quantification/xASL_quant_FSL.m
jan-petr: - [x] Rename xASL_quant_FSL to xASL_quant_External fd0633b5
- [x] Rename all FSL variables to External fd0633b5
- [x] Make the VABY/FABBER/FSL choices cleaner

VABY is a full viable option. Not a single-time exception that we will remove. Most likely this is going to replace FSL. The original code is unchanged except for tabs. So I try to keep things organized into functional section.

Once we have our own tools, this part of code will remain, our own tools won't be inserted here.

Functions/FSL/xASL_fsl_SetFSLdir.m
jan-petr: See above. VABY is now separated. And we can have both VABY (quantification) and FSL (TopUp).

Functions/Administration/xASL_adm_GetDeprecatedFields.m
HenkMutsaerts: Nice! Not sure if VABY is an external directory that can automatically be detected like FSL? But you can resolve this.

Functions/External/b02b0.cnf
HenkMutsaerts, line 1: Bit confusing that we have /Functions/External and /External. Can we add a README.md to /Functions/External:

CONTENTS of /Functions/ExternalREADME.md:

<LICENSE STATEMENT EXPLOREASL>

This folder contains ExploreASL code, under the ExploreASL license, that acts as an "API" to external software. This software may or may not be located in ExploreASL/External

Functions/External/xASL_ext_VABYRun.m
```diff
@@ -0,0 +1,90 @@
+function [x, Result1] = xASL_ext_VABYRun(VABYCommand, x, NicenessValue, bVerbose)
+%xASL_ext_VABYRun Run VABY from Matlab (ExploreASL)
+%
+% FORMAT: [x] = xASL_adm_RunVABY(VABYCommand, x[, NicenessValue, bVerbose])
+%
+% INPUT:
+%   VABYCommand     - Command line job for VABY (REQUIRED)
+%   x               - structure containing fields with all information required to run this submodule (REQUIRED)
```
HenkMutsaerts: not a submodule

Functions/External/xASL_ext_VABYRun.m
```diff
@@ -0,0 +1,90 @@
...
+    bVerbose = true;
+end
+
+% Defaults
+Result1 = NaN;
+
+%% Find VABY directory
+[VABYdir, x] = xASL_ext_VABYSetDir(x);
+
+if min(isnan(VABYdir))
```
HenkMutsaerts: bit cryptic way of coding, 

contains(VABYdir, NaN)

reads easier

minor comment of course

Perhaps another question: why would you get a VABY path plus NaN in case of a valid VABY path?

Functions/External/xASL_ext_VABYRun.m
```diff
@@ -0,0 +1,90 @@
...
+    % installation found
+    warning('No VABY installation found, skipping VABY function');
+    return;
+end
+
+% Setup VABY quantification
+VABYinit = '';
+VABYoutput = '';
+OutputString = '.nii';
+if strcmp(VABYCommand(1:4),'vaby')
```
HenkMutsaerts, line 67: So you can also start vaby without needing the vaby folder?

Should this not give a warning?

What about a user running ExploreASL with VABY in the following folder:

/vaby/RunExploreASL/data/...

note that some users run multiple pipelines and give their data an additional folder layer with the name of the pipeline. So perhaps this (1:4) is not completely robust?

Also, should this be strcmpi?

Functions/External/xASL_ext_VABYRun.m
```diff
@@ -0,0 +1,90 @@
...
+	VABYCommand = fullfile(VABYdir, VABYCommand);
+end
+
+%% Be nice
+NiceString = ['nice -' num2str(NicenessValue) ' '];
+fprintf('%s\n', ['VABY: NiceNess=' num2str(NicenessValue) ', output=' OutputString]);
```
HenkMutsaerts: Why do you output 

output=.nii

?

Functions/External/xASL_ext_VABYRun.m
```diff
@@ -0,0 +1,90 @@
...
+
+%% Run VABY
+if bVerbose
+    Result1 = system([VABYinit VABYoutput NiceString VABYCommand], '-echo');
+else
+    Result1 = system([VABYinit VABYoutput NiceString VABYCommand]);
+end
+if Result1~=0
+    warning('VABY command didnt work nicely:');
```
HenkMutsaerts, line 83: Why : colon at the end, when you don't print a warning? Shouldn't you catch the output of system and print the warning (especially if bVerbose == 0)

Also, if I remember correctly, Result1 will now give you the screen output of the system command, not if it ran correctly or not. So you want to have something like:

[outputMessage, Result1] = system(...)

Functions/External/xASL_ext_VABYRun.m
```diff
@@ -0,0 +1,90 @@
...
+	VABYCommand = fullfile(VABYdir, VABYCommand);
+end
+
+%% Be nice
+NiceString = ['nice -' num2str(NicenessValue) ' '];
+fprintf('%s\n', ['VABY: NiceNess=' num2str(NicenessValue) ', output=' OutputString]);
+
+%% Run VABY
+if bVerbose
+    Result1 = system([VABYinit VABYoutput NiceString VABYCommand], '-echo');
```
HenkMutsaerts: VABYinit and VABYoutput are always empty?

Check the FSL stuff I wrote, I think VABYinit should have the system initialization, e.g. defining the VABY folder to the path etc. Note that when you run system from Matlab, it doesn't initialize the same as when you start a default session within the OS, because it doesn't run the bash_profile, bash_rc etc.

Or alternatively, we have that in xASL_system, you can check.

Functions/External/xASL_ext_VABYSetDir.m
```diff
@@ -0,0 +1,67 @@
...
+% https://github.com/ExploreASL/ExploreASL/blob/main/LICENSE
+% you may only use this file in compliance with the License.
+% __________________________________
+
+
+%% Admin
+if nargin<1
+    x = struct;
+end
+if nargin<2 || isempty(bAutomaticallyDetectVABY)
```
HenkMutsaerts: Why do we have this both as an x table parameter and a parameter that you can give as second input argument, also: you want the x table (~dataPar.json) to override ExploreASL defaults, not the other way around right?

Functions/External/xASL_ext_VABYSetDir.m
```diff
@@ -0,0 +1,67 @@
...
+
+if isfield(x,'VABYdir') && ~isempty(x.VABYdir)
+    % if we already have an VABY dir, skip this function
+    VABYdir = x.VABYdir;
+    return;
+end
+
+% For VABY quantification, we cannot run automatic detection
+if isfield(x, 'external') && isfield(x.external, 'ExternalQuantificationType') && strcmp(x.external.ExternalQuantificationType, 'VABY')
+	error('External quantification with VABY requested. You need to provide path in x.VABYdir');
```
HenkMutsaerts: "provide the path to VABY"

and you mean "the path to the VABY folder" right? Or to a function? And which folder inside VABY?

Functions/External/xASL_ext_VABYSetDir.m
```diff
@@ -0,0 +1,67 @@
...
+end
+
+
+%% Detect OS
```
HenkMutsaerts: I would print this only when running VABY, not when loading the directory from x.VABY into VABY ;)

So this function essentially only provides an error message when x.VABY was not defined? :)

Functions/QualityControl/qc_glossary.tsv
```diff
@@ -10,7 +10,8 @@ ASL_tSNR_Slope_Corr
 ASL_tSNR_WM_Ratio				
 ASL_tSNR_WMref_Ratio				
 BackgroundSuppressionNumberPulses	Number of Background Suppression Pulses			
-bUseBasilQuantification				
+bUseExternalQuantification
```
HenkMutsaerts: In the QC output I would only add the ExternalQuantificationType.
BTW: does this parameter default to "disabled" or something, when we don't use external quantification?

Functions/Quantification/xASL_quant_ASL.m
```diff
@@ -88,8 +86,8 @@
 	end
 end
 
-if x.modules.asl.bQuantifyMultiPLD && ~bUseBasilQuantification
-	error('Multi-PLD quantification currently works only with BASIL');
+if x.modules.asl.bQuantifyMultiPLD && ~bUseExternalQuantification
```
HenkMutsaerts, line 91: Do we want the same error message for x.modules.asl.bQuantifyMultiTE ?

Functions/Quantification/xASL_quant_ASL.m
```diff
@@ -135,14 +133,14 @@
...
+    % For BASIL/FABBER/VABY the x.Q.SliceReadoutTime is used internally, otherwise
     % x.Q.SliceReadoutTime is added to ScaleImage
     
     switch lower(x.Q.MRAcquisitionType)
         case '3d'
             fprintf('%s\n','3D sequence, not accounting for SliceReadoutTime (homogeneous PLD for complete volume)');
             x.Q.SliceReadoutTime = 0;
-            if bUseBasilQuantification
+            if bUseExternalQuantification
                 x.Q.BasilSliceReadoutTime = 0;
```
HenkMutsaerts: Should this parameter be renamed to x.Q.ExternalSliceReadoutTime ?
Because it it ignorant of the type of external quantification right.

I propose we either define quantification-ignorant parameters (and call them that way, make this clear) OR have separate sections for BASIL/FABBER/VABY (I don't mind the structure for now, just as long as it is clear which parameters are quantification-type-independent or quantification-type-dependent

Or clarify that this parameter really only is used for BASIL and the other external quantifications use the same SliceReadoutTime as ExploreASL

Functions/Quantification/xASL_quant_External.m
```diff
@@ -0,0 +1,706 @@
...
+    % Define defaults
+	Tex_map = [];
+	ATT_map = [];
+	ABV_map = [];
+
+	if ~isfield(x.modules.asl, 'bCleanUpExternal') || isempty(x.modules.asl.bCleanUpExternal)
+		x.modules.asl.bCleanUpExternal = true;
+	end
+    
+    %% 1. Define temporary paths for FSL
```
HenkMutsaerts: Really only for FSL? I'm fine with not having separate structure for the different external quantifications for now, but clarify what is quantification-type-independent (i.e. "external quantification") or what is quantification-type-dependent (i.e. FSL/BASIL/FABBER/VABY)

Functions/Quantification/xASL_quant_External.m
```diff
@@ -0,0 +1,706 @@
...
+	% Create FSL output directory
```
HenkMutsaerts: Really only for FSL? I'm fine with not having separate structure for the different external quantifications for now, but clarify what is quantification-type-independent (i.e. "external quantification") or what is quantification-type-dependent (i.e. FSL/BASIL/FABBER/VABY)

Functions/Quantification/xASL_quant_External.m
```diff
@@ -0,0 +1,706 @@
...
+
+    %% 2. Delete previous output
+    xASL_adm_DeleteFileList(x.dir.SESSIONDIR, ['(?i)^' dirExternalOutput '.*$'], 1, [0 Inf]);
+    FolderList = xASL_adm_GetFileList(x.dir.SESSIONDIR, ['(?i)^' dirExternalOutput '.*$'], 'FPList', [0 Inf], 1);
+    for iFolder=1:numel(FolderList)
+        xASL_delete(FolderList{iFolder}, 1);
+    end
+    fprintf('%s\n', 'Note that any file not found warnings can be ignored, this pertains to the use of symbolic links by BASIL/FABBER/VABY');
+    
+    % Remove residual BASIL-related files
```
HenkMutsaerts: Really only for BASIL? I'm fine with not having separate structure for the different external quantifications for now, but clarify what is quantification-type-independent (i.e. "external quantification") or what is quantification-type-dependent (i.e. FSL/BASIL/FABBER/VABY)

Functions/Quantification/xASL_quant_External.m
```diff
@@ -0,0 +1,706 @@
...
+    xASL_delete(pathExternalOptions);
+    xASL_delete(pathExternalInput);
+	xASL_delete(pathExternalOutput, 1);
+    
+    %% 3. Write the PWI4D as Nifti file for BASIL/FABBER to read as input
```
HenkMutsaerts: Really only for BASIL/FABBER? I'm fine with not having separate structure for the different external quantifications for now, but clarify what is quantification-type-independent (i.e. "external quantification") or what is quantification-type-dependent (i.e. FSL/BASIL/FABBER/VABY)

Functions/Quantification/xASL_quant_External.m
```diff
@@ -0,0 +1,706 @@
...
+		bQuantifyMultiTE = false;
+	end
+
+    %% 4. Create option_file that contains options which are passed to the FSL command
+    % ExternalOptions is a character array containing CLI args for the BASIL/FABBER/VABY command
+	
+	% Define if BASIL or FABBER or VABY is used - multiTE needs FABBER. VABY is an alternative
+	% Allow external input of quantification type
+	if isfield(x, 'external') && isfield(x.external, 'ExternalQuantificationType')
+		strQuantificationType = x.external.ExternalQuantificationType;
```
HenkMutsaerts: You really only want this string here in this function? Perhaps better to call it

localQuantificationType

or

QuantificationTypeHere

Functions/Quantification/xASL_quant_External.m
```diff
@@ -0,0 +1,706 @@
...
+		if (isfield(x.modules.asl, 'bQuantifyMultiTE') && x.modules.asl.bQuantifyMultiTE) || bQuantifyMultiTE
+			strQuantificationType = 'FABBER';% Default for multi-TE
+		else
+			strQuantificationType = 'BASIL';% Default for single-TE
+		end
+	end
+
+	ExternalOptions = xASL_sub_ExternalOptions(pathExternalOptions, x, strQuantificationType, PWI4D_json, pathExternalInput, pathExternalOutput);
+
+    %% 5. Run BASIL and retrieve CBF output
```
HenkMutsaerts: Really only for BASIL? I'm fine with not having separate structure for the different external quantifications for now, but clarify what is quantification-type-independent (i.e. "external quantification") or what is quantification-type-dependent (i.e. FSL/BASIL/FABBER/VABY)

Functions/Quantification/xASL_quant_External.m
```diff
@@ -0,0 +1,706 @@
...
+			[~, resultExternal] = xASL_ext_FSLRun([ExternalFunctionName ' ' ExternalOptions], x);
+		case 'fabber'
+			ExternalFunctionName = 'fabber_asl';
+			[~, resultExternal] = xASL_ext_FSLRun([ExternalFunctionName ' ' ExternalOptions], x);
+		case 'vaby'
+			ExternalFunctionName = 'vaby_asl';
+			[~, resultExternal] = xASL_ext_VABYRun([ExternalFunctionName ' ' ExternalOptions], x);
+	end
+    
+    % Check if FSL failed
```
HenkMutsaerts: Really only for FSL? I'm fine with not having separate structure for the different external quantifications for now, but clarify what is quantification-type-independent (i.e. "external quantification") or what is quantification-type-dependent (i.e. FSL/BASIL/FABBER/VABY)

Functions/Quantification/xASL_quant_External.m
```diff
@@ -0,0 +1,706 @@
...
+			pathExternalCBF = xASL_adm_GetFileList(pathExternalOutput, '^mean_cbf\.nii$', 'FPListRec');
+			pathExternalATT = xASL_adm_GetFileList(pathExternalOutput, '^mean_att\.nii$', 'FPListRec');
+			pathExternalITT = xASL_adm_GetFileList(pathExternalOutput, '^mean_itt\.nii$', 'FPListRec');
+			pathExternalABV = xASL_adm_GetFileList(pathExternalOutput, '^mean_fblood\.nii$', 'FPListRec');
+			pathExternalTex = xASL_adm_GetFileList(pathExternalOutput, '^mean_texch\.nii$', 'FPListRec');
+	end
+
+        % Check and load all output files
+	if isempty(pathExternalCBF)
+        error([ExternalFunctionName ' failed']);
```
HenkMutsaerts, line 178: extra indent

Functions/Quantification/xASL_quant_External.m
```diff
@@ -0,0 +1,706 @@
...
+		ABV_map = xASL_io_Nifti2Im(pathExternalABV{end}); % we assume the latest iteration (alphabetically) is optimal. also converting cell to char array
+	end
+    
+    % Tex
+	if ~isempty(pathExternalTex)
+		Tex_map = xASL_io_Nifti2Im(pathExternalTex{end}); % we assume the latest iteration (alphabetically) is optimal. also converting cell to char array
+	end
+	
+    %% 6. Scaling to physiological units
+    % Note different to xASL_quant_ASL since Fabber has T1 in seconds
```
HenkMutsaerts: Really only for FABBER? I'm fine with not having separate structure for the different external quantifications for now, but clarify what is quantification-type-independent (i.e. "external quantification") or what is quantification-type-dependent (i.e. FSL/BASIL/FABBER/VABY)

Functions/Quantification/xASL_quant_External.m
```diff
@@ -0,0 +1,706 @@
...
+    % and does not take into account labeling efficiency
+    
+    CBF_nocalib = CBF_nocalib .* 6000 .* x.Q.Lambda ./ x.Q.LabelingEfficiency;
+    % (For some reason, GE sometimes doesn't need the 1 gr->100 gr conversion)
+    % & old Siemens sequence also didn't need the 1 gr->100 gr conversion
+
+	ABV_map = ABV_map ./ x.Q.LabelingEfficiency;
+    
+    %% 7. Householding
+	% Basils Output is in the subfolder '/FSL_Output' which contains multiple steps if there are multiple iterations, and always contains
```
HenkMutsaerts: Really only for BASIL? I'm fine with not having separate structure for the different external quantifications for now, but clarify what is quantification-type-independent (i.e. "external quantification") or what is quantification-type-dependent (i.e. FSL/BASIL/FABBER/VABY)

Functions/Quantification/xASL_quant_External.m
```diff
@@ -0,0 +1,706 @@
...
+%
+% __________________________________
+
+
+%% 0. Admin
+if nargin<6 
+	error('Require 6 input parameters.');
+end
+
+% Set BASIL dataPar options and their defaults
```
HenkMutsaerts: Really only for BASIL? I'm fine with not having separate structure for the different external quantifications for now, but clarify what is quantification-type-independent (i.e. "external quantification") or what is quantification-type-dependent (i.e. FSL/BASIL/FABBER/VABY)

Functions/Quantification/xASL_quant_External.m
```diff
@@ -0,0 +1,706 @@
...
+
+% Set basic parameters newly as they might differ in case of a merged sequence
+%bQuantifyMultiPLD = x.modules.asl.bQuantifyMultiPLD;
+if length(unique(jsonPWI4D.Q.Initial_PLD))>1 || length(unique(jsonPWI4D.Q.LabelingDuration))>1
+	bQuantifyMultiPLD = true;
+else
+	bQuantifyMultiPLD = false;
+end
+
+if length(unique(jsonPWI4D.Q.EchoTime)) > 1
```
HenkMutsaerts: We do this already in the main function above right? I would remove it here then, for consistency within the code. We rather want something to crash than run incorrectly/inconsistently.

Functions/Quantification/xASL_quant_External.m
```diff
@@ -0,0 +1,706 @@
...
+	% On Low quality settings, turn off all extra processing options
+	if isfield(x, 'settings') && isfield(x.settings, 'Quality') && ~x.settings.Quality
+		x.modules.asl.bSpatialBASIL = false;
+		x.modules.asl.bInferT1BASIL = false;
+		x.modules.asl.bInferArtBASIL = false;
+		x.modules.asl.ExchBASIL = 'simple';
+		x.modules.asl.DispBASIL = 'none';
+	end
+
+	% Setting defaults for BASIL specific options
```
HenkMutsaerts: You can remove this comment

Functions/Quantification/xASL_quant_External.m
```diff
@@ -0,0 +1,706 @@
...
+		x.modules.asl.DispBASIL = 'none';
+	end
+
+	if ~isfield(x.modules.asl, 'ATTSDBASIL') || isempty(x.modules.asl.ATTSDBASIL)
+		x.modules.asl.ATTSDBASIL = 1.0;
+	end
+end
+
+%% 1. Create the options file
+% ExternalOptions is a character array containing CLI args for the Basil command
```
HenkMutsaerts: Really only for BASIL? I'm fine with not having separate structure for the different external quantifications for now, but clarify what is quantification-type-independent (i.e. "external quantification") or what is quantification-type-dependent (i.e. FSL/BASIL/FABBER/VABY)

Functions/Quantification/xASL_quant_External.m
```diff
@@ -0,0 +1,706 @@
...
+		% Path to input and output
+		ExternalOptions = [ExternalOptions ' -o ' xASL_adm_UnixPath(pathExternalOutput, ispc)];
+		ExternalOptions = [ExternalOptions ' -i ' xASL_adm_UnixPath(pathExternalInput, ispc)];
+end
+
+% Define masking
+if x.modules.asl.bMaskingExternal
+	% Check for uninitialized Mask variable or file
+	if ~isfield(x.P, 'Path_BrainMaskProcessing')
+		warning('BASIL masking set to TRUE, but the mask variable x.P.Path_BrainMaskProcessing is not initialized.');
```
HenkMutsaerts: Really only for BASIL? I'm fine with not having separate structure for the different external quantifications for now, but clarify what is quantification-type-independent (i.e. "external quantification") or what is quantification-type-dependent (i.e. FSL/BASIL/FABBER/VABY)

Functions/Quantification/xASL_quant_External.m
```diff
@@ -0,0 +1,706 @@
...
+	elseif ~xASL_exist(x.P.Path_BrainMaskProcessing, 'file')
+		warning('BASIL masking set to TRUE, but the mask is missing: %s\n', x.P.Path_BrainMaskProcessing);
```
HenkMutsaerts: same

Functions/Quantification/xASL_quant_External.m
```diff
@@ -0,0 +1,706 @@
...
+				fprintf(FIDoptionFile, '--ti%d=%.2f\n', iTI, TIs(iTI));
+			end
+		else
+			fprintf(FIDoptionFile, '--ti=%.2f\n', TIs);
+		end
+
+		% Either print bolus duration or unspecify it
+		if isfield(jsonPWI4D.Q, 'LabelingDuration') && ~isempty(jsonPWI4D.Q.LabelingDuration) 
+			if length(unique(jsonPWI4D.Q.LabelingDuration))>1
+				warning('PASL multi-PLD currently supports only a single Labeling Duration');
```
HenkMutsaerts: in ExploreASL or in general?

Functions/Quantification/xASL_quant_External.m
```diff
@@ -0,0 +1,706 @@
...
+			end
+			fprintf(FIDoptionFile, '--tau=%.2f\n', jsonPWI4D.Q.LabelingDuration(1)/1000);
+		else
+			% Bolus duration not know. If multi-TI, then try to infer it
```
HenkMutsaerts: not know -> unknown

Functions/Quantification/xASL_quant_External.m
```diff
@@ -0,0 +1,706 @@
...
+			if length(TIs) > 1
+				fprintf(FIDoptionFile, '--infertau\n');
+				fprintf('BASIL: Infer bolus duration component\n')
```
HenkMutsaerts, line 469: Really only for BASIL? I'm fine with not having separate structure for the different external quantifications for now, but clarify what is quantification-type-independent (i.e. "external quantification") or what is quantification-type-dependent (i.e. FSL/BASIL/FABBER/VABY)

Functions/Quantification/xASL_quant_External.m
```diff
@@ -0,0 +1,706 @@
...
+	case {'casl','pcasl'}
+		% Prepare unique PLDs+LabDur combinations
+		
+		% First create a labeling duration vector of the same length
+		LabDurs = jsonPWI4D.Q.LabelingDuration/1000;
+		PLDs = jsonPWI4D.Q.Initial_PLD/1000;
+
+		switch (lower(strQuantificationType))
+			case 'fabber'
+				% For FABBER and multi-TE, we have to group TEs
```
HenkMutsaerts, line 483: What do we do if we have PASL multi-TE?
Perhaps issue a warning as well above that we don't support PASL multi-TE yet. Now you issue a warning that we don't support FABBER/VABY with PASL, but this is perhaps not specific for those with PASL multi-TE (I know that probably nobody did this, but ok:)

BTW: what happens with dual-echo ASL, does all this go correctly? If I remember correctly, we have a parameter somewhere for dual-echo, taking the first for ASL. Do we need a warning that if you request multi-TE quantification with nTE<4, that we will do it, but that it may not provide robust results?

Functions/Quantification/xASL_quant_External.m
```diff
@@ -0,0 +1,706 @@
...
+				TEs = round(jsonPWI4D.Q.EchoTime'/1000,3);
+
+				% We go through the LabDurs, PLDs, and TEs and keep grouping similar LD+PLD with increasing TE into groups
+				LabDursGrouped = [];
+				PLDsGrouped = [];
+				nTEGrouped = [];
+				iTE = 1;
+
+				while iTE<=length(TEs)
+					% We define a block
```
HenkMutsaerts: Are all the nomenclature changes we did in the harmonization of the Hadamard decoding/PWI subtraction etc also applied here?

Functions/Quantification/xASL_quant_External.m
```diff
@@ -0,0 +1,706 @@
...
+				end
+
+				LabDurs = LabDursGrouped;
+				PLDs = PLDsGrouped;
+				nTE = nTEGrouped;
+			case 'vaby'
+				TEs = round(jsonPWI4D.Q.EchoTime'/1000,3);
+				nTE = length(TEs);
+			case 'basil'
+				% Normal multi-timepoint without multi-TE
```
HenkMutsaerts: multi-timepoint -> multi-PLD. Let's keep nomenclature the same throughout t avoid confusion and misunderstanding

Functions/Quantification/xASL_quant_External.m
```diff
@@ -0,0 +1,706 @@
...
+				end
+
+				for iTE = 1:length(TEs) %So for each volume, we print a TE value
+					if iTE == 1
+						ExternalOptions = [ExternalOptions ' --tes=' sprintf('%.3f', TEs(iTE))];
+					else
+						ExternalOptions = [ExternalOptions ',' sprintf('%.3f', TEs(iTE))];
+					end
+				end
+			case 'basil'
```
HenkMutsaerts, line 569: Just out of curiosity, if BASIL also runs FABBER internally for fitting, would it be easier for our code if we would replace "BASIL" here also by FABBER but then with different parameters? It would reduce the path differences etc. Not for now of course.

Functions/Quantification/xASL_quant_External.m
```diff
@@ -0,0 +1,706 @@
...
+				% Specify that we run the PCASL/CASL model
+				fprintf(FIDoptionFile, '--casl\n');
```
HenkMutsaerts, line 571: Here we output feedback, should we do this as well for the other quantification types?

Functions/Quantification/xASL_quant_External.m
```diff
@@ -0,0 +1,706 @@
...
+					for iPLD = 1:length(PLDs)
+						fprintf(FIDoptionFile, '--pld%d=%.2f\n', iPLD, PLDs(iPLD));
+					end
+				else
+					fprintf(FIDoptionFile, '--pld=%.2f\n', PLDs(1));
+				end
+		end
+
+		switch (lower(strQuantificationType))
+			case {'fabber','basil'}
```
HenkMutsaerts, line 584: Do we somewhere need a check that strQuantificationType is either fabber/basil/vaby and issue a warning otherwise?

Functions/Quantification/xASL_quant_External.m
```diff
@@ -0,0 +1,706 @@
...
+					if iLD == 1
+						ExternalOptions = [ExternalOptions ' --taus=' sprintf('%.2f', LabDurs(iLD))];
+					else
+						ExternalOptions = [ExternalOptions ',' sprintf('%.2f', LabDurs(iLD))];
+					end
+				end
+		end
+end
+
+if ~isempty(regexpi(strQuantificationType, 'basil', 'once'))
```
HenkMutsaerts: We should be consistent. Here, you could have "fabber_basil" as you're checking containing only. In the other parts, you're checking for an exact "basil" (though case independent).

Functions/Quantification/xASL_quant_External.m
```diff
@@ -0,0 +1,706 @@
...
+	% Act as if we do not have repeats
+	%fprintf(FIDoptionFile, '--repeats=%i\n', size(PWI, 4)/PLDAmount);
+	fprintf(FIDoptionFile, '--repeats=1\n');
+
+	% Slice-timing
+	fprintf(FIDoptionFile, '--slicedt=%f\n', x.Q.BasilSliceReadoutTime/1000);
+
+	if isfield(x.Q,'LookLocker') && x.Q.LookLocker
```
HenkMutsaerts: You don't want to print LookLocker-related warnings etc for non-basil? Even if we cannot run it yet with FABBER or VABY, it would be good to give the user feedback about LookLocker not being inplemented yet. Or specify somewhere that for LookLocker you need BASIL. E.g., imagine somebody doing LookLocker with multi-TE. Even if this is hypothetical, extra warnings don't hurt.

Functions/Quantification/xASL_quant_External.m
```diff
@@ -0,0 +1,706 @@
...
+			end
+			fprintf(option_file, '--FA=%f\n', x.Q.FlipAngle(1));
+			fprintf('BASIL: Flip angle for Look-Locker readout: %f\n', x.Q.FlipAngle(1));
+		else
+			warning('BASIL: Unknown flip angle for Look-Locker\n');
+		end
+	end
+end
+
+%% 4. Model fiting parameters
```
HenkMutsaerts: Should this header be 4. BASIL fitting parameters?

Functions/Quantification/xASL_quant_External.m
```diff
@@ -0,0 +1,706 @@
...
+if ~isempty(regexpi(strQuantificationType, 'basil', 'once'))
```
HenkMutsaerts: Same here, we want exact basil, not contains basil (to avoid fabber_basil)

Functions/Quantification/xASL_quant_External.m
```diff
@@ -0,0 +1,706 @@
...
+			% Default initial ATT for PASL is 0.7
+			fprintf(FIDoptionFile, '--bat=0.7\n');
+		case {'pcasl','casl'}
+			% Default initial ATT for PCASL is 1.3
+			fprintf(FIDoptionFile, '--bat=1.3\n');
+	end
+
+	if bQuantifyMultiPLD
+		% Multi-PLD or Time Encoded data allows to fit arrival times
+		fprintf(FIDoptionFile, '--batsd=%f\n', x.modules.asl.ATTSDBASIL);
```
HenkMutsaerts, line 643: Note for the future: we should distinguish between BAT (macrovascular/non-crushed ATT) and microvascular/crushed ATT (e.g. FEAST).

Functions/Quantification/xASL_quant_External.m
```diff
@@ -0,0 +1,706 @@
...
+	end
+end
+
+%% 5. Extra BASIL fitting options
+if ~isempty(regexpi(strQuantificationType, 'basil', 'once'))
```
HenkMutsaerts: is exact basil. not contains basil, avoiding fabber_basil

Functions/Quantification/xASL_quant_External.m
```diff
@@ -0,0 +1,706 @@
...
+			fprintf(FIDoptionFile, '--bat=1.3\n');
+	end
+
+	if bQuantifyMultiPLD
+		% Multi-PLD or Time Encoded data allows to fit arrival times
+		fprintf(FIDoptionFile, '--batsd=%f\n', x.modules.asl.ATTSDBASIL);
+	end
+end
+
+%% 5. Extra BASIL fitting options
```
HenkMutsaerts: Shouldn't we combine this with 4 as both are BASIL-specific?

Functions/Quantification/xASL_quant_External.m
```diff
@@ -0,0 +1,706 @@
...
+				warning('Look-Locker quantification with multiple flip angles, e.g. QUASAR, is not implemented yet');
+			end
+			fprintf(option_file, '--FA=%f\n', x.Q.FlipAngle(1));
+			fprintf('BASIL: Flip angle for Look-Locker readout: %f\n', x.Q.FlipAngle(1));
+		else
+			warning('BASIL: Unknown flip angle for Look-Locker\n');
+		end
+	end
+end
+
```
HenkMutsaerts, line 629: In section 4 & 5, we only have BASIL fitting parameters. Do we have that many more fitting parameters for BASIL than for FABBER/VABY? We are those for FABBER/VABY in the big if above, and for BASIL not in the same if but separate below?

Functions/Quantification/xASL_quant_FEAST.m
```diff
@@ -34,8 +34,8 @@ function xASL_quant_FEAST(x)
 %% 1. Provide feedback
 fprintf('%s\n', 'Running FEAST quantification, assuming single compartment CBF quantification');
 
-if x.modules.asl.bUseBasilQuantification
-    warning('BASIL quantification detected, FEAST ATT computation might be inaccurate');
+if x.modules.asl.bUseExternalQuantification
+    warning('External quantification detected, FEAST ATT computation might be inaccurate');
```
HenkMutsaerts: Replace by:

warning('External quantification detected, FEAST ATT computation is not tested yet with external quantification');

Modules/xASL_module_ASL.m
```diff
@@ -949,12 +949,22 @@
     end
 end
 
-%% 2. Manage bUseBasilQuantification parameter that activates BASIL quantification
-if ~isfield(x.modules.asl, 'bUseBasilQuantification') || isempty(x.modules.asl.bUseBasilQuantification)
-	x.modules.asl.bUseBasilQuantification = false;
+%% 2. Manage bUseExternalQuantification parameter that activates External quantification quantification
+if ~isfield(x.modules.asl, 'bUseExternalQuantification') || isempty(x.modules.asl.bUseExternalQuantification)
+	if ~isfield(x.modules.asl, 'ExternalQuantificationType') || ~isempty(regexpi(x.modules.asl.ExternalQuantificationType, '(basil|fabber|vaby)'))
```
HenkMutsaerts: Should we also test for:

if x.modules.asl.bUseExternalQuantification && isempty(x.modules.asl.ExternalQuantificationType)?

and 

if x.modules.asl.bUseExternalQuantification && isempty(regexpi(x.modules.asl.ExternalQuantificationType, '(basil|fabber|vaby)')?

BTW we want the regexp to include ^ $ to avoid the possibility for basil_fabber

HenkMutsaerts (changes requested):
What a beast! Lots of work, kudos, really nice!
I gave a lot of minor comments that are easy to fix, a few more important ones.
Sorry for the repetitions, just checked everywhere that we are consistent in warnings and comments so it's easier to continue with the code later on :)

Functions/External/xASL_ext_VABYRun.m
jan-petr: Done in 15cf7ec9

Functions/External/xASL_ext_VABYRun.m
jan-petr: xASL_ext_VABYSetDir outputs NaN in case a valid path doesn't exist. But don't ask _me_ why, because you programmed it like that :) I would have preferred an empty string.

We have to use min(isnan(VABYdir)) or sum(isnan(VABYDIR)) to make sure it works for both strings/chars/ and numbers(because NaN) is a number.
Contains(VABYdir, NaN) will not work because contains requires string on input and can't have NaNs as patterns.

Going back to the original comment -> I would never use NaNs as output because then you really mix chars and numbers and this creates issues.

Functions/External/xASL_ext_VABYRun.m
jan-petr: Colon was there already. But I removed it. 15cf7ec9

The message is printed on the screen already - no need to change this.

Functions/External/xASL_ext_VABYRun.m
jan-petr: see above - VABY doesn't need these system variables. So I removed it as described above.

Functions/External/xASL_ext_VABYSetDir.m
jan-petr: I have actually removed it completely. This is only interesting for FSL because we have WSL. For VABY, the user doesnt have to be reminded what system is using :D

Yes - it has so far limited function. But to make it similar to FSL, I've kept this and maybe we will be able to add automatic function location later. 15cf7ec9

Functions/External/b02b0.cnf
jan-petr: Agree. Though I would rather add this description into Functions/README.md that we already and that already had an FSL section rather than to the subdirectory as none of the subdirectories had a README.me. I agree with the content. Done in 15cf7ec9

Functions/External/xASL_ext_VABYRun.m
jan-petr: Sorry, VABYinit and VABYoutput should be removed. Nice catch. That's a copy paste from FSL, which is not necessary from VABY so it was set empty. With separate functions, we should of course remove it...

You either provide full path to "vaby" or just the command and the path is added here. Same as for FSL. You always need VABY folder. It's just that you can provide it yourself to the function. So this is OK. I added a comment about this. Done in 15cf7ec9

It should be strcmp, because the command "vaby" is provided by us.

This is not about path to data but path to the command, so 1:4 is robust...

Functions/External/xASL_ext_VABYRun.m
jan-petr: that was FSL related and wasn't used. Removed in 15cf7ec9

Functions/External/xASL_ext_VABYSetDir.m
jan-petr: Good question. I just copied what you wrote for FSL ;)

But checking the code - this extra option is either unused. Or we provide the dataPar's external.bAutomaticallyDetectFSL. So it is completely superfluous and I've removed it for both VABY and FSL in 15cf7ec9

Functions/External/xASL_ext_VABYSetDir.m
jan-petr: to the VABY command. Fixed in 15cf7ec9

Functions/QualityControl/qc_glossary.tsv
jan-petr: Agree with only printing the type. No it doesn't default - I will default it to "none" if bUseBasilQuantification is 0. Agree?

Functions/Quantification/xASL_quant_External.m
jan-petr: Yes. As above...

Functions/Quantification/xASL_quant_External.m
jan-petr: Yes

Functions/Quantification/xASL_quant_External.m
jan-petr: yes

Functions/Quantification/xASL_quant_External.m
jan-petr: Yes. Fixed in f2b38559

Functions/Quantification/xASL_quant_External.m
jan-petr: Yes. Done in f2b38559

Functions/Quantification/xASL_quant_External.m
jan-petr: Yes. Fixed in f2b38559

Functions/Quantification/xASL_quant_External.m
jan-petr: Fixed in f2b38559

Functions/Quantification/xASL_quant_External.m
jan-petr: Yes done in f2b38559

Functions/Quantification/xASL_quant_External.m
jan-petr: Agree. Fixed in f2b38559

Functions/Quantification/xASL_quant_External.m
jan-petr: Yes. Done in f2b38559

Functions/Quantification/xASL_quant_External.m
jan-petr: Agree. I forgot about this one. Changed in f2b38559

Functions/Quantification/xASL_quant_ASL.m
jan-petr: Fully agree. And I was changing that. But forgot about this one... I Propose SliceReadoutTimeDifference as this is not about BASIL but about a different representation. Added in f2b38559

Functions/Quantification/xASL_quant_ASL.m
jan-petr: Good point. Added in f2b38559

Functions/Quantification/xASL_quant_External.m
jan-petr: Yes. BASIL only. This is in the BASIL branch.

Functions/Quantification/xASL_quant_External.m
jan-petr: They are extremely secretive about the settings and I can't configure FABBER to run with full BASIL functionality. So I mainly use FABBER for DEBBIE because it has a completely different module and uses very different config parameters. So lets keep it as it is.

Functions/Quantification/xASL_quant_External.m
jan-petr: I removed the feedback. It was superfluous for BASIL as well ;)

Functions/Quantification/xASL_quant_External.m
jan-petr: Sure. It was basic and advanced parameters. But since we now have also VABY and FABBER then indeed better to merge. Done in a 385a0d4d

Functions/Quantification/xASL_quant_External.m
jan-petr: Done in 385a0d4d

Functions/Quantification/xASL_quant_External.m
jan-petr: Done in 385a0d4d

Functions/Quantification/xASL_quant_External.m
jan-petr: Yes. Done in 385a0d4d

Functions/Quantification/xASL_quant_External.m
jan-petr: Yes. changed on all occurences in 385a0d4d

Functions/Quantification/xASL_quant_External.m
jan-petr: Good point. I added that at the start of xASL_quant_External in 385a0d4d

Functions/Quantification/xASL_quant_External.m
jan-petr: Done in 385a0d4d

Functions/Quantification/xASL_quant_External.m
jan-petr: Specified in 385a0d4d

Functions/Quantification/xASL_quant_External.m
jan-petr: Good points. 
It would be simple to implement multi-TE PASL, but we don't have such an example, so I prefer not to do it, because I'm not able to test it. So I added a warning about this in 385a0d4d

Of course if we do internal ExploreASL quantification. Or if multi-TE is removed from PWI4D.json, then we won't notice it here and everything will run fine. So no edit need for that.

Functions/Quantification/xASL_quant_External.m
jan-petr: Done in 385a0d4d

Functions/Quantification/xASL_quant_External.m
jan-petr: BASIL. Specified in 385a0d4d

Functions/Quantification/xASL_quant_External.m
jan-petr: Done in 385a0d4d

Functions/Quantification/xASL_quant_External.m
jan-petr: Done in 385a0d4d

Functions/Quantification/xASL_quant_External.m
jan-petr: Done in 385a0d4d

Functions/Quantification/xASL_quant_External.m
jan-petr: Done in 385a0d4d

Functions/Quantification/xASL_quant_External.m
jan-petr: Yes. I was a bit hesitant to have it doubled. So it was for double safety. But I'm fine removing it. Done in 385a0d4d

Functions/Quantification/xASL_quant_External.m
jan-petr: Good point to improve later.

Functions/Quantification/xASL_quant_External.m
jan-petr: Yes. We have all those for BASIL, but not for FABBER/VABY.

Modules/xASL_module_ASL.m
jan-petr: Empty Type is tested at the start of the population module.

we test already for having an unknown quantification type as described in a comment above.

Functions/Quantification/xASL_quant_FEAST.m
jan-petr: Done in 160d2997

Functions/Quantification/xASL_quant_External.m
jan-petr: Yes. done in 160d2997

Functions/External/xASL_ext_VABYRun.m
HenkMutsaerts: We always use NaN as a number, never as a string or char. I know I programmed this, I'm fine using [] instead. We just need consensus on when to use [] and when NaN. One is e.g. missing data and the other is incorrect code/processing.

I now understand your point, NaN is a number and the desired output is a char. OK, then we should change it to either 'n/a' (the BIDS equivalent of NaN) or just and empty char.

You can resolve for now

Functions/External/xASL_ext_VABYRun.m
HenkMutsaerts: I don't get it. The comment says "check correct path" but you now say vaby command. How can a correct path start with 'vaby'? A correct path is e.g. /opt/vaby

or something similar

Also note: there are users that have:

/vaby/RunExploreASL/data/...
/vaby/RunExploreASL/vabycode/

OK now I get it.
So it should be:

if strcmp(VABYCommand(1:4),'vaby') ~exist(VABYCommand, 'dir')

right?

Functions/QualityControl/qc_glossary.tsv
HenkMutsaerts: You mean default to "none" if bUseExternalQuantification is false right?
Yes that makes sense.

Functions/Quantification/xASL_quant_External.m
HenkMutsaerts: OK then only my question remains if we proceed to multi-TE quantification and the user has only 3 TEs or something, do you want to issue a warning that this is perhaps not robust, or not? What is the minimum number of TEs required for a decent fit?

We could do the same for PLD of course.

But you can also ignore this point and resolve

Functions/External/b02b0.cnf
HenkMutsaerts: No I would (also) add a README.md to the External folder. I understand that none of the other subdirectories have a README.md but this is just potentially confusing with the whole license thingy. You can copy the content from Functions/README.md, I'm fine having it twice.

You can choose and resolve :)

Functions/External/xASL_ext_VABYSetDir.m
HenkMutsaerts: I probably added it there for users using FSL from Matlab without ExploreASL or something similar. Fine with what you propose, you can resolve.

HenkMutsaerts (changes requested):
Minor stuff

Functions/External/b02b0.cnf
jan-petr: OK. I made a copy

Functions/External/xASL_ext_VABYRun.m
jan-petr: Yes - it's the mixing of types that makes it tricky.

OK - I'm changing to []. Because this is function output not directly related to BIDS and it doesn't get out to the user - so nobody sees that N/A. 

You can resolve, I just kept it here to show that I indeed changed it to an empty char already.

Functions/External/xASL_ext_VABYRun.m
jan-petr: Yes - it's checking if VABYcommand has the command without path or with path ("vaby file.nii' or "/usr/local/vaby1.2/vaby file.nii"). I adapted the comment and code..

Also - we don't need to add the "exist" command. Because we already have a check on directory existence above

Functions/Quantification/xASL_quant_External.m
jan-petr: Good point. 3 is a bare minimum, less than 3 cannot work. 

For less than 3, I issue an error.
For the rest, I would not care - a bit of user responsibility. These data are not so easy to come around, so people won't run it on 3TEs by a mistake ;)

Functions/QualityControl/qc_glossary.tsv
jan-petr: We only print the type, not the boolean.
the type is initialized as none when teh boolean is false...

HenkMutsaerts (approved):

PR #1827, opened 2024-11-30
Bug #1826 native space lesions

### LINKED ISSUE #1826: Native space lesions stats
### Description

In this issue, we try to fix some issues with the native space processing as report in #58 and we are also trying to make sure that native-space stats are not only run in GM/WM, but also in the Lesions.

### Tasks

- [x] Add header explanations & comments to xASL_wrp_PreparePV (6th step) #1705 
- [x] Transform Lesion_T1 and ROI_T1 into ASL space in xASL_wrp_PreparePV #1705 
- [x] xASL_im_CreateAnalysisMasks.m use more clear booleans - bROIStats_NativeSpace & bROIStats_StandardSpace, instead of bSkip* for xASL_im_CreateAnalysisMask & bNativeSpaceAnalysis (bNativeSpaceAnalysis is too vague, bSkip vague and negative) - bSkip renamed 9964e36d
- [x] 1. remove `x.S.InputDataStrNative`. Only use x.S.InputDataStr and always add (q|r|) in front of xASL_adm_GetFileList for searching the datatypes. This should work in both/all spaces. Alternatively, "if native space" removes a q or r inside xASL_wrp_GetROIstatistics in the separate native space function. b1e9f656
- [x] Merge code that is similar across native/standard space b1e9f656
- [x] Create stats tables also on Lesion_T1 and FLAIR_T1 files. They already exist in ASL module. Be careful to binarize them first. This is relatively easy task as all native space ROIs are already subject specific. 908a2e53
- [x] ROIs should be created for WB. 908a2e53
- [x] When creating native space masks by transformation from standard space, we put the code to a subfunction b647b0de
- [x] PVLesion is for some reason not created and is then missing in xASL_stat_GetRoiStats on line 113
- [x] PVLesion is created as T1 to native ASL space as a direct transformation and creates PV-volume. This is done in ASL module. We can discuss if this is useful or not, but for now we can keep it. See below:
- [x] We want to also create Lesion including peritumoral + contralateral etc. All that ROI mirroring is done population module. So the Lesion transformation needs to be done from standard to native space inside the population module when creating atlases. So we need to add extra code for this to replicate this for Lesion as is done for atlases. And we reuse a subfunction for atlases also for lesions. 54693675
- [x] Stop using PVLesion for calculating population stats for native space lesions and use the new Lesion-ROI. 54693675
- [x] PVLesion and DeepWM_Atlas aren't removed when restarting the ASL module

### Release notes

Fix native space processing of lesions.

jan-petr:
### Linked issue

Close #1826


Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
```diff
@@ -633,7 +635,7 @@
 				pvPrimary = ones(size(DataIm)); % no masking
 				bSkipPVC = 1;
             else
-                switch x.S.TissueMaskingLocal{1}
+                switch x.S.TissueMaskingLocal
```
HenkMutsaerts, line 657: If you change this, you should also change this in xASL_module_Population. We don't need this as cell still because different atlases can have different TissueMaskingLocal? E.g. an atlas can contain GM and WM regions. But we can also keep that as a potential future feature.

Modules/xASL_module_Population.m
```diff
@@ -235,16 +235,6 @@
     % Iterate over DataTypes
     for iDataType=1:length(x.S.DataTypes)
         x.S.InputDataStr = x.S.DataTypes{iDataType};
-
```
HenkMutsaerts, line 248: This is still valid, did you move this elsewhere?

Modules/xASL_module_Population.m
```diff
@@ -312,12 +293,26 @@
 		% Obtain a unique list of lesion names without the subject name
 		LesionUniqueROIList = unique(LesionROIList);
 
+		% Standard space analyzis in a specific ROI with no tissue restriction
```
HenkMutsaerts: analysis

HenkMutsaerts: Perhaps we should create a new code section here for clarity, that this concerns lesions. And move all the lesion-specific stuff here. Something like:

%% ===============================================================
%% 7.b Lesion ....
%% ===============================================================

Modules/xASL_module_Population.m
```diff
@@ -286,18 +276,9 @@
                 x.S.InputNativeSpace = 1;
                 x.S.InputAtlasNativeName = [x.S.Atlases{iAtlas} '_Atlas'];
                 xASL_wrp_GetROIstatistics(x);
-                % ROI statistics (optional: native space)
```
HenkMutsaerts, line 299: Where did the native space analysis with x.S.bSubjectSpecificROI==false go? This should be the default if you want native space. The addition for subject-specific lesions is good, but is a specific situation. E.g., the use case of a standard atlas with native space PVC is much more common.

HenkMutsaerts (changes requested):
Few points

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
jan-petr: In xASL_module_Population - this is always a string and never a cell. That's why this had to be fixed here...

Modules/xASL_module_Population.m
jan-petr: Yes of course - this is now inside the GetRoiStatistics functions

Modules/xASL_module_Population.m
jan-petr: I don't understand. 
this change here isn't about bSubjectSpecificROI

Modules/xASL_module_Population.m
jan-petr: Fixed

Functions/ImageProcessing/xASL_im_Lesion2Mask.m
```diff
@@ -196,6 +196,10 @@
 % First create the individual mask overlays with different colors
 for iIm=1:6
     OutIm{iIm} = xASL_vis_CreateVisualFig(x, {x.P.Pop_Path_rT1 VisualizeImage==iIm}, [], [0.75 0.35], [], {x.S.gray x.S.colors_ROI{iIm}});
+	if numel(OutIm{iIm}) == 1 && isnan(OutIm{iIm})
+		% In case the lesion did not make it to the visualization, we make an empty visualization with backgorund only
```
HenkMutsaerts: But this is what xASL_vis_CreateVisualFig already does? If you add an empty or NaN overlay, it just prints the background. 

BTW: isnan(OutIm{iIm}) will return a matrix back with ones, that's not what you want right?
If something crashed here with an empty lesionmask, I would create a warning instead and print what you suggest below, and 

else
     do what we already have
     
But essentially, that should also already be a general feature inside xASL_vis_CreateVisualFig (or else we should add it)

So you can remove your addition?

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
```diff
@@ -187,11 +187,12 @@
 namesROIlocal = x.S.NamesROI;
 
 if x.S.InputNativeSpace
-
-    
 	inputAtlasTmp = xASL_io_Nifti2Im(fullfile(x.dir.xASLDerivatives,x.SUBJECTS{1},listSessions{1},[x.S.InputAtlasNativeName '.nii']));
+	if x.S.bSubjectSpecificROI
+		inputAtlasTmp = round(inputAtlasTmp);
```
HenkMutsaerts: This shouldn't be necessary because atlases/masks should be resampled using nearest neighbor? Which resampling option did you use? Because round(linearResampling) might be OKish but round(B-spline) probably not :)

We also have a step where we resample atlases to our 1.5x1.5x1.5 mm resolution if the provided atlas was not in this resolution; this should also be nearest-neighbor.

In fact, upon loading atlases (in standard or whatever space) we should have a check that the unique(atlases) are integers only, and otherwise issue a warning, and do the rounding upon loading the atlases.

This is very tricky because the atlases can be 4D where each image volume is a single mask, but can also be 3D where each label integer is a single mask.

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
```diff
@@ -339,6 +340,9 @@
 				x.LeftMask = xASL_im_IM2Column(x.LeftMask, x.S.masks.WBmask);
 
 				inputAtlasTmp = xASL_io_Nifti2Im(fullfile(x.dir.xASLDerivatives,x.SUBJECTS{iSubject},listSessions{iSess},[x.S.InputAtlasNativeName '.nii']));
+				if x.S.bSubjectSpecificROI
+					inputAtlasTmp = round(inputAtlasTmp);
```
HenkMutsaerts: Same here

Modules/Module_Processing/Module_Structural/xASL_wrp_SegmentT1w.m
```diff
@@ -287,7 +287,7 @@
 rLesionList = xASL_adm_GetFileList(x.D.PopDir, ['(?i)^rLesion.*' x.P.SubjectID '\.nii'], 'FPList', [0 Inf]);
 LesionIM = zeros([121 145 121]); % assuming 1.5 mm MNI
 for iL=1:length(rLesionList)
-    LesionIM(xASL_im_ConvertMap2Mask(xASL_io_Nifti2Im(rLesionList{iL}))) = 1;
+    LesionIM(logical(xASL_im_ConvertMap2Mask(xASL_io_Nifti2Im(rLesionList{iL})))) = 1;
```
HenkMutsaerts: Shouldn't we add the logical inside xASL_im_ConvertMap2Mask instead?
Assuming that map should be converted to a single mask and not labels of multiple masks, which seems a safe assumption ;)

Modules/Module_Processing/Module_Structural/xASL_wrp_VisualQC_Structural.m
```diff
@@ -318,6 +318,10 @@ function xASL_adm_VisualCheckLesionRemoval(x, Lesion_list)
         
 		ImOut1 = xASL_vis_CreateVisualFig( x, {x.P.Pop_Path_rc1T1_ORI x.P.Pop_Path_rc2T1_ORI}, [], [1 0.8], [], []);
 		ImOut2 = xASL_vis_CreateVisualFig( x, {x.P.Pop_Path_rT1 LesionIM},[], [1 0.8], [], []);
+		% Since the lesion is really small, it can happen that it's not visible and the visualization will fail. In that case, display rT1 only
+		if numel(ImOut2) == 1 && isnan(ImOut2)
```
HenkMutsaerts: Why not 

if numel(ImOut2) == 1 && isnan(ImOut2)
    
else
     default visualization as we had it

Because now it runs twice.

Also, this functionality should be inside xASL_vis_CreateVisualFig, perhaps with a warning. I also commented this in the other PR

Modules/xASL_module_Population.m
HenkMutsaerts: OK as long as it is similar for both standard space & native space (because both can have this prefix). Then you can resolve

Modules/xASL_module_Population.m
HenkMutsaerts: OK I couldn't see it anymore but I see now that it is still there, my bad

HenkMutsaerts (changes requested):
Few minor points. Nice work!

Functions/ImageProcessing/xASL_im_Lesion2Mask.m
jan-petr: Not really. If the input is an empty matrix, then the output if the visualization is NaN and then it crashes everything. So I had to do this fix.

But I agree that we can integrate this check. So an empty input is now replaced with matrix with zeros inside the visualization function.

jan-petr: Done in 8026b5a7
OK?

Modules/Module_Processing/Module_Structural/xASL_wrp_SegmentT1w.m
jan-petr: Good point. Done.

Modules/Module_Processing/Module_Structural/xASL_wrp_VisualQC_Structural.m
jan-petr: Done as commented above.

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
jan-petr: Good point. Yes, atlases are binary. But while downsampling to native space, we can't use nearest neighbor - that would just not resample to the low-resolution properly. So that's why we smooth and interpolate.
Afterwards, rounding is perfectly valid to judge if a voxel is above of below 50% of the partial volume. This is the most correct approach. But I agree that we might do this thresholding elsewhere.

So I'll simply do this for atlases right in the resampling function. OK?

External/SPMmodified/xASL/xASL_im_ConvertMap2Mask.m
```diff
@@ -76,7 +76,7 @@
     % Bypassed the previous, to keep fair across all segmentations
     IMout           = IMin>SimpleHalfValue;
 else
-    IMout           = IMin;
+    IMout           = IMin > 0;
```
HenkMutsaerts: I would change as follows:

replace

else
    IMout           = IMin > 0;

by 

IMout = logical(IMout);

Because if you already created a logical, this won't change it.
Now it is confusing to the reader as you first check if it has more than zeroes and then you apply that, whereas it is just a cryptic way of writing logical(IMout) :)

Minor comment of course.

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
HenkMutsaerts: That's fine, but note that we also allow MNI atlases to be a single 3D matrix, where each label is a different ROI.
E.g. 1 == caudate 2 == putamen etc.
Your approach is fine for a 4D matrix where each 3D matrix is a binary mask. But for a label image this won't work.

So, that means that we always need to convert a 3D label matrix to a 4D matrix with multiple binary 3D image volumes.

Additionally, nearest neighbor warping of course goes much faster than higher DoF. So if we could smooth and threshold in high res and interpolate with nearest neighbor that would be great, but you probably argue that that is incorrect :)

Most important is the label use case explained above, you can try /ExploreASL/External/Atlases4ROIs/LicenseLimited/Hammers.nii.gz

in native space as use case.

HenkMutsaerts (changes requested):
It's mainly the 4D label matrix that needs checking. You can test with /ExploreASL/External/Atlases4ROIs/LicenseLimited/Hammers.nii.gz in native space. See comment.

External/SPMmodified/xASL/xASL_im_ConvertMap2Mask.m
jan-petr: Done

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
jan-petr: We now transform Lesions not from T1, but from standard space. So we have the perilesional and contralateral ROIs created in standard space.

Moreover, this now works nicely with binary 4D masks. For 3 masks with multiple labels, we use nearest neighbor interpolation. We might want to change that in the future and do the transformation properly by binarizing the mask for each label.

Regarding your comment. Taking 3D multilabel and smoothing it and the interpolating with nearest neighbor isn't just incorrect, it won't work at all :) We would have to split it into a 4D binary. Which is rather easy and not even computationally difficult. I can easily implement that...

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
jan-petr: Yes. All this is fixed. I now make sure that Lesions and ROIs are now properly binarized during the transformation. So we don't need any rounding here or on the other location. So this code is no longer needed and I have removed it.

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
HenkMutsaerts: OK, then what I would do, is change xASL_spm_Deformations. This is where we should do the downsampling, presmoothing etc. Not elsewhere. The advantage is that we then always do it the same and correct way, and it simplifies code/debugging.

1. Implement an option inside xASL_spm_Deformations to do the proper downsampling with presmoothing (or automatically detect if presmoothing is needed)
2. Automatically also process 4D images correctly
3. Automatically also process 3D labels correctly

4. Check everywhere this was done outside xASL_spm_Deformations, and revamp this such that we do it inside xASL_spm_Deformations

You probably want to do this in a new issue?

HenkMutsaerts (approved):
See last comment for new issue

Modules/xASL_module_Population.m
```diff
@@ -235,16 +236,6 @@
     % Iterate over DataTypes
     for iDataType=1:length(x.S.DataTypes)
```
jan-petr, line 247: For native space and Lesion/ROI, we remove the protection for too small ROIs.

jan-petr (commented):
One last thing to fix

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
jan-petr: Agree. New issue #1864

PR #1829, opened 2024-12-03
Feature #1828 temporal variability

### LINKED ISSUE #1828: Temporal variability of a metric in the ASL QC module
### Description

In general, we want to be able to take features from the PWI4D image (such as CBF or spatial CoV) per deltaM, and output their variance/SD/MAD.

As a use case, we take the pvGM>0.7 spatial CoV of each deltaM of the PWI4D.nii and output its sCoV.

In standard space, we have this feature in #594.

### Tasks

- [x] Always save CBF4D in quantification for single-PLD and single-TE  0670e96e
- [x] Create new function in xASL_wrp_VisualQC_ASL, between step 8 and 9. JAN: I added it to step 9 as it already deals with CBF and masks, so best to keep this unified and together.  0670e96e
- [x] Calculate mean, median, spatial sCoVfor each subtraction repetition from PWI4D. Using pvGM>0.7.  0670e96e JAN: diffCoV 
- [x] Calculate diffCoV. JAN: Note that diffCoV requires spatial relations and thus cannot be calculated on masked images like is done for GM. So I had to use the original images before masking. 86a8c775
- [x] I calculated temporal SD over GM and WM in CBF. and temporal mean and SD in ScoV - the normalization can be done afterwards.
- [x] delete CBF4D after the end of ASL module unless save CBF4D is specified fe0e7457

### How to test

temporal sCoV -> do we want to exclude the cerebellum?
Also, we should actually include the CoW; so for now perhaps the full brain (pGM+pWM) works best.

Pasted Graphic
Pasted Graphic 1

### Release notes

Report sCoV in QC based on CBF4D

jan-petr:
### Linked issue

Close #1828


Modules/xASL_module_ASL.m
```diff
@@ -472,21 +472,21 @@
...
-	if x.modules.asl.SaveCBF4D
-		if size(xASL_io_Nifti2Im(x.P.Path_PWI4D_used), 4) == 1
-			warning('x.modules.asl.SaveCBF4D was requested but only one volume exists, skipping');
-		else
-			fprintf('%s\n','Quantifying CBF4D in native space');
-			xASL_wrp_Quantify(x, x.P.Path_PWI4D_used, x.P.Path_CBF4D, x.P.Path_rM0, x.P.Path_SliceGradient, true);
 
-			fprintf('%s\n','Quantifying CBF4D in standard space');
-			xASL_wrp_Quantify(x, x.P.Pop_Path_PWI4D_used, x.P.Pop_Path_qCBF4D, [], [], true);
+		% allow 4D quantification as well, storing CBF4D. This is currently not implemented for multi-PLD/multi-TE (BASIL/FABBER)
```
HenkMutsaerts: We don't have to change this yet so might want to revert this.
But let's see.
Also: we now do this in both native & standard space.

HenkMutsaerts (commented):
Checking now, wait with changing

Modules/xASL_module_ASL.m
jan-petr: Why reverting? This was there already for a very long time...

HenkMutsaerts: I meant that you are now defaulting to calculating CBF4D, we can revert this defaulting, and only start defaulting this when we indeed find out that this is useful.

HenkMutsaerts (approved):
See single comment for you to understand what I meant, but I'm fine for now. I just thought that we didn't need to change the SaveCBF4D yet, until we know for sure that it is useful as an extra vascular parameter or for QC.

HenkMutsaerts (approved):

Issue #594, opened 2021-05-22
Expand functionality of xASL_wrp_GetROIStatistics


HenkMutsaerts:
### Description

I would like to expand the functionality of the ROI statistics in the population module by a temporal value, e.g. the temporal variability of spatial CoV, and a ratio, e.g. the ratio of average CBF in the borderzone divided by average CBF in the proximal parts of vascular territories.

### Requirements

We need to add some flexibility without losing the modularity of these functions. Also, we might want to do this for the standard space output only, as @jan-petr still needs to fix the broken modularity for the native space output.

Suggested changes:
1. `xASL_stat_GetROIStatistics` calculates a function - e.g. mean/median/SD/etc - for a single ROI, for a single image volume - e.g. `qCBF.nii`. This function should allow doing this with a 4D input, e.g. `qCBF4D.nii`, with a vector as output instead of a scalar, per subject/ROI/function.
2. `xASL_wrp_GetROIStatistics` detects this vector and performs a temporal function on this, e.g. SD. This temporal function should be an input argument to `xASL_wrp_GetROIStatistics` just like the current ROI-level input function(s).
3. `xASL_wrp_GetROIStatistics` should be able to compare ROI values, depending on the input argument, e.g.:
    * ratio: ratio between two ROIs, e.g. borderzone / proximal
    * asymmetry index: relative difference between two ROIs, e.g. left vs right for lateralization
    * the same but for multiple ROIs, e.g. relative SD between multiple ROIs

PR #1837, opened 2024-12-06
Feature #1831 temporal var population

### LINKED ISSUE #1831: Temporal variability in population module
### Description

We want to run the temporal variability stats that we defined in #1828 in `xASL_qc_CollectQC_ASL` also in the population module, because I will generate a new NIfTI mask that has both the supratentorial WB mask and the Circle of Willis.

### Tasks
- [x] Cherry pick from #594 and remove it there 7808fb23
- [x] Load data 4D 7808fb23
- [x] inside Population module, inside native space analysis, when SaveCBF4D is on (so that we have CBF4D.nii in the native space) we calculate 4D stats 7808fb23
- [x] Using any atlas provided in dataPar.json and called for population analysis. Move it to the native space and get the same temporal variability etc. parameters as defined in xASL_qc_CollectQC_ASL, but then within this mask 
-> HENK: I thought of doing this in a separate function, because otherwise all the ROIstats calculations have to accommodate an extra dimension (for 4D instead of 3D, add the different ROIs etc, so multi-dim, so quite an overhaul of the ROIstats functions, which I started in #594 but will create much extra work. And we want to change the native space stuff. So I thought easiest for now is to create a function separate from GetROIstats and do this all there, but you can also do it with GetROIstats, but then please create separate sections or subfunctions for this . 7808fb23
- [x] If CBF4D image is found, then it will do all analysis (all atlases and all contrasts like scov and CBF) in native space in 4D
-> I meant in 3D per deltaM, so it will do it nDeltaM times, and then after we take the mean/median/sCoV of these parameters, like in #1828
-> HENK: note that  this has to be per deltaM calculated, and then the temporal mean or SD of that. So 1. spatial 2. temporal function (not the other way around). 11005589
- [x] We calculate spatialCoV4D and diffCoV4D over an atlas from MNI transformed to native space per 1 deltaM. From that, we take mean/median/SD.
-> for each deltaM separately indeed, so we calculate spatialCoV3D and diffCoV3D for each deltaM separately, and then get the mean/median/SD etc over those nDeltaM spatialCoV3D's etc. 11005589
- [x] Fix calculate sptialCoV function to work in 4D

- [x] 1. Add function for within-mask calculation of the Z-direction center-of-mass (slice with the average label position: first take average of each slice within the mask, then the center-of-mass across the Z-direction); center-of-mass is the "intensity-weighted center", so not the geometric center (for which you would binarize first).
- [x] Case A) Function (1) is performed per deltaM, and the temporal SD and mean are calculated of this. JAN: Called LabLoc4D_mean and LabLoc4D_sd. I've additionally did the same for sCoV- first calculate per slice/repetition, then center per repetition, then mean/sd over repetitions.
- [x] Case B) Temporal SD is first calculated voxel-wise and function (1) is ran after it. JAN: I calculated temporal SD voxelwise, then slice-wise mean and weighted-slice-mean. Called StdLoc4D.
- [x] xASL_stat_ComputeDifferCoV doesn't work in the current implementation, the input arguments don't correspond. JAN: Fixed, minor error in the header that I forgot to update.
- [x] HENK: I did the following changes in my last commit for xASL_stat_GetROIstatistics:
* Some calculations do not need the second dimension because they already had 1 dimension
* Also initialize sCoV4D_SliceWise & tempSD4D_SliceWise (this fixes a bug if different subjects have different FoV/nSlices)
* do not set NaNs to zeroes (to avoid bias), though we may want to improve vectorSlice such that it doesn’t take slices that only contain NaNs
* flip negative values by abs() instead of setting them to 0
* Add new parameters to output
* renamed 
mean4D -> mean4D_SliceWise
sCoV4D -> sCoV4D_SliceWise
tempSD4D -> tempSD4D_SliceWise
vecSlice -> vectorSlice
mean4Dcenter -> mean4DcenterSlice
sCoV4Dcenter -> sCoV4DcenterSlice
x.S.DAT_LabLoc4D_mean_PVC0 -> x.S.DAT_LabelSliceLocation4D_tMean_PVC0
x.S.DAT_LabLoc4D_sd_PVC0 -> x.S.DAT_LabelSliceLocation4D_tSD_PVC0
x.S.DAT_sCoVLoc4D_mean_PVC0 -> x.S.DAT_LabelsCoV_SliceLocation_tMean_PVC0
x.S.DAT_sCoVLoc4D_sd_PVC0 -> x.S.DAT_LabelsCoV_SliceLocation_tSD_PVC0
x.S.DAT_StdLoc4D_PVC0 -> x.S.DAT_SlicetSD_SliceLocation_PVC0

### Release notes

New population statistics on CBF4D data.

jan-petr:
### Linked issue

Close #1831

### DIFF
--- a/Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
+++ b/Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
@@ -815,31 +815,37 @@ for iSubject=1:x.dataset.nSubjects
 					x.S.DAT_diffCoV4D_mean_PVC0(SubjSess, iROI) = xASL_stat_MeanNan(diffCoV4D);
 					x.S.DAT_diffCoV4D_sd_PVC0(SubjSess, iROI) = xASL_stat_StdNan(diffCoV4D);
 
-					mean4D = zeros(size(Data4D, 3), size(Data4D, 4));
+                    % Compute average label position stats
+                    % Initialization
+					mean4D_SliceWise = zeros(size(Data4D, 3), size(Data4D, 4));
+                    sCoV4D_SliceWise = zeros(size(Data4D, 3), size(Data4D, 4));
+                    tempSD4D_SliceWise = zeros(1, size(Data4D, 3));
+
 					%diffCoV4DCenter = zeros(1, size(Data4DIm, 2));
 					% Precalculate the temporal values in 2D for each slice and repetition
 					for iSlice=1:size(Data4D, 3)
 						for iRepetition=1:size(Data4DIm, 2)
-							mean4D(iSlice, iRepetition) = xASL_stat_ComputeMean(Data4D(:, :, iSlice, iRepetition), CurrentMaskNotVascularFull(:, :, iSlice), 0, 0, 1);
-							sCoV4D(iSlice, iRepetition) = xASL_stat_ComputeSpatialCoV(Data4D(:, :, iSlice, iRepetition), CurrentMaskNotVascularFull(:, :, iSlice), 0, 0);
+							mean4D_SliceWise(iSlice, iRepetition) = xASL_stat_ComputeMean(Data4D(:, :, iSlice, iRepetition), CurrentMaskNotVascularFull(:, :, iSlice), 0, 0, 1);
+							sCoV4D_SliceWise(iSlice, iRepetition) = xASL_stat_ComputeSpatialCoV(Data4D(:, :, iSlice, iRepetition), CurrentMaskNotVascularFull(:, :, iSlice), 0, 0, 1);
 						end
 						% First STD across repetitions, the slice-wise mean
-						tempSD4D(iSlice) = xASL_stat_ComputeMean(xASL_stat_StdNan(Data4D(:, :, iSlice, :), [], 4), CurrentMaskNotVascularFull(:, :, iSlice), 0, 0, 1);
+						tempSD4D_SliceWise(iSlice) = xASL_stat_ComputeMean(xASL_stat_StdNan(Data4D(:, :, iSlice, :), [], 4), CurrentMaskNotVascularFull(:, :, iSlice), 0, 0, 1);
 					end
 
 					% For each repetition - calculate the best slice
-					vecSlice = repmat((1:size(Data4D, 3))', [1 size(mean4D, 2)]); % Vector with slice numbers
-					mean4D(isnan(mean4D)) = 0; mean4D(mean4D<0) = 0; % Clean NaNs and negative values
-					mean4Dcenter = xASL_stat_SumNan(mean4D.*vecSlice, 1)./xASL_stat_SumNan(mean4D, 1); % Calculate the weighted mean
-					x.S.DAT_LabLoc4D_mean_PVC0(SubjSess, iROI) = xASL_stat_MeanNan(mean4Dcenter);
-					x.S.DAT_LabLoc4D_sd_PVC0(SubjSess, iROI) = xASL_stat_StdNan(mean4Dcenter);
+					vectorSlice = repmat((1:size(Data4D, 3))', [1 size(mean4D_SliceWise, 2)]); % Vector with slice numbers
+					mean4D_SliceWise = abs(mean4D_SliceWise); % flip negative values
+					mean4DcenterSlice = xASL_stat_SumNan(mean4D_SliceWise.*vectorSlice, 1)./xASL_stat_SumNan(mean4D_SliceWise, 1); % Calculate the weighted mean
+					x.S.DAT_LabelSliceLocation4D_tMean_PVC0(SubjSess, iROI) = xASL_stat_MeanNan(mean4DcenterSlice);
+					x.S.DAT_LabelSliceLocation4D_tSD_PVC0(SubjSess, iROI) = xASL_stat_StdNan(mean4DcenterSlice);
 
-					sCoV4D(isnan(sCoV4D)) = 0; sCoV4D(sCoV4D<0) = 0; % Clean NaNs and negative values
-					sCoV4Dcenter = xASL_stat_SumNan(sCoV4D.*vecSlice, 1)./xASL_stat_SumNan(sCoV4D, 1); % Calculate the weighted mean
-					x.S.DAT_sCoVLoc4D_mean_PVC0(SubjSess, iROI) = xASL_stat_MeanNan(sCoV4Dcenter);
-					x.S.DAT_sCoVLoc4D_sd_PVC0(SubjSess, iROI) = xASL_stat_StdNan(sCoV4Dcenter);
+					sCoV4D_SliceWise = abs(sCoV4D_SliceWise); % flip negative values
+                    % Note that sCoV4D is only negative for negative means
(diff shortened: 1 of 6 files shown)


Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
```diff
@@ -520,6 +518,7 @@
 
 		%% 4.c Load data
         Data3D = []; % initialize empty variable
+		Data4D = []; % initialize empty variable
```
HenkMutsaerts, line 521: The way I would write it, we always have Data4D (which can be either 3D or 4D, depending on the input), which would get us cleaner code. 

I agree that the way you started this is quicker for now, but we could revamp this later. So fine for now

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
```diff
@@ -754,13 +761,13 @@
 					end
 
                     %% CBF (now remove vascular artifacts)
-                    if x.S.bMasking(2)==1 % apply vascular mask
-                        CurrentMask = CurrentMask & VascularMask;
+                    if x.S.bMasking(2)==1 % apply vascular mask, but rename this to CurrentVascular mask, because we still need the original CurrentMask for 4D temporal calculations
```
HenkMutsaerts: I like this, perhaps for clarity change this such that:

* rename CurrentMask -> CurrentMaskNotVascular
* rename CurrentVascularMask -> CurrentMaskVascular

to be sure. So we won't have CurrentMask anymore. Good idea?

As long as for CBF we always have CurrentMaskVascular (except when x.S.bMasking(2)==0 ) and for non-CBF (e.g. spatial CoV) we always have CurrentMaskNoVascular

That's probably why I did it by overwriting CurrentMask, that CurrentMask is always overwritten with new edits, rather than multiplying the number of CurrentMasks for each edits (which could lead to a lot of masks down the line)

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
```diff
@@ -786,12 +793,38 @@
...
+					if sum(CurrentMask(:)) > MinVoxels
+						CurrentMaskFull = xASL_im_Column2IM(CurrentMask, x.S.masks.WBmask);% We need the full mask for the Diff-CoV calculation
+
+						for iRepetition=1:size(Data4DIm, 2)
+							diffCoV4D(iRepetition) = xASL_stat_ComputeDifferCoV(Data4D(:, :, :, iRepetition), CurrentMaskFull);
+						end
+					else
+						diffCoV4D = NaN;
+					end
+					% Compuate the temporal values - we add here the type of statistics and PVC status, the contrast type (CBF/ATT/Tex) is assigned outside of this function
```
HenkMutsaerts: Compuate -> Compute

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
```diff
@@ -786,12 +793,38 @@
...
+					sCoV4D = zeros(1, size(Data4DIm, 2));
+					diffCoV4D = zeros(1, size(Data4DIm, 2));
+
+					% Precalculate the temporal values - use functions that work in 3D only to do it for each temporal point
+					for iRepetition=1:size(Data4DIm, 2)
+						sCoV4D(iRepetition) = xASL_stat_ComputeSpatialCoV(Data4DIm(:, iRepetition), CurrentMask, MinVoxels, 0);
+					end
 
+					if sum(CurrentMask(:)) > MinVoxels
+						CurrentMaskFull = xASL_im_Column2IM(CurrentMask, x.S.masks.WBmask);% We need the full mask for the Diff-CoV calculation
```
HenkMutsaerts: Why don't we do this inside xASL_stat_ComputeDifferCoV ? To keep this code cleaner?

HenkMutsaerts (changes requested):
Some minor stuff.

Note that the way I started it in #594, is that you can give any function, and that function would be computed per deltaM, and later averaged etc. That would be more future-proof, because soon I would like to replace the spatial CoV by the proximal-to-distal voxel ranking and computing the correlation coefficient for each deltaM. Perhaps we can get that done for the OHBM abstract, as that would be very novel. But I agree that it is easier to hard-code the functions for now, that will also simplify your wish to remove im2column here (if I remember correctly)

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
jan-petr: Done

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
jan-petr: Fixed

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
jan-petr: Yes of course. MinVoxels is now inside.

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
```diff
@@ -769,29 +775,77 @@
...
+							mean4D(iSlice, iRepetition) = xASL_stat_ComputeMean(Data4D(:, :, iSlice, iRepetition), CurrentMaskNotVascularFull(:, :, iSlice), 0, 0, 1);
+							sCoV4D(iSlice, iRepetition) = xASL_stat_ComputeSpatialCoV(Data4D(:, :, iSlice, iRepetition), CurrentMaskNotVascularFull(:, :, iSlice), 0, 0);
+						end
+						% First STD across repetitions, the slice-wise mean
+						tempSD4D(iSlice) = xASL_stat_ComputeMean(xASL_stat_StdNan(Data4D(:, :, iSlice, :), [], 4), CurrentMaskNotVascularFull(:, :, iSlice), 0, 0, 1);
+					end
+
+					% For each repetition - calculate the best slice
+					vecSlice = repmat((1:size(Data4D, 3))', [1 size(mean4D, 2)]); % Vector with slice numbers
+					mean4D(isnan(mean4D)) = 0; mean4D(mean4D<0) = 0; % Clean NaNs and negative values
```
HenkMutsaerts: NaNs don't need cleaning I would say, as we use our nonNan stats code throughout.

Negative values are tricky. They are either background suppression artifacts or vascular artifacts (which can also turn negative annoyingly; this is perhaps why the sCoV per slice could indeed work). In any case what we do now assumes vascular artifacts only, so I will flip them for this artifact.

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
jan-petr: Agree. Lets do the in the follow-up issues. and close this now.

jan-petr: New issue #1849 is created

HenkMutsaerts (approved):

PR #1842, opened 2024-12-08
Closes #1841 xASL_adm_CleanUpBeforeRerun: only add locked dirs to CurrentDir…

### LINKED ISSUE #1841: xASL_adm_CleanUpBeforeRerun crashes with empty lock dirs
### Description

With a complete rerun of the structural module, we first also check which locked dirs need to be deleted, and flexibly add these to a list. However, when locked dirs are empty, this flexible list adding doesn't work as it cannot add an empty cell structure to CurrentDir(end+1) (left & right side of assignment have different sizes).

### Tasks

- [x] xASL_adm_CleanUpBeforeRerun: only add locked dirs to CurrentDir list if they are not empty

### Release notes

Bugfix cleaning structural module before rerunning

HenkMutsaerts:
### Linked issue

Closes #1841


jan-petr (approved):
All ok

PR #1851, opened 2024-12-17
Closes #1850 illegal chars import inconsistency

### LINKED ISSUE #1850: Import inconsistency with illegal character-name changes
### Description

When we have a subject token in sourcedata with an illegal name, e.g. `subi_105`, this is changed in DCM2NIIX to `subi105`, but this is not transferred to NII2BIDS (at least not when both parts are run separately as [1,0,0] and [0,1,0] imports).

### Tasks

- [x] xASL_module_Import: Only create 999_ready if both states 1 (dcm2nii) & 2 (nii2bids) are finished
- [x] Correct illegal characters in `subjectName` as well (listSubjectsSessions gets the already corrected names from derivatives/Temp). Alternatively (Henk's favorite): we only use the listSubjectsSessions: i.e. everything that is in derivatives/Temp, for NII2BIDS.

### Release notes

Fix minor issue in the import of subjects with special characters in the name.

HenkMutsaerts:
### Linked issue

Closes #1850


Modules/Module_Import/xASL_wrp_NII2BIDS.m
```diff
@@ -90,6 +90,19 @@
 	listSubjectsSessions = xASL_adm_GetFileList(x.modules.import.imPar.TempRoot,[],false,[],true);
     for iSubjectSession = 1:length(listSubjectsSessions)
         % Only run it for the current subject (maybe we can do this more elegantly in the future)
+
+        % In case there were illegal characters in subjectName, we get rid
+        % of them here (because the same was done in DCM2NIIX as well)
+        % And if we run first only DCM2NIIX, and on the next instance only
+        % run NII2BIDS, this name correction will not be transferred here
+        % otherwise
+        subjectName = xASL_adm_CorrectName(subjectName,2);
```
jan-petr: I am a bit confused. First of all, Skyler provided the data without a subject name. So any subjectName special characters would have to be provided by you. Or am I missing something.

I have tried subjectname "sub+\!1" and when arriving to this line, subjectName was perfectly free of any special characters.

So I'd need to know which special characters you had a problem with and why. Just to find out if this is Luna related etc. Because for me it work perfectly for "sub1" and for "sub+\!1" and this code change isn't needed for me.

I've tried of course running steps [1 1 0] or [1 0 0] with [0 1 0] and it was always doing well even without this edit. So I need to first reproduce this error and see if this really is a problem and if we can't find a better solution.

Modules/xASL_module_Import.m
```diff
@@ -183,7 +183,10 @@
     x = xASL_adm_CleanUpX(x);
     
     % We need to terminate the module correctly
-    x.mutex.AddState('999_ready');
+    if x.mutex.HasState(StateName{1}) && x.mutex.HasState(StateName{2})
```
jan-petr, line 186: This part is of course perfectly fine.

jan-petr (changes requested):
I can't reproduce the error and therefore cannot confirm the solution works/is necessary

Modules/Module_Import/xASL_wrp_NII2BIDS.m
HenkMutsaerts: OK try reproducing it for me:

1. pull last develop version
2. Run: ExploreASL('/Temp4Jan/SkylerZhangUser',[1,0,0],0) which contains one subject with the name subi_105)
3. Verify that this gave '/Temp4Jan/SkylerZhangUser/derivatives/ExploreASL/temp/subi105' (so it removed the _ )
4. Run: ExploreASL('/Temp4Jan/SkylerZhangUser',[1,1,0],0) with a breakpoint in xASL_wrp_NII2BIDS at line 90
listSubjectsSessions will give you the new name: subi105
subjectName will give you the old name: subi_105
Hence, it will skip running NII2BIDS for this subject.

Solution: also run xASL_adm_CorrectName on subjectName

This is platform independent: because the old names (with the illegal character) in sourcedata is not changed
and the new name (without the illegal character) in temp is changed.

jan-petr: OK. Got it I tried to run ExploreASL as [1 0 0] and then [0 1 0], which works. I haven't tried [1 0 0] followed by [1 1 0].

Anyway - the problem happens on an earlier level, so we should fix this early on and not here. See my newest commit.

HenkMutsaerts: Looks good. Though at first sight, it seemed this code belonged to DCM2NII only, and not also to NII2BIDS, which is confusing (as this is initialization code).

Is it a small effort for you to create 

/ExploreASL/Modules/Module_Import/Initialization/

and move all import initialization functions (that works for all dcm2nii nii2bids scripts) to this folder?

jan-petr: Done

jan-petr (approved):
OK

PR #1863, opened 2024-12-24
Closes #1862 b use mni as dummy structural

### LINKED ISSUE #1862: Minor feedback improvement for bUseMNIAsDummyStructural
### Description

Minor feedback improvement for bUseMNIAsDummyStructural. Some users keep this option enabled. They should be aware that if some subjects don't have a structural module yet processed, it will continue processing ASL data with dummy MNI images. If you then reprocess modules or something else, this can potentially lead to registration issues as the NIfTI orientation matrix is not always restored upon rerunning.

### Tasks
- [x] xASL_module_ASL: improve bUseMNIAsDummyStructure feedback
- [x] xASL_wrp_RegisterASL: improve bUseMNIAsDummyStructure feedback
- [x] Good practice with rerunning is first deleting the derivatives folder, perhaps we can also add a warning or something for this somewhere

### How to test

### Release notes

Improving user feedback with bUseMNIAsDummyStructural

HenkMutsaerts:
### Linked issue

Closes #1862

### DIFF
--- a/Modules/Module_Processing/Module_ASL/xASL_wrp_RegisterASL.m
+++ b/Modules/Module_Processing/Module_ASL/xASL_wrp_RegisterASL.m
@@ -129,10 +129,6 @@ if x.modules.asl.bDCTRegistration
 	x.modules.asl.bRegistrationContrast = 3;
 end
 
-% By default, don't use dummy structural even if the structural image is missing
-if ~isfield(x.modules.asl,'bUseMNIasDummyStructural') || isempty(x.modules.asl.bUseMNIasDummyStructural)
-	x.modules.asl.bUseMNIasDummyStructural = false;
-end
 
 %% B. Manage OtherList
 % Define OtherList for registration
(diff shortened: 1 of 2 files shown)


Modules/Module_Processing/Module_ASL/xASL_wrp_RegisterASL.m
```diff
@@ -217,10 +213,16 @@
 
 
 %% E. Allow registration without structural data
+%% PM: This part is partly equal to part xASL_module_ASL > section A2
 StructuralDerivativesExist = xASL_exist(x.P.Path_y_T1, 'file') && xASL_exist(x.P.Path_c1T1, 'file') && xASL_exist(x.P.Path_c2T1, 'file');
 StructuralRawExist = xASL_exist(x.P.Path_T1, 'file') || xASL_exist(x.P.Path_T1_ORI, 'file');
 
-% In case that we don't have the structural derived data, we need to check the reason
+% By default, don't use dummy structural even if the structural image is missing
```
jan-petr: Since we are modifying this, wouldn't it make more sense to move this to the ASL module where we deal with all dataPar parameters?!

jan-petr (changes requested):
Fine with all the new comments and warnings. But we might consider moving the parameter defaults to the same place all other defaults are assigned in the ASL module.

Modules/Module_Processing/Module_ASL/xASL_wrp_RegisterASL.m
HenkMutsaerts: Not sure, we have 6 parts A-6 that are more "special/specific" that parsing ASL processing parameters.
This is about if we will run xASL_module_ASL yes or no (in section A2 in xASL_module_ASL).

Here in xASL_wrp_RegisterASL, we repeat this for the case that somebody runs this wrapper separately or does debugging. I thought about standardizing it, but it the sections in both functions are not completely identical; they have a different function.

jan-petr: Sure - for debugging and separate execution, we can enable/disable the boolean. 
But all dataPar processing booleans are initialized in ASL module. It makes no sense to do it elsewhere. I of course only talk about this boolean.

HenkMutsaerts: I didn't change that. I only improved its behavior here and at the start of the ASL module.

HenkMutsaerts: Fixed in 0248a3ea5f9cce491b593c7befbbaf5af5055b64

jan-petr (approved):
OK

PR #1876, opened 2025-01-09
#1875 adm_SortDicomToFolders minor fix

### LINKED ISSUE #1875: DICOM sorting script with series number only
### Description

The xASL_adm_SortDicom script works fine, but it isn't able to process the data in case there are highly anonymized scans that don't even have protocol description, but only series number. 

In that case - use twice the series number for renaming.

While suboptimal and people should properly export their data, some PACS have this issue, so we need this for certain sites.

### Release notes

DICOM sorting with series number only.

jan-petr:
### Linked issue

Close #1875


HenkMutsaerts (approved):
Improved the warnings a bit for clarity & make our life easier, you can edit if I made a mistake (didn't test it). Do you agree with this?

PR #1880, opened 2025-01-11
Feature #1516 g ee asl

### LINKED ISSUE #1516: GE enhanced ASL 7 PLDs check
### Description

Figure out correct parameter reading for GE eASL. We have an example flavor here, and we know the correct PLD and labeling duration from a sequence description file. But this doesn't match the DICOM information at all.

See example here. The flavor is added and working, but automatic DICOM parameter parsing not.
GE_PCASL_3Dspiral_DV29.1_eASLPLD7_1

### Tasks

- [x] Read GE private fields 015387a0
- [x] distinguish eASL and normal ASL GE version 015387a0
- [x] Convert correctly GE private fields to PLD values 015387a0
- [x] GE and DICOM fields for labeling duration and PLD are a mess and do not correspond to the real values at all. But there's no clear way how to recalculate this. 015387a0
- [x] NII2BIDS crashes without ASLcontext, which it should be able to detect automatically with latest GE scans? (M0 is always at the end and the rest is just deltaM). See if we can extract this also for eASL GE. 015387a0
- [x] First do GE special tags and only then merge dcm2nii and studyPar.json
- [x] Empty x.S added to dataPar during BIDS2Legacy
- [x] Compile FLUX version
- [x] Compile DCMTK windows
- [x] Compile DCMTK MacOS
- [x] Update flavors
- [x] test flavors for old DCMTK - GE_PCASL_3Dspiral_DV25.0_eASL_1 generates 1 PLD instead of adding 0 to m0scan.
- [x] test flavors for new DCMTK (both should work as we still provide studyPar.json info until all is recompiled)

### Release notes

Read GE private fields.

jan-petr:
### Linked issue

Close #1516


External/DCMTK/xASL_mex_DcmtkRead.cpp
```diff
@@ -38,6 +38,13 @@
  *    RescaleSlopeOriginal - 0x2005, 0x140a or 0x2005, 0x110a
  *    GELabelingType - 0x0019, 0x109C
  *    GELabelingDuration - 0x0043, 0x10A5
+ *    GESequenceName - 0x0019, 109e
```
HenkMutsaerts, line 41: Perhaps nice to mention here what each CV contains?

External/SPMmodified/spm_dicom_dict.txt
```diff
@@ -3671,6 +3671,13 @@ FFFE	E0DD	SequenceDelimitationItem	UN	1
 0043	192C	EffectiveEchoSpacing	SS	1
 0019	1028	BandwidthPerPixelPhaseEncode	SS	1
 0019	109C	GELabelingType	CS	1
+0019	109E	GESequenceName	CS	1
```
HenkMutsaerts, line 3674: Can you add this to SPM_modified text so we can keep track what we need to tweak when we update SPM?

Functions/InputOutput/xASL_io_DcmtkRead.m
```diff
@@ -173,6 +184,35 @@
 	header.AssetRFactor = str2double(header.AssetRFactor);
 end
 
+% Adapt the field type
```
HenkMutsaerts, line 187: What do you want if these fields exist but don't have the format you expected? Do you want a warning for the user to report back to you?

Do we, in general, want warnings that tell the user that they can contact us, and we collect more flavors? That's always better than just a simple crash or non-satisfied user :)

Modules/Module_Import/DCM2NII/xASL_imp_DCM2NII_ConvertScan.m
```diff
@@ -200,7 +200,7 @@
                 warning(['JSON sidecar missing for ' nii_files{iFile}]);
             end
         end
-        if ~isempty(jsonFiles) && isempty(regexp(first_match, '[nii|nii\.gz]$', 'once'))
+        if ~isempty(jsonFiles) && isempty(regexp(first_match, '(nii|nii\.gz)$', 'once'))
```
HenkMutsaerts, line 203: I have a deja vu :) But that could be another function. Should we check find and replace, if Michael made this mistake at multiple locations?

Modules/Module_Import/NII2BIDS/xASL_bids_BIDSifyASLJSON.m
```diff
@@ -104,31 +105,178 @@
 	end
 end
 
-%% 5. Prioritize DICOM fields over the manually provided studyPar fields
+%% 4b. GE parameter parsing
```
HenkMutsaerts, line 108: Sorry for asking extra work, and I know this is not something you changed essentially here, but such cases it seems to me it is worth it to move this to a specific GE subfunction, and then the separate steps we do

%% 1. Conversion of eASL acquisition parameters

etc, and then we can summary this list in the subfunction header (doesn't have to be a fancy function header, but it's nice to have this summary) -> also chatGPT does this nicely :)

HenkMutsaerts (changes requested):
Minor stuff

Functions/InputOutput/xASL_io_DcmtkRead.m
jan-petr: Not really needed. They are either read as char or numeric and then we're fine with it as this is the format we want. Or they are in number/char and that's what we are catching here and converting.

So here, we are already catching the special examples and converting. No warning is needed as this is a common situation, nothing exceptional. So we are just making sure it correctly converts them.

A good comment you say, but right now, we are either able to convert this. And if not, it's hard to recognize it - or else, it's hard to recognize that this is simply not an anonymized sequence. So I wouldn't put a message to contact us. No flavor, even unknown ones should be crashing.

Modules/Module_Import/DCM2NII/xASL_imp_DCM2NII_ConvertScan.m
jan-petr: Haha. Either that. Or you fixed that in another branch :) And I'll deal with that in conflict resolving :)

Modules/Module_Import/NII2BIDS/xASL_bids_BIDSifyASLJSON.m
jan-petr: Of course. no prob. we have a subfunction with headers now...

HenkMutsaerts (approved):

PR #1887, opened 2025-01-23
#1886 Add verbosity option to Gzip

### LINKED ISSUE #1886: Suppress the printing of 'file replaced by its .gz' by default
### Description
 
At the end of the processing, it is quite overwhelming to have such an immense number of lines printed saying, "This file was replaced by its .gz." It's so much that it's not even possible to scroll up and check the last steps  –  of course, this is possible to check in the log files, but it's still better to suppress this by default:

Image

### Release notes
Minor fix of suppressing Ziping output in the end of the population module.

jan-petr:
### Linked issue

close #1886

### DIFF
--- a/Modules/Module_Processing/Module_Population/xASL_adm_GzipAllFiles.m
+++ b/Modules/Module_Processing/Module_Population/xASL_adm_GzipAllFiles.m
@@ -51,6 +51,8 @@ function xASL_adm_GzipAllFiles(ROOT, bFolder, bUseLinux, pathExternal, bVerbose)
 
     exit_code = NaN;
     
+    fprintf('\n%s\n', 'G-zZzZipping .nii files, this can take a while...');
+
     %% ----------------------------------------------------
     %% 1) Faster unix version, using OS file system
     if bUseLinux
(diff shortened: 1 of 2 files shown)


Modules/Module_Processing/Module_Population/xASL_adm_GzipAllFiles.m
```diff
@@ -54,7 +57,11 @@ function xASL_adm_GzipAllFiles(ROOT, bFolder, bUseLinux, pathExternal)
 		if exist(ROOT,'dir') % Check if directory exists, otherwise a random current directory would be zipped
 			PathToSearch = xASL_adm_UnixPath(ROOT);
             oldPath = pwd;
-			[exit_code, system_result] = system(['cd ' PathToSearch '; for i in `find * | grep -E \.nii$`; do gzip -1 -f -q -v "$i"; done'], '-echo');
+			if bVerbose
```
HenkMutsaerts, line 62: Is removing the -v option sufficient, as we still have the -echo option? (I mean did you test it)?

HenkMutsaerts (approved):
Single question, that somebody should test it. But otherwise fine. I added an extra print to warn the user that it can take a while (my reason to add -echo was that the user sees what's happening instead of looking 30 min at a busy ExploreASL without knowing if it froze)

PR #1889, opened 2025-01-23
Closes #1888 fsl path spaces

### LINKED ISSUE #1888: When a folder has a space (EduCloud), FSLCommand crashes
### Description

The problem is that not all linux commands work well with spaces escaped with a backslash \.

E.g., this occurs when the path has any space in a directory (filenames are usually not created with spaces), e.g. /Users/Beatriz/My Project/derivatives/ExploreASL/ etc )

For an FSL command, the solution is to use quotes "" without escaping spaces for the FSL command.
For the FSL options file, none of the normal solutions work. What does work, is ensuring the system() command starts in the ASL folder, and have the files without the folder paths added as options in the option file.

### Tasks
- [x] xASL_adm_UnixPath: add option not to escape spaces but use quotes ""
- [x] xASL_ext_FSLRun: always cd to the session dir
- [x] xASL_quant_External: redefine paths for/in FSL option file

@jan-petr : how does this work with Vaby, should we also go to the session dir (change in xASL_ext_FSLRun) and have filenames rather than folders in the option file? If Vaby doesn't use an option file, we can use the extra input option in xASL_adm_UnixPath

### Beatriz' original question

There two issues:
First, the FSLModelOptions.txt created in line 120 and used in line 123, as shown below, includes the paths of the data and doesn't deal with spaces:

Image

Image

Henk-Jan had the idea to add " " in the paths, as shown here:

Image

But then this was still not enough:

Image

as this FSLCommand also includes a path with this 'My Project' folder with a space.

*Of course* this could be solved by removing that space, and I've tried, but this folder is a default folder of the server with a specific format (e.g. 'Home' folder), but if I do that, the whole server doesn't respond / work.

### Release notes
Bugfix in FSL call

HenkMutsaerts:
### Linked issue

Closes #1888

### DIFF
--- a/External/SPMmodified/xASL/xASL_adm_UnixPath.m
+++ b/External/SPMmodified/xASL/xASL_adm_UnixPath.m
@@ -1,12 +1,15 @@
-function [PathIs] = xASL_adm_UnixPath(PathIs, bTryWSL)
+function [PathIs] = xASL_adm_UnixPath(PathIs, bTryWSL, bUseQuotes)
 %xASL_adm_UnixPath Convert paths to Unix-compatible paths
 %
-% FORMAT: [PathIs] = xASL_adm_UnixPath(PathIs[, bTryWSL])
+% FORMAT: [PathIs] = xASL_adm_UnixPath(PathIs[, bTryWSL, bUseQuotes])
 %
 % INPUT:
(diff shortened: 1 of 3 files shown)


jan-petr (approved):
OK

PR #1891, opened 2025-01-23
#1890 Fix AAL atlas

### LINKED ISSUE #1890: Fix AAL atlas
### Description

The AAL atlas is incorrectly saved an aligned.

### Tasks

- [x] The main issue is that the original atlas AAL3v1 isn't aligned to any of the MNI2009 or ICBM152 atlases, but it is rather created on colin27 atlas. So we need to segment AAL3v1 using CAT12 to obtain the colin27->IXI512 alignment. And then this must be applied to AAL3v1.
- [x] Furthermore, AAL3v is split to left and right regions. We need to merge this and the splitting to hemispheres is done in ExploreASL
- [x] Fix the script
- [x] Perform the transformation and save the new atlas

### Release notes

Nothing. Report together with #1855

jan-petr:
### Linked issue

Close #1890


WorkInProgress/Administration/xASL_adm_CoregisterAtlas.m
```diff
@@ -1,4 +1,4 @@
-% Copyright 2015-2024 ExploreASL (Works In Progress code)
+% Copyright 2015-2025 ExploreASL (Works In Progress code)
```
HenkMutsaerts, line 1: Please change the name to xASL_adm_RegisterAtlas or something, coregistration has the same meaning, and if it differs, coregistration is used for linear/intra-subject registration.

WorkInProgress/Administration/xASL_adm_CoregisterAtlas.m
```diff
@@ -10,9 +10,13 @@
 pathTPMAtlasFile = 'mni_icbm152_gm_tal_nlin_asym_09c.nii';
 pathTPMAtlasFileWM = 'mni_icbm152_wm_tal_nlin_asym_09c.nii';
 pathTPMAtlasFileGMWM = 'mni_icbm152_gmwm_tal_nlin_asym_09c.nii';
+pathColinDir = 'mni_colin27_2008';
```
HenkMutsaerts, line 13: We don't save these in ExploreASL right? (at least we shouldn't if we don't use them in the pipeline), should we save them in CustomScripts for ease of use?

WorkInProgress/Administration/xASL_adm_CoregisterAtlas.m
```diff
@@ -122,12 +191,38 @@
...
 
 % Create MAT file
-IM = xASL_io_Nifti2Im(fullfile(pathTPM, 'oAAL3v1_1mm.nii'));
-IM = uint8(IM);
-save(fullfile(pathTPM, 'oAAL3v1_1mm.nii.mat'),'IM');
-xASL_adm_GzipNifti(fullfile(pathTPM, 'oAAL3v1_1mm.nii'));
-
 IM = xASL_io_Nifti2Im(fullfile(pathTPM, 'oWMPM_Type_III.nii'));
-IM = uint8(IM);
+IM = uint8(ceil(IM/2.0)); % Make sure it's divided by 2 and rounded to deal with the L-R and merge them to bilateral
```
HenkMutsaerts: This doesn't work for each atlas, most have left and right but also single bilateral regions; so this needs checking per ROI atlas!

HenkMutsaerts (approved):
Few points for you to check but I don't need to see them :)

WorkInProgress/Administration/xASL_adm_CoregisterAtlas.m
jan-petr: Yes. Agree. Will do...

WorkInProgress/Administration/xASL_adm_CoregisterAtlas.m
jan-petr: Haha. No of course not - these are all on my disc in a temp dir. I only save the result to ExploreASL dir.
But it's a good idea to add also the input files to CustomScripts so that this can be re-run easier.

WorkInProgress/Administration/xASL_adm_CoregisterAtlas.m
jan-petr: Yes. Fully agree!!! But for these two cases, it works right?

jan-petr: Fix with the correct script.

PR #1892, opened 2025-01-26
Closes #1476 report generator swap

### LINKED ISSUE #1476: Swap out default Report Generator to  xASL_qc_GenerateReport
### Description

#1225 implemented a new PDFGenerator, #1552 updated the PDF generator to the recent multi-session x.Output format.

### Tasks
- [x] adapting configuration to existing PDF (was already done by Maarten in the config generator
- [x] Change ` xASL_qc_GenerateReport` to run module-wise
- [x] Change ` xASL_qc_GenerateReport` to run session-wise
- [x] Use this new report generator in xASL_module_Structural & xASL_module_ASL
- [x] Move `xASL_qc_CreatePDF.m` to the legacy ExploreASL folder in CustomScripts repository
- [x] qc_glossary.tsv: improve aliases, including units behind the key-name instead of behind the values, where the ranges should be (seems a more economical use of the page space)
- [x] xASL_qc_GeneratePDFConfig: sort key-names
- [x] Do not store the PDF config file, we recreate this per module, per subject. We temporarily store this in the subject-folder
- [x] xASL_qc_ParsePdfConfig: Instead of the footer "created by ExploreASL", which is clear from the ExploreASL logo, I added date and time
- [x] xASL_qc_ParsePdfConfig: cosmetic revamp: added subfunction lists and explanations, removed subfunctions that fit within another subfunction because they were very short, moved settings subfunctions to text and image subfunctions, and text subfunctions to image subfunctions, if they were used there only

#### miscellaneous **
- [x] xASL_adm_GzipAllFiles: add feedback to unix zipping. With unix zipping, there was no feedback to the user that it takes quite long, and there was no progress shown on the files that were zipped. This is fixed now. Very small change that doesn't need a separate issue imho.

#### Add reference ranges (old #1440 now integrated here)
- [x] xASL_qc_CreateQCreference: created new reference range values from EPAD, created a new WIP function for this that crawls reference values from a dataset (EPAD VUmc site in this case, which is Philips 2D EPI with Bsup and a long PLD), and stores them in separate .mat files with similar structure as x.mat, so it can be matched in qc_glossary.tsv key-by-key and added as ranges
- [x] xASL_qcGeneratePdfConfig: some bugfixes for module parsing
- [x] xASL_qcGeneratePdfConfig: manually fixed PDF text & image positions for better layout & having the text nicely spread out
- [x] xASL_qcGeneratePdfConfig: print subject in the title
- [x] xASL_qc_ParsePdfConfig: don't print subject as QC key-value
- [x] xASL_qc_ParsePdfConfig: fix positions of logo & date/time
- [x] xASL_qc_ParsePdfConfig: -fillpage gives a better layout than -bestfit
- [x] xASL_qc_ParsePdfConfig_sub_Generate_QC_String: full revamp

#### Add QC images to fieldName where fieldName is fileName of the printed NIfTI, to allow printing a figure legend
- [x] x.Output_im.(module).(session).[image matrix] -> x.Output_im.(module).(session).(imageName).[image matrix]
- [x] qc_glossary: Cleanup
- [x] xASL_wrp_VisualQC_Structural: store QC image filenames
- [x] xASL_wrp_VisualQC_ASL: store QC image filenames & fix TrackProgress
- [x] xASL_wrp_VisualQC_ASL: fix TrackProgress
- [x] xASL_vis_AddIM2QC: add QC image filenames as a separate field containing the image cell
- [x] xASL_qc_CollectParameters: store software versions also for sessions
- [x] xASL_qc_CollectParameters: rename ScanType->Module
- [x] xASL_qc_ParsePdfConfig: abbreviate Git commit to 7 length-format in PDF
- [x] xASL_qc_ParsePdfConfig: use QC image filename field
- [x] xASL_qc_ParsePdfConfig: add caption (header) below QC images

#### Manual for ourselves:
1. inside modules, output is printed to x.Output & x.Output_im
2. xASL_qc_GenerateReport -> xASL_qc_LoadPdfConfig (move as subfunction to former)
    configReportPDF.json is created by user (or automatically in the case of ExploreASL)
  containing all pages/canvases/images/text etc to be printed to the PDF
3. xASL_qc_ParsePdfConfig then goes over the config, and creates the PDF based on the content 
 	of this config, using the content of x.Output & x.Output_im (in the case of ExploreASL)
 	using the nomenclature, units, and ranges stored in QC_glossary.tsv
4. QC_glossary.tsv ranges are created from EPAD using xASL_qc_CreateQCreference
 
#### ToDo
- [ ] HJ: Compute ranges with NOVICE & Sleep as well, is more robust than only with EPAD
- [ ] @jan-petr Create list of meaning of parameters (already in the documentation online?)

#### PM
- [ ] Add CBF ranges for MindBoggle atlas ROIs (if mediri will still implement this?)
- [ ] Add sagittal QC images
- [ ] PM: Merge xASL_adm_formatWithRoundedMagnitude with xASL_num2str
- [ ] PM: Figure alignment is slightly off -> perhaps Jan can fix this? may have to do with the PDF resampling?

### How to test
- [x] Continuously tested by Henk with /External/TestDataSet
- [x] PDF output checked by Henk with TestDataSets
- [ ] Are parameters missing?
- [ ] PM: which parameters shall we not print? Should we separate into QC & output measurements?
- [ ] PM: Should we recalculate the ranges with new releases, if quantification changes significantly? (perhaps not really)
- [x] JAN: Run UnitTests
- [x] JAN: Run 10 TestDataSets

### Release notes
Implementation of a new optimized PDF generator

HenkMutsaerts:
### Linked issue

Closes #1476


Functions/Administration/xASL_adm_formatWithRoundedMagnitude.m
```diff
@@ -0,0 +1,61 @@
+function [stringValue] = xASL_adm_formatWithRoundedMagnitude(value, sigFigs)
```
jan-petr, line 1: Do you really need this? 
sprintf('%.3g') is doing a decent job for both <0 and >0. 1234000 would be printed as 1.23e6, which is not as nice, but printing 1234000 as 1230000 is opening a very large room for confusion.

Functions/QualityControl/xASL_qc_CollectParameters.m
```diff
@@ -1,4 +1,4 @@
-function x = xASL_qc_CollectParameters(x, iSubject, ScanType, iSession)
+function x = xASL_qc_CollectParameters(x, iSubject, Module, iSession)
 %xASL_qc_CollectParameters Collect all parameters from structural & ASL, spread over the derivative folders
 %
 % FORMAT: x = xASL_qc_CollectParameters(x, iSubject, ScanType, CollectQCFunction [, iSession])
```
jan-petr: Please update header on line 4 and 9 - parameter "Module"

Functions/QualityControl/xASL_qc_CollectParameters.m
```diff
@@ -96,21 +100,33 @@
...
 end
 
-% General software versions (put this in structural only, to avoid redundant output
-x.Output.(ScanType).Version_ExploreASL = x.Output.SoftwareVersion.ExploreASL; % ExploreASL version
-x.Output.(ScanType).Version_ExploreASL_git = x.Output.SoftwareVersion.ExploreASL_git; % ExploreASL git commit tag
-x.Output.(ScanType).Version_Matlab = x.Output.SoftwareVersion.Matlab; % Matlab
-x.Output.(ScanType).Version_SPM12 = x.Output.SoftwareVersion.SPM12; % SPM
-x.Output.(ScanType).(x.SESSIONS{iSession}).Version_FSL = x.Output.SoftwareVersion.FSL; % FSL
+% General software versions
+if isfield(x, 'SESSION')
```
jan-petr, line 114: It might happen that the module is structural and SESSIOn field exists. Do we then really want to write .(x.SESSION)? Or will this never happen? 

I'm just speculating, I'm not completely sure.

jan-petr (changes requested):
Minor things...

Functions/Administration/xASL_adm_formatWithRoundedMagnitude.m
HenkMutsaerts: Well a doctor doesn't like reading 1.23e6, why is 1230000 confusing? The rounding is the same. 

I see the point though that this is a big function for a small improvement, but it is worth it for the clinical PDF. I used chatGPT to create it :)

Functions/QualityControl/xASL_qc_CollectParameters.m
HenkMutsaerts: When the structural module is run, I don't think x.SESSION exists, from xASL_Iteration. I tested this and it comes out as x.Output.Structural.Version_ExploreASL.

Functions/QualityControl/xASL_qc_CollectParameters.m
HenkMutsaerts: Done in 5f5073dee8cb13a3937862e1d63741aa3a8b8964

jan-petr (approved):
OK

PR #1897, opened 2025-02-12
#1825 NII import without JSON

### LINKED ISSUE #1825: Nifti import doesn't work without JSON
### Description

If JSON is missing when trying to import NII, then it crashes.

### Tasks

- [x] Allow missing JSONs for anat files
- [x] Allow missing JSONs for perf files
- [x] Handle missing JSON parameters for ASL
- [x] Test and update flavor Siemens_VE11_NiftiOnly_1

### Release notes

Allow NIfTI import without JSON.

jan-petr:
### Linked issue

Close #1825

### DIFF
--- a/Functions/Administration/xASL_adm_OrderFields.m
+++ b/Functions/Administration/xASL_adm_OrderFields.m
@@ -44,7 +44,7 @@ function outStruct = xASL_adm_OrderFields(inStruct,orderStruct)
 
 orderArray = fieldnames(orderStruct);
 inArray = fieldnames(inStruct);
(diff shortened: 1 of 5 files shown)


HenkMutsaerts (approved):
If this really only changes the behavior for having NIfTI's without JSON, this is great.
Though if we can do without generating an empty JSON in raw, that would even be better, simply because it requires an extra validation step to the user (instead of seeing the missing JSON, the user gets a warning from BIDS validator, from ExploreASL, and still has to open the JSON)

PR #1903, opened 2025-02-18
Closes #1742 xASL_qc_CollectQC_Structural: add warnings for incomplete CAT12…

### LINKED ISSUE #1742: Structural QC warnings
### Description

When re-running structural module of an already processed data, it fails in the structural QC as 

ERROR: Structural module terminated for subject 1: sub-Vas001_1

ans =

    'Unrecognized field name "S".
     
     Error in xASL_qc_CollectQC_Structural (line 83)
             Struct.T1w_IQR_Perc = min(100,max(0,105 - curr.S.qualityratings.IQR*10));

JAN: That's quite weird. Because it means that the CAT12 QC file cat_T1_<SUBJECT>.mat was present, but somehow that file was corrupted. Not really something we can do anything about. We already catch if this file is missing, we could also catch if this file is corrupt (i.e. doesn't contain this QC field exported by CAT12), but something went wrong with the rerunning. I would recommend the user to simply rerun everything.

HENK: OK I fixed it according to your warning suggestion.

### Tasks

- [x] xASL_qc_CollectQC_Structural: Add warnings for incomplete CAT12 QC file

### Release notes

Improve warnings in Structural QC collection.

HenkMutsaerts:
### Linked issue

Closes #1742

### DIFF
--- a/Functions/QualityControl/xASL_qc_CollectQC_Structural.m
+++ b/Functions/QualityControl/xASL_qc_CollectQC_Structural.m
@@ -59,7 +59,7 @@ function [x] = xASL_qc_CollectQC_Structural(x, iSubject)
         bProcess = true;        
     elseif isempty(PathLST)
         bProcess = false;
-        if xASL_exist(x.P.Path_FLAIR,'file')
+        if xASL_exist(x.P.Path_FLAIR, 'file')
             warning('Didnt find any LST volumetric results');
             Struct.FLAIR_WMH_vol_mL = NaN;
             Struct.FLAIR_WMH_n = NaN;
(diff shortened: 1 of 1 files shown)


jan-petr (changes requested):
I fixed a minor bug. Otherwise it looks good.

It is very weird that you even had this error. I can understand a missing file, but a corrupt file is very weird. Can you post the example so that I can investigate?! Was it corrupt on the first run or only on the re-run and the first run was OK?

If you can't reproduce this, we can merge this fix, but we might potentially learn more about the reasons for this error.

jan-petr (approved):
OK

PR #1920, opened 2025-03-26
Bug #1918 minor bugs

### LINKED ISSUE #1918: Testing-discovered bugs
### Description

Full testing of the develop branch uncovered some errors that crashed the pipeline.

### Tasks

- [x] Unit testing of JSON check has errors. Fixed in 64e426d6
- [x] Error in Philips_2DEPI_FLAIR_Segm_PWI4D_Quant_Parameters_LoQ SliceReadoutTime missing. Fixed in 6ce7c661
- [x] Error in Philips_2DEPI_Lesion_NoFLAIR_NoM0_HiQ CurrentMaskVascular in population module. Fixed in 4e0a88a4

### Release notes

Minor errors.

jan-petr:
### Linked issue

Close #1918

### DIFF
--- a/Modules/xASL_module_ASL.m
+++ b/Modules/xASL_module_ASL.m
@@ -606,6 +606,14 @@ function [x] = xASL_module_ASL_ParseParameters(x, bOutput)
 % First ensure that we don't use quantification parameters from other sequences
 x.Q = struct;
 
+% Load x.Q information from dataPar.json if available
+if isfield(x, 'dir') && isfield(x.dir, 'dataPar') && exist(x.dir.dataPar, 'file')
+    xDataPar = xASL_io_ReadJson(x.dir.dataPar);
+	if isfield(xDataPar, 'x') && isfield(xDataPar.x, 'Q')
+		x.Q = xDataPar.x.Q;
+	end
+end
+    
 [~, x] = xASL_adm_LoadParms(x.P.Path_ASL4D_parms_mat, x, bOutput);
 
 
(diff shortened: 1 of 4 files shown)


Modules/xASL_module_ASL.m
```diff
@@ -606,6 +606,14 @@
 % First ensure that we don't use quantification parameters from other sequences
 x.Q = struct;
 
+% Load x.Q information from dataPar.json if available
```
HenkMutsaerts: So you say that we need this because of specific quantification settings; which are not part of BIDS rawdata. That is fine, but there is probably a reason that we cleaned x.Q two lines back; not to reuse the settings from other sequences.

So we either remove both 607 and this new part (and always use the x.Q from dataPar, loaded when loading dataPar at the start), or we need some way to make this session/subject specific (if we even need this)?

HenkMutsaerts (changes requested):
Minor comment in xASL_module_ASL

Modules/xASL_module_ASL.m
jan-petr: This is fully up to you. We need to load it somehow.

At the default location, it is really loaded only once at the start. So it can indeed accumulate and retain information. That's why you previously decided to clean it here. 

So I'm fine cleaning it at every subject and reloading x.Q here afterwards (i.e. the current solution).
I'm also fine not cleaning it and loading it only once (i.e. removing 607 and this new part). 

I don't think there's a good subject specific way without writing fully subject specific dataPar.json like we have for studyPar - which we can do in the future. So I wouldn't do this now as that would require considerable time.

HenkMutsaerts (approved):

PR #1932, opened 2025-04-03
#1931 vis_AddIM2QC fix bug

### LINKED ISSUE #1931: Fix major bug in xASL_vis_AddIm2QC
### Description

Fix major bug

### Tasks

- [x] Indexing is no longer needed as the index was by definition always 1 (probably stayed there historically when you modified the code).
- [x] It crashes when older processed data are loaded and you want to add more images in the x-struct

### Release notes

Bug fix, report with #1476

jan-petr:
### Linked issue

Close #1931

### DIFF
--- a/Functions/Visualization/xASL_vis_AddIM2QC.m
+++ b/Functions/Visualization/xASL_vis_AddIM2QC.m
@@ -60,7 +60,6 @@ function [x] = xASL_vis_AddIM2QC(x, parms)
 
 
     %% Create the field
-    IndexIm = 1;
     if ~isfield(x, 'Output_im') || isempty(x.Output_im)
         x.Output_im = struct;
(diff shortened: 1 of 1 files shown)


HenkMutsaerts (changes requested):
I'm hesitant with this, because I thought we were about to release 2.0.0 which has no backward compatibility anymore. So fine to do this now, for the workshop, but for 2.0.0 we should remove all this backward compatibility because it makes the code increasingly complex. We changed so much, and I would anyway recommend reprocessing if somebody has partly processed data from a while back.

Can you put the old code in comments, add a comment that we will put this back when releasing 2.0.0?

HenkMutsaerts (approved):
yes this is super!

PR #1952, opened 2025-06-03
#1930 Smooth before FSL

### LINKED ISSUE #1930: Smoothing before FSL quantification
### Description

Add an option to smooth before the quantification using External algorithm.

HENK: very good idea, this could really improve the reliability of the fit I believe. But to do this correctly, we would have to smooth taking GM-WM and ROIs into account. When I would create a map with per voxel the amount that we want to smooth (e.g. 16 FWHM in WM, 8 FWHM in GM, and perhaps also ROI or vascular territory-separated in a later more complicated stage), could you then let the smoothing kernel vary spatially based on such a map?

Or perhaps easier to smooth once with 16 FWHM, multiply this with pvWM, and once with 8 WHM, multiply this with pvGM, and combine these two linearly?

JAN: We can of course do all that, but I'd add that in a later stage and now keep this simple option only. 

### Tasks

- [x] HENK: we may need to tweak the processingmask first #1871 JAN: Agree, and important to do it soon, but can be done also after merging this commit.
- [x] Add an option to smooth before quantification
- [x] Apply this before calling FSL
- [x] Correct parameter name - so far it is ExternalSmoothGaussianMM. @HenkMutsaerts are you OK with this name, or you want a better one.
- [x] Explain this in the documentation

### Release notes

An option to presmooth PWI maps before external quantification.

jan-petr:
### Linked issue

Close #1930


HenkMutsaerts (approved):
Nice, I would only change the name perhaps a bit, to mention that it is about quantification (e.g., TopUp is also an external process)

x.modules.asl.QuantExtSmoothGaussianMM

PR #1956, opened 2025-06-12
#1954 Fix cat_main lesion exclusion

### LINKED ISSUE #1954: Fix Lesions in CAT
### Description

In CAT segmentation, we have the option to load and exclude lesions from the segmentation.
But if a lesion was loaded empty (e.g. incorrect T1 file), it crashes

### Tasks

- [x] Several external and internal lesions are assumed that are summed for exclusion. We have created a variable YlesionFull. However, if one lesion was empty, this variable was failed to be initialized.
- [x] Lesion2Mask function crashed when one of the sublesions was empty. So I added an extra check. Also note that the line `OutImFinal(OutIm{1} == OutIm{2}) = OutImFinal(OutIm{1} == OutIm{2})` because it truly doesn't do anything.

### How to test

Tested in GLIOCARE on a subject that failed previously.

### Release notes

Fix Lesion exclusion from CAT12 segmentation.

jan-petr:
### Linked issue

Close #1954

### DIFF
--- a/External/SPMmodified/toolbox/cat12/cat_main.m
+++ b/External/SPMmodified/toolbox/cat12/cat_main.m
@@ -222,14 +222,14 @@ if ~isfield(res,'spmpp')
     end
     if job.extopts.new_release && 0 
       % improvements for large ventricles ... not working now (RD201911)
-      if isfield(res,'Ylesion') && sum(res.Ylesion(:)>0)
+      if isfield(res,'YlesionFull') && sum(res.YlesionFull(:)>0)
         % [trans,res.ppe.reginitp] = cat_main_registration2(job2,res2,Ycls(1:2),Yy,tpm.M,res.Ylesion); 
         [trans,res.ppe.reginitp] = cat_main_registration2(job2,res2,Ycls(1:2),Yy,tpm.M,res.YlesionFull); 
       else
         [trans,res.ppe.reginitp] = cat_main_registration2(job2,res2,Ycls(1:2),Yy,tpm.M); 
       end
     else
-      if isfield(res,'Ylesion') && sum(res.Ylesion(:)>0)
+      if isfield(res,'YlesionFull') && sum(res.YlesionFull(:)>0)
         % [trans,res.ppe.reginitp] = cat_main_registration(job2,res2,Ycls(1:2),Yy,tpm.M,res.Ylesion); 
         [trans,res.ppe.reginitp] = cat_main_registration(job2,res2,Ycls(1:2),Yy,tpm.M,res.YlesionFull);
       else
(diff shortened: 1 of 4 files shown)


External/SPMmodified/toolbox/cat12/cat_main.m
```diff
@@ -222,14 +222,14 @@
     end
     if job.extopts.new_release && 0 
       % improvements for large ventricles ... not working now (RD201911)
-      if isfield(res,'Ylesion') && sum(res.Ylesion(:)>0)
+      if isfield(res,'YlesionFull') && sum(res.YlesionFull(:)>0)
```
HenkMutsaerts, line 225: Should we update /ExploreASL/External/README_SPMmodified.txt accordingly for the two CAT12 changes?

External/SPMmodified/toolbox/cat12/cat_main_updateSPM.m
```diff
@@ -84,12 +84,14 @@
   %% Some error handling
   %    ds('l2','',vx_vol,Ysrc./WMth,Yp0>0.3,Ysrc./WMth,Yp0,80)
   Yp0  = single(P(:,:,:,3))/255/3 + single(P(:,:,:,1))/255*2/3 + single(P(:,:,:,2))/255;
-  if isfield(res,'Ylesion') && sum(res.Ylesion(:)>0)
-    res.YlesionFull = res.Ylesion; % ExploreASL fix
-    res.Ylesion = cat_vol_ctype( single(res.Ylesion) .* (Yp0>0.2) ); 
-    for k=1:size(P,4), Yl = P(:,:,:,k); Yl(res.Ylesion>0.5) = 0; P(:,:,:,k) = Yl; end  
-    Yl = P(:,:,:,3); Yl(res.Ylesion>0.5) = 255; P(:,:,:,3) = Yl; clear Yl; 
-    Yp0  = single(P(:,:,:,3))/255/3 + single(P(:,:,:,1))/255*2/3 + single(P(:,:,:,2))/255;
+  if isfield(res,'Ylesion')
```
HenkMutsaerts, line 87: I wonder if the behavior differs between Ylesion and YlesionFull, so why not just keeping one lesion parameter?

You probably don't want the changes here applied to YlesionFull, then perhaps better naming would be YlesionMasking.

Otherwise, if you replace all Ylesion by YlesionFull, this is tricky. Probably this code here is there for the automatically detected lesions (e.g., WMH) by CAT12, and you don't want to change this behavior in the rest of CAT12 (which we now do as you replaced Ylesion by YlesionFull in the rest).

Shouldn't we instead detect user-specific lesions, for user-specific lesions we skip this part, otherwise we run this part but we keep a single Ylesion parameter throughout?

Functions/ImageProcessing/xASL_im_Lesion2Mask.m
```diff
@@ -198,17 +198,18 @@
     OutIm{iIm} = xASL_vis_CreateVisualFig(x, {x.P.Pop_Path_rT1 VisualizeImage==iIm}, [], [0.75 0.35], [], {x.S.gray x.S.colors_ROI{iIm}});
 end
 
-% Initialize final image
-OutImFinal = zeros(size(OutIm{1}));
-OutImFinal(OutIm{1}==OutIm{2}) = OutImFinal(OutIm{1}==OutIm{2});
+% Initialize final image - make sure it has the proper dimensions when an individual image is returned empty
+OutImFinal = zeros(max(cellfun(@(x) size(x,1),OutIm)), max(cellfun(@(x) size(x,2),OutIm)), max(cellfun(@(x) size(x,3),OutIm)));
```
HenkMutsaerts: I would replace cellfun(@(x) by cellfun(@(y) as we use x as a Table throughout ExploreASL. It might work this way but never hurts to be careful. Also, I use y throughout, so we are more consistent.

Functions/ImageProcessing/xASL_im_Lesion2Mask.m
```diff
@@ -198,17 +198,18 @@
...
 
 for iIm=1:length(OutIm)
-    % Create masks with non-grey voxels (i.e. the overlay)
-    OutImMask{iIm} = OutIm{iIm}(:,:,1)~=OutIm{iIm}(:,:,2) | OutIm{iIm}(:,:,2)~=OutIm{iIm}(:,:,3)  | OutIm{iIm}(:,:,1)~=OutIm{iIm}(:,:,3);
-    OutImMask{iIm} = repmat(OutImMask{iIm}, [1 1 3]);
-    
-    % Add the overlay from each mask
-    OutImFinal(OutImMask{iIm}) = OutIm{iIm}(OutImMask{iIm});
+	if isequal(size(OutImFinal), size(OutIm{iIm}))
```
HenkMutsaerts, line 205: What do we do if they don't have equal size? Now they are skipped, at least we should provide a warning to the user?

Are you sure this is correct usage? The user needs to provide Lesion_T1.nii (same space as the T1.nii). If this is not the case, we should create an error earlier in the pipeline! The user will otherwise just continue the code, even if we put a warning here, they won't read all the output.

Plus, if you create this error earlier, we don't need these changes?

HenkMutsaerts (changes requested):
Please first read my suggestions before changing, to save you time.

External/SPMmodified/toolbox/cat12/cat_main.m
jan-petr: Yes. Done

External/SPMmodified/toolbox/cat12/cat_main_updateSPM.m
jan-petr: The point of this edit was to make sure YLesionFull is initialized with YLesion even if YLesion was empty.

YLesionFull simply is a concatenation of our own manual Lesions and the automatic YLesion from cat12, so we don't want to disrupt what is done with YLesion, but we want to have our joint YLesionFull. So the code changes might not be 100% optimal, but they are minimally invasive.

Functions/ImageProcessing/xASL_im_Lesion2Mask.m
jan-petr: It definitively isn't a problem from the technical standpoint. But no prob to change to improve readability. Done

Functions/ImageProcessing/xASL_im_Lesion2Mask.m
jan-petr: Good point done.

External/SPMmodified/toolbox/cat12/cat_main_updateSPM.m
HenkMutsaerts: We should keep the original variable naming of CAT12 throughout, except if we add our own Ylesion (e.g. yLesionCostMasking). I thought you now renamed all CAT12 Ylesion's? If you only renamed our variable, I agree.

Just checking, as we want to update SPM & CAT12 at some point.

HenkMutsaerts (changes requested):
One conversation open, should be easy to fix

HenkMutsaerts (approved):

PR #1958, opened 2025-06-25
#1957 stat_GetROIstatistics Fix lesion masks

### LINKED ISSUE #1957: Population masks are turned of for Lesions and ROIs
### Description

Susceptibility masks are applied automatically for manually inputed Lesions and ROIs - which should not be the case. These masks are manually created and the user takes the responsibility for those.

### Tasks

- [x] Make sure that by default extra masks are turned of for Lesions and ROIs
- [x] Make sure that Whole brain (WB=GM+CSF+WM) is used

### Release notes

Manual Lesion and ROI statistics are unrestricted with general population masks.

jan-petr:
### Linked issue

Close #1957

### DIFF
--- a/Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
+++ b/Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
@@ -141,7 +141,7 @@ end
 if ~isfield(x.S,'bMasking') || isempty(x.S.bMasking)
 	if x.S.bSubjectSpecificROI
 		% For Specific ROIs - Lesion and ROI provided on the input, we do not superimpose additional masks by default
-		fprintf('\n\n%s\n\n', 'Calculating statistics for subject-specific lesions or ROIs -> disabling all masks: susceptibility regions will not be masked, vascular artifacts may be included.');
+		warning('Calculating statistics for subject-specific lesions or ROIs -> disabling all masks: susceptibility regions will not be masked, vascular artifacts may be included.');
 		bMasking = [0 0 0 0];
 	else % The standard default applies all masks
 		bMasking = [1 1 1 1];
@@ -159,8 +159,8 @@ if ~x.S.IsASL
 end
 
 if x.S.InputNativeSpace
-	bMasking(1) = 0; % disable susceptibility masking always for native space analyzis
-else
+	bMasking(1) = 0; % disable susceptibility masking always for native space analysis
+end
 
 if ~isfield(x.S, 'bWMH')
     x.S.bWMH = false; % by default, WMH are excluded from all ROIs
@@ -748,7 +748,7 @@ for iSubject=1:x.dataset.nSubjects
 				end
             else
                 % Provide some feedback for debugging                
-                if xASL_stat_SumNan(DataIm(:)) == 0
+                if ~xASL_stat_SumNan(DataIm(:))
                     % Check for empty CBF map first
                     fprintf('%s\n', ['Warning: Empty image for ' x.SUBJECTS{iSubject} '_ASL_' xASL_num2str(iSess)]);
 				end
@@ -756,7 +756,7 @@ for iSubject=1:x.dataset.nSubjects
                 % pvPrimary is the main tissue type that is investigated
                 % pvSecondary is the other tissue type that is used with PVC
 
-				if bMasking(3)==0 % no tissue-masking
+				if ~bMasking(3) % no tissue-masking
                     pvPrimary = ones(size(DataIm));
                     pvSecondary = ones(size(DataIm));
 					bSkipPVC = 1;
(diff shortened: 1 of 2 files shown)


HenkMutsaerts (changes requested):
I'm fine with this, but would add warnings as this changes the analysis between the default — which still is the same for non-subject-specific ROIs — and subject-specific ROIs.

The susceptibility mask is there to remove the susceptibility regions only, let me know if it does more than that, then I should change it.
The vascular artifact mask is there to remove outliers, which we want by default for CBF but not for spatial CoV.
Etc.

We can of course reconsider this, but I would at least add a warning that the user now gets output that is biased and potentially less robust.

If you agree, please add the following lines in this if-statement:

fprintf('\n\n\n%s\n\n\n\n', 'Subject-specific lesions or ROIs detected, disabling all masks. Note that this may bias your analysis: susceptibility regions will not be masked, vascular artifacts may be included, all tissue classes will be included (GM, WM, CSF), and extracranial voxels may be included, depending on the subject-specific lesion mask/ROI.');

We could also add these warnings per disabled mask, to generalize better.

BTW: I disagree that with subject-specific ROIs you always want to use WB. E.g., when a user uses ROIs defined by fMRI (which I did with Carolina in Chile), the user wants only the GM inside the ROIs, and use a form of PVC).

jan-petr:
We have agreed that we will do bMasking defaulting differently for normal mask and subject-specific mask. But we will only change the defaulting. We will not overide the inserted bMasking value - if a value is provided, then this will be applied unchanged also to Lesions.

Therefore, the defaulting has to be done for every Atlas and not in general. Also - it makes sense that the defaulting is at a single location and not spread around several files. Because previously, we defaulted once very early in the code. And we couldn't change the defaults afterwards. Therefore, I had to remove general defaulting from Init-visualization and only did the localized defaulting that was needed for the local purpose. [7476e94](https://github.com/ExploreASL/ExploreASL/pull/1958/commits/7476e94e218a6032012450a56b35c5614c9b2d61)

I have then changed the defaulting method for normal atlas/Lesions

I will not add warning per disabled masks - as we already report all masking. So warning for disabled is not needed.

I have added a shortened version of your proposed warning. I do not mention the tissue types as these are done elsewhere. So the warning can be later added there, once we deal with tissue types for Lesions.

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
```diff
@@ -139,37 +139,52 @@
...
+	bMasking = x.S.bMasking;
 end
 
 if ~x.S.IsASL
-    x.S.bMasking(3) = 0; % disable tissue masking
+    bMasking(3) = 0; % disable tissue masking - this is done in all cases
 end
 
+if x.S.InputNativeSpace
+	bMasking(1) = 0; % disable susceptibility masking always for native space analyzis
```
HenkMutsaerts: analyzis -> analysis

1. Furthermore, are you sure you want to disable susceptibility masking for native space analysis? It doesn't hurt... And now you make the native and standard space analysis even more different?
2. I miss the "end" of this if-statement?
3. We do still want the same behavior for native space right, the WMH masking and printing of the masking?

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
```diff
@@ -139,37 +139,52 @@
 
 %% 0.a Manage masking
 if ~isfield(x.S,'bMasking') || isempty(x.S.bMasking)
-    x.S.bMasking = [1 1 1 1]; % default
+	if x.S.bSubjectSpecificROI
+		% For Specific ROIs - Lesion and ROI provided on the input, we do not superimpose additional masks by default
+		fprintf('\n\n%s\n\n', 'Calculating statistics for subject-specific lesions or ROIs -> disabling all masks: susceptibility regions will not be masked, vascular artifacts may be included.');
```
HenkMutsaerts: I would still make it a warning, otherwise users won't read this. The point is that by default the user can trust the masks, but now there is different behavior between the user-provided ROIs/lesions and the atlas-provided ROIs. Otherwise fine.

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
```diff
@@ -743,7 +756,7 @@
                 % pvPrimary is the main tissue type that is investigated
                 % pvSecondary is the other tissue type that is used with PVC
 
-				if x.S.bMasking(3)==0 % no tissue-masking
+				if bMasking(3)==0 % no tissue-masking
```
HenkMutsaerts: For consistency, we can change these statements throughout into:

if bMasking

and 

if ~bMasking

Modules/Module_Processing/xASL_init_VisualizationSettings.m
```diff
@@ -141,12 +133,22 @@
     end
 
     ImageWB = xASL_io_Nifti2Im(x.D.Atlas.WholeBrain);
-    if x.S.bMasking(4)==0
+
+    if ~isfield(x.S,'bMasking') || isempty(x.S.bMasking)
+        bMasking = [1 1 1 1];
```
HenkMutsaerts, line 139: Here they need defaulting to x.S as otherwise these parameters won't travel to the GetROIstats function. Alternatively, you remove it completely in this function (if we don't use x.S.bMasking elsewhere)?

It seems only necessary for the WBmask below, you can also just specify that one.

HenkMutsaerts (changes requested):
Minor changes

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
jan-petr: 0. Done
1. This was always done like that because we would have to transfer the susceptibility mask to native space. We would have to implement this.
2.Good catch. error during copy-paste. Fixed
3.Yes, otherwise this should be the same.

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
jan-petr: Fine with me. I just copy-pasted what you wrote :) Done.

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
jan-petr: Sure. Done

Modules/Module_Processing/xASL_init_VisualizationSettings.m
jan-petr: That's the point. That we only use this locally and we don't set the initialization for the later functions here. We only need it for WBmask. See the reasonaning for that in my comment way above.

I added a comment explaining that bMasking is onyl used locally.

HenkMutsaerts (approved):
Note that eventually everything should be exactly the same for both native & standard space; with just a subfunction that warps first in the case of native space.

PR #1969, opened 2025-10-08
Feature #1965 subcortical atlases

### LINKED ISSUE #1965: Subcortical atlases
### Description

Add the following atlases
https://freesurfer.net/fswiki/ThalamicNuclei 
https://surfer.nmr.mgh.harvard.edu/fswiki/BrainstemSubstructures 
https://www.nmr.mgh.harvard.edu/resources/aan-atlas

### Tasks

- [x] Download atlases
- [x] Combine them in a single atlas
- [x] Create a version without subdivision and with subdivision
- [x] Create labels
- [x] Transform atlases to IXI512
- [x] Add them to ExploreASL
- [x] Edlow Atlas (Harvard Ascending Arousal Network, AAN) into ExploreASL
- [x] Resample as NN
- [x] Add them to the ExploreASL documentation
- [x] Remove AAN as too small
- [x] inform Nico+Dominik
- [x] Prepared merged version of Thalamus+BrainStem
- [x] Test - prepare config files
- [x] New compilation+docker
- [x] Inform mediri

### Release notes

Add thalamus and brainstem atlas from Freesurfer.

### Release notes

Add thalamus and brainstem atlas from Freesurfer.

jan-petr:
### Linked issue

Close #1965


External/Atlases4ROIs/LicensePermissive/FreesurferThalamus.tsv
```diff
@@ -0,0 +1,26 @@
+LGN
```
HenkMutsaerts, line 1: Just looked at the atlas, I see you merged both left and right into bilateral.
However, it seems you have to redo this and resample nearest neighbor, there are some edges with erroneous regions from linear interpolation, it seems. And the regions are generally too small for ASL's resolution, so we might want to merge some regions.

WorkInProgress/Administration/xASL_adm_RegisterAtlas.m
```diff
@@ -5,11 +5,19 @@
 % __________________________________
 
 %% Please add here the path to the atlas files and to ExploreASL
+
```
HenkMutsaerts, line 8: I would move this function to our WorkInProgress folder, it is not a real function but a script depending on local paths. Useful script though, did you check that DARTEL works better than GS or oldnormalise for registering atlases?

Can't remember what I did previously exactly :)

BTW: are all regions mutually exclusive, then you can add them to a single volume

HenkMutsaerts (changes requested):
Few issues, comments are valid for both atlases.

You provided details under a license file, I have renamed this according to our existing atlases and added the license file; see last commit

WorkInProgress/Administration/xASL_adm_RegisterAtlas.m
jan-petr: It is already in WIP :)
Yes - it's IXI512, so we need DARTEL/GS instead of old-normalise. 
DARTEL likely does almost the same job as GS, so I'd say DARTEL is good.

In theory, we can merge the two atlases, they should be indeed mutually exclusive. But it has two different sources, so I'd prefer to keep them apart for this license/source reason. Technically, they can indeed be merged. What do you think?

External/Atlases4ROIs/LicensePermissive/FreesurferThalamus.tsv
jan-petr: Thanks for fixing the licenses.

Yes, they are small I know. That's really annoying, but they wanted those atlases, so I'd only merge them afterwards. Kind of taking regions and averaging across the results.

I will of course fix the edge. it should be indeed NN interpolation

HenkMutsaerts (changes requested):
Conversations closed, I'll approve once the ROIs are interpolated with nearest neighbor (I can approve without it, but then it is difficult to keep overview of what still needs to be done right?)

jan-petr:
Updated Freesurfer atlases. Added AAN atlas. Freesurfer should be fine, but please check the Atlas. And check AAN in detail.

HenkMutsaerts (approved):
Interesting that the Freesurfer atlas is (relatively) open-source, I thought that all Freesurfer atlases were non-commercial.

Registration and NN-interpolation look good, only the AAN brainstem ROIs are too small for ASL; not sure who wanted them but I would delete them; don't give this as default to users (they will sniff around in our folders)

PR #1998, opened 2026-01-15
Feature #1987 CSF analysis

### LINKED ISSUE #1987: CSF analysis
### Description

Currently, we cannot calculate CSF signal in any reasonable way. Unless we provide it as ROI and Lesion which is then tissue nonspecific. 
The easiest reasonable way is to add CSF tissue and then use some atlases that actually have CSF regions

#### Tissue segmentations are deleted at the end of the population module
HENK:
TissueMasking uses subject-specific templates; early versions of ExploreASL didn't save c3T1 (and the SPM backup may not yet still, so we could check this).  And check if it is removed from the MNI space/population folder, if David needs to repeat his analysis; we could add the deletion of the MNI segmentations to the `DeleteTemp` option, so that David can disable this.

SOLUTION: Not a problem, c3T1 is part of analysis and we only delete c3T1 and keep pvCSF.

#### Combining masks either by providing extra mask names (A) or as allowing combinations as input (B)
HENK: CSF + cortical ROIs -> this is tricky, it should then be TissueMasking ICV (or we allow a mix of TissueMasking, such as CSF+GM in this case, and CSF+WM+GM in the ICV case)
-> BTW, don't we already have WB (wholebrain) as an option (which is ICV here, BTW ICV = intracranial volume = not a good term)
JAN: WB(wholeBrain) is WM+GM. I would definitively want to have CSF. ICV was just a suggestion - I'm happy with any other definition - WM+GM+CSF or GM+CSF - just think of a nice name to use.
-> HENK: Then `WB` should be called `parenchyma` for clarity. My point was we should choose A) we only have different masks (e.g., `GM+WM+CSF` is one mask) or B) we allow the user to combine masks
HENK: Another easy solution, is that we allow for multiple TissueMasking & ROI combinations (we have this already as possibility in dataPar.json).
JAN: That's what I'm of course talking about. Allowing CSF in TissueMasking. But we cant use GM+CSF, unless we define an option called "GM+CSF" or "GMCSF" etc.
HENK: What I meant here, is that if David doesn't want both GM & CSF masks to become 1 mask, but just want one ROI to be within the CSF tissue and another ROI inside the GM tissue, then we already have this (TissueMask & ROIs can be a vector)

SOLUTION: combine masks into a single mask, then I propose the following 7 TissueMasking option names (strings):
* GM
* WM
* CSF
* GM+WM
* GM+CSF
* WM+CSF
* GM+WM+CSF

#### Region-growing in case of small regions/PVC
HENK: For the evaluation below: we need to be careful that we keep coding modular and similar to what we already have; I only foresee issues if we need region growing or shrinking that we have for the PVC, to ensure that a ROI has sufficient "other tissue types" (so WM in the case of GM, but in the case of CSF this would be WM+GM as both tissues are/can be juxtaposed) -> so that would require reorganizing such that the region-growing always include "all other juxtaposed tissuetypes"
JAN: I'm not proposing any changes there. Just adding CSF as tissue mask. Region growing is not needed as PVC is turned off in such situations. We can of course turn it on for CSF vs GM+WM. 
HENK: OK so currently we only keep region-growing for GM ROIs? Then we should add an if statement for this part (if we don't already have it)

JAN: do GM+WM rather than parenchyma. sometimes do GM+ROI and sometimes CSF+ROI or GM+CSF
3.⁠ ⁠⁠I fully agree to do GM+WM, GM+WM+CSF etc. That’s what I proposed in the start. Of course I was proposing a concept, the name ICV was not really fitting, just a quick label. In practice, I would keep it easy and would allow only predefined strings like “GM+WM” - so we can actually very easily program all options like “GM+WM”, “WM+GM”, “CSF+GM”, “CSF+WM” etc. I can do some proper parsing of substrings and pluses, but maybe it’s easier to just list the combinations - there are a few only. Anyway - parsing “+” or listing the options will be fully equivalent for the user, so only the parsing function will differ, the rest of the functionality will have the same code
HENK: Ok I agree, let’s do the “Tissue+Tissue” parsing, we remove the name “WB” then

### Tasks
- [x] Make sure pvCSF is created for native space as well - was OK
- [x] Make sure pvCSF is created and not deleted - done in 4204a0ea
- [x] Allow CSF and GM+WM strings in all input functions 06c686ca
- [x] Ensure these options are allowed everywhere 06c686ca
- [x] Currently, ROI expansion was extreme and risked biasing - it was adding 17 voxels or surrounding WM and doing that twice. So it potentially added over voxels with over 5cm radius with as much as 90% GM. That was too extreme - despite that we were checking of not adding voxels closer to GM, it was still very risky and I reduced the maximum distance to 1/3 of what it was. fc52d15f
- [x] Add option TissueMasking = CSF fc52d15f
- [x] Add option TissueMasking = GM+CSF+WM and combinations (GM, WM, CSF, GM+WM, GM+CSF, WM+CSF, GM+WM+CSF) fc52d15f
- [x] Make sure these options are order independent fc52d15f
- [x] Adjust the evaluation that it will really work (PVC for CSF is GM+WM, for WM is GM, for GM is WM etc) fc52d15f
- [x] Remove option WB -> Or rather WB will be made equivalent to GM+WM and a background compatibility warning will be issued fc52d15f
- [x] Do region growing only for GM and write a warning about this e3e79a0d
- [x] Add these options to the documentation
- [x] Write in manual that we can do CSF + cortical ROIs. Or CSF+ventricle ROI from Hammers or DKA. Or ICV+ventricle anatomical ROI.

### Release notes

Add various options to analyze CSF signal.

jan-petr:
### Linked issue

Close #1987

### DIFF
--- a/Functions/InputOutput/xASL_io_ReadDataPar.m
+++ b/Functions/InputOutput/xASL_io_ReadDataPar.m
@@ -73,8 +73,8 @@ function [x] = xASL_io_ReadDataPar(pathDataPar, bStudyPar)
 %
 %   ...
 %   "S":
-%	{"Atlases": ["Total","Total","DeepWM","Mindboggle_OASIS_DKT31_CMA"],
-%    "TissueMasking": ["GM","WM","WM","GM+WM"]}}
+%	{"Atlases": ["Total","Total","DeepWM","Mindboggle_OASIS_DKT31_CMA","Total","Total","Total","Total","Total"],
+%    "TissueMasking": ["GM","WM","WM","GM", "CSF", "GM+WM, "GM+CSF", "WM+CSF", "GM+WM"CSF"]}}
 %   ...
 %
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
--- a/Functions/InputOutput/xASL_io_ReadDataPar.m
+++ b/Functions/InputOutput/xASL_io_ReadDataPar.m
@@ -74,12 +74,12 @@ function [x] = xASL_io_ReadDataPar(pathDataPar, bStudyPar)
 %   ...
 %   "S":
 %	{"Atlases": ["Total","Total","DeepWM","Mindboggle_OASIS_DKT31_CMA"],
-%    "TissueMasking": ["GM","WM","WM","GM"]}}
+%    "TissueMasking": ["GM","WM","WM","GM+WM"]}}
 %   ...
 %
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
 % __________________________________
-% Copyright 2015-2024 ExploreASL
+% Copyright 2015-2026 ExploreASL
 % Licensed under Apache 2.0, see permissions and limitations at
 % https://github.com/ExploreASL/ExploreASL/blob/main/LICENSE
 % you may only use this file in compliance with the License.
--- a/Modules/Module_Processing/Module_ASL/xASL_wrp_PreparePV.m
+++ b/Modules/Module_Processing/Module_ASL/xASL_wrp_PreparePV.m
@@ -67,10 +67,8 @@ else
         % resolution, if the resolutions are identical
         xASL_Copy(x.P.Path_c1T1, x.P.Path_PVgm, true);
         xASL_Copy(x.P.Path_c2T1, x.P.Path_PVwm, true);
-		if xASL_exist(x.P.Path_c3T1)
-            xASL_Copy(x.P.Path_c3T1, x.P.Path_PVcsf, true);
-		end
-		
+		xASL_Copy(x.P.Path_c3T1, x.P.Path_PVcsf, true);
+				
 		if xASL_exist(x.P.Path_WMH_SEGM, 'file')
             xASL_Copy(x.P.Path_WMH_SEGM, x.P.Path_PVwmh, true);
 		end
@@ -78,11 +76,8 @@ else
         % also do this in standard space
         xASL_Copy(x.P.Pop_Path_rc1T1, x.P.Pop_Path_PV_pGM, true);
         xASL_Copy(x.P.Pop_Path_rc2T1, x.P.Pop_Path_PV_pWM, true);
+		xASL_Copy(x.P.Pop_Path_rc3T1, x.P.Pop_Path_PV_pCSF, true);
 		
-		if xASL_exist(x.P.Pop_Path_rc3T1)
-            xASL_Copy(x.P.Pop_Path_rc3T1, x.P.Pop_Path_PV_pCSF, true);
-		end
-
 		if xASL_exist(x.P.Path_rWMH_SEGM, 'file')
             xASL_Copy(x.P.Path_rWMH_SEGM, x.P.Pop_Path_PV_WMH_SEGM, true);
 		end
--- a/Modules/Module_Processing/Module_ASL/xASL_wrp_PreparePV.m
+++ b/Modules/Module_Processing/Module_ASL/xASL_wrp_PreparePV.m
@@ -19,9 +19,9 @@ function x = xASL_wrp_PreparePV(x, bStandardSpace)
 % If bStandardSpace:
 %
 %   1. Create dummy upsampled ASL scan, for registration
-%   2. Reslice pGM & pWM to hi-res ASL
+%   2. Reslice pGM & pWM & pCSF to hi-res ASL
 %   3. Estimate effective spatial resolution of ASL
-%   4. Smooth pGM & pWM to this spatial resolution
+%   4. Smooth pGM & pWM & pCSF to this spatial resolution
 %   5. Move smoothed tissue posteriors to MNI space
 %
 % else: run step 3 only, which will use the effective spatial resolution
@@ -34,7 +34,7 @@ function x = xASL_wrp_PreparePV(x, bStandardSpace)
 %
 % EXAMPLE: xASL_wrp_PreparePV(x);
 % __________________________________
-% Copyright (C) 2015-2024 ExploreASL
+% Copyright (C) 2015-2026 ExploreASL
 % Licensed under Apache 2.0, see permissions and limitations at
 % https://github.com/ExploreASL/ExploreASL/blob/main/LICENSE
 % you may only use this file in compliance with the License.
@@ -67,7 +67,7 @@ else
         % resolution, if the resolutions are identical
         xASL_Copy(x.P.Path_c1T1, x.P.Path_PVgm, true);
         xASL_Copy(x.P.Path_c2T1, x.P.Path_PVwm, true);
-        if xASL_exist(x.P.Path_c3T1)
+		if xASL_exist(x.P.Path_c3T1)
             xASL_Copy(x.P.Path_c3T1, x.P.Path_PVcsf, true);
 		end
 		
@@ -79,6 +79,10 @@ else
         xASL_Copy(x.P.Pop_Path_rc1T1, x.P.Pop_Path_PV_pGM, true);
         xASL_Copy(x.P.Pop_Path_rc2T1, x.P.Pop_Path_PV_pWM, true);
 		
+		if xASL_exist(x.P.Pop_Path_rc3T1)
+            xASL_Copy(x.P.Pop_Path_rc3T1, x.P.Pop_Path_PV_pCSF, true);
+		end
+
 		if xASL_exist(x.P.Path_rWMH_SEGM, 'file')
             xASL_Copy(x.P.Path_rWMH_SEGM, x.P.Pop_Path_PV_WMH_SEGM, true);
 		end
@@ -105,6 +109,11 @@ if bStandardSpace
 	% because we then go directly to standard space and only apply the T1->MNI and not ASL->T1 nonlinear transformation
 	xASL_spm_reslice([x.P.Path_rPWI ',1'], [x.P.Path_c1T1 ',1'], [],[], x.settings.Quality);
 	xASL_spm_reslice([x.P.Path_rPWI ',1'], [x.P.Path_c2T1 ',1'], [],[], x.settings.Quality);
+
+	if xASL_exist(x.P.Path_c3T1)
+		xASL_spm_reslice([x.P.Path_rPWI ',1'], [x.P.Path_c3T1 ',1'], [],[], x.settings.Quality);
+	end
+
 	if xASL_exist(x.P.Path_WMH_SEGM, 'file')
 		xASL_spm_reslice([x.P.Path_rPWI ',1'], [x.P.Path_WMH_SEGM ',1'], [],[], x.settings.Quality);
 	end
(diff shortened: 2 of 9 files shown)


Functions/InputOutput/xASL_io_ReadDataPar.m
```diff
@@ -74,12 +74,12 @@
 %   ...
 %   "S":
 %	{"Atlases": ["Total","Total","DeepWM","Mindboggle_OASIS_DKT31_CMA"],
-%    "TissueMasking": ["GM","WM","WM","GM"]}}
+%    "TissueMasking": ["GM","WM","WM","GM+WM"]}}
```
HenkMutsaerts: I would do as follows, throughout the code:

GM
WM
CSF
GM+WM
WM+CSF
GM+CSF
GM+WM+CSF

(just all possibilities, and clean)

Modules/Module_Processing/Module_ASL/xASL_wrp_PreparePV.m
```diff
@@ -67,7 +67,7 @@
         % resolution, if the resolutions are identical
         xASL_Copy(x.P.Path_c1T1, x.P.Path_PVgm, true);
         xASL_Copy(x.P.Path_c2T1, x.P.Path_PVwm, true);
-        if xASL_exist(x.P.Path_c3T1)
+		if xASL_exist(x.P.Path_c3T1)
```
HenkMutsaerts: With the standard ExploreASL run, c3T1 should always exist. 

There is the possibility that CAT12 fails, in which we fall back to SPM; there we have to tweak the SPM settings such that we always save c3T1 as well. Then we can remove the "xASL_exist" throughout in this code.

With very old ExploreASL runs, it could be that c3T1 didn't exist, but there will be more chances, as this was very very old (c3T1 has been there quite a while I think), so I would just change this without backward compatibility.

(Plus, note that xASL_Copy already checks if something exists, and doesn't crash if it doesn't, it will just give a warning if I'm not mistaken)

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
```diff
@@ -387,48 +383,63 @@
...
-					for iMask=1:size(x.S.InputMasks,3)
-						if sum(SumList(iROI,iMask))~=0 % skip empty ROIs
-							x.S.InputMasks(:,iROI,iMask) = xASL_im_CreatePVEcROI(x,x.S.InputMasks(:,iROI,iMask), pGM_MNI, pWM_MNI);
-                        else
-                            numVoxels = sum(x.S.InputMasks(:,iROI,iMask));
-                            fprintf('\n');
-                            warning('%s\n', ['Current ROI ' namesROIlocal{iROI} ' only contains ' xASL_num2str(numVoxels) ' voxels, so this ROI will be skipped']);
-                            fprintf('%s\n\n', 'Consider reducing the minimal ROI volume by lowering x.S.MinimalROIVolume to also evaluate this ROI');
+				
+				if strcmp(x.S.TissueMaskingLocal, 'GM')
```
HenkMutsaerts, line 403: So you changed it that we only do the region growing for GM-only ROIs right? Is this what we should have? The reason for the region-growing, is that with PVC you need a sufficient amount of voxels (also in case of slight misalignments). If you look only at the CSF surrounding the brain, and use PVC for this, you may also want to do region-growing? Only if WM is included, you may not want region-growing, as it is sufficiently large.
But if you have GM+WM, then you may also want some region-growing to include some CSF in the PVC.

Though I think the region-growing is now only programmed for GM, to grow inwards, including more WM. So perhaps what you did now is good as a temporary solution, but then in the near future we want region-growing also for other tissue-type (combinations)?

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
```diff
@@ -700,15 +725,42 @@
...
+				        pvSecondary = pWM;
+						pvPrimaryName = 'GM+CSF';
+						pvSecondaryName = 'WM';						
+					case {'WM+CSF', 'CSF+WM'}
+                        % we want WM+CSF, so GM is used for PVC
+                        pvPrimary = pWM+pCSF;
+				        pvSecondary = pGM;
+						pvPrimaryName = 'WM+CSF';
+						pvSecondaryName = 'GM';						
+                    case {'WB', 'GM+WM', 'WM+GM'}
```
HenkMutsaerts, line 750: Would remove WB here and below, we don't need backward compatibility with release 2.0.0 right?

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
```diff
@@ -700,15 +725,42 @@
...
+                        % we want GM+WM, using CSF as "WM" in PVC
+						if strcmp(x.S.TissueMaskingLocal, 'WB')
+							warning('TissuMasking==WB was used. This options was discontinued and was replaced by GM+WM. TissueMasking==GM+WM is used now, but please modify your input for the next time.')
+						end
 				        pvPrimary = pGM+pWM;
-				        pvSecondary = ones(size(pGM)) - pGM - pWM; % CSF
-						pvPrimaryName = 'WB';
+				        pvSecondary = pCSF; % CSF
```
HenkMutsaerts, line 756: We don't want to skip PVC for pGM+pWM ?

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
```diff
@@ -681,7 +706,7 @@
             %                        (loop outside this function for multiple tissue types)
             % 'GM' = gray matter
             % 'WM' = white matter
-            % 'WB' = whole brain (= GM+WM)            
+            % 'GM+WM' = whole brain GM+WM, previously was 'WB'
```
HenkMutsaerts: Like above, mention all here, and check that we have all below:

GM
WM
CSF
GM+WM
WM+CSF
GM+CSF
GM+WM+CSF

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
```diff
@@ -1034,29 +1086,26 @@
...
 
 ROI = xASL_im_Column2IM(ROI, x.S.masks.WBmask); % convert to image, decompress
 
-ROI = xASL_im_PVC_ROIexpansion(ROI, pGM, pWM, 17); % Dilate ROI
-pCSF = max(0,1-pGM-pWM);
-ROI(pCSF>0.35) = 0; % Remove pCSF from the mask
-ROI = xASL_im_PVC_ROIexpansion(ROI, pGM, pWM, 17); % Dilate ROI
+if max(size(pGM,1),size(pGM,2) < 130) 
+	% native space
+	ROI = xASL_im_PVC_ROIexpansion(ROI, pGM, pWM, 2); % Dilate ROI
```
HenkMutsaerts, line 1098: Can we not make `xASL_im_PVC_ROIexpansion` space-dependent instead; that it checks the voxel-size and determines itself what to dilate with? Because now in native space you have an arbitrary dilation distance, dependent on voxel-size.

And `xASL_stat_GetROIstatistics` should become space-ignorant

Modules/Module_Processing/xASL_init_LoadDataPar.m
```diff
@@ -182,7 +182,7 @@
 			if ~isempty(regexpi(x.S.Atlases{iAtlas}, 'WM')) || ~isempty(regexpi(x.S.Atlases{iAtlas}, 'whitematter'))
 				textRecommendation = [textRecommendation 'WM'];
 			elseif ~isempty(regexpi(x.S.Atlases{iAtlas}, 'WB')) || ~isempty(regexpi(x.S.Atlases{iAtlas}, 'wholebrain'))
```
HenkMutsaerts, line 184: If we add GM+WM, should we remove WB (which was GM+WM right)?
I would use the same nomenclature here, so GM+WM+CSF throughout?

Modules/xASL_module_Population.m
```diff
@@ -267,7 +268,7 @@
             % We use the specified tissue type
             % 'GM' = gray matter
             % 'WM' = white matter
-            % 'WB' = whole brain (= GM+WM)
+            % 'GM+WM' = whole brain
```
HenkMutsaerts: I would call GM+WM "whole brain parenchyma" or similar, as some would think that whole brain = GM+WM+CSF

Testing/UnitTests/xASL_ut_master_xASL_ExploreASL.m
```diff
@@ -530,7 +530,7 @@
 % Create dataPar.json
 dataParStruct.x.settings.Quality = 0;
 dataParStruct.x.S.Atlases = {'Total','DeepWM','Hammers','HOcort_CONN','HOsub_CONN','Mindboggle_OASIS_DKT31_CMA'};
-dataParStruct.x.S.TissueMasking = {'GM','WM','WB','GM','GM','GM'};
+dataParStruct.x.S.TissueMasking = {'GM','WM','GM+WM','GM','GM','GM'};
```
HenkMutsaerts: We don't want to test CSF as well?

HenkMutsaerts (changes requested):
Some minor suggestions

Functions/InputOutput/xASL_io_ReadDataPar.m
jan-petr: Done. But sometimes less is more ;)

Modules/Module_Processing/Module_ASL/xASL_wrp_PreparePV.m
jan-petr: Agree. No need to check path for COPY. But otherwise, we want to be cautious and avoid crashes.

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
jan-petr: Yes - growing for GM only. That's what you've concluded in the issue.

In general, we want to avoid region growing as this will inadverently include more of the primary tissue as well. So it will always slightly spatially bias the analysis.

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
jan-petr: Done

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
jan-petr: Strictly speaking not. But I do issue a warning to the user and it will bring unnecessary problems if we remove it without a warning. I find that this is a clean approach and much more user oriented than simply removing it. Despite the version update.

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
jan-petr: We could indeed keep it.

Modules/Module_Processing/xASL_init_LoadDataPar.m
jan-petr: Not really. This thing isn't really altering the settings but it is giving only helpful instructions for the user based on the atlas name.

So based on the atlas name, we propose what kind of TissueMasking they should use if they forget to add the TissueMasking parameter.

We have introduced this helpful message because users were confused and asking for help.

So if somebody adds DeepWM atlas without adding TissueMasking, it will give an advice. Same as an atlas including WB - then it will propose using GM+WM. 

So this is fully consistent with the latest verison and not a background compatibility thing.

Modules/xASL_module_Population.m

Testing/UnitTests/xASL_ut_master_xASL_ExploreASL.m
jan-petr: Done

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
jan-petr: Good point. Fixed.

Functions/InputOutput/xASL_io_ReadDataPar.m
```diff
@@ -73,13 +73,13 @@
 %
 %   ...
 %   "S":
-%	{"Atlases": ["Total","Total","DeepWM","Mindboggle_OASIS_DKT31_CMA"],
-%    "TissueMasking": ["GM","WM","WM","GM"]}}
+%	{"Atlases": ["Total","Total","DeepWM","Mindboggle_OASIS_DKT31_CMA","Total","Total","Total","Total","Total"],
```
HenkMutsaerts, line 76: Should we rename "total" with "WholeBrain"?

Modules/Module_Processing/Module_ASL/xASL_wrp_PreparePV.m
```diff
@@ -105,6 +104,11 @@
 	% because we then go directly to standard space and only apply the T1->MNI and not ASL->T1 nonlinear transformation
 	xASL_spm_reslice([x.P.Path_rPWI ',1'], [x.P.Path_c1T1 ',1'], [],[], x.settings.Quality);
 	xASL_spm_reslice([x.P.Path_rPWI ',1'], [x.P.Path_c2T1 ',1'], [],[], x.settings.Quality);
+
+	if xASL_exist(x.P.Path_c3T1)
```
HenkMutsaerts, line 108: We don't need to say "if exists" right, as we don't need backward compatibililty?

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
HenkMutsaerts: OK then we should add a PM comment that the region-growing is only for GM, and may be needed for other tissues as well.

You didn't answer "what if you have GM+WM, then you may also want region growing to include CSF in the PVC" -> there may be more use cases where we need the region-growing for sufficient PVC.

I understand your concern about including more of the primary tissue; we need to build the region growing such that it remains mutually exclusive with the other ROIs of the primary tissue; and growing only to the important direction (e.g. only inward if you want more WM for GM PVC) -> perhaps we should just move this to a new issue (if we don't have it already?)

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
HenkMutsaerts: Sure; just requires more code, same for on the top; that you still check if c3T1 exists. It is fine, just if we keep doing it we keep growing in if-statements and the code becomes more complicated.

Modules/Module_Processing/xASL_init_LoadDataPar.m
HenkMutsaerts: I see, but then I would clarify:

"GM+WM" is (whole-brain) parenchyma, "GM+WM+CSF" is the full whole-brain intracranial; just that the word "whole-brain" may not always be clear.

HenkMutsaerts (approved):
The backward code compatability is fine with me, I suggest removing some of it because it makes the code less readable (and the whole point of a full release is that it is not backward compatible anymore, per semantic versioning -> and users should not mix ExploreASL versions with their data anyways).

"WB" is potentially confusing, I would replace "total" by "whole-brain" (which should include CSF) and the GM, WM, CSF is more clear

We should check that we have a region-growing issue, and we can move my comment/ideas there.

Modules/Module_Processing/Module_ASL/xASL_wrp_PreparePV.m
jan-petr: that's error prevention. not background compatibility as such ;)

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
jan-petr: Done

Modules/Module_Processing/xASL_init_LoadDataPar.m
jan-petr: Absolutely. I added the comments into code to explain

PR #2000, opened 2026-01-19
Bug #1974 q rprint structures

### LINKED ISSUE #1974: QC PDF crashes with x.Q.BASIL fields
### Description

x.Q.BASIL has subfields, which breaks the PDF printing as it also wants to print sequence parameters aside from the QC output parameters.

It seems that quantification settings sneaked into the acquisition parameters x field, while they should be strictly separated as sequence parameters are fixed but processing/quantification settings are user-defineable. Processing and quantification settings should be in x.module.asl, and acquisition parameters should be in x.Q. It is fine to have a separate quantification settings field in x.module.asl, but quantification settings are processing settings, not acquisition parameters.

JAN: Fully agree, but all BASIL/FABBER/TOPUP processing options should be already in x.modules.asl and not in x.Q. I'm not aware of any other processing parameters. You are probably using some old settings. Can you tell me which settings you use, I can then help to propose a fix: if I see your settings, I can 1) write a code that issues warnings, 2) modify the QC function so that it ignores nested structures. But the tasks belong are no longer relevant.

### Tasks
Problem is when x.Q has inside structures. This then crashes the QC report.

- [x] Move x.Q.BASIL to x.modules.asl.BASIL - this is already moved. X.Q.BASIL is incorrectly set in the config
- [x] Check if there are other quantification settings that live in x.Q that should belong in x.modules.asl - this isn't the case for any other variables.
- [x] Substructures in x.Q are skipped when creating QC-output list and then this won't generate any issues for the QC printing to PDF.

### Release notes

Avoid printing substructs in QC PDF report.

jan-petr:
### Linked issue

Close #1974

### DIFF
--- a/Functions/QualityControl/xASL_qc_CollectQC_ASL.m
+++ b/Functions/QualityControl/xASL_qc_CollectQC_ASL.m
@@ -48,7 +48,7 @@ function [x] = xASL_qc_CollectQC_ASL(x, iSubject, iSession)
 %
 % EXAMPLE: x = xASL_qc_CollectQC_ASL(x, 10, 1);
 % __________________________________
-% Copyright (c) 2015-2024 ExploreASL
+% Copyright (c) 2015-2026 ExploreASL
 % Licensed under Apache 2.0, see permissions and limitations at
 % https://github.com/ExploreASL/ExploreASL/blob/main/LICENSE
 % you may only use this file in compliance with the License.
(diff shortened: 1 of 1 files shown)


HenkMutsaerts (approved):
That's fine, but I consider this a quick/dirty fix (fixing the symptom). I would rather fix the cause (that BASIL.Infer is not a x.Q but a setting).

But that's perhaps nitpicking :)

PR #2013, opened 2026-02-12
Feature #2006 weighted mean

### LINKED ISSUE #2006: Weighted mean and PV threshold for each atlas
### Description

Add an extra feature to define PV threshold for each atlas/tissueType in dataPar.json and enable calculated weighted mean on top of just median and mean of CBF values.

### Tasks

- [x] Weighted mean cf9e809
- [x] PV threshold added as x.S.TissueThreshold - an optional parameter that can be changed for each Atlas 41da1b3 and 41da1b3
- [x] Lesions are transformed to native space as PV maps and not as binary masks. For normal atlas, we need binary mask that is modulated by PV of GM (or other tissue). For ROIs, we need the PV of the ROI. 3023333
- [x] Load Subject specific lesions as PV 33a14e5b
- [x] Avoid binarizing subject specific lesions 33a14e5b
- [x] Weighted mean for subject specific ROIs should be weighted by the ROI and truly weighted. For normal anatomical ROI, we assume the weighting by tissue is there inherently.   f3fa4c8b
- [x] x.S.LesionROIThreshold program
- [x] x.S.LesionROIThreshold documentation
- [x] x.S.TissueThreshold to documentation
- [x] Example of Thresholds in the processing examples in the documentation together with Atlases and Tissue masking.
- [x] Describe ROIThresholds and how to provide Lesion_t1_1 and ROIs in the processing documentation.

### Release notes

Individual PV threshold and weighted mean.

jan-petr:
### Linked issue

Close #2006

### DIFF
--- a/Functions/Statistics/xASL_stat_ComputeMean.m
+++ b/Functions/Statistics/xASL_stat_ComputeMean.m
@@ -10,7 +10,7 @@ function [CBF_GM, CBF_WM] = xASL_stat_ComputeMean(imCBF, imMask, nMinSize, bPVC,
 %            - ignore when 0 (OPTIONAL, default = 0)
 %   bPVC   - perform PV-correction (OPTIONAL, DEFAULT = 0)
 %            0 - don't do partial volume correction, just calculate a mean or median on imMask
-%            1 - simple partial volume correction by normalizaton by the GM volume - see Petr et al. 2018
+%            1 - simple partial volume correction by normalizaton by the GM volume, a simple weighted mean - see Petr et al. 2018
 %            2 - partial volume correction using linear regression and imGM, imWM maps according to Asllani et al. 2008
 %   bParametric - performs parametric statistics (1 mean) or non-parametric when turned off (0 median) (OPTIONAL, DEFAULT 1) 
 %   imGM   - GM partial volume map with the same size as imCBF
--- a/Functions/Statistics/xASL_stat_ComputeMean.m
+++ b/Functions/Statistics/xASL_stat_ComputeMean.m
@@ -12,7 +12,6 @@ function [CBF_GM, CBF_WM] = xASL_stat_ComputeMean(imCBF, imMask, nMinSize, bPVC,
 %            0 - don't do partial volume correction, just calculate a mean or median on imMask
 %            1 - simple partial volume correction by normalizaton by the GM volume - see Petr et al. 2018
 %            2 - partial volume correction using linear regression and imGM, imWM maps according to Asllani et al. 2008
-%            3 - weighted mean across the ROI with weighting with the GM volume - this is mainly used for Lesions/ROIs
 %   bParametric - performs parametric statistics (1 mean) or non-parametric when turned off (0 median) (OPTIONAL, DEFAULT 1) 
 %   imGM   - GM partial volume map with the same size as imCBF
 %            (OPTIONAL, REQUIRED for bPVC==2 and bPVC==1)
@@ -181,12 +180,6 @@ switch (bPVC)
 		gwcbf                      = (imCBF')*pinv(gwpv');
 		CBF_GM                     = gwcbf(1);
 		CBF_WM                     = gwcbf(2);
-	case 3
-		%% 3e. Weighted mean
-		if isempty(imGM)
-			error('imGM needs to be provided for bPVC == 1');
-		end
-		CBF_GM = xASL_stat_SumNan(imCBF.*imGM, 1)/xASL_stat_SumNan(imGM, 1);
 end
 
     % % Print histograms to check validity
--- a/Functions/Statistics/xASL_stat_ComputeMean.m
+++ b/Functions/Statistics/xASL_stat_ComputeMean.m
@@ -10,8 +10,9 @@ function [CBF_GM, CBF_WM] = xASL_stat_ComputeMean(imCBF, imMask, nMinSize, bPVC,
 %            - ignore when 0 (OPTIONAL, default = 0)
 %   bPVC   - perform PV-correction (OPTIONAL, DEFAULT = 0)
 %            0 - don't do partial volume correction, just calculate a mean or median on imMask
-%            1 - simple partial volume correction by normalizaton by the GM volume, a simple weighted mean - see Petr et al. 2018
+%            1 - simple partial volume correction by normalizaton by the GM volume - see Petr et al. 2018
 %            2 - partial volume correction using linear regression and imGM, imWM maps according to Asllani et al. 2008
+%            3 - weighted mean across the ROI with weighting with the GM volume - this is mainly used for Lesions/ROIs
 %   bParametric - performs parametric statistics (1 mean) or non-parametric when turned off (0 median) (OPTIONAL, DEFAULT 1) 
 %   imGM   - GM partial volume map with the same size as imCBF
 %            (OPTIONAL, REQUIRED for bPVC==2 and bPVC==1)
@@ -48,7 +49,7 @@ function [CBF_GM, CBF_WM] = xASL_stat_ComputeMean(imCBF, imMask, nMinSize, bPVC,
 %             volume errors on the estimation of gray matter cerebral blood flow with arterial spin labeling MRI. Magnetic Resonance Materials in 
 %             Physics, Biology and Medicine. 2018 Dec 1;31(6):725-34.
 % __________________________________
-% Copyright (C) 2015-2024 ExploreASL
+% Copyright (C) 2015-2026 ExploreASL
 % Licensed under Apache 2.0, see permissions and limitations at
 % https://github.com/ExploreASL/ExploreASL/blob/main/LICENSE
 % you may only use this file in compliance with the License.
@@ -180,6 +181,12 @@ switch (bPVC)
 		gwcbf                      = (imCBF')*pinv(gwpv');
 		CBF_GM                     = gwcbf(1);
 		CBF_WM                     = gwcbf(2);
+	case 3
+		%% 3e. Weighted mean
+		if isempty(imGM)
+			error('imGM needs to be provided for bPVC == 1');
+		end
+		CBF_GM = xASL_stat_SumNan(imCBF.*imGM, 1)/xASL_stat_SumNan(imGM, 1);
 end
 
     % % Print histograms to check validity
--- a/Modules/Module_Processing/Module_Population/xASL_im_CreateGroupAnalysisMask.m
+++ b/Modules/Module_Processing/Module_Population/xASL_im_CreateGroupAnalysisMask.m
@@ -118,7 +118,19 @@ for iAtlas = 1:length(x.S.Atlases)
 	% Note that since #900, all atlases are stored in the x.D.Atlas field, with extension .nii only
 end
 
-%% B2) Save FOV mask for each subject
+%% B2. Transform atlases to native space for each ASL session
+
+%% >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
+%% >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
+%% >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
+%% >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
+%% >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
+%% >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
+%% >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
+
+%% PM: THIS IS A TEMPORARY HACK, THIS FUNCTION IS ABOUT CREATING A SINGLE GROUP ANALYSIS MASK, NOT ABOUT ROIs OR ATLASES.
+%% THIS PART SHOULD BE MOVED TO xASL_stat_AtlasForStats
+
 if x.modules.population.bNativeSpaceAnalysis
 	for iSession=1:x.dataset.nSessions
 		%x.SESSIONS{iSession}
@@ -179,6 +191,15 @@ end
 if ~bCreateStandardSpaceMasks
 	return;
 end
+
+%% >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
+%% >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
+%% >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
+%% >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
+%% >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
+%% >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
+%% >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
+%% >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
 
 %% C) Create & save VBA mask
 MaskAnalysis = MaskSusceptibility;
--- a/Modules/Module_Processing/Module_Population/xASL_im_CreateGroupAnalysisMask.m
+++ b/Modules/Module_Processing/Module_Population/xASL_im_CreateGroupAnalysisMask.m
@@ -143,7 +143,7 @@ if x.modules.population.bNativeSpaceAnalysis
 						[~, filenameAtlas, extensionAtlas] = xASL_fileparts(pathMaskInputList{iAtlas});
 						pathMaskOutput = fullfile(x.dir.SESSIONDIR, [filenameAtlas '_Atlas' extensionAtlas]);
 						% Run the transformation function
-						xASL_im_CreateGroupAnalysisMask_Transform(pathMaskInput, pathMaskOutput, x.P.Path_PWI, x);
+						xASL_im_CreateGroupAnalysisMask_Transform(pathMaskInput, pathMaskOutput, x.P.Path_PWI, x, 1);
 					end
 				end
 
@@ -161,7 +161,7 @@ if x.modules.population.bNativeSpaceAnalysis
 						LesionROINativePaths{iROI} = fullfile(x.dir.SESSIONDIR, [LesionROIStandardPaths{iROI}(2:iEnd), '.nii']);
 						
 						% Transform the Lesion or ROI to native space
-						xASL_im_CreateGroupAnalysisMask_Transform(fullfile(x.D.PopDir, LesionROIStandardPaths{iROI}), LesionROINativePaths{iROI}, x.P.Path_PWI, x);
+						xASL_im_CreateGroupAnalysisMask_Transform(fullfile(x.D.PopDir, LesionROIStandardPaths{iROI}), LesionROINativePaths{iROI}, x.P.Path_PWI, x, 0);
 					end
 					
 				end
@@ -251,7 +251,7 @@ function [outputPath, bCreateStandardSpaceMasks] = xASL_sub_CheckTemplatePath(pr
 	end
 end
 
-function xASL_im_CreateGroupAnalysisMask_Transform(pathMaskInput, pathMaskOutput, pathReference, x)
+function xASL_im_CreateGroupAnalysisMask_Transform(pathMaskInput, pathMaskOutput, pathReference, x, bBinary)
 % Function that transforms the input mask from standard space to native space
 % pathMaskInput is the input file in standard space
 % pathMaskOutput is the output file name in native space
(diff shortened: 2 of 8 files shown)


Functions/Statistics/xASL_stat_ComputeMean.m
```diff
@@ -180,6 +181,12 @@
 		gwcbf                      = (imCBF')*pinv(gwpv');
 		CBF_GM                     = gwcbf(1);
 		CBF_WM                     = gwcbf(2);
+	case 3
+		%% 3e. Weighted mean
+		if isempty(imGM)
+			error('imGM needs to be provided for bPVC == 1');
```
HenkMutsaerts: bPVC ==3

Functions/Statistics/xASL_stat_ComputeMean.m
```diff
@@ -180,6 +181,12 @@
...
+		end
+		CBF_GM = xASL_stat_SumNan(imCBF.*imGM, 1)/xASL_stat_SumNan(imGM, 1);
```
HenkMutsaerts: I don't get it, with any PVC we assume that within an ROI/lesion, GM CBF is the same everywhere.
So if this is not the case for (3), this should at least be clarified.

Also, this won't give you CBF but CBF*pvGM?
It seems a variant on case (1).

Let's assume a single voxel ROI of CBF 25 and pvGM = 0.5. "Single-compartment" PVC (case 1) should say that a 100% GM-filled voxel should be CBF 50 (definition of PVC). Case 3 gives (25 * 0.5) / 0.5 = 25?
Why would one want this, and it is not partial volume correction? (which by definition gives the CBF of a voxel filled 100% with GM?

Modules/Module_Processing/Module_Population/xASL_im_CreateGroupAnalysisMask.m
```diff
@@ -275,11 +275,11 @@ function xASL_im_CreateGroupAnalysisMask_Transform(pathMaskInput, pathMaskOutput
 	xASL_delete(pathTmpPreSmooth);
 
 	% Threshold at 50%
-	imMaskTmp = xASL_io_Nifti2Im(pathMaskOutput);
-	imMaskTmp = imMaskTmp > 0.5;
-	xASL_io_SaveNifti(pathMaskOutput, pathMaskOutput, imMaskTmp);
+	%imMaskTmp = xASL_io_Nifti2Im(pathMaskOutput);
+	%imMaskTmp = imMaskTmp > 0.5;
+	%xASL_io_SaveNifti(pathMaskOutput, pathMaskOutput, imMaskTmp);
 else % multilabel file
```
HenkMutsaerts: should be "binary or (multi)label file" right?

Modules/Module_Processing/Module_Population/xASL_im_CreateGroupAnalysisMask.m
```diff
@@ -275,11 +275,11 @@ function xASL_im_CreateGroupAnalysisMask_Transform(pathMaskInput, pathMaskOutput
...
 	% Threshold at 50%
-	imMaskTmp = xASL_io_Nifti2Im(pathMaskOutput);
-	imMaskTmp = imMaskTmp > 0.5;
-	xASL_io_SaveNifti(pathMaskOutput, pathMaskOutput, imMaskTmp);
+	%imMaskTmp = xASL_io_Nifti2Im(pathMaskOutput);
+	%imMaskTmp = imMaskTmp > 0.5;
+	%xASL_io_SaveNifti(pathMaskOutput, pathMaskOutput, imMaskTmp);
 else % multilabel file
-	% multi-label masks - no presmooth, nearest-neighbor interpolation, no thresholding
+	% multi-label masks - no presmooth, nearest-neighbor interpolation, no thresholding, keep binary
```
HenkMutsaerts: binary or multi-label also allowed?

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
```diff
@@ -204,25 +205,35 @@
 %% 0.b Native space atlas input
 if x.S.InputNativeSpace
 	inputAtlasTmp = xASL_io_Nifti2Im(fullfile(x.dir.xASLDerivatives,x.SUBJECTS{1},listSessions{1},[x.S.InputAtlasNativeName '.nii']));
-	atlasN = max(inputAtlasTmp(:));
-	x.S.InputMasks = zeros(length(x.LeftMask), atlasN);
-	for kk = 1:atlasN
-		x.S.InputMasks(:,kk) = xASL_im_IM2Column(inputAtlasTmp == kk,x.S.masks.WBmask);
-	end
+	if x.S.bSubjectSpecificROI
```
HenkMutsaerts, line 208: Note that every exception for native space will soon need to be refactored in having the same code here space-independent (resampling every step if needed)

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
```diff
@@ -204,25 +205,35 @@
 %% 0.b Native space atlas input
 if x.S.InputNativeSpace
 	inputAtlasTmp = xASL_io_Nifti2Im(fullfile(x.dir.xASLDerivatives,x.SUBJECTS{1},listSessions{1},[x.S.InputAtlasNativeName '.nii']));
-	atlasN = max(inputAtlasTmp(:));
-	x.S.InputMasks = zeros(length(x.LeftMask), atlasN);
-	for kk = 1:atlasN
-		x.S.InputMasks(:,kk) = xASL_im_IM2Column(inputAtlasTmp == kk,x.S.masks.WBmask);
-	end
+	if x.S.bSubjectSpecificROI
+		% For subject specific atlases - they are a 4D collection of non-binary PV maps
```
HenkMutsaerts: No, why not the same in each space? As we want to refactor this into identical space-independent code. So either always use IM2Column, or not, if this is isn't possible (IM2Column contains the full brain mask so should be able also for subject-specific atlases). Also doesn't matter if they are binary or continuous.

Or do you mean that depending on the space, you won't use the IM2Column?

Or why do you still use the IM2Column for the standard atlases in native space?

Why should subject-specific atlases be non-binary? Any input atlas is non-binary first, and can be masked into binary; if we want this as general option inside ExploreASL, this should go to the atlas management (the previous function), not to GetROIStatistics; as this would get too convoluted.

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
```diff
@@ -204,25 +205,35 @@
...
 
-	if ~isempty(strfind(x.S.InputAtlasNativeName,'Hammers'))
-		namesROIs2Merge{1} = {'middle_frontal_gyrus','superior_frontal_gyrus','inferior_frontal_gyrus'};
-		namesROIs2Merge{2} = {'lateral_orbital_gyrus','medial_orbital_gyrus','posterior_orbital_gyrus','anterior_orbital_gyrus'};
-		namesROIsJoint = {'middle+superior+inferior_frontal_gyrus','lateral+medial+posterior+anterior_orbital_gyrus'};
-	else
-		namesROIs2Merge = [];
-		namesROIsJoint = [];
-	end
+		if ~isempty(strfind(x.S.InputAtlasNativeName,'Hammers'))
```
HenkMutsaerts, line 223: GetROIstatistics should be atlas-ignorant. Any ROI names to merge should be in the Atlas management function. And why should Hammers always be merged in native space? Perhaps not everybody wants this?

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
```diff
@@ -204,25 +205,35 @@
...
+		else
+			namesROIs2Merge = [];
+			namesROIsJoint = [];
+		end
 
-	% Amend the new atlas names, change the actual atlas for each subject alone
-	for rr = 1:length(namesROIs2Merge)
-		atlasN = atlasN + 1;
-		namesROIlocal{atlasN} = namesROIsJoint{rr};
+		% Amend the new atlas names, change the actual atlas for each subject alone
```
HenkMutsaerts, line 232: Also goes to atlas management, not ROIstatistics

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
```diff
@@ -276,7 +287,7 @@
 
 
 	if ~x.S.InputNativeSpace
-		SumList = squeeze(sum(x.S.InputMasks,1));
+		SumList = squeeze(sum(x.S.InputMasks>0.5,1));
```
HenkMutsaerts, line 290: Keep the same for each space, so if you want more functionality, program this space independently (or we do this later, but be aware that the refactoring gets more complicated)

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
```diff
@@ -362,19 +372,28 @@
 				x.LeftMask = xASL_im_IM2Column(x.LeftMask, x.S.masks.WBmask);
 
 				inputAtlasTmp = xASL_io_Nifti2Im(fullfile(x.dir.xASLDerivatives,x.SUBJECTS{iSubject},listSessions{iSess},[x.S.InputAtlasNativeName '.nii']));
-				atlasN = max(inputAtlasTmp(:));
-				x.S.InputMasks = zeros(length(x.LeftMask),atlasN);
-				for kk = 1:atlasN
-					x.S.InputMasks(:,kk) = xASL_im_IM2Column(inputAtlasTmp == kk,x.S.masks.WBmask);
-				end
+				if x.S.bSubjectSpecificROI
+					% For subject specific atlases - they are a 4D collection of non-binary PV maps
```
HenkMutsaerts, line 376: Same as above

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
```diff
@@ -404,9 +423,9 @@
 					for iROI=1:size(x.S.InputMasks,2)
 						for iMask=1:size(x.S.InputMasks,3)
 							if sum(SumList(iROI,iMask))~=0 % skip empty ROIs
-								x.S.InputMasks(:,iROI,iMask) = xASL_im_CreatePVEcROI(x,x.S.InputMasks(:,iROI,iMask), pGM_MNI, pWM_MNI);
+								x.S.InputMasks(:,iROI,iMask) = xASL_im_CreatePVEcROI(x,x.S.InputMasks(:,iROI,iMask)>0.5, pGM_MNI, pWM_MNI);
```
HenkMutsaerts, line 426: Keep atlas management in the atlas function

Modules/Module_Processing/Module_Population/xASL_wrp_GetROIstatistics.m
```diff
@@ -142,14 +142,14 @@ function xASL_wrp_GetROIstatistics(x)
 end
 
 % Add subject/run-wise input data (e.g., qCBF) and subject/run-wise tissue mask (e.g., GM or GM+WM)
-x.S.output_ID = [x.S.output_ID x.S.InputDataStr '_' strrep(x.S.TissueMaskingLocal,'+','_')];
+x.S.output_ID = [x.S.output_ID x.S.InputDataStr '_' strrep(x.S.TissueMaskingLocal,'+','_') '_PV' num2str(x.S.TissueThresholdLocal)];
```
HenkMutsaerts, line 145: So you want a subject-specific threshold or subject-specific weighting? Or both?

Modules/xASL_module_Population.m
```diff
@@ -65,10 +65,40 @@
...
+if length(x.S.TissueMasking) ~= length(x.S.TissueThreshold)
+	error('x.S.TissueThreshold has to match the length of x.S.TissueMasking');
+end
+
+if max(x.S.TissueThreshold) > 1
+	error('Maximum value for x.S.TissueThrehold is 1');
+end
+
+if min(x.S.TissueThreshold) < 0
+	error('Minimum value for x.S.TissueThrehold is 0');
```
HenkMutsaerts: Threshold spelling (you can check "resolved" for such minor conversations after you fixed them, I don't need to check them)

Testing/UnitTests/xASL_ut_master_xASL_ExploreASL.m
```diff
@@ -1048,6 +1048,7 @@
 dataParStruct.x.settings.Quality = 0;
 dataParStruct.x.S.Atlases = {'TotalGM','DeepWM','Hammers','HOcort_CONN','HOsub_CONN','Mindboggle_OASIS_DKT31_CMA'};
 dataParStruct.x.S.TissueMasking = {'GM','WM','GM','GM','GM','GM'};
+dataParStruct.x.S.TissueThreshold = [0.5, 0.5, 0.5, 0.5, 0.5, 0.5];
```
HenkMutsaerts, line 1051: 0.5 or 0.7? elsewhere we default to 0.7?

HenkMutsaerts (changes requested):
I think we should have discussed this before programming. Main points: 1. atlas/ROI management happens in xASL_stat_AtlasForStats, not in GetROIstatistics. 2. Native/standard space and both subject-specific and standard atlases/ROIs should be managed identically; many atlases are first probablistic, also standard atlases, before they are binarized into masks, if we want this to be a general feature, we should implement it as general feature (also why I suggested to carefully think before we implement a feature wish that somebody else could do theirselves as well; though I get your wish to do it the right way

Functions/Statistics/xASL_stat_ComputeMean.m
jan-petr: This is now removed.

Functions/Statistics/xASL_stat_ComputeMean.m
jan-petr: I removed this option as I decided not to use it anyway.

Modules/Module_Processing/Module_Population/xASL_im_CreateGroupAnalysisMask.m
jan-petr: I explain the terminology. Either 3D file with integers or 4D with 1 or more binary masks.

Modules/Module_Processing/Module_Population/xASL_im_CreateGroupAnalysisMask.m
jan-petr: that's the same. it's always binary. See my new explanation.

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
jan-petr: Yes. We have a master issue for this and I'm simplifying this over time.

Testing/UnitTests/xASL_ut_master_xASL_ExploreASL.m
jan-petr: This is Unit testing -> we need a number to make sure it works. Doesn't matter which number.

Modules/xASL_module_Population.m
jan-petr: Good catch. I did a fulltext search and it's luckily just these two messages, not really anywhere in the real code.

Modules/Module_Processing/Module_Population/xASL_wrp_GetROIstatistics.m
jan-petr: Not sure if I understand. 
We didn't have any option to change GM-threshold. This was fixed. We can now change this threshold in dataPar.json. And we can change it for any atlas provided. So you can do CBF+GM+Tatu+0.5 or sCoV+GM+Total+0.7 etc etc. And set any combination - so we need to report the PV value with the TSV file for providence.

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
jan-petr: Not sure if I can answer that. We have decided that together in the past :) Probably because the regions were too small. I'm happy changing this in any way - doing this for both native+standard, deleting it, doing in atlas management. 

But its not my choice to do this :)

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
jan-petr: as said above - we've decided to do this in the past....

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
jan-petr: We have decided to keep atlases as 3D files with regions indexed as integers.

For Lesions/ROIs, we have 4D files with these 4 automatically created lesions - contralateral hemisphere, contralateral ROI etc. And I want to keep them as probabilistic as they are not (unlike atlases) coupled with tissue masking.

So we need to treat atlases and Lesions/ROIs differently.

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
jan-petr: This is not atlas management. If we have probabilistic ROIs, we need to threshold them for these support functions. Otherwise, we convert them to binary as ==0 or >0. So a mask with 0.00001 will be treated as 1 -> and this can lead to big issues.

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
jan-petr: See similar comment below

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
jan-petr: Fully agree that IM2Column should be used for all options - and it is the case.

See the explanation below why ROIs/Lesions are probabilistic and atlases binary. But in general - all atlases are binary and when downsampling, we are keeping them binary. Which is fine, because atlases are always coupled with tissue masking - usually GM. 

ROI and Lesions are binary in their native space, but I don't want to keep them binary when resample to ASL voxels, because ROIs/Lesions are evaluated as they are without combining them with tissue-masking.

Regarding Atlas management -> this is indeed all done in atlas management. ROIs are downsampled in atlas management - here we are only applying the IM2Column to them...

The reason for having this difference is that ROIs come as 4D probabilistic (ROI, perilesion, contralateral etc - your idea ;)). And Atlases come as 3D with integers (also your decision). So we have to have this if we want to run IM2Column.

Previously, we have binarized the ROIs using "logical" and this made quite some mess or even 0.001 being considered as inside ROI.

Modules/Module_Processing/Module_Population/xASL_im_CreateGroupAnalysisMask.m
HenkMutsaerts: No, to be clear, I would say:

binary == 0 and 1
multilabel == 0 background 1 putamen 2 white matter 3 CSF etc

So you mean multiple 3D binary images, that's fine, but then please don't call it multi-label as this can be confusing.

So in this case, it probably shouldn't be a boolean binary but rather single or multiple masks?

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
HenkMutsaerts: I thought we had two options (ignoring PVC1)

1. PVC0: >0.7 (or >0.5 whatever)
2. PVC2: not thresholding, first region growing/fixing so we have sufficient amount of voxels for each tissue type, then >0.5 to avoid 0.00001 inside PVC, then run PVC?

Is this still the case?

Modules/xASL_module_Population.m
```diff
@@ -65,10 +65,40 @@
 	error('You need to provide x.S.Atlases and x.S.TissueMasking with the same length');
 end
 
+if ~isfield(x.S, 'TissueThreshold')
+	% The default threshold is 0.7
+	x.S.TissueThreshold = ones(1,length(x.S.TissueMasking)) * 0.7;
+end
```
HenkMutsaerts: Do you mind "elseif-ing" these new if-statements for readability and shorter code?

Modules/xASL_module_Population.m
```diff
@@ -65,10 +65,40 @@
...
+
+if min(x.S.LesionROIThreshold) < 0
+	error('Minimum value for x.S.LesionROIThreshold is 0');
+end
+
 % Print the used atlases	
 fprintf('\nThe following ROI atlases have been selected with the following tissue:\n')
 for iAtlas=1:length(x.S.Atlases)
-    fprintf('%s\n', [x.S.TissueMasking{iAtlas} ' tissue within ' x.S.Atlases{iAtlas} ' ROIs']);
+    fprintf('%s\n', [x.S.TissueMasking{iAtlas} ' tissue within ' x.S.Atlases{iAtlas} ' ROIs, and Tissue threshold ' num2str(x.S.TissueThreshold(iAtlas))]);
```
HenkMutsaerts: Can we rephrase this as:

fprintf('%s\n', [x.S.TissueMasking{iAtlas} '>' num2str(x.S.TissueThreshold(iAtlas)) ' tissue within ' x.S.Atlases{iAtlas} ' ROIs, and Tissue threshold ' num2str(x.S.TissueThreshold(iAtlas))]);

or similar?

Most important that it is clear, but we can make it nicely readable as this may generate many sentences

HenkMutsaerts (changes requested):
Some minor still, rest that I would like to have improved we discuss in the ExploreASL whatsapp and I add it to the native space cleanup issue

Modules/Module_Processing/Module_Population/xASL_im_CreateGroupAnalysisMask.m
jan-petr: Sure - "multi-label" is not what I wrote in the code :) I've renamed it to 3D non-binary and 4D binary.
I've improved the descriptions, but it's pointless to say that binary is 0 and 1 - literally everybody knows that. Including medical doctors.

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
jan-petr: Yes. But that's related to what I said in the other comments. PV-threshold is thresholding the tissue mask (e.g. GM). But atlas mask was segmented as ==0 or >0. And for Lesions+ROIs, we don't add the additional tissue mask - so PV-threshold wasn't applied and the thresholding of the ROI wasn't really ideal.

Modules/xASL_module_Population.m

HenkMutsaerts (approved):
Please check my last commit and agree with it before merging, see also addition to the top of the description in #58.

PR #2019, opened 2026-02-20
Closes #1439 wmh alignment rerunning

### LINKED ISSUE #1439: WMH alignment in rerunning ExploreASL
### Description

When the structural module is repeated without removing the derivatives folder, the WMH_SEGM becomes misaligned because ExploreASL assumes that it is aligned with FLAIR if provided by someone else, but ExploreASL's structural module aligns it with the T1w.

### Tasks
- [x] 1. Create a function that detects if two NIfTIs have the same dimensions, and then the same orientation (outputting if this is mat or mat0): `xASL_im_IsEqualOrientations`
- [X] Ensure that we don't revert to nii.mat0
    -> we actually do revert to nii.mat0 in several instances, if the orientation matrix nii.mat is corrupt (e.g., contains NaNs), 
         then we issue a warning.
- [X] 2. xASL_adm_CleanUpBeforeRerun: line 198 (restore back-upped ORI files):
This doesn't take anything from mat0
- [x] 3. xASL_im_CenterOfMass: add a warning when defaulting to mat0
- [x] 4. xASL_wrp_LinearReg_FLAIR2T1w: add ROI's to registration list
- [x] 5. xASL_wrp_LST_Segment_FLAIR_WMH: correct mat0 behavior
- [x] 6. Disable changeMat0 warning in xASL_wrp_LST_Segment_FLAIR_WMH

### How to test

1. Run the structural module with a T1w & FLAIR but not WMH_SEGM; check registration WMH_SEGM with T1w & FLAIR at the end of the structural module
2. Rerun the structural module after deleting the structural lock files (but keeping the BIDS2Legacy lock files); check registration WMH_SEGM with T1w & FLAIR at the end of the structural module
3. Run the structural module with a T1w & FLAIR and WMH_SEGM registered to the FLAIR (e.g., Twins population, with FLAIR from Carole); check registration WMH_SEGM with T1w & FLAIR at the end of the structural module
4. Run the structural module with a T1w & FLAIR and Lesion_T1_1.nii & Lesion_T1_2.nii

(add these examples to the TestDataSet examples)

Example by Mervin: `/Users/hjmutsaerts/ExploreASL/ASL/Structural_Registration_Test_HJM`

### Release notes

Fix WMH realignment issues after rerunning ExploreASL. Note that this doesn't fix rerunning datasets that were already processed before this fix.

HenkMutsaerts:
### Linked issue

Closes #1439

### DIFF
--- a/Functions/ImageProcessing/xASL_im_IsEqualOrientations.m
+++ b/Functions/ImageProcessing/xASL_im_IsEqualOrientations.m
@@ -1,0 +1,49 @@
+function [isEqualdim, isEqualMat0, isEqualMat] = xASL_im_IsEqualOrientations(pathNifti1, pathNifti2)
+%xASL_im_IsEqualOrientations Check if NIfTI images have equal orientations
+%
+% FORMAT: [isEqualdim, isEqualMat0, isEqualMat] = xASL_im_IsEqualOrientations(pathNifti1, pathNifti2)
+% 
+% INPUT:
+%
+%   Definitions:
+%
+%   pathNifti1 - Path to the first NIfTI (REQUIRED, STRING)
+%   pathNifti2 - Path to the second NIfTI (REQUIRED, STRING)
+%
+% OUTPUT:
+%
+%   isEqualdim  - Boolean for equal image matrix dimensions (true = equal, false = not equal)
+%   isEqualMat0 - Boolean for equal orientation mat0 (the original image orientation, i.e. before realignments)
+%   isEqualMat  - Boolean for equal orientation mat (the current image orientation, i.e. after realignments)
+%                         
+% -----------------------------------------------------------------------------------------------------------------------------------------------------
+% DESCRIPTION:          This function checks is NIfTI images have equal orientations, which can be useful e.g., when derivatives and original
+%                       images need to stay aligned (e.g., FLAIR and WMH_SEGM)
+% 
+% -----------------------------------------------------------------------------------------------------------------------------------------------------
+%
+% EXAMPLE: [isEqualdim, isEqualMat0, isEqualMat] = xASL_im_IsEqualOrientations('pathToMyNifti', 'pathToMyOtherNifti');
+%
+% __________________________________
+% Copyright (c) 2015-2025 ExploreASL
+% Licensed under Apache 2.0, see permissions and limitations at
+% https://github.com/ExploreASL/ExploreASL/blob/main/LICENSE
+% you may only use this file in compliance with the License.
+% __________________________________
+
+
+%% 1. Load NIfTIs
+objectNifti1 = xASL_io_ReadNifti(pathNifti1);
+objectNifti2 = xASL_io_ReadNifti(pathNifti2);
+
+
+%% 2. Compare image dimensions
+isEqualdim = isequal(objectNifti1.hdr.dim, objectNifti2.hdr.dim);
+
+
+%% 3. Compare orientations
+isEqualMat0 = isequal(objectNifti1.mat0, objectNifti2.mat0);
+isEqualMat = isequal(objectNifti1.mat, objectNifti2.mat);
+
+
+end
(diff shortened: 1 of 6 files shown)


Modules/Module_Processing/Module_Structural/xASL_wrp_LST_Segment_FLAIR_WMH.m
jan-petr, line 189: Resampling should not be based on the images size, but also on the similarity of the image matrix. In theory, you can have the same size but different orientation

Modules/Module_Processing/Module_Structural/xASL_wrp_LST_Segment_FLAIR_WMH.m
```diff
@@ -209,11 +209,20 @@ function xASL_wrp_LST_Segment_FLAIR_WMH(x, rWMHPath, WMHsegmAlg)
...
 % Create a dummy WMH_SEGM in case of a FLAIR without WMH_SEGM
-if xASL_exist(x.P.Path_rFLAIR, 'file') && ~xASL_exist(x.P.Path_WMH_SEGM, 'file')
-    imageFLAIR = xASL_io_Nifti2Im(x.P.Path_rFLAIR);
-    imageWMH = zeros(size(imageFLAIR), 'logical');
-    xASL_io_CreateNifti(x.P.Path_WMH_SEGM, imageWMH, [], [], 0);
+% >>>>>>>>>>> PM: THIS IS TRICKY, we assume here that if LST did not create a WMH_SEGM, there is no lesion load,
+% >>>>>>>>>>> although LST could also have not run correctly (but it would have crashed)
+if xASL_exist(x.P.Path_FLAIR, 'file') && ~xASL_exist(x.P.Path_WMH_SEGM, 'file')
+    warning('LST did not segment WMH, so we create a dummy WMH_SEGM.nii, assuming no WMH load');
+    imageT1 = xASL_io_Nifti2Im(x.P.Path_T1);
```
jan-petr, line 218: Is there a reason why you switch from rFLAIR to T1. I'm fine with it, just double checking if that's intentional. You can resolve.

Functions/ImageProcessing/xASL_im_IsEqualOrientations.m
```diff
@@ -0,0 +1,49 @@
+function [isEqualdim, isEqualMat0, isEqualMat] = xASL_im_IsEqualOrientations(pathNifti1, pathNifti2)
```
jan-petr, line 1: The function looks OK. But I wonder why have you even created it when it's not used anywhere. 
You have also talked about fixing a bug for Mervin with non-revereted WMH - and I don't see this or the used of this function anywhere-  is that OK?

jan-petr (changes requested):
A few comments and submitted a few minor fixes.

Modules/Module_Processing/Module_Structural/xASL_wrp_LST_Segment_FLAIR_WMH.m
HenkMutsaerts: True, how should we implement this? Just checking the voxel size of the orientation matrix?

Functions/ImageProcessing/xASL_im_IsEqualOrientations.m
HenkMutsaerts: True, I created it but then decided not to use it; initially I wanted to do a check for equal mat/mat0 between NIfTIs for rerunning, but then I changed my mind that fixing the cause is better than dealing with the symptom :)

jan-petr: OK :) That's fine. Lets keep it. You can resolve

Modules/Module_Processing/Module_Structural/xASL_wrp_LST_Segment_FLAIR_WMH.m
jan-petr: Yes exactly - check both the voxel size, but also read "mat" and check if this is equal. First check "mat" and then image size.

jan-petr: will do in #2024

jan-petr (approved):
OK. One last fix is moved to #2024

PR #2022, opened 2026-02-26
Bug #1747 resolution estimation

### LINKED ISSUE #1747: Resolution estimation
### Description

The resolution estimation needs to be revisited inside xASL_init_DefaultEffectiveResolution.m

### Tasks

- [x] Fix code on line 136 of xASL_adm_DefineASLResolution to qnt = sort(imCBF(~isnan(imCBF),'ascend'))) a39081d3
- [x] If AcquisitionVoxelSize is provided, use this value instead of guessing a39081d3
HENK: we need to check that in BIDS this is the effective spatial resolution, not the nominal one
JAN:AcquisitionVoxelSize is the nominal one. So not effective one. But still better than the reconstructed resolution :) We should indeed add those extra effects to the VoxelSize - which we are doing.
HENK2: Nominal can be reconstructed resolution?
JAN2: No, nomimal is defined by acquisition matrix and acquisition FOV. It's close to the acquired voxel size (but not equal to), which might be still different from effective resolution. Reconstructed is what you choose to upsample to - can be kept at nominal, or can be upsampled.

- [x] Input resolutions only for <= 3T, we don't know them for 7T (as blurring in 3D ASL depends on T2, which depends on fieldstrength) a39081d3
- [x] GE resolution lower than 2mm in-plane needs to be adjusted as two times upsampled for 3T sequence. But do this only for 8 spirals. For 4 spirals, the resolution is worse. For no spiral information, we skip this. 5717eefc
HENK: why only lower than 2mm? I would definitely not skip this when not knowing the spiral information, often we don't know the spiral information, this is not a required BIDS field, and with GE we can safely assume that most users have 8 spirals. 80%20% rule: it is better to have better quantification in the majority of the users than to have better quantification in the minority of experts testing something exotic
JAN: Spiral information is saved in 95% of GE DICOMs, so we have this information in nearly all cases. But I'm fine setting the default to 8 when the DICOM info isn't available - added a task below.
HENK2: Great, and we could issue a warning. 99% of all GE data is 8 spirals. BTW: this is in the import right, because it is a BIDS parameter? (we should do data-dependent parameters mostly in import, and only processing settings/choices in the processing, as much as we can)
JAN2: We should be careful in setting unknown things to defaults in import. It is simply unknown and reported unknown, but we shouldn't invent values. Instead, during processing, we can make educated guess, but we can't assume things during import which is supposed to be rigorous. When spiral info isn't known - default to 8 during resolution estimation.  3a4cd7c
- [ ] GRASE - lower resolution than 3mm is fine, we set the threshold to 2mm for 3T and 1.5T but not 7T and we only know these for Siemens
HENK: why? with lower resolution than 3 mm you still have the PSF right?
JAN: The point here is to identify the difference between nominal and reconstructed resolution. GE never has lower resolution than 3mm unless this is a reconstructed resolution by a factor of 2. For GRASE and Siemens, nominal resolution between 2-3mm is not uncommon, so we don't raise any red flags. But when the resolution is below 2mm, then we need to sound the alarm.
HENK2: OK; afaik the nominal resolution IS the reconstruction resolution (at how it is reported and seems to be for many users).
JAN2: No, reconstruction might be equal to acquisition and nominal. but nominal IS NOT reconstruction and IS NOT acquisition. These are 3 distinct terms. Most people report "resolution" and do not care about the difference between them - but that's either simplification, or is true for a particular sequence only, or is plainly wrong. The fact the most people use terms incorrectly doesn't make the terms right and doesn't justify incorrect usage :)
Reconstruction = whatever grid you choose to use for final data display
Nominal = theoretical voxel size based on FOV and matrix
Acquisition = nominal, but taking into account partial Fourier etc.
Effective = taking into account blurring and imperfections etc

### Release notes

Update resolution defaults for GRASE and Spiral sequences.

jan-petr:
### Linked issue

Close #1747

### DIFF
--- a/Modules/Module_Processing/xASL_init_DefaultEffectiveResolution.m
+++ b/Modules/Module_Processing/xASL_init_DefaultEffectiveResolution.m
@@ -71,7 +71,7 @@ elseif x.MagneticFieldStrength > 3
 	% For more than 3T scanners, we don't have this estimated as the resolutions are much higher
 	EffectiveResolution = NativeResolution;
 	fprintf('%s\n', ['Assume effective resolution equal to voxel size ' num2str(EffectiveResolution(1)) ' ' num2str(EffectiveResolution(1)) ' ' num2str(EffectiveResolution(3))]);
-elseif ~isfield(x.Q, 'PulseSequenceType') || ~isfield(x.Q, 'MRAcquisitionType') || ~isfield(x.Q.Vendor)
+elseif ~isfield(x.Q, 'PulseSequenceType') || ~isfield(x.Q, 'MRAcquisitionType') || ~isfield(x.Q, 'Vendor')
 	% If sequence or vendor are still missing we skip this function
 	error('Settings of x.Q.PulseSequenceType or x.Q.MRAcquisitionType or x.Q.Vendor are missing');
 else	
--- a/Modules/Module_Processing/xASL_init_DefaultEffectiveResolution.m
+++ b/Modules/Module_Processing/xASL_init_DefaultEffectiveResolution.m
@@ -70,7 +70,7 @@ if ~isempty(x.Q.AcquisitionVoxelSize) && numel(x.Q.AcquisitionVoxelSize) == 3
 elseif x.MagneticFieldStrength > 3
 	% For more than 3T scanners, we don't have this estimated as the resolutions are much higher
 	EffectiveResolution = NativeResolution;
-	fprintf('%s\n', ['Assume effective resolution equal to voxel size ' num2str(EffectiveResolution(1)) ' ' num2str(EffectiveResolution(1)) ' ' num2str(EffectiveResolution(3)) )
+	fprintf('%s\n', ['Assume effective resolution equal to voxel size ' num2str(EffectiveResolution(1)) ' ' num2str(EffectiveResolution(1)) ' ' num2str(EffectiveResolution(3))]);
 elseif ~isfield(x.Q, 'PulseSequenceType') || ~isfield(x.Q, 'MRAcquisitionType') || ~isfield(x.Q.Vendor)
 	% If sequence or vendor are still missing we skip this function
 	error('Settings of x.Q.PulseSequenceType or x.Q.MRAcquisitionType or x.Q.Vendor are missing');
@@ -105,9 +105,22 @@ else
 
 	%% ----------------------------------------------------------------------------------------
 	%% 2) Attempt accounting for in-plane interpolation in reconstruction
-	if strcmpi(x.Q.PulseSequenceType, 'spiral') && strcmpi(x.Q.MRAcquisitionType, '3D') && NativeResolution(1)<4
-		NativeResolution(1:2) = max(NativeResolution(1:2),[3.75 3.75]);
-		% to account for in-plane interpolation
+	if strcmpi(x.Q.PulseSequenceType, 'spiral') && strcmpi(x.Q.MRAcquisitionType, '3D')
+		% GE tends to upsample their spiral acquisitions to 1.6-1.9mm voxels
+		% For non-GE, or in-plane resolution higher than 2mm, we can't assume that the reconstruction was upsampled and we leave the native resolution intact
+		if regexpi(x.Q.Vendor, 'GE') && NativeResolution(1) < 2
+			% The individual resolution will also depend on the FOV, so we cannot set a fixed resolution but rather modify it based on the native resolution
+			if isempty(x.Q.NumberOfArms)
+				warning('Number of spirals not defined for a GE sequence');
+			elseif x.Q.NumberOfArms == 8
+				% For the standard number of spirals, we use 2 times the native resolution
+				NativeResolution(1:2) = 2 * NativeResolution(1:2);
+			elseif x.Q.NumberOfArms == 4
+				NativeResolution(1:2) = 3 * NativeResolution(1:2);
+			else
+				warning('Number of spirals for a GE sequence is not in the common range of 4 or 8')
+			end
+		end
 	elseif strcmpi(x.Q.PulseSequenceType, 'GRASE') && strcmpi(x.Q.MRAcquisitionType, '3D') && NativeResolution(1)<3
 		NativeResolution(1:2) = max(NativeResolution(1:2),[3.8 3.8]);
 	end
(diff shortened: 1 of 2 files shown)


Modules/Module_Processing/xASL_init_DefaultEffectiveResolution.m
```diff
@@ -51,50 +46,91 @@
...
-%% ----------------------------------------------------------------------------------------
-%% 2) Attempt accounting for in-plane interpolation in reconstruction
-if strcmpi(x.Q.PulseSequenceType, 'spiral') && strcmpi(x.Q.MRAcquisitionType, '3D') && NativeResolution(1)<4
-	NativeResolution(1:2) = max(NativeResolution(1:2),[3.75 3.75]);
-	% to account for in-plane interpolation
-elseif strcmpi(x.Q.PulseSequenceType, 'GRASE') && strcmpi(x.Q.MRAcquisitionType, '3D') && NativeResolution(1)<3
-	NativeResolution(1:2) = max(NativeResolution(1:2),[3.8 3.8]);
-end
+%% Obtain the effective resolution
+%% 1. Basic cases when resolution is not calculated
```
HenkMutsaerts: Not sure what you mean by this? "basic cases when resolution is not calculated"

Modules/Module_Processing/xASL_init_DefaultEffectiveResolution.m
```diff
@@ -51,50 +46,91 @@
...
+if ~isempty(x.Q.AcquisitionVoxelSize) && numel(x.Q.AcquisitionVoxelSize) == 3
+	% If estimated effective resolution is given in JSON-sidecar, then we use it
+	EffectiveResolution = x.Q.AcquisitionVoxelSize;
+	fprintf('%s\n', ['Assume effective resolution ' num2str(EffectiveResolution(1)) ' ' num2str(EffectiveResolution(1)) ' ' num2str(EffectiveResolution(3)) ' as provided in the JSON-sidecar'])
```
HenkMutsaerts: But AcquisitionVoxelSize never has the effective resolution, but rather the voxel size? Only for 2D EPI they are equal. So if we know this parameter, we still need to do step 2 (estimating effective resolution)

Modules/Module_Processing/xASL_init_DefaultEffectiveResolution.m
```diff
@@ -51,50 +46,91 @@
...
+elseif x.MagneticFieldStrength > 3
+	% For more than 3T scanners, we don't have this estimated as the resolutions are much higher
```
HenkMutsaerts: I don't get this. There are 3T scans with higher spatial resolution, and 7T scans with lower spatial resolution (though generally, with 7T there is higher spatial resolution, but doesn't really need to be the case). 

Or do you mean that higher Tesla gives different blurring? Then still, it would be safer to assume any blurring than to assume no blurring...

Modules/Module_Processing/xASL_init_DefaultEffectiveResolution.m
```diff
@@ -51,50 +46,91 @@
...
+	end
 
-%% ----------------------------------------------------------------------------------------
-%% 3) Calculate and report effective spatial resolution
-EffectiveResolution = NativeResolution.*Estimated_FWHM;
-fprintf('%s\n',[x.Q.PulseSequenceType ' NIfTI has native resolution ' num2str(NativeResolution(1)) ' ' num2str(NativeResolution(2)) ' ' num2str(NativeResolution(3)) ', assuming PSF ' num2str(Estimated_FWHM(1)) ' ' num2str(Estimated_FWHM(2)) ' ' num2str(Estimated_FWHM(3)) ' this gives estimated effective resolution ' num2str(EffectiveResolution(1)) ' ' num2str(EffectiveResolution(1)) ' ' num2str(EffectiveResolution(3))])
+	%% ----------------------------------------------------------------------------------------
+	%% 2) Attempt accounting for in-plane interpolation in reconstruction
+	if strcmpi(x.Q.PulseSequenceType, 'spiral') && strcmpi(x.Q.MRAcquisitionType, '3D')
+		% GE tends to upsample their spiral acquisitions to 1.6-1.9mm voxels
```
HenkMutsaerts: In my experience (product sequence), GE tends to double the resolution (going from ~3.8 to ~1.9mm)

HenkMutsaerts (changes requested):

Modules/Module_Processing/xASL_init_DefaultEffectiveResolution.m
jan-petr: clarified

Modules/Module_Processing/xASL_init_DefaultEffectiveResolution.m
jan-petr: Yes. Good that you caught this. AcquisitionVoxelSize indeed isn't effective. But we should still check this as it might provide the correct acquisition resolution in case somebody upsampled this during reconstruction. But we still need to calculate the effective resolution. This is fixed now

Modules/Module_Processing/xASL_init_DefaultEffectiveResolution.m
jan-petr: Yes. This was to skip this for 7T. But I agree to use similar assumptions as for 3T. Though we still need to consider 7T and their higher resolution of ASL and avoid the automatic correction of upsampled resolution during reconstruction. I have fixed the code.

Modules/Module_Processing/xASL_init_DefaultEffectiveResolution.m
jan-petr: Yes exactly. that's what I meant. I have fixed the comment.

jan-petr: For GRASE, it has recalculate the voxel size for anything below 3x3 to 3.8x3.8. This was on top of the effective resolution correction. I believe that this was fully wrong and I have no clue where is this comming from. I would do the same as for GE. If resolution is below 2mm inplane, then we expect 2 times upsampling during reconstruction. 

But taking 2.5 or 2.7mm inplane and setting it to 3.8 inplane and then still applying the effective resolution correction is very wrong.

HenkMutsaerts (approved):

Issue #1992, opened 2025-12-22
Additional ROI/Lesion


jan-petr:
### Description

If a users adds a new ROI/Lesion file (tied to an existing T1/FLAIR image) after the structural module was already executed, it will not transform it to the standard space and won't be able to run the population module on it. 

HENK: We could ensure that the ROI/lesion warping is a separate function within Structural module>Resampling submodule, which is then called again at the start of the population module if new ROIs/lesions are detected.

HENK: The problem is that you have Lesion_T1 and Lesion_FLAIR, which should be registered with T1_ORI and FLAIR_ORI before the structural module is run. The first steps of the structural module, are registering T1w (and others, including lesions) to MNI, and then FLAIR to T1w. The same for ROIs. This step then would also need to be redone, making it tricky.

JAN: Fully agree with the proposed solution and the issues as well. See if you agree with the brake-down to smaller tasks.

HENK2: To start with: the user should use correct nomenclature. `ROI` = user-provided masks used in the population module. `Lesion` is the same but also does cost-function masking. So by design, the user cannot add extra lesions while wanting to skip rerunning the Structural module. But the user can add extra ROIs.

So to keep it general/simple/modular, I would completely move the ROI ->standard space warping steps from the structural module to the population module, and not touch the lesion code. This way, this is nicely kept separate. (otherwise, your defined tasks below look OK)

### Tasks

- [ ] Based on existing code, write a specific function that handles all registration steps to ROI/Lesion that are done between T1_ORI-T1. 
- [ ] Based on existing code, write a specific function that transforms Lesions/ROI to standard space.
- [ ] Replace Lesion/ROI registration and spatial normalization steps with the above functions in the structural module.
- [ ] Replace Lesion/ROI registration steps for native space with dedicated functions in the ASL/population module.
- [ ] In population module, go through all lesions/ROIs in standard space. Check if they were transformed already into standard space. If not, rerun the functions above.

Doing that - we will have all the functionality on single space (robust to modifications and no double code), but called from two different places. But if the structural module is re-run and registration modified, it will already recalculate it for all lesions as well, so there's no risk that there are non-updated Lesions in standard space. And all missing lesions will be supplied during the population module.

### How to test

Optional: insert description about how to test the code changes here

### Release notes

Required: summarize the changes for the release notes here

PR #2027, opened 2026-02-27
#2024 Transform T1 to native space

### LINKED ISSUE #2024: Fix several issues with Lesions and ROIs
### Description

Make sure that T1 image is transformed to ASL native space (along with pGM) and not deleted later as we need it for QCing lesions etc.

### Tasks
- [x] Save T1-PV in native space 8c96fc57a40ed37fd5c7db3395cd8ae0349ce09d
- [x] T1-PV in native space is only saved for subjects with Lesions and ROIs fdbd8bd6
- [x] write a function that resets Lesion/ROI orientation. If MAT0 and size of T1/FLAIR and Lesions corresponds, but MAT differs, then we throw a warning and update Lesions/ROIs 1a93420e
- [x] xASL_wrp_LST_Segment_FLAIR_WMH.m, line 189 - follow-up of #1439 - check not only size, but MAT before resampling.  13a29f1d
- [x] Fix #2023 line 435 of the structural module 13a29f1d

- [x] pCSF needs to be kept in MNI space for the population module which now uses it. We now clean all r1T1, rc1T1 etc, so this is nice, but we need to keep all PVs just in case the user needs it (our new calculations). Double check if this works now

### Release notes

Fix issues with Lesions and ROIs.

jan-petr:
### Linked issue

Close #2024


HenkMutsaerts (changes requested):
I have no detailed change requests, but I wonder if we should generate a new file for everyone, while we only need it for a very specific use case (only those wanting to review lesions, and then out of that group only those that want this in native space). So 99% of the users will use 1 Mb more data to help 1% of the users.

So that warrants an if-statement right? We already have an if statement for WMH_SEGM, which is used much more frequently.

jan-petr:
> I have no detailed change requests, but I wonder if we should generate a new file for everyone, while we only need it for a very specific use case (only those wanting to review lesions, and then out of that group only those that want this in native space). So 99% of the users will use 1 Mb more data to help 1% of the users.
> 
> So that warrants an if-statement right? We already have an if statement for WMH_SEGM, which is used much more frequently.

Good point. Though it's rather insignificant compared to all the 25MB files in the population module. 
The question is - what if-statement you want? Do you want to have an option to turn this on/off in dataPar.json. I'm fine with that

HenkMutsaerts (changes requested):
I would make it dependent on having a Lesion/ROI (or at least the use cases where it is needed), and delete it at the end

Functions/Administration/xASL_adm_AlignLesionROItoStructural.m
```diff
@@ -0,0 +1,62 @@
...
+%
+% FORMAT: xASL_adm_AlignLesionROItoStructural(Lesion_ROI_list)
+%
+% INPUT:
+%   Lesion_ROI_list - list of Lesions and ROIs (REQUIRED)
+%
+% OUTPUT: n/a
+%
+% -----------------------------------------------------------------------------------------------------------------------------------------------------
+% DESCRIPTION: Go through all the Lesions and ROIs in the list. Check the corresponding FLAR/T1 files. If the MAT of both files are similar, then do not do anything.
```
HenkMutsaerts: FLAIR not FLAR

Functions/Administration/xASL_adm_AlignLesionROItoStructural.m
```diff
@@ -0,0 +1,62 @@
...
+% if they differ, but the matrix size and MAT0 are equal, then reset the MAT for the Lesion or ROI and save again
+%
+% EXAMPLE: xASL_adm_AlignLesionROItoStructural(Lesion_ROI_list)
+% -----------------------------------------------------------------------------------------------------------------------------------------------------
+% __________________________________
+% SPDX-License-Identifier: Apache-2.0
+% https://github.com/ExploreASL/ExploreASL/blob/main/LICENSE
```
HenkMutsaerts, line 19: These two lines will be deleted (in the other issue)

Functions/Administration/xASL_adm_AlignLesionROItoStructural.m
```diff
@@ -0,0 +1,62 @@
...
+	[startIndex,endIndex] = regexp(fName, '(T1|FLAIR)'); % Extract the name of the structural file
+	fStructName = fullfile(fPath,[fName(startIndex:endIndex) fExt]);
+
+	if xASL_exist(fStructName, 'file')
+		% Load NIfTI header of both files
+		lesionHeader = xASL_io_ReadNifti(Lesion_ROI_list{iS});
+		structHeader = xASL_io_ReadNifti(fStructName);
+
+		% Check if the MAT are OK
+		if ~all(abs(lesionHeader.mat - structHeader.mat) < 1e-3, 'all')
```
HenkMutsaerts, line 40: This assumes that ROI_T1_1.nii etc have the same orientation as T1.nii, but this doesn't have to be true right? As long as the voxel-size is identical, the ROI or lesion could be aligned with T1.nii while being eg shifted with 3 voxels and the orientation matrix having the same translation in real-world dimension, in the opposite direction?

Functions/Administration/xASL_adm_AlignLesionROItoStructural.m
```diff
@@ -0,0 +1,62 @@
...
+			% The transformation matrices differ, we have to fix this
+			
+			if ~isequal(size(lesionHeader.dat), size(structHeader.dat))
+				% Option 2 - sizes differ, we report a difference that we cannot fix
+				warning('%s\n%s\n%s', 'The transformation matrix MAT of Lesion/ROI and the corresponding structural files differ, and matrix sizes differ as well, so I cannot reset the orientation. Please check the files:', Lesion_ROI_list{iS}, fStructName);
```
HenkMutsaerts: change comment into "and image matrix sizes (resolution vs field-of-view) differ as well"

or something similar, to explain to the user (not everybody understands that image==matrix)

Functions/ImageProcessing/xASL_im_Modulation.m
```diff
@@ -51,6 +51,8 @@ function xASL_im_Modulation(x)
 %tNii.dat(:,:,:,:,:)  = single(ones(tNii.dat.dim));
 
 % Run the SPM script for creating the determinants
+xASL_delete(x.P.Path_j_T1);
```
HenkMutsaerts, line 52: Are you not sure that we should remove this only at the end of the code? (you can resolve)

Modules/Module_Processing/Module_ASL/xASL_wrp_PreparePV.m
```diff
@@ -56,6 +56,14 @@
     x.P.Path_despiked_ASL4D = x.P.Path_ASL4D;
 end
 
+% Check if there are Lesions or ROIs saved for the structural images. This then influences the code below for deciding which files to save
+bLesion_ROI = xASL_adm_GetFileList(x.dir.SUBJECTDIR, ['(?i)^(Lesion|ROI)_(' x.P.STRUCT '|' x.P.FLAIR ')_\d*\.nii$'], 'FPList', [0 Inf]);
+if isempty(bLesion_ROI)
+	bLesion_ROI = 0;
```
HenkMutsaerts: false and true, not 0 and 1?

Functions/Administration/xASL_adm_LesionResliceList.m
```diff
@@ -1,5 +1,5 @@
-function [INname, OUTname] = xASL_wrp_LesionResliceList(x,bLesion_T1,bLesion_FLAIR,bROI_T1,bROI_FLAIR)
-%xASL_wrp_LesionResliceList Creates list of structural image paths to
+function [INname, OUTname] = xASL_adm_LesionResliceList(x,bLesion_T1,bLesion_FLAIR,bROI_T1,bROI_FLAIR)
```
HenkMutsaerts, line 1: Where is this function run? I would check that the orientation matches only once in the whole pipeline

Modules/Module_Processing/Module_Structural/xASL_wrp_LinearReg_T1w2MNI.m
```diff
@@ -61,7 +61,10 @@ function xASL_wrp_LinearReg_T1w2MNI(x, bAutoACPC)
 for iS=1:length(Lesion_ROI_list)
     OtherList{end+1,1} = Lesion_ROI_list{iS};
 end
-    
+xASL_adm_AlignLesionROItoStructural(Lesion_ROI_list);
```
HenkMutsaerts: Why here as well? And why only the ROI list?

Shouldn't we run this function only once in the whole pipeline?

Modules/Module_Processing/Module_Structural/xASL_wrp_LST_Segment_FLAIR_WMH.m
```diff
@@ -186,7 +188,7 @@ function xASL_wrp_LST_Segment_FLAIR_WMH(x, rWMHPath, WMHsegmAlg)
 				end
 
 				% Resample if needed
-				if ~isequal(size(LesionIM), size(WMHim))
+				if ~all(abs(LesionHeader.mat - WMHheader.mat) < 1e-3, 'all') || ~isequal(size(LesionIM), size(WMHim))
```
HenkMutsaerts, line 181: Resampling doesn't fix differences in orientation (misalignment), it only fixes differences in matrix size. So your first argument here should then only be the diagonal of the orientation matrix, right?

Modules/Module_Processing/xASL_init_FileSystem.m
```diff
@@ -23,7 +23,7 @@
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
 % EXAMPLE: x = xASL_init_FileSystem(x);
 % __________________________________
-% Copyright 2015-2024 ExploreASL
+% Copyright 2015-2026 ExploreASL
```
HenkMutsaerts: Line will be removed

Modules/xASL_module_Structural.m
```diff
@@ -432,8 +432,8 @@
 %% 8    Resample to standard space
 iState = 8;
 FileMissing = false; % Check availability NIfTIs in standard space, else redo warping
-List2CheckNativeSpace = {x.P.Path_T1    , x.P.Path_T1      , x.P.Path_T1,        x.P.Path_FLAIR,     x.P.Path_T2,     x.P.Path_T1c};   % Source path
-List2Check =            {x.P.Pop_Path_rT1 x.P.Pop_Path_rc1T1 x.P.Pop_Path_rc2T1, x.P.Pop_Path_rFLAIR x.P.Pop_Path_rT2 x.P.Pop_Path_rT1c}; % Population path
+List2CheckNativeSpace = {x.P.Path_T1     , x.P.Path_T1,       x.P.Path_T2,     x.P.Path_T1c};   % Source path
```
HenkMutsaerts: These lists don't match?

HenkMutsaerts (changes requested):
There is a reason that I didn't program this, in the other issue, where I rather checked carefully that we always realign all dependent NIfTI files equally.

This is just to tricky, it assumes that T1 and ROI_T1_1 should have identical orientation matrices when ROI_T1_1 was created, and that ExploreASL is the only program that could change it.

BTW: there are more dependent NIfTIs than only ROIs and Lesions (e.g. M0 is dependent on ASL4D for orientation, FLAIR from T1, etc); do we now have a different behavior for ROIs and Lesions than the others?

jan-petr:
> There is a reason that I didn't program this, in the other issue, where I rather checked carefully that we always realign all dependent NIfTI files equally.
> 
> This is just to tricky, it assumes that T1 and ROI_T1_1 should have identical orientation matrices when ROI_T1_1 was created, and that ExploreASL is the only program that could change it.
> 
> BTW: there are more dependent NIfTIs than only ROIs and Lesions (e.g. M0 is dependent on ASL4D for orientation, FLAIR from T1, etc); do we now have a different behavior for ROIs and Lesions than the others?

This is a misunderstanding. We of course treat all files the same, we don't make exceptions and we do things carefully. This is part of the ongoing discussion of what happens when we want to add new ROIs to an already processed pipeline - this is something that can happen rather often in practice.

1. We do not do this for M0 or ASL4D, because these do not change after processing and they should be always correctly aligned.
2. ROIs and Lesions are of course always correctly realigned with the corresponding structural file at all times.
3. The reason why we need this checking is that when we want to add ROIs and Lesions additionally after the structural module was already executed. As discussed in #1992 . When this happens, ROIs have incorrect MAT and the registration fails. So this fixes it.
4. Do not forget that ROIs are supposed by definition to be aligned to ROI. We require them to be created with the same matrix and same orientation as the T1 file - otherwise, this would have been a mess. 
5. So the only thing that my fix does is verifying -> does the ROI has the exact matrix as T1, does the ROI has the exact MAT0 as T1 -> then obviously, it needs to have the same MAT as T1. Because ROIs are never coregistered with T1 (that is impossible) as they are supposed to be aligned from the start. So my fix only fixes the issue of not having the ROI properly synced because it was added later. But it doesn't do anything to a ROI that wasn't aligned at the start or that has a different matrix or diffrent MAT0.

To sum up -> this fix helps to fix issues without having any disadvantages.
If the ROI has a different MAT0, different matrix etc -> it leaves it alone.
If the ROI was provided aligned to T1 with the same matrix and M0 -> it syncs MAT in case it wasn't synced.

I do not see any scenario where this could go wrong.

Modules/Module_Processing/xASL_init_FileSystem.m
jan-petr: Yes. This conflict will be resolved later.

Modules/xASL_module_Structural.m
jan-petr: They do - check the final version, not the difference.

Functions/Administration/xASL_adm_AlignLesionROItoStructural.m
jan-petr: fixed

Functions/Administration/xASL_adm_AlignLesionROItoStructural.m
jan-petr: See the other discussion. By removing these lines, we loose provenance.

Functions/Administration/xASL_adm_AlignLesionROItoStructural.m
jan-petr: See my longer comment below. The assumption (that cannot ever be broken) is that ROI_T1 and T1 are provided aligned at the start. Then every transformation to T1 is applied to ROI_T1 as well.

So ROI can be slightly shifted in the matrix with respect to T1. And MAT of ROI will compensate for that. But if that would be the case, the MAT0 of ROI and MAT0 of T1 would also differ. So if MAT0 of ROI and T1 are equivalent, then the same needs to apply to MAT.

That's why - in case MAT of ROI and T1 are misaligned, I report a warning, but I only fix MAT when the original MAT0 of ROI and T1 are the same. But of course I do not do anything if the MAT0 are equal. This fixes the most likely case of ROI being created and saved on T1 in the same space, but added only after T1 was realigned to AC-PC.

So you are right, but my code avoid fixing the described scenario.

Functions/Administration/xASL_adm_AlignLesionROItoStructural.m
jan-petr: done

Functions/ImageProcessing/xASL_im_Modulation.m
jan-petr: Very sure. SPM creates j_TI.nii based on y_T1.nii. 
So when re-runing a pipeline, j_T1.nii.gz exists and then j_T1.nii is created (because SPM doesn't check for j_T1.nii.gz existence), which creates a conflict.

So I added this line so that before j_T1.nii is created, any existing version is removed.

It's a critical fix that was crashing the pipeline.

Modules/Module_Processing/Module_ASL/xASL_wrp_PreparePV.m
jan-petr: fixed

Functions/Administration/xASL_adm_LesionResliceList.m
jan-petr: I have no clue. You created this function for the QC. But the function name did not match filename. So I fixed that.

Modules/Module_Processing/Module_Structural/xASL_wrp_LinearReg_T1w2MNI.m
jan-petr: Only ROI -> see my answer below. We only do this for ROI, because they are manually inserted by the user and not copied automatically from rawdata. This creates possible issues, that I'm trying to fix.

In theory, somebody can add ROIs and then remove lock-files at random for re-runing. So that's why I prefer this check to be run more often (takes minimal time). But I'm open to discussion.

Modules/Module_Processing/Module_Structural/xASL_wrp_LST_Segment_FLAIR_WMH.m
jan-petr: We have agreed to do this in #1439.

If we want to read image data that correspond voxel by voxel between two images, then we can only do this when the sizes are equal and the transformation matrices are equal. 
If MAT or size differs, there isn't a one-by-one correspondence in the voxels.

So if the goal is that the two images are aligned in world-coordinates, we don't need any resampling. If we want them to be aligned in native space -> then both the orientation matrix and image size need to be equivalent.

Functions/Administration/xASL_adm_AlignLesionROItoStructural.m
HenkMutsaerts: I cannot find a longer comment below?

I don't think it is true that mat0 and mat are always that related. This is the design, and how SPM treats it. But if somebody treats this wrong (as I did with WMH, as we recently fixed in #1439), then mat0 are equal but mat is not, or vice versa. I mean that I generate a new NIfTI, based on the T1 (so incorrectly copying T1's mat0).

So:
Use case 1: fresh data from rawdata, the user has provided ROI/Lesion that is aligned with T1/FLAIR
Use case 2: the user reruns the pipeline
Use case 3: the user adds ROI/Lesion during/after the pipeline run

How our workflow is now (at least my philosophy for #1439), is that we always keep all dependent NIfTIs aligned, with any orientation matrix change we do in the pipeline. And we never reset to mat0, because we assume it is always aligned.

Perhaps you are fixing backward compatibility, e.g. for David's lesions having processed stuff BEFORE #1439? 

The point of a develop/beta version, is that it doesn't have to be stable; having to rerun data is not important. It is more important that the release is stable. 

So we should avoid adding something that fixes backward compatibility (previous code error), that potentially backfires.

Modules/Module_Processing/Module_Structural/xASL_wrp_LST_Segment_FLAIR_WMH.m
HenkMutsaerts: Good point

HenkMutsaerts (changes requested):
3 minor questions left

HenkMutsaerts (changes requested):
1 last point

Functions/Administration/xASL_adm_AlignLesionROItoStructural.m
jan-petr: I agree about "not fixing backwards compatibility". But that's not the case. The process for keeping lesions aligned was working correctly already before. So we are not fixing Use case 1 or 2 - these worked before. We are providing solution for use case 3.

I agree with the Philosophy and we are in line with it.

My fix helps to resolve the prospective error of "Use case 3", while it will detect "Use case 1 and 2" and will not do anything.

HenkMutsaerts: BTW: only now I read your "long comment/answer" below, for some reason it wasn't attached to code (so I can also not answer it, so I'll answer it here).

Do I understand it correctly, that you're only checking the situation where mat0 is equal between ROI and T1 but mat is not?

I agree that this is likely incorrect, but 1) we cannot assume that is always incorrect, and 2) we cannot assume that mat0 was correct (see my last argument:

I don't think it is true that mat0 and mat are always related. This is the design, and how SPM treats it. But if somebody treats this wrong (as I did with WMH, as we recently fixed in https://github.com/ExploreASL/ExploreASL/issues/1439), then mat0 are equal but mat is not, or vice versa. I mean that I generate a new NIfTI, based on the T1 (so incorrectly copying T1's mat0).

So -> if mat0 are indentical between T1 and ROI_T1, and mat is not, I agree in issuing a warning or error, I just do not agree automatically reverting mat to mat0.

Modules/Module_Processing/Module_Structural/xASL_wrp_LinearReg_T1w2MNI.m
HenkMutsaerts: I would still only do this once in the pipeline, so then I would do this later in the structural module, only once?

Functions/Administration/xASL_adm_AlignLesionROItoStructural.m
jan-petr: HENK: Do I understand it correctly, that you're only checking the situation where mat0 is equal between ROI and T1 but mat is not?
JAN: Yes. Exactly!!

Yes - but in WMH, you have messed up. That was not a normal situation. We have to assume that when user provides us data, it's not messed up. 

In correct files, mat0 and mat are always related. Unless you mess up. But we need a predictable behavior, we can't count with people messing up.

You do the exact same thing for WMH - you're saving to mat, then changing mat0 to T1 mat0 and then using this to control if WMH is T1 or FLAIR based. 
So you are arguing differently for your solution and for my solution. Think about this - this isn't consistent.

Modules/Module_Processing/Module_Structural/xASL_wrp_LinearReg_T1w2MNI.m
jan-petr: OK. I'm fine with this and can do this only just before the final transformation. OK?

Functions/Administration/xASL_adm_AlignLesionROItoStructural.m
jan-petr: Additionally as discussed on Whatsapp. We can fix MAT0. Under very clearly defined conditions -> everything matching, but a MAT difference. 

There has to be a clear signal that this happened. Henk proposes to end with an error to inform the user. Jan thinks that user might not spot an error and will do the re-run, which would then work. It will thus be a negligible improvement over warning only. Is there something more visible to do? I'm open to more solutions - just want to pick the best one.

Modules/Module_Processing/Module_Structural/xASL_wrp_LinearReg_T1w2MNI.m
jan-petr: Done. It's now only done once at the end.

Functions/Administration/xASL_adm_AlignLesionROItoStructural.m
HenkMutsaerts: Like I said, that mat and mat0 are related depends on the software, and if the user creates a new NIfTI manually, then there is the opportunity to do this wrongly. But you're also right that this may occur less frequently than the case where you invented this for.

So happy to fix this and then issue an error, with this warning. The error will crash the processing, and the user will check why the processing was crashed.

If we catch errors subject-wise, and the processing continues with the next subject, we would need an error and warning summary at the end, which we don't have yet (would be nice to have, not sure if we have this created as issue yet?) but for now, crashing the processing is the best we can do if there is something potentially harmful.

The text should be very clear to the user:

fprintf(  "\n\n\n Note that mat... differed but mat0 did not, so it seems as if the alignment between XXX and YYY is wrong, but was initially (==mat0) correct. We have now set the orientations (==mat) back to the original orientations (==mat0) for both NIfTIs. Ensure to check if these are now still aligned!!! \n\n\n"

And then 

error('Misalignment noticed, tried to fix this, needs a user-check before rerunning!!!');

jan-petr: Done

HenkMutsaerts (approved):

PR #2028, opened 2026-03-02
Bug #2025 fix test data set

### LINKED ISSUE #2025: Fix aggregation script for TestDataset
### Description

When running the testdataset10 test, it returns no CBF - likely the script for aggregating is broken.
Fix also other minor details

### Tasks

- [x] Fix GM CBF and CoV values are reported.
- [x] Run test-dataset on main to get the reference data
- [x] GE_3Dspiral_T1prefilled_Cyst_M0_noFLAIR_2ndRun_LoQ - fixed by updating dataPar.json
- [x] Siemens3DGRASE_Structural_AlreadyDone_HiQ - fixed by updating dataPar.json
- [x] Fix WM CBF in Philips_2DEPI_FLAIR_Segm_PWI4D_Quant_Parameters_LoQ, Philips_2DEPI_Lesion_NoFLAIR_NoM0_HiQ
- [x] Rename to 2.0.0_beta
- [x] Rename tables to 2.0.0_beta once we fix bugs

### Release notes
Update TestDataSet-11 reference to new PV-threshold results.

jan-petr:
### Linked issue

Close #2025


HenkMutsaerts (changes requested):
Did you check if, before doing the pv>0.7 change, the test values were sufficiently close to the old reference values? Because otherwise, it is possible that we affected the values without knowing, with an earlier code change (we usually only test this with a new release right?)

Alternatively, you can validate that the current and old reference values are not too far apart (if I look at them it looks OK, but it makes sense to validate this formally; otherwise, what is the point of doing this testing :)

jan-petr:
> Did you check if, before doing the pv>0.7 change, the test values were sufficiently close to the old reference values? Because otherwise, it is possible that we affected the values without knowing, with an earlier code change (we usually only test this with a new release right?)
> 
> Alternatively, you can validate that the current and old reference values are not too far apart (if I look at them it looks OK, but it makes sense to validate this formally; otherwise, what is the point of doing this testing :)

Yes I did. 3 datasets created very large changes, so I've re-run them with 0.7 and 0.5 activated and the 0.5 was <1% close to the previous version. so it's only the threshold difference.

So for most datasets, I changed the values and updated the reference. For 2 datasets, I changed dataPar to PV=0.5 to keep the result similar to what it was.

HenkMutsaerts (approved):

PR #2030, opened 2026-03-03
#1833 DEBBIE motion correction

### LINKED ISSUE #1833: Motion correction for multi-PLD and multi-TE goes wrong
### Description

Similarly to control vs label images, images acquired at different PLD or TE will differ because of different TR after saturation, different amount of labeling inside, or different contrast because of different TE.
So if we perform motion correction, we need to be sure that it correct for motion and doesn't try to overcompensate for expected signal differences.

In DEBBIE sequence, This is mostly a problem for different TEs: they produce very different contrast leading to artificial estimated motion. For Time-encoding -> this doesn't generate large false motion. See images below for HAD4 and HAD8

image
This image shows for control and 3 PLDs and 8 TEs (DEBBIE FME HAD4, here still Hadamard encoded) that the assumed position change based on intensity-based alignment incorrectly thinks of pseudo-motion due to echo time changes.

Below the same for DEBBIE FME HAD8: single TE 7 PLDs & 1 control, two repetitions
image

### Tasks

- [x] During motion correction, identify multi-TE acquisition, time-encoded acqusition, multi-PLD without time-encoded, or standard acquisition.
- [x] For multi-TE: Perform motion correction only on the shortest TE. Turn of Zig-zag. Register to first image. Apply estimated motion in shortest TE for the other TEs within the same block.
- [x] For Time-encoded: Turn off Zig-zag. Register to first image. Perform normal motion correction as the images have the same TR.
- [x] For simple multi-PLD (without multi-TE and without time-encoding): Use Zig-zag. Register to first. For each PLD - register the images to the first for that PLD. Do not register across different PLDs
- [x] For standard acquisition. Use Zig-zag, register to mean. The standard way.
- [x] For multi-TE and multi-PLD - make sure the estimated parameters are put together correctly and saved to TXT and NII files as if a normal  motion correction would be executed. That means -> we only change the motion estimation and don't have to change any of the other functions.
- [x] Functions for calculating mean motion and plotting motion are moved to a sub-function for readability.

### How to test
Corrected version - see how motion doesn't change across TEs, but only between blocks.
Image

### Release notes

Fix multi-TE motion correction.

MimiE447:
### Linked issue

Close #1833

### DIFF
--- a/External/SPMmodified/spm_realign.m
+++ b/External/SPMmodified/spm_realign.m
@@ -114,7 +114,9 @@ end
 %--------------------------------------------------------------------------
 def_flags          = spm_get_defaults('realign.estimate'); % Loads SPM's default realignment settings.
 def_flags.PW       = '';
-if bZigzag % Zig-zag modifies defaults — zig-zag regression assumes a fixed reference image, not a moving mean.
+if bZigzag 
+	% bZigZag is an option introduced for ASL-MRI, because the consistent control-label intensity differences in ASL can create a false zig-zag motion pattern.
+	% To avoid this false motion estimation, bZigZag regresses out consistent intensity differences between odd and even repetitions that are attirbute to control-label intensity difference and not due to motion
 	def_flags.graphics = 1; % Graphics on
 	def_flags.quality  = 1; % Highest quality
 	def_flags.rtm      = 0; % Disables rference to mean
@@ -165,10 +167,9 @@ if numel(P)==1 % Single session
         P{1}=cleanandsave_parameters(P{1},ref); % This is where zig-zag regression happens
     end 
 elseif numel(P) > 1
-      %  Multiple sessions
-      %  This block:
-      %  Aligns the first volume of each session
-      %  Applies that transform to all volumes in each session
+      %  Multiple sessions:
+      %  Aligns the first volumes of each session together. 
+      %  Applies the same transformation of the first volume to all remaining volumes in each session
       %  Runs realignment within each session
       %  Applies zig-zag cleaning per session
     Ptmp = P{1}(1);
@@ -627,9 +628,9 @@ Vo=V;
 Q = zeros(n,6);
 clQ=zeros(n,6);
 for j=1:n
-	qq     = spm_imatrix(V(j).mat/V(1).mat); % Extract motion parameters
-	Q(j,:) = qq(1:6); % original motion (6 columns)
-    clQ(j,:)=qq(1:6); % copy to be cleaned
+	qq     = spm_imatrix(V(j).mat/V(1).mat); % Extract motion parameters from the MAT of NII. Relative motion compared to the first frame
+	Q(j,:) = qq(1:6); % Create a 6 column motion matrix
+    clQ(j,:)=qq(1:6); % Motion matrix for later cleaning
 end
 for j=1:6
     refval=clQ(:,j);
--- a/External/SPMmodified/spm_realign.m
+++ b/External/SPMmodified/spm_realign.m
@@ -112,18 +112,18 @@ end
 
 %-Flags
 %--------------------------------------------------------------------------
-def_flags          = spm_get_defaults('realign.estimate');
+def_flags          = spm_get_defaults('realign.estimate'); % Loads SPM's default realignment settings.
 def_flags.PW       = '';
-if bZigzag
-	def_flags.graphics = 1;
-	def_flags.quality  = 1;
-	def_flags.rtm      = 0;
-else
+if bZigzag % Zig-zag modifies defaults — zig-zag regression assumes a fixed reference image, not a moving mean.
+	def_flags.graphics = 1; % Graphics on
+	def_flags.quality  = 1; % Highest quality
+	def_flags.rtm      = 0; % Disables rference to mean
+else 
 	def_flags.graphics = ~spm('CmdLine');
 end
 def_flags.lkp      = 1:6;
 
-if nargin < 2
+if nargin < 2 % Merge user flags with defaults - Any missing fields in flags are filled with defaults.
     flags = def_flags;
 else
     fnms = fieldnames(def_flags);
@@ -144,26 +144,33 @@ end
 
 %-Images
 %--------------------------------------------------------------------------
-if ~iscell(P), P = {P}; end
-for i=1:numel(P), if ischar(P{i}), P{i} = spm_vol(P{i}); end; end
-P(cellfun(@isempty,P)) = [];
+if ~iscell(P), P = {P}; end % Always treat P as a cell array (one cell per session).
+for i=1:numel(P), if ischar(P{i}), P{i} = spm_vol(P{i}); end; end % Convert filenames → spm_vol structs.
+P(cellfun(@isempty,P)) = []; % Remove empty sessions.
 if ~isempty(flags.PW) && ischar(flags.PW), flags.PW = spm_vol(flags.PW); end
 
 if isempty(P), warning('Nothing to do.'); end
 
 %-Perform realignment
 %==========================================================================
-if numel(P)==1
-    P{1} = realign_series(P{1}, flags);
-	if bZigzag
-        if nargout==0, save_parameters(P{1}); end
+if numel(P)==1 % Single session
+    P{1} = realign_series(P{1}, flags); % Perform rigid-body realignment against the first image.
+    if bZigzag
+        if nargout==0, save_parameters(P{1}); end % Save raw motion parameters first.
         n = length(P{1});
         ref=-ones(n,1);           %% ZW  should be switched later to allow different label control order, now assume first image is label
         ref(2:2:end)=1;
         if(ctrfirst), ref=-ref; end
-        P{1}=cleanandsave_parameters(P{1},ref);
-	end
+        % Builds zig-zag reference vector
+        P{1}=cleanandsave_parameters(P{1},ref); % This is where zig-zag regression happens
+    end 
 elseif numel(P) > 1
+      %  Multiple sessions
+      %  This block:
+      %  Aligns the first volume of each session
+      %  Applies that transform to all volumes in each session
+      %  Runs realignment within each session
+      %  Applies zig-zag cleaning per session
     Ptmp = P{1}(1);
     for s=2:numel(P)
         Ptmp = [Ptmp ; P{s}(1)];
@@ -192,7 +199,7 @@ end
 %-Save results
 %==========================================================================
 
-if ~nargout
+if ~nargout % If the function is called without output:
     for s=1:numel(P)
         %-Save parameters as rp_*.txt files
         %------------------------------------------------------------------
@@ -200,7 +207,7 @@ if ~nargout
 			save_parameters(P{s});
 		end
 			
-        %-Update voxel to world mapping in images header
+        %-Update voxel to world mapping in images header - Update NIfTI headers
         %------------------------------------------------------------------
         for i=1:numel(P{s})
             spm_get_space([P{s}(i).fname ',' num2str(P{s}(i).n)], P{s}(i).mat);
@@ -230,6 +237,21 @@ function P = realign_series(P,flags)
 % The scaling (and offset) parameters are also set to contain the
 % optimum scaling required to match the images.
 %__________________________________________________________________________
+
+% This is pure SPM and unchanged by zig-zag.
+% At a high level, it:
+    % Samples voxels from the reference image
+    % Smooths images
+    % Computes image gradients
+    % Solves a least-squares rigid-body registration
+    % Iterates until convergence
+
+% x1,x2,x3 → sampled voxel coordinates
+% G → intensity values
+% dG1,dG2,dG3 → spatial gradients
+% A → Jacobian of intensity wrt motion parameters
+% Solve (AᵀA)⁻¹Aᵀb for motion updates
+% This is classic Friston 1995 rigid-body registration.
 
 if numel(P)<2, return; end
 
@@ -605,16 +627,17 @@ Vo=V;
 Q = zeros(n,6);
 clQ=zeros(n,6);
 for j=1:n
-	qq     = spm_imatrix(V(j).mat/V(1).mat);
-	Q(j,:) = qq(1:6);
-    clQ(j,:)=qq(1:6);
+	qq     = spm_imatrix(V(j).mat/V(1).mat); % Extract motion parameters
+	Q(j,:) = qq(1:6); % original motion (6 columns)
+    clQ(j,:)=qq(1:6); % copy to be cleaned
 end
 for j=1:6
     refval=clQ(:,j);
-    clQ(:,j)=refval-ref/(ref'*ref)*ref'*refval;
-    clQ(:,j)=clQ(:,j)-clQ(1,j);
+    clQ(:,j)=refval-ref/(ref'*ref)*ref'*refval; % This is linear regression removal.
+    clQ(:,j)=clQ(:,j)-clQ(1,j); % Re-reference to first volume. Ensures: First image has zero motion and that motion remains relative to first volume
 end
 for j=1:n
+    % Rebuild transformation matrices - This replaces the original motion with the cleaned motion.
     nmat = spm_matrix(clQ(j,:));
     Vo(j).mat=nmat*V(1).mat;
 end
--- a/Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
+++ b/Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
@@ -78,26 +78,31 @@ minVoxelSize = double(min(tempnii.hdr.pixdim(2:4)));
 %% Define motion correction options
 % bMoCoPossible boolean states if it is possible to perform motion correction with the given data 
 % x.asl.module.motionCorrection states if the motion correction is wanted by the user
-if x.modules.asl.bContainsSubtracted
+if isfield(x.Q,'LookLocker') && x.Q.LookLocker
+	bMoCoPossible = false;
+	fprintf('%s\n',['Skipping motion correction for ' x.P.SubjectID '_' x.P.SessionID ' as Look-Locker correction is not implemented.']);
+elseif x.modules.asl.bContainsSubtracted
 	% Motion correction is disabled, potentially insufficient contrast
 	bMoCoPossible = false;
 	fprintf('%s\n',['Skipping motion correction for ' x.P.SubjectID '_' x.P.SessionID ' because it only has DeltaM volumes, which may contain insufficient contrast']);
 elseif nFramesPerTE > 1
+	% We only do motion correction when the number of frames is higher than one. For multi-TE, we count number of frames per TE.
 	bMoCoPossible = true;
 else
 	bMoCoPossible = false;
 	fprintf('%s\n',['Skipping motion correction for ' x.P.SubjectID '_' x.P.SessionID ' because it had only ' num2str(nFramesPerTE) ' 3D frames.']);
 end
 
-if isfield(x.Q,'LookLocker') && x.Q.LookLocker
-	bMoCoPossible = false;
-	fprintf('%s\n',['Skipping motion correction for ' x.P.SubjectID '_' x.P.SessionID ' as Look-Locker correction is not implemented.']);
-end
-
 % Set flags for multi-TE and multi-PLD datasets
-if x.Q.nUniqueInitial_PLD>1 && (~isfield(x.modules.asl, 'bTimeEncoded') || ~x.modules.asl.bTimeEncoded)
-	% Here, we consider only simple multiPLD for averaging across PLDs. TimeEncoded is treated as singlePLD - because there are usually not many repetitions
+
+if x.Q.nUniqueInitial_PLD>1
+	% Here, we consider only simple multiPLD for averaging across PLDs
     bMultiPLD = 1;
+
+	% Note that TimeEncoded is a special case and needs to be treated separately. Therefore, we make sure this flag is initialized
+	if ~isfield(x.modules.asl, 'bTimeEncoded')
+		x.modules.asl.bTimeEncoded = false;
+	end
 else
     bMultiPLD = 0;
 end
@@ -139,7 +144,7 @@ end
 % So, here we check if it exists, but otherwise we default to disabling it
 
 if bMultiPLD || bMultiTE
-    % outlier exclusion is temporarily disabled for multiPLD/TE
+    % outlier exclusion is temporarily disabled for multiPLD (including Time-encoded) or multi-TE
     % as we are still developing this feature
     fprintf('%s\n', 'multi-PLD or multi-TE detected, disabling outlier exclusion, not yet implemented');
     bSpikeRemoval = false;
@@ -183,6 +188,7 @@ end
 if bMultiTE || x.modules.asl.bTimeEncoded
     % ZigZag are temporarily disabled for multiTE and TimeEncoded
     % as we are still developing this feature
+	% Note that for standard multi-PLD, Zig-zag can be applied because standard multi-PLD still has controls and labels
     fprintf('%s\n', 'multi-PLD or multi-TE detected, disabling zig-zag motion estimation, not yet implemented');    
     bZigZag = false;
 
@@ -207,7 +213,11 @@ end
 if bMultiTE
     fprintf('SPM motion estimation in multi-TE data: Aligning shortest TEs only');
 elseif bMultiPLD
-    fprintf('SPM motion estimation in multi-PLD data: Aligning within PLDs only');
+	if x.modules.asl.bTimeEncoded
+		fprintf('SPM motion estimation in single-TE Time-encoded data: Aligning all')
+	else
+		fprintf('SPM motion estimation in multi-PLD data: Aligning within PLDs only');
+	end
 else
     fprintf('Standard SPM motion estimation');
 end
@@ -222,15 +232,11 @@ switch x.settings.Quality
 	case 1 % normal quality
 		flags.quality = 1;
 		flags.sep = minVoxelSize;
+		flags.rtm = 1; % realign to mean
 	case 0 % low quality for fast try-out
 		flags.quality = 0.01;
+		flags.rtm = 0; % disable realign to mean
 		flags.sep = minVoxelSize*2;
-end
-
-if bMultiTE || bMultiPLD 
-	flags.rtm = 0; % disable realign to mean
-else
-	flags.rtm = 1; % realign to mean
 end
 
 flags.interp = 1;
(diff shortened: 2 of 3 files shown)


External/SPMmodified/spm_realign.m
```diff
@@ -112,18 +112,18 @@
...
 %--------------------------------------------------------------------------
-def_flags          = spm_get_defaults('realign.estimate');
+def_flags          = spm_get_defaults('realign.estimate'); % Loads SPM's default realignment settings.
 def_flags.PW       = '';
-if bZigzag
-	def_flags.graphics = 1;
-	def_flags.quality  = 1;
-	def_flags.rtm      = 0;
-else
+if bZigzag % Zig-zag modifies defaults — zig-zag regression assumes a fixed reference image, not a moving mean.
```
HenkMutsaerts: To me, this is not a clear explanation of zig-zag. Consider replacing with something like:

bZigZag is called this way because of the consistent control-label intensity differences, which can create a zig-zag motion pattern. To avoid this, bZigZag regresses out consistent intensity differences that are not due to motion. Another example is intensity differences with different TEs; these should not be removed by motion correction.

External/SPMmodified/spm_realign.m
```diff
@@ -144,26 +144,33 @@
...
         ref(2:2:end)=1;
         if(ctrfirst), ref=-ref; end
-        P{1}=cleanandsave_parameters(P{1},ref);
-	end
+        % Builds zig-zag reference vector
+        P{1}=cleanandsave_parameters(P{1},ref); % This is where zig-zag regression happens
+    end 
 elseif numel(P) > 1
+      %  Multiple sessions
+      %  This block:
```
HenkMutsaerts: "block" means "block of code"? consider replacing by

% This code section:

External/SPMmodified/spm_realign.m
```diff
@@ -144,26 +144,33 @@
...
+      %  Aligns the first volume of each session
+      %  Applies that transform to all volumes in each session
+      %  Runs realignment within each session
+      %  Applies zig-zag cleaning per session
```
HenkMutsaerts, line 174: What does session mean here?

External/SPMmodified/spm_realign.m
```diff
@@ -605,16 +627,17 @@ function save_parameters(V)
 Q = zeros(n,6);
 clQ=zeros(n,6);
 for j=1:n
-	qq     = spm_imatrix(V(j).mat/V(1).mat);
-	Q(j,:) = qq(1:6);
-    clQ(j,:)=qq(1:6);
+	qq     = spm_imatrix(V(j).mat/V(1).mat); % Extract motion parameters
+	Q(j,:) = qq(1:6); % original motion (6 columns)
+    clQ(j,:)=qq(1:6); % copy to be cleaned
```
HenkMutsaerts: If all these comments are added, perhaps clarify which part adds the parameters (the rp*.txt file has both without and with zig-zag regression), and which part prints the with zig-zag regression parameters to the .mat sidecar

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
```diff
@@ -83,18 +80,31 @@ function xASL_wrp_RealignASL(x, bASL)
...
 end
 
 if isfield(x.Q,'LookLocker') && x.Q.LookLocker
 	bMoCoPossible = false;
 	fprintf('%s\n',['Skipping motion correction for ' x.P.SubjectID '_' x.P.SessionID ' as Look-Locker correction is not implemented.']);
 end
 
+% Set flags for multi-TE and multi-PLD datasets
+if x.Q.nUniqueInitial_PLD>1 && (~isfield(x.modules.asl, 'bTimeEncoded') || ~x.modules.asl.bTimeEncoded)
+	% Here, we consider only simple multiPLD for averaging across PLDs. TimeEncoded is treated as singlePLD - because there are usually not many repetitions
```
HenkMutsaerts: I don't get this. TimeEncoded could have many repetitions, but moreover the opposite: singlePLD can easily have many repetitions, so saying that TimeEncoded is treated as singlePLD because TimeEncoded has not many repetitions is confusing.

Furthermore, skipping outlier detection should done with Hadamard (this is checked based on bMultiPLD below; so this means that a single TE Hadamard will have bMultiPLD=0 and bMultiTE=0 and outlier detection/ENABLE will not be skipped)

Essentially/simply speaking; Hadamard is a subcategory under MultiPLD, so suggestion if bMultiPLD should include Hadamard (or change boolean definitions)?

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
```diff
@@ -83,18 +80,31 @@ function xASL_wrp_RealignASL(x, bASL)
...
+    bMultiPLD = 1;
+else
+    bMultiPLD = 0;
+end
+
+if x.Q.nUniqueEchoTime>1
```
HenkMutsaerts, line 110: What is the difference between bMultiTE and nFramesPerTE > 1?
Or can we use one of these parameters?

Perhaps defining bMultiPLD and bMultiTE earlier above, and use throughout

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
```diff
@@ -162,19 +172,19 @@ function xASL_wrp_RealignASL(x, bASL)
...
-if nFrames < 10 % Only execute ENABLE if we have at least 5 control-label pairs (==10 volumes)
+if nFramesPerTE < 10 % Only execute ENABLE if we have at least 5 control-label pairs (==10 volumes)
 	bENABLE = false;
 end
 
 
 %% Manage zig-zag in motion correction
-if x.Q.nUniqueInitial_PLD>1 || x.Q.nUniqueEchoTime>1
-    % ZigZag are temporarily disabled for multiPLD/TE
+if bMultiTE || x.modules.asl.bTimeEncoded
```
HenkMutsaerts, line 188: Why not for MultiPLD?

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
```diff
@@ -192,114 +202,158 @@ function xASL_wrp_RealignASL(x, bASL)
...
+	case 1 % normal quality
+		flags.quality = 1;
+		flags.sep = minVoxelSize;
+	case 0 % low quality for fast try-out
+		flags.quality = 0.01;
+		flags.sep = minVoxelSize*2;
+end
+
+if bMultiTE || bMultiPLD 
+	flags.rtm = 0; % disable realign to mean
```
HenkMutsaerts: Well theoretically you would still want this, but then within the PLD, or within the first TE only.

The idea is that realigning to the mean is always better than realigning to the first, as this equally spreads the interpolations, and there won't be any bias towards the awake or sleepy phase (head position/motion can be correlated with brain state). The realigning to the first (first-pass) is only needed to get the estimates to get the mean position.

So theoretically we should still enable realign to mean (the second-pass). In practice the difference may be small because of the low numbers of repeats; but then again the number of repeats (with equal acquisition duration) shouldn't matter; we can still have the situation in which all acquisitions were quite far from the first, because the pt got scared after the start of the sequence but then managed to stay still in the new position.

So I would always keep the realign to the mean (second pass)

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
```diff
@@ -192,114 +202,158 @@ function xASL_wrp_RealignASL(x, bASL)
...
+
+% Run motion correction for corresponding case
+% Note that this is the adapted spm_realign, including zig-zag
+% regression to account for ASL's potential control-label difference in
+% average head position
+
+V = spm_vol(InputPath); 
+
+if bMultiTE || bMultiPLD
+	% Prepare for updating the volumes - this is not necessary for the simple case where all is handeled by spm_realign
```
HenkMutsaerts: So why don't we always do this; this will create negligible extra computation time, but it makes the code easier (one if statement less), and potentially more reproducible?

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
```diff
@@ -192,114 +202,158 @@ function xASL_wrp_RealignASL(x, bASL)
...
+	Y = spm_read_vols(V); % Read the image
+	rp_all = zeros(nFrames, 6); % The final aggregate RP-file
+	mat_all = zeros(4, 4, size(Y,4)); % The final aggregate MAT-file
+end
+
+if bMultiTE
+    % Handles Multi-TE dataset regardless of PLD
+	% Registers all frames with the shortest PLDs and then applies the same transformation to all the longer PLDs
```
HenkMutsaerts: I assume this should be TEs in this sentences instead of PLDs?

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
```diff
@@ -192,114 +202,158 @@ function xASL_wrp_RealignASL(x, bASL)
...
+	% It assumes that the volume is sorted in the order of acquisition with blocks of increasing TEs
+    
+	% Indexes that flags the first/shortest TEs frames
+    idx_firstTE = find((x.Q.EchoTime == min(x.Q.EchoTime)));
```
HenkMutsaerts: Strictly speaking, it should be idx_minTE and idx_maxTE, e.g. with Thijs' TEprep sequence you could have larger TEs being earlier than smaller TEs?

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
```diff
@@ -192,114 +202,158 @@ function xASL_wrp_RealignASL(x, bASL)
...
+	idx_lastTE =  find((x.Q.EchoTime == max(x.Q.EchoTime)));
+    
+	if length(idx_firstTE) ~= length(idx_lastTE)
+		% TEs should form blocks and thus have the same lengths
```
HenkMutsaerts: Consider replacing by:

% TEs should all have the same number of blocks (e.g., control-label repetitions and/or PLDs)

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
```diff
@@ -192,114 +202,158 @@ function xASL_wrp_RealignASL(x, bASL)
...
+		error('Number of shortest TEs and longest TEs do not match');
```
HenkMutsaerts: Consider rephrasing:

Number of shortest TEs and longest TEs do not match, check if there are missing frames/volumes.

Also: is this an error that we can make more general, e.g., move above, where we check if all "dimensions" (control-label, repetitions, TEs, PLDs) are completely acquired?

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
```diff
@@ -192,114 +202,158 @@ function xASL_wrp_RealignASL(x, bASL)
...
-		if ii==2
-			NDV{2} = [0; NDV{2}]; % add leading zero difference
+
+	% Realigns only the flagged frames
+    spm_realign(V(idx_firstTE), flags, bZigZag); 
+	% This affects the ASL4D.mat file, rp_ASL4D.txt file and MAT within the ASL4D.nii volume
+    
+    rp_temp = load(rpfile); % Load the rp file written by spm_realign
+
+	V = spm_vol(InputPath); % Read the updated volumes
```
HenkMutsaerts, line 276: Move out of this if-statement -> you can also do this with a single TE, it will be negligible extra processing time but easier to read code and more reproducible.

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
```diff
@@ -192,114 +202,158 @@ function xASL_wrp_RealignASL(x, bASL)
...
-				title(['Motion plot of ' x.P.SubjectID '-' x.P.SessionID]);
-				ylabel('NDV/frame (mm//frame)');
-			end
-			
-			xlabel('frame#');
-			axis([1 length(NDV{ii}) 0 ceil(max(NDV{ii}))]); % fix X-axes to be same for subplots
+        
+    end
+    
+elseif bMultiPLD
```
HenkMutsaerts: So bMultiTE and bMultiPLD are mutually exclusive in this if-statement; they should not be right?

And the case of TimeEncoded is again different; then I would use bTimeEncoded for this

Or at least manage bZigZag separately, as we do for the simple case

(in the near future, you could do bMultiPLD for Time Encoded as well, because you can decode which images belong to which PLD)

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
```diff
@@ -192,114 +202,158 @@ function xASL_wrp_RealignASL(x, bASL)
...
+	% Realigns only the flagged frames
+    spm_realign(V(idx_firstTE), flags, bZigZag); 
+	% This affects the ASL4D.mat file, rp_ASL4D.txt file and MAT within the ASL4D.nii volume
+    
+    rp_temp = load(rpfile); % Load the rp file written by spm_realign
+
+	V = spm_vol(InputPath); % Read the updated volumes
+    % Loop through TE groups
+    for idxTE = 1:length(idx_firstTE)
+        % Repeats the motion estimates extracted from the rp file for the rest of the TEs and writes to the corresponding rows in rp_all
```
HenkMutsaerts, line 279: Consider adding

(rp_all is what ends up in the rp*.txt sidecar, rp=realign parameters)

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
```diff
@@ -192,114 +202,158 @@ function xASL_wrp_RealignASL(x, bASL)
...
+        rp_all(idx_firstTE(idxTE):idx_lastTE(idxTE), :) = repmat(rp_temp(idxTE,:), x.Q.nUniqueEchoTime, 1); 
+    
+		for idxAllTE = idx_firstTE(idxTE):idx_lastTE(idxTE)
```
HenkMutsaerts: Consider adding

(mat_all is what ends up in the ASL*.mat sidecar, containing the orientation matrices for each frame/volume)

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
```diff
@@ -192,114 +202,158 @@ function xASL_wrp_RealignASL(x, bASL)
...
-			axis([1 length(NDV{ii}) 0 ceil(max(NDV{ii}))]); % fix X-axes to be same for subplots
+        
+    end
+    
+elseif bMultiPLD
+    % Handles only Multi-PLD datasets - aligns only between the same PLDs
+	% Motion correction across all PLDs is not really necessary as that can be done with the simple motion correction
+	% Note that we handle normal mutli-PLD (not TimeEncoded), so there are still control and label images and we can thus do ZigZag
+
+    for pld = x.Q.uniqueInitial_PLD(:)'
```
HenkMutsaerts: do this for each case, also for singlePLD, and also for multi-TE multi-PLD

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
```diff
@@ -34,12 +34,10 @@ function xASL_wrp_RealignASL(x, bASL)
...
-% Licensed under Apache 2.0, see permissions and limitations at
-% https://github.com/ExploreASL/ExploreASL/blob/main/LICENSE
-% you may only use this file in compliance with the License.
-% __________________________________
-
-
+% Licensed under Apache 2.0, see permissions and limitations at
+% https://github.com/ExploreASL/ExploreASL/blob/main/LICENSE
+% you may only use this file in compliance with the License.
+% __________________________________
```
HenkMutsaerts, line 42: Add this to description (we can improve later):

+% Note that:^M
+% A. This submodule distinguishes ASL (control-label) pairs from non-ASL (e.g. fMRI/DTI), as with control-label pairs,^M
+% the zig-zag approach is used, in which the average control vs label intensity differences are disregarded (not erroneously seen as motion).^M
+% B. If multiple TEs are acquired, only the first TE (with the highest SNR) is used for motion estimation^M
+% C. Motion outlier detection is skipped for multiple PLDs and/or multiple TEs

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
```diff
@@ -69,12 +68,10 @@ function xASL_wrp_RealignASL(x, bASL)
 
 %% Read basic image information
 tempnii = xASL_io_ReadNifti(InputPath);
-nFrames = double(tempnii.hdr.dim(5));
-if length(x.Q.EchoTime)>1
-    nFrames=nFrames/numel(unique(x.Q.EchoTime));
-end
-minVoxelSize = double(min(tempnii.hdr.pixdim(2:4)));
+nFrames = double(tempnii.hdr.dim(5)); % Total number of frames
+nFramesPerTE=nFrames/numel(unique(x.Q.EchoTime)); % Number of frames per unique TE
```
HenkMutsaerts: Add comments:

% (note that for most cases, where we only have a single TE, nFramesPerTE==nFrames

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
```diff
@@ -83,18 +80,31 @@ function xASL_wrp_RealignASL(x, bASL)
...
-elseif nFrames > 1
+elseif nFramesPerTE > 1
 	bMoCoPossible = true;
 else
 	bMoCoPossible = false;
-	fprintf('%s\n',['Skipping motion correction for ' x.P.SubjectID '_' x.P.SessionID ' because it had only ' num2str(nFrames) ' 3D frames.']);
+	fprintf('%s\n',['Skipping motion correction for ' x.P.SubjectID '_' x.P.SessionID ' because it had only ' num2str(nFramesPerTE) ' 3D frames.']);
 end
 
 if isfield(x.Q,'LookLocker') && x.Q.LookLocker
```
HenkMutsaerts: Move this to the if-statement right before here, as first

(if bLookLocker, bContainsSubtracted, bMultiTE)

HenkMutsaerts (changes requested):
1. xASL_wrp_RealignASL: main comment is that we should simplify the code such that singlePLD will be processed as a multiPLD, and singleTE will be processed as a multiTE. And to make these flavor categories independent, e.g. at the top we define bZigZag, instead of saying "Time Encoded cannot be bMultiPLD because we do zig-zag for multi PLD but not for time encoded"

2. I assume that 2. Calculate and plot and 3. outlier detection weren't changed significantly (except for skipping in certain cases), so didn't check this (the github diff doesn't work here unfortunately)

3. So `spm_realign` doesn't seem to change, only comments are added for readability?

jan-petr:
> 1. xASL_wrp_RealignASL: main comment is that we should simplify the code such that singlePLD will be processed as a multiPLD, and singleTE will be processed as a multiTE. And to make these flavor categories independent, e.g. at the top we define bZigZag, instead of saying "Time Encoded cannot be bMultiPLD because we do zig-zag for multi PLD but not for time encoded"

Agree - will improve the comments

> 2. I assume that 2. Calculate and plot and 3. outlier detection weren't changed significantly (except for skipping in certain cases), so didn't check this (the github diff doesn't work here unfortunately)

Yes - this was only moved to a subfunction. No functionality edits.

> 3. So `spm_realign` doesn't seem to change, only comments are added for readability?
Yes - just comments added by Mimi to help her understand it.

External/SPMmodified/spm_realign.m
jan-petr: Done

External/SPMmodified/spm_realign.m
jan-petr: Done

External/SPMmodified/spm_realign.m
jan-petr: Session is internal name inside spm_realign which means that the registration is split into blocks. I improved the comment, but won't change the terminology as this is the terminology of SPM and it explains this in the header.

External/SPMmodified/spm_realign.m
jan-petr: Done

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
jan-petr: Yes. Good catch :)

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
jan-petr: Done

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
jan-petr: Done.


For PLDs and repetitions, this is already checked before. Here, we don't check consistency of data, we kind of check if it will crash or not with the current code. So we prevent crashing without an explanation.

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
jan-petr: Done on a line where rp_all is initialized

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
jan-petr: added also on the initialization line.

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
jan-petr: Done

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
jan-petr: Done

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
jan-petr: nFramesPerTE is effectively the number of repetitions irrespective of TE. 

bMultiTE does not necessarily have nFramesPerTE > 1. 

I'll improve the definitions and comments

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
jan-petr: Agree. For transparency, we treat all as multi-PLD - even multi-TE and Time-encoded.

Then we modify the exact choices based on this... Like switching ENABLE etc. Modified throughout the code.

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
jan-petr: This is correct. Standard multi-PLD has controls and labels, so we can run Zig-zag. I've added a comment about this.

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
jan-petr: Agree. I'm deleting this.

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
jan-petr: For single-PLD, we don't use these parameters at all. That's handeled within SPM. So it would make the code more convoluted.

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
jan-petr: Good point. Fixed.

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
jan-petr: It can't be moved out of the if-statement. It has to be run after spm_realign for multi_TE, but before we reassemble it for the other TEs. So it has to be within this loop.

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
jan-petr: They kind of are.
The options -> sorted by priority are
MultiTE-multiPLD
multiTE-singlePLD
singleTE-multiPLD
singleTE-singlePLD

We just skip certain options because we don't have them, so I can't really nicely test them. Shall I add these options and say that they are implemented?

Timenecoded option is now properly fixed. And Zigzag as well.

MultiPLD-timeencoded won't really have a specific option because we don't have sufficient repetitions. So for that, we will need the boundary based registration.

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
jan-petr: Not sure what should we do - see the other comments.

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
jan-petr: Done

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
HenkMutsaerts: The way I would like it, is that singleTE is processed as multiTE, and singlePLD is processed as multiPLD. This is how we changed the rest of the pipeline as well. Creates negligible processing overhead but makes the code easier and more reproducible (fewer if-statements).

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
```diff
@@ -162,19 +181,20 @@ function xASL_wrp_RealignASL(x, bASL)
...
 
 
 %% Manage zig-zag in motion correction
-if x.Q.nUniqueInitial_PLD>1 || x.Q.nUniqueEchoTime>1
-    % ZigZag are temporarily disabled for multiPLD/TE
+if bMultiTE || x.modules.asl.bTimeEncoded
+    % ZigZag are temporarily disabled for multiTE and TimeEncoded
     % as we are still developing this feature
+	% Note that for standard multi-PLD, Zig-zag can be applied because standard multi-PLD still has controls and labels
     fprintf('%s\n', 'multi-PLD or multi-TE detected, disabling zig-zag motion estimation, not yet implemented');
```
HenkMutsaerts: Change into:

fprintf('%s\n', 'Time-encoded or multi-TE detected, disabling zig-zag motion estimation, not yet implemented');

HenkMutsaerts (changes requested):
Can you please change it such that:

1. Process singleTE similar as MultiTE (extra overhead, but programming-wise more consistent. I'm fine with keeping the if-statement for updating the transformation matrices, but would like to remove the if statement for the realignment estimation split of singleTE and multiTE
2. Process singlePLD similar as multiPLD (for same reason as above).

I'm also happy to implement this, but perhaps faster if you do this.
We do this as well with our other code; makes it also easier to expand later on.

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
jan-petr: Done

jan-petr (approved):
OK

HenkMutsaerts (approved):

PR #2043, opened 2026-04-16
#2041 module_population New ASL PV defaults

### LINKED ISSUE #2041: Default PV should include 0.5
### Description

Currently, default PV when no atlas is included is 0.7. Make sure that GM PV at 0.5 is added as well.

### Release notes

Update PV default thresholds to both 0.5 and 0.7

jan-petr:
### Linked issue

Close #2041


Modules/xASL_module_Population.m
```diff
@@ -35,7 +35,7 @@
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
 % SPDX-License-Identifier: Apache-2.0
 % ExploreASL; see permissions and limitations at https://github.com/ExploreASL/ExploreASL/blob/main/LICENSE
-% __________________________________
+% __________________________________ti
```
HenkMutsaerts: Something going wrong here?

Modules/xASL_module_Population.m
```diff
@@ -53,9 +53,19 @@
 % Check again for Atlases (main checking is done when loading dataPar)
 if ~isfield(x.S,'Atlases') && ~isfield(x.S, 'TissueMasking')
 	% If no values are provided, then we provide the defaults
-	x.S.Atlases = {'Total','DeepWM','Tatu_ACA_MCA_PCA'}; % Default
-    x.S.TissueMasking = {'GM' 'WM','GM'}; % GM WM GM, fits with the Total & DeepWM & Tatu_ACA_MCA_PCA above
-    % Note that this should be in the same order as the atlases/ROIs
+	% GM WM GM tissues with the Total & DeepWM & Tatu_ACA_MCA_PCA
```
HenkMutsaerts, line 56: Always having both 0.7 and 0.5 is too confusing to the user, we should give explorative results as default output, not many results necessarily.

The alternative is 0.6, or anyway recommending 0.5 PVC.

BTW: we don't catch all here, e.g. if Atlases exists but the others don't, we won't catch it. Is this on purpose?

HenkMutsaerts (changes requested):

Modules/xASL_module_Population.m
jan-petr: This is really tricky! 
Because literature unanimously wants 0.7. But 0.7 is really sparse and for PV, you really need 0.5.

Doing 0.5 only is wrong. Doing 0.7 only is also wrong. The user simply has to understand it and not just blindly take 0.5 without even knowing it. Or blindly take 0.7.

0.6 doesn't solve anything - we can't use 0.6 and write 70% PV in the paper.

We catch all -> If atlases are provided and the rest not, it throws and error and explanation+recommendation. It doesn't fix anything, but tells user to fix it and re-run -> this is on prupose, you have insisted on making it like this :D

Modules/xASL_module_Population.m

Modules/xASL_module_Population.m
HenkMutsaerts: OK then we generate both 0.5 and 0.7 for now, and let's see how this is taken up by the users.

I see now that we catch all, I was only looking at line 54.

HenkMutsaerts (approved):
See new comment, that you can resolve.

BTW: pvWM>0.7 is not really defined in literature, the majority of the WM has pvWM near 1, so the pVWM doesn't matter that much, the erosion matters more (especially for deepWM).

So if we previously had a different pvWM, I would consider keeping this; as I did some comparisons to check which combination pvWM with nLayers erosion worked best.

PR #2045, opened 2026-04-16
#2042 Differentiate T2GM and T2tissue

### LINKED ISSUE #2042: T2tissue in mutli-TE fitting
### Description

For BBB-ASL with multi-TE acquisition, we use T2 of blood and T2 of tissue. In certain subjects/diseases/regions, we want to use different T2 (tumor, choroid plexus, etc). 

For that, we allow modifying T2 tissue in dataPar.json. However, T2tissue was also used for M0 quantification where it is assumed that T2 tissue has GM value. 

To simplify, we create a new parameter T2GM that is used for GM specific tasks. And TissueT2 is changed to T2tissue - this is used for multi-TE BBB-ASL and by default, it is initialized to GM value.

### Tasks

- [x] For M0 calculation. Create a new parameter T2GM. Rename this everywhere
- [x] For Multi-TE quantification, keep T2tissue.
- [x] Inform Ayse/Gulce/Bea/Janka
- [x] Manual - Rename T2tissue to T2GM in the manual
- [x] Manual - Add T2tissue as extravascular compartment T2

### Release notes

Distinguish between T2GM and T2tissue for multiTE. Change names of T1 and T2 parameters.

jan-petr:
### Linked issue

Close #2042


HenkMutsaerts (changes requested):
1. Background: why should M0 and ASL fitting use different tissue T2 times? They shouldn't.
JAN: 1. For M0, you need to do signal conversion from tissue to blood. For that, we hand-pick GM voxels and extrapolate. Therefore, you need to use T2-GM. For ASL, we need to use the correct T2-tissue value in each voxel. Mutli-TE ASL isn't run in just GM voxels (like M0 calibration), but in all voxels. In case of tumors, we want to do ASL with tumor-T2, but M0 is done using GM-T2 in GM voxels only.

H: I see your point, but you are comparing two use cases, whereas I would want to keep things general.
If you keep it general, you have T2GM and T2WM, and which one is used further down in the pipeline, if the M0 is correct based on tumor T2, GM T2, or WM T2, that is then all decided later on, or should be clarified with booleans.

If you want to keep it simple for now, then rename the parameters.

E.g., T2_tissue_M0 T2_GM_M0 T2_WM_M0 T2_tissue_fitting T2_GM_fitting T2_WM_fitting

or whatever

2. T2tissue should be renamed to what it is, because T2GM is also T2tissue. Do you mean that T2tissue is T2WM? Or T2M0?
Jan: T2-tissue is only used for the mutli-TE model because we want to be able to run it for different tissue types. In the future, we will use a T2-map for multi-TE.

H: note that a T2map for multi-TE is tricky because misalignments (which are larger in patients than in controls) will generate additional bias (potentially gaining accuracy by loosing precision), but perhaps if we smooth it this is less of an issue. We wanted to smooth before fitting anyway; so maybe this solves it.

I suggest renaming it to T2_tissue_fitting then, or something similar to clarify that this is used for the fitting only.

3. We assume GM analyses (even though our M0 processing strips to mostly WM only)
J: OK, so we strip to WM, but use GM value -> this was then wrong in the code and has to be fixed. I only did renaming, but didn't change functionality

H: Sure

Henk: So I'm fine with anything that 1) makes sense 2) is consistent 3) clear to the user

Jan:
I'm fine renaming T2-tissue to something else to specify the purpose.


Henk: so we will have?
T2_tissue_fitting
T2_GM_M0
T2_WM_M0


JAN: So what is the final solution?
1. I'm happy to change the name to T2_tissue_fitting, but I'd call it T2_Tissue_Extravascular.
H: 
tissue == extravascular, and GM==extravascular WM==extravascular.
If this is for custom tissue T2, why not T2_tissueCustom or something like this?

Final: T2_multi-TE_Extravascular

2. What about M0 calibration. Should we use T2_WM_M0? I'm happy to do that, but we would need to revisit the calibration process. I propose that I do the renaming, keep it at T2_GM_M0 and create a new issue to fix the calibration.
H:
Agreed

3. T2_GM_M0 - I do not like that as T2_GM is T2_GM and it won't differ if it is for M0 or Multi-TE fitting. That's different for T2_tissue_fitting which is not tissue specific. So I would avoid double-definition by purpose+tissue type.
But I agree with your general arguments.
H:
Agreed

jan-petr:
1. T2_GM and T2_WM defined, with reasonable backwards compatibility
2. I added a new issue #2047 to fix the T2 time for M0 calibration
3. New parameter is named T2_MultiTE_Extravascular

HenkMutsaerts (approved):

PR #2046, opened 2026-04-16
#2044 Save residults of BASIL

### LINKED ISSUE #2044: Save BASIL residuals
### Description

Make sure that BASIL residuals are saved

 - Residuals: For this, we need to run FABBER with '--save-residuals' in the Options.txt to save the difference between the measured data and the model prediction. We cna check these together, but ideally these residuals should be small and we can check how these change across time/echoes.
- Model fit: We need to add --save-model-fit and then compare it against the original data(good visual check to see whether the model is following the data well).

### Tasks

- [x] Save FABBER/BASIL residuals
- [x] Save FABBER/BASIL model fit
- [x] Remove model fit, residuals, parameters before re-running
- [x] DeleteTemp - delete model fit
- [x] save BASIL/FABBER aprameter file for provenance

### Release notes

Save BASIL/FABBER residuals and input parameters

jan-petr:
### Linked issue

Close #2044

### DIFF
--- a/Functions/Quantification/xASL_quant_ASL.m
+++ b/Functions/Quantification/xASL_quant_ASL.m
@@ -1,6 +1,6 @@
-function [ScaleImage, CBF, ATT, ABV, Tex, ITT, Res, Fit, Log] = xASL_quant_ASL(PWI4D_Path, M0_im, imSliceNumber, x, bUseExternalQuantification, bSaveCBF4D)
+function [ScaleImage, CBF, ATT, ABV, Tex, ITT, Residuals, Fit, Log] = xASL_quant_ASL(PWI4D_Path, M0_im, imSliceNumber, x, bUseExternalQuantification, bSaveCBF4D)
 %xASL_quant_ASL Perform a multi-step quantification of single or multi-PLD with or without external quantification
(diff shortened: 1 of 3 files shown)


Functions/Quantification/xASL_quant_ASL.m
```diff
@@ -1,6 +1,6 @@
-function [ScaleImage, CBF, ATT, ABV, Tex, ITT] = xASL_quant_ASL(PWI4D_Path, M0_im, imSliceNumber, x, bUseExternalQuantification, bSaveCBF4D)
+function [ScaleImage, CBF, ATT, ABV, Tex, ITT, Res, Fit, Log] = xASL_quant_ASL(PWI4D_Path, M0_im, imSliceNumber, x, bUseExternalQuantification, bSaveCBF4D)
```
HenkMutsaerts: Res can also be results, write out for readability?

HenkMutsaerts (approved):
1 micro suggestion for renaming.

Should these new options always be saved, or only for debugging (DELETETEMP)?

Functions/Quantification/xASL_quant_ASL.m
jan-petr: Done.

PR #2050, opened 2026-04-24
#2049 module_ASL: Set defaults

### LINKED ISSUE #2049: T1WM default value
### Description

Default value of parameters like T1WM are now set centrally in a function and not in submodules.
So T1WM is not set anymore inside function for M0-quantification.

Therefore, we need to call the function for setting defaults in the main xASL_module.

### Release notes

Minor bug fix in setting T1WM defaults.

jan-petr:
### Linked issue

close #2049


HenkMutsaerts (approved):
Unsure if we should call it at the beginning of the ASL module if we only use it upon M0 processing (or even only at quantification, depending on the sequence).

PR #2052, opened 2026-04-27
#1010 first draft

### LINKED ISSUE #1010: Boundary-based registration (BBR)
### Description

Implement boundary-based registration (BBR) which may improve registration for ASL scans, provided there is sufficient spatial contrast/resolution.

### Requirements

### How to test

### Release notes

jan-petr:
### Linked issue

Close #1010


Functions/Administration/xASL_adm_GetDeprecatedFields.m
```diff
@@ -150,6 +150,7 @@
 	'bCleanUpExternal'              'modules'   'asl'                                'bCleanUpExternal';...
 	'SaveCBF4D'                     'modules'   'asl'                                'SaveCBF4D';...
 	'bHct2BloodT1'                  'modules'   'asl'                                'bHct2BloodT1';...
+	'bMoCoEdgeEnhanced'             'modules'   'asl'                                'bMoCoEdgeEnhanced';...
```
HenkMutsaerts, line 153: I would prefer calling it bMocoBoundaryBased (unless that's not what it is): keeping things easy by using the same terminology

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
```diff
@@ -270,15 +270,78 @@ function xASL_wrp_RealignASL(x, bASL)
...
+				imLoG(:,2,2) = 1;
+				imLoG(2,:,2) = 1;
+				imLoG(2,2,:) = 1;
+				imLoG(2,2,2) = -6; 
+
+				imEdge = convn(imSmooth, imLoG, 'same');
+			case 5
+				% LoG - Laplacian of Gaussian filter, 0.5STD Gaussian, then Laplacian, zero-crossings of second derivative
+				sigmaEdge = 0.5; % STD of smoothing kernel in voxels
+				imSmooth = imgaussfilt3(Yorig(:,:,:,iVol), sigmaEdge); % Gaussian smoothing
```
HenkMutsaerts: check carefully with all these image-based functions that the user doesn't require the Matlab image toolbox for these

As far as I can see, only imgaussfilt3 cannot be used, e.g. use  xASL_im_Smooth* instead

Functions/Administration/xASL_adm_GetDeprecatedFields.m
jan-petr: Indeed, it's not really boundary based as boundary based really segments the boundary. We truly only enhance the edges. 

We have discussed boundary based initially, but it would have been too tricky and wouldn't have any large benefit. So that's why we chosed Edge enhanced.

The idea behind is the same, it's just the implementation that differs a bit

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
jan-petr: Good point! I'm replacing those

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
```diff
@@ -246,11 +262,93 @@ function xASL_wrp_RealignASL(x, bASL)
...
+				% Sobel with 0.5STD presmoothing, thresholding and further smoothing
+				sigmaEdge = 0.5; % STD of smoothing kernel in voxels
+				imSmooth = imgaussfilt3(Yorig(:,:,:,iVol), sigmaEdge); % Gaussian smoothing
+				[Gx, Gy, Gz] = gradient(imSmooth); % Computes spatial gradients
+				imEdge = sqrt(Gx.^2 + Gy.^2 + Gz.^2); % Computes gradient magnitude
+				imThreshold = sort(imEdge(:)); % Calculate 95th percentile
+				imThreshold = imThreshold(ceil(numel(imThreshold)*0.95));
+				imEdge = imEdge > imThreshold; % Threshold to avoid intensity scale differences
+				imEdge = imgaussfilt3(double(imEdge), sigmaEdge); % Gaussian smoothing 
+			case 4
```
jan-petr, line 302: Method 4 seems to work the best - according to our tests with Mimi. I propose to keep that one as default and comment out all the rest. We want to keep them in the code for history - to be able to double-check our results later.

At the same time. It would be nice to check this fully separately and independently. Can you run the tests yourself as well and check the motion correction and alignment for DEBBIE?

Finally, we need to add this parameter to documentation and make sure this works nicely and is nicely integrated to the workflow.

Functions/Administration/xASL_adm_GetDeprecatedFields.m
HenkMutsaerts: Sounds good. You can write this explanation somewhere; perhaps at the top of the bMoCoEdgeEnhanced (sub)function?

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
HenkMutsaerts: You say here "Can you run the tests yourself as well and check the motion correction and alignment for DEBBIE?" to whom is this aimed?

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
```diff
@@ -55,12 +55,17 @@ function xASL_wrp_RealignASL(x, bASL)
...
 
 [Fpath, Ffile, Fext] = xASL_fileparts(InputPath);
+
 rpfile = fullfile( Fpath, ['rp_' Ffile '.txt']);
 rInputPath = fullfile( Fpath, ['r' Ffile Fext]);
 InputPathJson = fullfile( Fpath, [Ffile '.json']);
 rInputPathJson = fullfile( Fpath, ['r' Ffile '.json']);
 matFile = fullfile(Fpath, [Ffile '.mat']);
 
+% Path to a temporary file for registration
```
HenkMutsaerts, line 65: I think we have these paths already, if you check the xASL_wrp_RegistrationASL, e.g. something like PWI_temp. "reg" in the filename is good; but "temp" in the filename is perhaps better, because to the user it should be clear that it is a temporary file. "reg" could still be interpreted as "registered ASL image" (which can also be an end result)

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
```diff
@@ -268,69 +366,84 @@ function xASL_wrp_RealignASL(x, bASL)
...
+% 		% Loop through PLD groups
 % 		rp_all(idxSinglePLD, :) = rp_temp;
 % 
 % 		for idxPLD = idxSinglePLD(:)'
-% 			mat_all(:, :, idxPLD) = V(idxPLD).mat;
+% 			mat_all(:, :, idxPLD) = Vreg(idxPLD).mat;
 % 		end
 %     end
+
+elseif x.modules.asl.bMoCoEdgeEnhanced
```
HenkMutsaerts, line 407: Same criterion as if-statement?

Can we reprogram it such that we always do the multi-PLD motion correction (single-PLD simply being looping only once)?

Modules/Module_Processing/Module_ASL/xASL_wrp_RealignASL.m
```diff
@@ -268,69 +366,84 @@ function xASL_wrp_RealignASL(x, bASL)
...
     fprintf('Standard SPM motion estimation\n');
     % Handles simple datasets
-    spm_realign(V, flags, bZigZag);
+    spm_realign(Vorig, flags, bZigZag); % No reordering or edge-based registration, we can keep the results as they are
 end
- 
 
 % For these special cases, we need to save the updated transformation matrices
-if bMultiTE %|| (bMultiPLD && ~x.modules.asl.bTimeEncoded)
+if bMultiTE || x.modules.asl.bMoCoEdgeEnhanced %|| (bMultiPLD && ~x.modules.asl.bTimeEncoded)
```
HenkMutsaerts, line 425: But saving the updated transformation matrix never hurts, also for vanilla ASL.
So I would suggest always doing the same, saving the updated volume one by one will just be a single loop for single-PLD single-TE, and you can check if isequal to skip. But that would generalize,

HenkMutsaerts (changes requested):
#### bMoCoEdgeEnhanced
Why would edge enhanced be better necessarily for multiPLD or time encoded? I understand that the benefit may be larger than with plain vanilla ASL, but essentially the benefit yes/no is the same.

I mean that the possibility to help vs screw up the registration (if it hasn't been fully tested) is the same for all ASL flavors. So you can consider keeping it simple:

1. Currently default -> bMocoEdgeEnhanced == false
2. After thorough testing, the default -> bMocoEdgeEnhanced == true
(because if it helps multiPLD or time encoded, it also helps single PLD, just less)

The main point is to have it disabled before we have tested it with different readouts; so essentially the validation process is not concerning the labeling strategy, but concerning the readouts, e.g.:

with vascular PWI contrast, does it help?
with CBF PWI contrast, does it help?
with sharp control label images, does it help (probably yes)
with (very) smooth control label images, does it help (needs testing)

-> just to generalize this for both MoCo and registration

#### Separate function
We can use this both in moco and in registration, can you therefore move this to a separate function? This has the benefit that you can leave the other methods commented out in that separate function; and xASL_realign_ASL changes minimally

#### Check that this new edged enhanced registration function is not dependent of Matlab toolboxes

PR #2057, opened 2026-05-09
Fixes #2056 ucl bbb miscellaneous

### LINKED ISSUE #2056: Miscellaneous improvements UCL BBB data
### Description

I analyzed FME/DEBBIE BBB-ASL data from UCL, and came across the following issues, that I fixed.

### Tasks
Major:
- [x] xASL_imp_CheckImportSettings: add dicom extension .MR
- [x] xASL_ext_FSLRun, xASL_ext_FSLSetDir, xASL_ext_FSLTopUp: enable automatic FSL detection by default, if not specified

Minor:
- [x] bids-layout: hint about M0 warning in case of DEBBIE
- [x] xASL_imp_ReadSourceData: avoid incorrect extension warning (extension is in regular expression here, not normal wildcards)
- [x] xASL_imp_DCM2NII_ConvertScan & xASL_wrp_DCM2NII_Subject & xASL_wrp_NII2BIDS_Subject: improve visit feedback

### Release notes

Improve BBB-ASL workflow.

HenkMutsaerts:
### Linked issue

Fixes #2056

### DIFF
--- a/External/bids-matlab/+bids/layout.m
+++ b/External/bids-matlab/+bids/layout.m
@@ -453,7 +453,15 @@ function subject = parse_perf(subject)
                             % which is a safe option if no background suppression was used
                             if subject.perf(j).meta.BackgroundSuppression==true
                                 WarningID = 'BIDSLAYOUT:CautionControlAsM0Bsup';
-                                warning(WarningID, 'Caution when using control as M0, background suppression was applied'); % will print this warning unless WarningID has been set OFF
+                                warningState = warning('query',WarningID);
+
+                                if ~strcmp(warningState.state,'on')
+                                    fprintf('\n\n');
+                                    warning('Caution when using control as M0, background suppression was applied');
+                                    fprintf('%s\n', 'This warning may be ignored if a single M0 scan is used for multiple runs');
+                                    fprintf('%s\n', 'e.g., which is the case for FME/DEBBIE BBB-ASL');
+                                    fprintf('\n\n');
+                                end 
                                 warning('OFF', WarningID); % to print this warning only once
                             end
                             
--- a/External/bids-matlab/+bids/layout.m
+++ b/External/bids-matlab/+bids/layout.m
@@ -350,7 +350,7 @@ function subject = parse_perf(subject)
                 else
                     WarningID = ['BIDSLAYOUT:Missing' metafile];
                     warning(WarningID, ['Missing: ' metafile]); % will print this warning unless WarningID has been set OFF
-                    warning('OFF', WarningID); % to print this warning only once
+                    warning('off', WarningID); % to print this warning only once
                 end
                 
                 % Manage ASLCONTEXT-sidecar metadata (REQUIRED)
@@ -364,7 +364,7 @@ function subject = parse_perf(subject)
                 else
                     WarningID = ['BIDSLAYOUT:Missing' metafile];
                     warning(WarningID, ['Missing: ' metafile]); % will print this warning unless WarningID has been set OFF
-                    warning('OFF', WarningID); % to print this warning only once
+                    warning('off', WarningID); % to print this warning only once
                 end                
                 
                 % Manage M0 (REQUIRED)
@@ -374,7 +374,7 @@ function subject = parse_perf(subject)
                 if ~isfield(subject.perf(j).meta, 'M0Type')
                     WarningID = 'BIDSLAYOUT:MissingM0TypeField';
                     warning(WarningID, ['M0Type field missing in ' subject.perf(j).json_sidecar_filename]); % will print this warning unless WarningID has been set OFF
-                    warning('OFF', WarningID); % to print this warning only once
+                    warning('off', WarningID); % to print this warning only once
                 else
                     switch subject.perf(j).meta.M0Type
                         case 'Separate'
@@ -394,7 +394,7 @@ function subject = parse_perf(subject)
                             if isempty(m0_filename)
                                 % WarningID = 'BIDSLAYOUT:Missing_M0';
                                 warning(['Missing: M0.nii for ' subject.perf(j).filename]); % will print this warning unless WarningID has been set OFF
-                                % warning('OFF', WarningID); % to print this warning only once
+                                % warning('off', WarningID); % to print this warning only once
                             else
                                 % subject.perf(j).m0_filename = m0_filename;
                                 % -> this is included in the same structure for the m0scan.nii
@@ -414,7 +414,7 @@ function subject = parse_perf(subject)
                             if ~exist(fullfile(pth, m0_json_sidecar_filename), 'file')
                                 WarningID = ['BIDSLAYOUT:Missing' m0_json_sidecar_filename'];
                                 warning(WarningID, ['Missing: ' m0_json_sidecar_filename]); % will print this warning unless WarningID has been set OFF
-                                warning('OFF', WarningID); % to print this warning only once
+                                warning('off', WarningID); % to print this warning only once
                             else
                                 % subject.perf(j).m0_json_sidecar_filename = m0_json_sidecar_filename;
                                  % -> this is included in the same structure for the m0scan.nii
@@ -429,7 +429,7 @@ function subject = parse_perf(subject)
                                 if isempty(m0indices)
                                     WarningID = 'BIDSLAYOUT:MissingM0InASLcontext';
                                     warning(WarningID, 'No M0 volume found in aslcontext'); % will print this warning unless WarningID has been set OFF
-                                    warning('OFF', WarningID); % to print this warning only once
+                                    warning('off', WarningID); % to print this warning only once
                                 else
                                     subject.perf(j).m0type = 'within_timeseries';
                                     subject.perf(j).m0explanation = 'M0 is one or more image(s) in the *asl.nii[.gz] timeseries';
@@ -455,20 +455,20 @@ function subject = parse_perf(subject)
                                 WarningID = 'BIDSLAYOUT:CautionControlAsM0Bsup';
                                 warningState = warning('query',WarningID);
 
-                                if ~strcmp(warningState.state,'on')
+                                if strcmp(warningState.state,'on')
                                     fprintf('\n\n');
                                     warning('Caution when using control as M0, background suppression was applied');
                                     fprintf('%s\n', 'This warning may be ignored if a single M0 scan is used for multiple runs');
                                     fprintf('%s\n', 'e.g., which is the case for FME/DEBBIE BBB-ASL');
                                     fprintf('\n\n');
                                 end 
-                                warning('OFF', WarningID); % to print this warning only once
+                                warning('off', WarningID); % to print this warning only once
                             end
                             
                         otherwise
                             WarningID = 'BIDSLAYOUT:M0TypeUnknown';
                             warning(WarningID, ['Unknown M0Type:' subject.perf(j).meta.M0Type ' in ' subject.perf(j).json_sidecar_filename]); % will print this warning unless WarningID has been set OFF
-                            warning('OFF', WarningID); % to print this warning only once
+                            warning('off', WarningID); % to print this warning only once
                     end
                 end
                     
@@ -531,7 +531,7 @@ function subject = parse_perf(subject)
                 if ~exist(metafile, 'file')
                     WarningID = ['BIDSLAYOUT:Missing' metafile];
                     warning(WarningID, ['Missing: ' metafile]); % will print this warning unless WarningID has been set OFF
-                    warning('OFF', WarningID); % to print this warning only once
+                    warning('off', WarningID); % to print this warning only once
                 else
                     [~, Ffile] = fileparts(metafile);
                     subject.perf(j).json_sidecar_filename = [Ffile '.json'];
@@ -556,7 +556,7 @@ function subject = parse_perf(subject)
                         if ~exist(fullfile(fileparts(pth), path_intended_for{iPath}), 'file')
                             WarningID = 'BIDSLAYOUT:MissingIntendedForField';
                             warning(WarningID, ['Missing: ' fullfile(fileparts(pth), path_intended_for{iPath})]);  % will print this warning unless WarningID has been set OFF
-                            warning('OFF', WarningID); % to print this warning only once
+                            warning('off', WarningID); % to print this warning only once
                         else
                             % also check that this NIfTI aims to the same m0scan
                             [~, path2check, ext2check] = fileparts(path_intended_for{iPath});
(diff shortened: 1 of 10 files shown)


External/bids-matlab/+bids/layout.m
```diff
@@ -453,7 +453,15 @@ function tolerant_message(tolerant, msg)
...
+                                warningState = warning('query',WarningID);
+
+                                if ~strcmp(warningState.state,'on')
+                                    fprintf('\n\n');
+                                    warning('Caution when using control as M0, background suppression was applied');
+                                    fprintf('%s\n', 'This warning may be ignored if a single M0 scan is used for multiple runs');
+                                    fprintf('%s\n', 'e.g., which is the case for FME/DEBBIE BBB-ASL');
+                                    fprintf('\n\n');
+                                end 
                                 warning('OFF', WarningID); % to print this warning only once
```
jan-petr: should be 'off', not 'OFF'

External/bids-matlab/+bids/layout.m
```diff
@@ -453,7 +453,15 @@ function tolerant_message(tolerant, msg)
                             % which is a safe option if no background suppression was used
                             if subject.perf(j).meta.BackgroundSuppression==true
                                 WarningID = 'BIDSLAYOUT:CautionControlAsM0Bsup';
-                                warning(WarningID, 'Caution when using control as M0, background suppression was applied'); % will print this warning unless WarningID has been set OFF
+                                warningState = warning('query',WarningID);
+
+                                if ~strcmp(warningState.state,'on')
```
jan-petr: This won't work. 
it should be strcmp and not ~strcmp... Cause it will only print this when it's already OFF :)

Modules/Module_Import/DCM2NII/xASL_imp_ReadSourceData.m
```diff
@@ -89,12 +89,11 @@ function xASL_imp_ReadSourceData_CheckFolderHierarchy(x) % PM: this should be re
     if x.modules.import.imPar.bMatchDirectories
         % Check that there is no extension in the last folderHierachy element
         % This extension should only be there if bMatchDirectories is set to false
-        [~, ~, Fext] = xASL_fileparts(lastElement);
         
-        if ~isempty(regexpi(Fext, conditionFile, 'once'))
+        if contains(lastElement, conditionExtension)
```
jan-petr: You need to revert this edit. Effectively, this will take any directory containing '.' and report it as file. I don't think this is what we want.

Modules/Module_Import/DCM2NII/xASL_imp_ReadSourceData.m
```diff
@@ -89,12 +89,11 @@ function xASL_imp_ReadSourceData_CheckFolderHierarchy(x) % PM: this should be re
...
            warning('folderHierarchy includes a file extension but bMatchDirectories was set to true');
         end
-    elseif isempty(regexpi(lastElement, conditionFile, 'once'))
+    elseif ~isempty(regexpi(lastElement, conditionFile, 'once'))
```
jan-petr: This is also wrong. In case it looks for files and finds a known extension from the list, you start reporting it as unknown extension. this was correct previously.

Modules/Module_Import/xASL_wrp_DCM2NII_Subject.m
```diff
@@ -48,6 +48,17 @@
     
 	x.modules.import.imPar.visitNames = thisSubject.visitIDs;
 
+    if isempty(thisSubject.visitIDs) || isempty(thisSubject.visitIDs{1})
```
jan-petr: I don't understand why this was added. But the code should have been working without it. It seems very much like you have incorrectly configured sourcedata.json and are trying to correct for the bug here. 
Please show me the example or revert this entire edit - I don't think this code should ever be needed.

Modules/Module_Import/xASL_wrp_NII2BIDS_Subject.m
```diff
@@ -118,5 +118,10 @@
 		if bCorrected
 			warning([subjectName ', changed visit name: ' sessionName ' -> ' bidsLabel.visit]);
 		end
-	end
+    end
+
+    % If the session name is empty, we default to session 1
```
jan-petr: same here. I don't think this code is needed unless the input JSONs are wrong.

jan-petr (changes requested):

Modules/Module_Import/DCM2NII/xASL_imp_ReadSourceData.m
HenkMutsaerts: I see your point. The problem was that we would only flag file extensions for known file extensions. This should solve it:

length(lastElement)>1 && strcmp(lastElement(1:2), conditionExtension)

as we don't want the lastElement to start with '\.' because that clearly indicates an extension

Modules/Module_Import/xASL_wrp_DCM2NII_Subject.m
HenkMutsaerts: If you don't have a visit defined, in the Matlab GUI it would say:

`` Subject = dummytest, visit = , session = ASL_1, scan = T1w ``

With my changes, it will now correctly say

`` Subject = dummytest, session = ASL_1, scan = T1w ``

Modules/Module_Import/xASL_wrp_NII2BIDS_Subject.m
HenkMutsaerts: Same principle as above, code was not wrong for output but for output to the screen

External/bids-matlab/+bids/layout.m
HenkMutsaerts: Fixed in 2f35c836332eb483d36e8c247786715a76cb5ad9

External/bids-matlab/+bids/layout.m
HenkMutsaerts: Fixed in 2f35c836332eb483d36e8c247786715a76cb5ad9

Modules/Module_Import/DCM2NII/xASL_imp_ReadSourceData.m
HenkMutsaerts: Fixed in 2f35c836332eb483d36e8c247786715a76cb5ad9, and I now reverted the whole subfunction to be more correct and clear

Modules/Module_Import/xASL_wrp_DCM2NII_Subject.m
jan-petr: yes. of course. when visit is missing then don't print it. And I'm fine with that modification that prevented that. But there was another function where you were supplementing visits -> and that's what I didn't like.

Modules/Module_Import/DCM2NII/xASL_imp_ReadSourceData.m
```diff
@@ -74,33 +74,33 @@
...
     % Condition for file extension
-    conditionFile = '(zip|dcm|ima|xml|par|rec|nii|nii\.gz)';
-
-    % Other extension
     conditionExtension = '\.';
+    
+    % List of known file extensions (formatted as regular expression)
+    knownExtensions = '(zip|dcm|ima|xml|par|rec|nii|nii\.gz)';
+
+    hasExtension = length(lastElement)>1 && strcmp(lastElement(1:2), conditionExtension);
```
jan-petr: wrong use of strcmp using syntax for regexpi

you either do strcmp(lastElement(1), '.')
or regexpi(lastElement(1),'\.')
but not strcmp(lastElement(1:2),'\.')

Also - it is still wrong. you want to check if the entire filename/directory name has extensions 
so "file.zip" is a file and "file" is a directory.

But you are effectively checking if you have ".zip" - you'll never have that. 
You will need to check if "." is among the last 3 or 4 characters of the file name to distinguish if that is a directory or file.
But this will have still many false positive like "subject.01" will then be identified as file instead of a directory.

That's why it's better to go back to the original code with known extensions.

But the rest of the code and modifications is OK. However, this line won't do the right job as you want it to do.

Modules/Module_Import/xASL_wrp_DCM2NII_Subject.m
jan-petr: I'm fine with changes in xASL_imp_DCM2NII_ConvertScan.m that prevents printing the empty visit. But I don't see the reason to edit xASL_wrp_DCM2NII_Subject.m and xASL_wrp_NII2BIDS_Subject.m and add defaults there.

jan-petr (changes requested):
Changes still needed.

Modules/Module_Import/xASL_wrp_DCM2NII_Subject.m
HenkMutsaerts: This had to do with another location where otherwise visit wouldn't default to 1, in case there are no visits. If you really think this is potentially harmful, I would have to search why I implemented this

HenkMutsaerts: I checked and tested it extensively, also for how visits were printed in the TSV files etc; and this is the way it fixes it for all without breaking it.

If you're scared that my defaulting breaks something, I can check why I did this, but that will take me extra time.

Modules/Module_Import/DCM2NII/xASL_imp_ReadSourceData.m
HenkMutsaerts: You're right, fixed it in bfd1bd371fc2d5796bcbeaaf9dbb91147477e323. Solved now?

Note that previously, we checked for '.' and "example.foldername" gave a warning that it may be a file, that was the reason that I fixed this. Checking for '\.' is already more specific.

Modules/Module_Import/DCM2NII/xASL_imp_ReadSourceData.m
```diff
@@ -85,8 +85,12 @@ function xASL_imp_ReadSourceData_CheckLastElement(x)
     % List of known file extensions (formatted as regular expression)
     knownExtensions = '(zip|dcm|ima|xml|par|rec|nii|nii\.gz)';
 
-    hasExtension = length(lastElement)>1 && strcmp(lastElement(1:2), conditionExtension);
-    hasKnownExtension = ~isempty(regexpi(lastElement, knownExtensions, 'once'));
+    iExtension = regexp(lastElement, conditionExtension);
```
jan-petr, line 88: This won't work in general and without issues.
1) In case there's no '\.', iExtension will be []. It won't crash in this Matlab version, but I wouldn't rely on this behavior.
2) In case there are more than one '\.', it will create a warning that this will not work in the next Matlab release.

Modules/Module_Import/xASL_wrp_DCM2NII_Subject.m
jan-petr: It's not that I'm scared. It does brake something :) 
It suddenly names all visit/sessions and "ses-1". I don't think that this is a good collateral for correcting a minor issue with slightly incorrect printing of info in log files.

Modules/Module_Import/xASL_wrp_NII2BIDS_Subject.m
jan-petr: Yes - but by fixing a minor glitch in the output to the screen, you have altered all imports.

jan-petr (changes requested):
Minor errors

Modules/Module_Import/xASL_wrp_DCM2NII_Subject.m
HenkMutsaerts: agreed: default to run-1
agreed: session names vary, so don't default to session name

Modules/Module_Import/DCM2NII/xASL_imp_ReadSourceData.m
HenkMutsaerts: Fixed in 36bdea65090d2f409a5978f3786685c159c86a5e

Modules/Module_Import/xASL_wrp_DCM2NII_Subject.m
jan-petr: Perfect. Will you implement this, or shall I?

Modules/Module_Import/xASL_wrp_NII2BIDS_Subject.m
jan-petr: this is clarified - this is about session/visit, not session/run. So we fix this as agreed above. Should I do this? Or you will?

Modules/Module_Import/DCM2NII/xASL_imp_ReadSourceData.m
jan-petr: Better. But we probably want to use iExtension(end) rather than first. Assume
lastElement = 'longfile.withmany.extensions.nii';

also this won't work with .nii.gz unless we add ".gz" alone in thelist of known extensions.

Modules/Module_Import/xASL_wrp_DCM2NII_Subject.m
HenkMutsaerts: Current output of DICOM 2 NIFTI was:

`` Importing subject = GBR001011_: ``

Similar for NIFTI 2 BIDS.
Should be fixed now in 379c7ab9b385ab863ef32862e3b87d96fad818a6

Modules/Module_Import/xASL_wrp_NII2BIDS_Subject.m
HenkMutsaerts: Should be fixed now in 379c7ab9b385ab863ef32862e3b87d96fad818a6

Modules/Module_Import/DCM2NII/xASL_imp_ReadSourceData.m
HenkMutsaerts: That's the point, we want the whole extension. But I'll change it to what you want. Fixed in 379c7ab9b385ab863ef32862e3b87d96fad818a6

jan-petr: Yes, I know. but that's a bit tricky - last doesn't work, first doesn't work. So as long as we are detecting for '\.' it won't work perfectly. What I proposed is a workaround, I would have otherwise completely rehauled it...

Anyway - I'm happy with the current solution.

Modules/Module_Import/DCM2NII/xASL_imp_NII2BIDS_Run.m
```diff
@@ -32,7 +32,11 @@
 
     %% Print the run that is being converted
     xASL_adm_BreakString('CONVERT RUN');
-    fprintf('Converting subject %s, session %s, run %s, ', bidsLabel.subject, bidsLabel.visit, listRuns{iRun});
+    if ~isempty(bidsLabel.visit)
+        fprintf('Converting subject %s, session %s, run %s, ', bidsLabel.subject, bidsLabel.visit, listRuns{iRun});
+    else
+        fprintf('Converting subject %s, session %s, run %s, ', bidsLabel.subject, listRuns{iRun});
```
jan-petr: Good. But you forgot to remove the session :) on the second line :)

jan-petr (changes requested):
One small issue.

Modules/Module_Import/DCM2NII/xASL_imp_NII2BIDS_Run.m
HenkMutsaerts: Fixed in 6e742a3367c2eb6a750a7893de6e273276dfb4eb

jan-petr (approved):
OK

PR #2060, opened 2026-05-25
Closes #2059

### LINKED ISSUE #2059: TopUp disabling for bConcatenate
### Description

TopUp with multiple runs is typically applied to a single run only, which will reduce alignment between ASL runs. As alignment between ASL runs is more important than alignment of ASL with T1w, we temporary disable TopUp when multiple runs are concatenated, at least until this feature is fixed. The user can still override this by setting x.modules.asl.bTopUp

### Tasks

- [x] Create and manage bTopUp and x.modules.asl.bTopUp
- [x] if x.modules.asl.bMergingSessions, disable TopUp

### Release notes

Deactive TopUp for merged ASL sessions.

HenkMutsaerts:
### Linked issue

Closes #2059

### DIFF
--- a/Modules/xASL_module_ASL.m
+++ b/Modules/xASL_module_ASL.m
@@ -255,21 +255,41 @@ end
 
 
 %% ========================================================================================================================
-%% 1 TopUp (WIP, only supported if FSL installed)
+%% 1 TopUp (only supported if FSL installed)
 Path_RevPE = xASL_adm_GetFileList(x.dir.SESSIONDIR, '^(ASL4D|M0).*RevPE\.nii$', 'FPList', [0 Inf]);
 
+HasTopUpScans = xASL_exist(x.P.Path_M0, 'file') && ~isempty(Path_RevPE);
+
+if isfield(x.modules.asl, 'bTopUp')
+    bTopUp = x.modules.asl.bTopUp;
+    if bTopUp && ~HasTopUpScans
+        warning('TopUp requested but no TopUp scans detected');
+    end
+elseif HasTopUpScans
+    fprintf('\n%s\n', 'TopUp scans detected, enabling TopUp');
+    bTopUp = true;
+else
+    bTopUp = false;
+end
+
+if bTopUp && x.modules.asl.bMergingSessions
+        warning('TopUp for runs concatenation not implemented yet, disabled for now');
+        bTopUp = false;
+        % For now the default is to disable TopUp when bConcatenation
+end
+
 iState = 1;
-if xASL_exist(x.P.Path_M0,'file') && ~isempty(Path_RevPE)
+if bTopUp
     if ~x.mutex.HasState(StateName{iState}) || ~xASL_exist(fullfile(x.dir.SESSIONDIR, 'TopUp_fieldcoef.nii'),'file')
 
-        xASL_adm_DeleteFileList(x.dir.SESSIONDIR,'^(B0|Field|TopUp|Unwarped).*$',[],[0 Inf]); % delete previous TopUp stuff first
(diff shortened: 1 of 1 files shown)


Modules/xASL_module_ASL.m
```diff
@@ -255,21 +255,41 @@
...
 
 %% ========================================================================================================================
-%% 1 TopUp (WIP, only supported if FSL installed)
+%% 1 TopUp (only supported if FSL installed)
 Path_RevPE = xASL_adm_GetFileList(x.dir.SESSIONDIR, '^(ASL4D|M0).*RevPE\.nii$', 'FPList', [0 Inf]);
 
+HasTopUpScans = xASL_exist(x.P.Path_M0, 'file') && ~isempty(Path_RevPE);
+
+if isfield(x.modules.asl, 'bTopUp')
+    bTopUp = x.modules.asl.bTopUp;
```
jan-petr, line 264: Why do you define a new variable instead of keeping it inside x.modules.asl.bTopUp?

Modules/xASL_module_ASL.m
```diff
@@ -255,21 +255,41 @@
 
 
 %% ========================================================================================================================
-%% 1 TopUp (WIP, only supported if FSL installed)
+%% 1 TopUp (only supported if FSL installed)
 Path_RevPE = xASL_adm_GetFileList(x.dir.SESSIONDIR, '^(ASL4D|M0).*RevPE\.nii$', 'FPList', [0 Inf]);
 
+HasTopUpScans = xASL_exist(x.P.Path_M0, 'file') && ~isempty(Path_RevPE);
+
+if isfield(x.modules.asl, 'bTopUp')
```
jan-petr: also check if it is not empty

Modules/xASL_module_ASL.m
```diff
@@ -255,21 +255,41 @@
...
+        % For now the default is to disable TopUp when bConcatenation
+end
+
 iState = 1;
-if xASL_exist(x.P.Path_M0,'file') && ~isempty(Path_RevPE)
+if bTopUp
     if ~x.mutex.HasState(StateName{iState}) || ~xASL_exist(fullfile(x.dir.SESSIONDIR, 'TopUp_fieldcoef.nii'),'file')
 
-        xASL_adm_DeleteFileList(x.dir.SESSIONDIR,'^(B0|Field|TopUp|Unwarped).*$',[],[0 Inf]); % delete previous TopUp stuff first
+        xASL_adm_DeleteFileList(x.dir.SESSIONDIR,' ^(B0|Field|TopUp|Unwarped).*$',[],[0 Inf]); % delete previous TopUp stuff first
```
jan-petr: a bad idea to add the extra space inside the string.

jan-petr (changes requested):
Small details

Modules/xASL_module_ASL.m
HenkMutsaerts: Because x.modules.asl.bTopUp is the general setting (for the whole population), and bTopUp is the local setting (e.g., an individual subject may not have images for TopUp)

Modules/xASL_module_ASL.m
HenkMutsaerts: Fixed in e87f67583f21ed17833ffdfcaa618d5c2cf62803

Modules/xASL_module_ASL.m
HenkMutsaerts: Fixed in e87f67583f21ed17833ffdfcaa618d5c2cf62803

jan-petr (approved):
OK

PR #2065, opened 2026-06-02
#2061 DEBBIE import fix

### LINKED ISSUE #2061: PLD order inconsistent with visual order
### Description

There was a minor issue with DEBBIE import due to anonymization and one tagged was missed. All that's described below is because of the minor import error. Now fixed.

sub-SYN025 ASL_1 works well, but ASL_2 has an artifact-appearing PWI used for registration. This seems to be originating from the fact that the PLD vector of the PWIs of this HAD8 are inconsistent with their visual appearance.

Adding a breakpoint at line 312 xASL_im_ASLSubtractionAveraging:

Image

x.Q.InitialPLD_PWI4D AFTER xASL_quant_HadamardDecoding =
         600
        1000
        1400
        1800
        2200
        2600
        3000
         800
        1200
        1600
        2000
        2400
        2800
        3200

x.Q.InitialPLD_PWI4D BEFORE xASL_quant_HadamardDecoding =

       3400
         600
        1000
        1400
        1800
        2200
        2600
        3000
        3600
         800
        1200
        1600
        2000
        2400
        2800
        3200

### Tasks

- [x] @jan-petr It seems that the PWI4D (or PWI3D) images match with the PLD order BEFORE Hadamard decoding, not AFTER Hadamard decoding?

But then again, this also doesn't match fully...
The contrast is really weird, some instances could indeed be early/vascular, some late/noisy, but there are also instances that show only WM signal; could this be a background suppression issue?

E.g., this is one of the best-appearing ones within this AD cohort:
Image

### Release notes

Minor fix of DEBBIE import.

jan-petr:
### Linked issue

Close #2061

### DIFF
--- a/Modules/Module_Import/DCM2NII/xASL_imp_DCM2NII_ReorderTimeEncoded.m
+++ b/Modules/Module_Import/DCM2NII/xASL_imp_DCM2NII_ReorderTimeEncoded.m
@@ -45,7 +45,8 @@ if numel(nii_files)>=1
 				numberPLDs = timeEncodedMatrixSize;
 				numberRepetitions = int32(size(imASL,4)/numberTEs/numberPLDs);
 				% Check if the PLDs are interleaved or just repeated
-				% This is not done for gammastar version
+				% This is only relevant for non gammastar sequence versions. For gammaStar, this step can be skipped. 
+				% Note that there are two different DICOM tags that can be used to identify gammaStar - both are checked in case some were removed during data anonymization
 				if (~isfield(resultJSON, 'PulseSequenceDetails') || isempty(regexpi(resultJSON.PulseSequenceDetails, 'fme_gammastar', 'once'))) &&...
 			       (~isfield(resultJSON, 'SequenceName') || isempty(regexpi(resultJSON.SequenceName, 'gammastar', 'once')))
 					if numel(vectorPLD) > numberPLDs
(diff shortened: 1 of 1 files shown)


Modules/Module_Import/DCM2NII/xASL_imp_DCM2NII_ReorderTimeEncoded.m
```diff
@@ -46,7 +46,8 @@
 				numberRepetitions = int32(size(imASL,4)/numberTEs/numberPLDs);
 				% Check if the PLDs are interleaved or just repeated
 				% This is not done for gammastar version
-				if ~isfield(resultJSON, 'PulseSequenceDetails') || isempty(regexpi(resultJSON.PulseSequenceDetails, 'fme_gammastar', 'once'))
+				if (~isfield(resultJSON, 'PulseSequenceDetails') || isempty(regexpi(resultJSON.PulseSequenceDetails, 'fme_gammastar', 'once'))) &&...
```
HenkMutsaerts, line 50: Thanks, please improve the comment, now it is unclear why you do this.

So, if I understand correctly; there are two gammaSTAR versions with different PLD ordering:

Ordering 1: PulseSequenceDetails = fme_gammastar
Ordering 2: SequenceName = gammastar

Is this correct? I understand that it may be complicated but just put a bird-eyed view explanation like this in the comment above this change

HenkMutsaerts (approved):
See comment to please improve the code comment :)

PR #2073, opened 2026-06-10
Closes #2072

### LINKED ISSUE #2072: Minor improvements
### Description

Minor improvements in warnings/cosmetics

### Tasks

- [x] xASL_imp_ReadSourceData: regexp->strfind
- [x] xASL_wrp_DCM2NII_Subject: clarity warning duplicate scans
- [x] xASL_qc_SortBySpatialCoV: play nice with xASL_TrackProgress

### Release notes

Minor fixes in import.

HenkMutsaerts:
### Linked issue

Closes #2072

### DIFF
--- a/Modules/Module_Import/DCM2NII/xASL_imp_ReadSourceData.m
+++ b/Modules/Module_Import/DCM2NII/xASL_imp_ReadSourceData.m
@@ -79,19 +79,20 @@ function xASL_imp_ReadSourceData_CheckLastElement(x)
     % Get last element
     lastElement = lower(x.modules.import.imPar.folderHierarchy{end});
 
-    % Condition for file extension
-    conditionExtension = '\.';
-    
     % List of known file extensions (formatted as regular expression)
-    knownExtensions = '(zip|dcm|ima|xml|par|rec|nii|gz|nii\.gz)';
+    knownExtensions = '(zip|dcm|ima|xml|par|rec|nii|gz|nii\.gz|nii\\\.gz)';
 
-    iExtension = strfind(lastElement, conditionExtension);
-
-    if isempty(iExtension)
-        extensionIs = char;
-    else
-        extensionIs = lastElement(iExtension(end)+1:end);
-    end
+    iExtension = regexp(lastElement, '\\\.'); % First test extension defined properly for regexp as \.ext
+	if ~isempty(iExtension)
+		extensionIs = lastElement(iExtension(end)+2:end); % Need to count two characters \. before the extension
+	else
+		iExtension = regexp(lastElement, '\.'); % Afterwards, look for an extension defined .ext
+		if isempty(iExtension)
+			extensionIs = char;
+		else
+			extensionIs = lastElement(iExtension(end)+1:end); 
+		end
+	end
 
     hasExtension = ~isempty(extensionIs);
     hasKnownExtension = ~isempty(regexpi(extensionIs, knownExtensions, 'once'));
@@ -110,6 +111,5 @@ function xASL_imp_ReadSourceData_CheckLastElement(x)
               warning('Unknown extension in the last element of the folder hierarchy (%s)...',lastElement);
         end
     end
-
 
 end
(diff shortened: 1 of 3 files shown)


Modules/Module_Import/DCM2NII/xASL_imp_ReadSourceData.m
```diff
@@ -85,7 +85,7 @@ function xASL_imp_ReadSourceData_CheckLastElement(x)
     % List of known file extensions (formatted as regular expression)
     knownExtensions = '(zip|dcm|ima|xml|par|rec|nii|gz|nii\.gz)';
 
-    iExtension = regexp(lastElement, conditionExtension);
+    iExtension = strfind(lastElement, conditionExtension);
```
jan-petr: You can't use '\\.' with strfind as it looks for this exact string. 
Equivalent of regexp('file.nii','\\.') is strfind ('file.nii','.')
and strfind('filen.nii','\\.') won't return anything

Modules/Module_Import/xASL_wrp_DCM2NII_Subject.m
```diff
@@ -108,9 +108,18 @@
             
             %% 4. Iterate over scans
 			% Safety-check, we can't allow multiple matches for the same sequence
-			if numel(thisRun.scanIDs) ~= numel(unique(thisRun.scanIDs))
-				fprintf('Detected duplicate scans: %s\n',strjoin(thisRun.scanIDs, ', '));
-				error('We do not allow multiple token matches per scantype. Please adjust sourcestructure.json');
+			uniqueScanIDs = unique(thisRun.scanIDs);
```
jan-petr: This is a bit too complicated and can be simplified.
Try this testing scenario: thisRun.scanIDs = {'a','a','b','c','c','d'};
And replace your code with the following:

[uniqueScanIDs, ~, uniqueIndex] = unique(thisRun.scanIDs);
duplicateScans = cell(0);
if numel(thisRun.scanIDs) ~= numel(uniqueScanIDs)
                % Get the duplicate scanIDs
                for iID=1:numel(uniqueIndex)
                    if sum(uniqueIndex==iID)>1
                         duplicateScans{end+1} = uniqueScanIDs{iID};
                     end
                end
end

jan-petr (changes requested):
One error
one simplification

Modules/Module_Import/DCM2NII/xASL_imp_ReadSourceData.m
HenkMutsaerts: I don't get it. The problem was that we were looking for indices of any '.' whereas we only want indices of '\.' which clearly indicates a period used for an extension (in regular expression of the last element).

As in:

lastElement = 'jan.petr\\.nii';
conditionExtension = '\\.';
iExtension = regexp(lastElement, conditionExtension); -> [4 10] (gives any period in the sourceStructure regular expression, I don't want this)
iExtension = strfind(lastElement, conditionExtension); -> 9 (I only want the explicit '\\.' in the regular expression in sourceStructure

Modules/Module_Import/xASL_wrp_DCM2NII_Subject.m
HenkMutsaerts: Fixed in 46e83b383a4c3c0d864ffc4b3fa72f859d6a5056

Modules/Module_Import/DCM2NII/xASL_imp_ReadSourceData.m
jan-petr: Oops. The GUI escaped the characters in the comment. Read it again and this one as well.
Irrespective of this - the code was wrong.
regexp('file.nii','\\.') and strfind('file.nii','.') is fully equivalent.

But conditionExtension = '\\.'
so we are looking for '\\.' . This doesn't work with strfind.

Previously it was OK, now it's wrong

HenkMutsaerts: I had the same problem :) so read my comment again, and this one. I of course know the behavior of regexp vs strfind, and previously it was wrong, now it's ok.

So we want to search for the string '\\.' Not for the string '.' In the sourceStructure regular expression, a period '.' just means "anything", we don't care for that. We care for somebody potentially defining an extension (using '\\.')

jan-petr (approved):
OK

PR #2077, opened 2026-06-18
Closes #2076

### LINKED ISSUE #2076: Registration regularization
### Description

Some extreme ASL cases may lead to a 90 degrees rotation, when registering to T1w. This was found in the FACEHBI cohort, where some of these 3D PASL cases are a pure angiogram.

This leads me to thinking we should in general restrict the amount of rotations `spm_coreg` is allowed to do. With chatGPT, I have hacked `spm_coreg`. But @jan-petr you may want to check if this is fine, and/or if we need to stretch the allowed rotations.

I have hacked `spm_coreg` for this, as we call spm_coreg via the spm_jobman, and that makes it tricky to add extra flags. 

And restricting extreme rotations is something we want anyway. (we could also consider restricting extreme translations)

If you agree, I will also add this to `/External/README_SPMmodified.txt`

### Tasks

- [x] `spm_coreg`: flags.rotlim to limit rotations
- [x] Set to 20 degrees maximal rotations.

### How to test

Tested using 4 FACEHBI cases, but may need to stretch the allowed degrees.
See for example FACEHBI cases:
sub-15380711_ses-5
sub-17103413_ses-5
sub-17132877_ses-2
sub-17411419_ses-5

In cases where PWI-based registration is enforced in angiogram-appearing CBF maps, this now results in 20 degrees rotation instead of a full 90 degrees rotation. So it doesn't fix anything, it just limits how bad it gets.

Image

### Release notes
Limit maximal rotation between ASL and PWI.

HenkMutsaerts:
### Linked issue

Closes #2076

### DIFF
--- a/External/README_SPMmodified.txt
+++ b/External/README_SPMmodified.txt
@@ -173,8 +173,8 @@ IMAGE PROCESSING IMPROVEMENT
 
 DATE+NAME:2026_06_29, HM
 DESCRIPTION: Maximize spm_coreg rotations to 20 degrees
-FILE:
-spm_coreg.m @ 139, 159, 170, 185, 264
+FILE: spm_coreg.m @ 139, 159, 170, 185, 264
+ISSUE: #2076
 
 DATE+NAME:2020_07_01, HM
 DESCRIPTION: Remove feedback missing log-file (we don't use catlog_txt in xASL)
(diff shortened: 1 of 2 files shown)


jan-petr (approved):
Tested. Checked.

PR #2081, opened 2026-06-25
#2080 xASL_init_SubjectList fix subjectRegexp and inclusion logic

### LINKED ISSUE #2080: Fix BIDS2Legacy crash with ForceInclusionList and respect user-provided subjectRegexp
### Description

Two inter-related bugs in `xASL_init_SubjectList.m` prevent subject/session subsetting from BIDS rawdata.

**Bug 1 — BIDS2Legacy crash when `ForceInclusionList` is set with `/rawdata` being present:**

When `ForceInclusionList` is set in `dataPar.json` and `/rawdata` exists, `xASL_init_SubjectList` skips loading the BIDS layout. However, `ExploreASL_Process` still runs `xASL_module_BIDS2Legacy` because `bReadRawdata=true`, causing a crash at `xASL_module_BIDS2Legacy.m:91` due to the missing `x.modules.bids2legacy.BIDS` field.

Three mutually exclusive branches in `xASL_init_SubjectList.m` (lines 50, 66, 91):

| Branch | Condition | What it does |
|--------|-----------|-------------|
| 1 (line 50) | `isfield(x.dataset,'ForceInclusionList')` | Sets `TotalSubjects` directly, skips `bids.layout()` |
| 2 (line 66) | `elseif x.opts.bReadRawdata` | Calls `bids.layout()`, populates `x.modules.bids2legacy.BIDS` |
| 3 (line 91) | `else` | Scans derivatives folder for legacy subject dirs |

When `/rawdata` exists, `xASL_init_DataLoading.m:61` sets `bReadRawdata=true`. Setting `ForceInclusionList` triggers Branch 1, which skips `bids.layout()`. But `ExploreASL_Process.m:91-100` gates BIDS2Legacy on `bReadRawdata` (not on `bProcess` or `ForceInclusionList`), so BIDS2Legacy still runs and dereferences the missing `.BIDS` field at line 91.

This makes it impossible to process a subject/session subset from `/rawdata` without error. There is no other mechanism to select a subset of rawdata subjects — Branch 2 loads ALL subjects from the BIDS layout with no filtering.

**Bug 2 — `subjectRegexp` unconditionally overwritten in BIDS mode:**

In Branch 2 (line 84), `x.dataset.subjectRegexp` is set to `'^sub-.*$'` inside the per-subject loop, unconditionally overwriting any user-provided value:

```matlab
for iSubj=1:length(x.modules.bids2legacy.BIDS.subjects)
    ...
    x.dataset.TotalSubjects{iSubj} = [subjectName '_' visitName];
    x.dataset.subjectRegexp = '^sub-.*$'; % defaulting to BIDS subject regexp  % line 84
end
```

This is problematic:
- Any user-provided `subjectRegexp` in `dataPar.json` is silently discarded in BIDS mode
- The assignment is inside the loop (redundant — same value every iteration)
- The comment says "defaulting" but the assignment is unconditional, not a default

In the derivatives branch (Branch 3, line 108), `subjectRegexp` is respected and used to filter subjects via `xASL_adm_GetFileList`. BIDS mode should offer the same user control.

**Index coupling note:**

`xASL_module_BIDS2Legacy.m:73` computes `iSubjSess = find(strcmp(x.SUBJECTS, x.SUBJECT))` and uses it to index into `BIDS.subjects(iSubjSess)` (lines 91-92). This assumes 1:1 positional alignment between `x.SUBJECTS` and `BIDS.subjects`. Any fix must preserve this alignment — if `TotalSubjects` is filtered, `BIDS.subjects` must be filtered identically.

Both `xASL_init_SubjectList.m` and `xASL_module_BIDS2Legacy.m` only had license header updates since Sep 2024. These are pre-existing logic conflicts, not regressions.

### Tasks

- [x] Fix `xASL_init_SubjectList` to always call `bids.layout()` when `bReadRawdata=true`, even when `ForceInclusionList` is set
- [x] When `ForceInclusionList` is set and `bReadRawdata=true`, filter the `BIDS.subjects` array to only include entries matching `ForceInclusionList`, preserving 1:1 positional alignment with `x.SUBJECTS`
- [x] The `ForceInclusionList` matching must reconstruct the legacy name (`sub-XXX_N`) from `BIDS.subjects(i).name` + `BIDS.subjects(i).session` to compare against `ForceInclusionList` entries (which are in legacy format)
- [x] Move the `subjectRegexp` default assignment outside the per-subject loop (line 84) and only set it to `'^sub-.*$'` if the user has not already provided a `subjectRegexp` in `dataPar.json`
- [x] When `ForceInclusionList` is absent and `bReadRawdata=true`, apply the user-provided `subjectRegexp` as a filter on `TotalSubjects` (and `BIDS.subjects` identically) after the BIDS-to-legacy conversion, consistent with derivatives-mode behavior at line 108
- [x] Ensure `x.modules.bids2legacy.BIDS.subjects` and `x.dataset.TotalSubjects` remain length-aligned and position-aligned after any filtering
- [x] Verify the `rawdata`-absent path is unaffected (`xASL_init_DataLoading.m` sets `bReadRawdata=false` when no `/rawdata` exists, so BIDS2Legacy never runs and `ForceInclusionList` works with derivatives-only processing)
- [x] Standard Flavor and UnitTests run fine.

### How to test

**Test 1 — ForceInclusionList with rawdata (crash repro):**
1. Create a BIDS dataset with multiple subjects in `/rawdata`
2. Set `ForceInclusionList` in `dataPar.json` to a subset of subjects (legacy format `sub-XXX_N`)
3. Run `ExploreASL` with `bProcess = [1 1 0]` (Structural + ASL)
4. Verify `xASL_module_BIDS2Legacy` completes without error for only the selected subjects
5. Verify `x.modules.bids2legacy.BIDS.subjects` length matches `x.SUBJECTS` length (positional alignment)

**Test 2 — User-provided subjectRegexp with rawdata:**
1. Create a BIDS dataset with subjects matching different patterns (e.g., `sub-C9ORF007Philips`, `sub-C9ORF059Siemens`)
2. Set `subjectRegexp` in `dataPar.json` to a filter pattern (e.g., `^sub-C9ORF007.*$`) — do NOT set `ForceInclusionList`
3. Run `ExploreASL` with `bProcess = [1 1 0]`
4. Verify only subjects matching the regexp are processed
5. Verify `BIDS.subjects` is filtered to match `TotalSubjects` (alignment preserved)
6. Verify the user-provided `subjectRegexp` is not overwritten by `'^sub-.*$'`

**Test 3 — No ForceInclusionList, no subjectRegexp (default behavior, regression check):**
1. Create a BIDS dataset with multiple subjects in `/rawdata`
2. Do NOT set `ForceInclusionList` or `subjectRegexp` in `dataPar.json`
3. Run `ExploreASL` with `bProcess = [1 1 0]`
4. Verify all subjects are processed (default `'^sub-.*$'` applied)
5. Verify no regression vs. current behavior

**Test 4 — ForceInclusionList without rawdata (derivatives-only, regression check):**
1. Use a dataset with only `/derivatives/ExploreASL` (no `/rawdata`)
2. Set `ForceInclusionList` in `dataPar.json`
3. Run `ExploreASL`
4. Verify existing derivatives-mode behavior is unaffected

**Test 5 — Parallel workers (nWorkers > 1):**
1. Repeat Test 1 with `nWorkers = 2`
2. Verify both workers process their subject partitions without crash
3. Verify `BIDS.subjects` alignment holds per-worker (each worker builds its own `BIDS` layout, then partitions by `iWorker`/`nWorkers` via `xASL_init_Parallelization.m`)

### Release notes

Fixed a crash when using `ForceInclusionList` with rawdata processing. Additionally, `subjectRegexp` is no longer unconditionally overwritten in BIDS mode.

MauricePasternak:
### Linked issue

Closes #2080

### DIFF
--- a/Modules/Module_Processing/xASL_init_SubjectList.m
+++ b/Modules/Module_Processing/xASL_init_SubjectList.m
@@ -47,7 +47,64 @@ x.ROOT = subjectFolder;
 %% 2. Advanced parameter to enforce a subject list, without querying folder names
 x.dataset.TotalSubjects = cell(0);
 
-if isfield(x.dataset,'ForceInclusionList')
+if x.opts.bReadRawdata
+
+    %% 3. Load the BIDS structure of all subjects & translate to ExploreASL legacy
+    x.modules.bids2legacy.BIDS = bids.layout(subjectFolder);
+    
+    % Move the subjectRegexp default assignment outside the loop and respect user-provided one
+    if ~isfield(x.dataset, 'subjectRegexp')
+        x.dataset.subjectRegexp = '^sub-.*$';
+    else
+        % Do escapes and fixes (consistent with derivatives behavior)
+        x.dataset.subjectRegexp = strrep(x.dataset.subjectRegexp,'\\','\');
+        x.dataset.subjectRegexp = strrep(x.dataset.subjectRegexp,'$','');
+        x.dataset.subjectRegexp = [x.dataset.subjectRegexp '(|_\d+)$'];
+    end
+
+    % Now we translate the BIDS naming convention to ExploreASL's legacy name convention
+    translatedSubjects = cell(0);
+    for iSubj=1:length(x.modules.bids2legacy.BIDS.subjects)
+        subjectName = x.modules.bids2legacy.BIDS.subjects(iSubj).name;
+        visitName = x.modules.bids2legacy.BIDS.subjects(iSubj).session;
+        if isempty(visitName)
+            visitName = '1'; % we default to visit _1
+        elseif ~strcmp(visitName(1:4), 'ses-')
+            warning(['Illegal BIDS session definition for ' subjectName '_' visitName]);
+        else
+            visitName = visitName(5:end); % if we find a BIDS session, we add it as visit suffix in ExploreASL legacy format
+        end
+        
+        translatedSubjects{iSubj, 1} = [subjectName '_' visitName];
+    end
+
+    % Filter subjects if ForceInclusionList or subjectRegexp is set
+    keepIndices = true(length(translatedSubjects), 1);
+    if isfield(x.dataset, 'ForceInclusionList')
+        warning('Using custom list of subjects, on your own risk');
+        for iSubj = 1:length(translatedSubjects)
+            keepIndices(iSubj) = any(strcmp(translatedSubjects{iSubj}, x.dataset.ForceInclusionList));
+        end
+    elseif isfield(x.dataset, 'subjectRegexp')
+        for iSubj = 1:length(translatedSubjects)
+            keepIndices(iSubj) = ~isempty(regexp(translatedSubjects{iSubj}, x.dataset.subjectRegexp, 'once'));
+        end
+    end
+
+    % Apply filter to keep 1:1 positional alignment
+    keepIndices = find(keepIndices);
+    x.modules.bids2legacy.BIDS.subjects = x.modules.bids2legacy.BIDS.subjects(keepIndices);
+    x.dataset.TotalSubjects = translatedSubjects(keepIndices);
+
+    if isempty(x.dataset.TotalSubjects)
+        if isfield(x.dataset, 'ForceInclusionList')
+            error('No subjects found matching ForceInclusionList, check x.dataset.ForceInclusionList in dataPar.json');
+        else
+            error('No subjects found, check your rawdata folder for BIDS compatibility or subjectRegexp');
+        end
+    end
+
+elseif isfield(x.dataset,'ForceInclusionList')
     % This is an option if you want to select subjects yourself,
     % instead of using all the subjects that comply with the regular expression
     x.dataset.TotalSubjects = x.dataset.ForceInclusionList(:);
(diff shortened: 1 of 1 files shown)


Modules/Module_Processing/xASL_init_SubjectList.m
```diff
@@ -47,28 +47,23 @@
...
     if ~isfield(x.dataset, 'subjectRegexp')
         x.dataset.subjectRegexp = '^sub-.*$';
-        warning('Ensure that the ForceInclusionList is in accordance with the BIDS subjectRegexp');
-        fprintf('%s\n', 'Otherwise, this will go wrong further down the road');
-    end
-
-    if isempty(x.dataset.TotalSubjects)
-        error('No subjects found, check x.dataset.ForceInclusionList in dataPar.json');
+    else
+        % Do escapes and fixes (consistent with derivatives behavior)
```
jan-petr, line 59: Do we really need to escape those instead of leaving that as the responsibility of the user? What if the users adds the additional escapes?

jan-petr (changes requested):
One comment about the necessity to escape subjectRegexp.
Apart from that, the code looks fine and I'll just quickly run our basic tests.

Modules/Module_Processing/xASL_init_SubjectList.m
MauricePasternak: This logic was present in `develop`. I think it should stay since it provides a consistency pattern for users who come from different languages (i.e. in Python we are so used to the raw string utility that we never double-escape).

Both `'sub-\\d+` and `'sub-\d+'` would be correctly interpreted.

As for the edge case that someone really, really needs the literal backslash in there, I don't think this is an issue in practice:
- on Windows, it's an illegal character for folder names.
- on Unix it is highly non-standard and mostly likely a typo someone made in editing ExploreASL manually. If anything, this would help bring their attention to the non-standard practice.

In other words, I think the convenience outweighs one very unlikely edge case.

jan-petr: Sorry. It's of course all correct. Obviously, we won't ever have backslashes in the folder names. 

You are replacing '\\\\' with '\\'. And I read the comment "do escapes" and thought you are escaping all '\\' and not the other way around. 

So all good and we keep this.

jan-petr (approved):
All is OK for me. And all tests completed successfully. Should I merge it right away?

PR #2090, opened 2026-07-09
Closes #2089 xASL_init_Parallelization doesn't filter x.SUBJECTS

### LINKED ISSUE #2089: xASL_init_Parallelization doesn't filter x.SUBJECTS
### Description

`xASL_init_Parallelization` truncates `x.dataset.TotalSubjects` to the subset of subjects assigned to the current worker, but it never updates `x.SUBJECTS`. Later, `xASL_init_Iteration` iterates over `x.SUBJECTS`, which still contains the full subject list. So every worker attempts to process every subject. This generates a flood of "mutex is locked" errors in log files from non-owning workers repeatedly failing to acquire the lock.

### Tasks

- [x] Filter `x.SUBJECTS` and update `x.dataset.nSubjects` in `xASL_init_Parallelization` after truncating `x.dataset.TotalSubjects`

### How to test

Run ExploreASL with `nWorkers > 1` on a dataset with multiple subjects. Before the fix, every worker's log shows all subjects plus "mutex is locked" errors for subjects owned by other workers. After the fix, each worker should process only its assigned subjects, with no cross-worker mutex conflicts.

### Release notes

Fix iWorker/nWorkers subject partitioning: `x.SUBJECTS` was never filtered after parallelization, causing all workers to attempt processing all subjects and flooding logs with mutex-locked errors.

MauricePasternak:
Closes #2089

### DIFF
--- a/Modules/Module_Processing/xASL_init_Parallelization.m
+++ b/Modules/Module_Processing/xASL_init_Parallelization.m
@@ -16,6 +16,10 @@ function [x] = xASL_init_Parallelization(x)
 %               started multiple times on a server, or alternatively running the
 %               ExploreASL compilation, and doesn't require the Matlab parallel toolbox.
 %
+%               IMPORTANT: xASL_init_Iteration processes x.SUBJECTS (not
+%               x.dataset.TotalSubjects). Both lists must be sliced for this worker,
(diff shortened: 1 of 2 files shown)


jan-petr (approved):
OK

PR #2093, opened 2026-07-12
Bug #2092 fix debbie population

### LINKED ISSUE #2092: Missing Lesions can cause an error
### Description

A few issues are crashing DEBBIE processing with variable number of sessions and Lesions.

1. For subject-specific ROI processing, we need to load a lesion to initialize the structures. So the lesion for the first subject was loaded. However, subject specific Lesions do not need to be given for all subjects, so this can crash when the first subject has it missing (which is not incorrect).

### Tasks

- [x] Deal with ROI-stats when Lesions are missing for the first session.

### Release notes

Fix ROI-stats when the Lesions are missing for the first subject.

jan-petr:
### Linked issue

Close #2092

### DIFF
--- a/Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
+++ b/Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
@@ -107,6 +107,7 @@ if nSessions~=x.dataset.nSessions
     warning('Not all sessions/runs seem correctly processed in the ASL module');
 end
 
+%%% Dealing with native space specific issues
 if x.S.InputNativeSpace
 	% For native space, it could happen that ROIs/Atlases are not saved for all subjects/sessions, so while setting up, 
 	% we need to load the first existing one
@@ -283,6 +284,7 @@ fprintf('%s\n',['Preparing ROI-based ' x.S.output_ID ' statistics:']);
         x.S.MinimalROIVolume = 1; % 1 mL
     end
     
+	% Initialize the voxel size, which is always the same for standard space, but study-specific for native space
     if x.S.InputNativeSpace
 		VoxelSize = xASL_io_ReadNifti(fullfile(x.dir.xASLDerivatives,x.SUBJECTS{indExistSubject},listSessions{indExistSession},[x.S.InputAtlasNativeName '.nii']));
 		VoxelSize = [norm(VoxelSize.mat(1:3,1)), norm(VoxelSize.mat(1:3,2)), norm(VoxelSize.mat(1:3,3))];
--- a/Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
+++ b/Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
@@ -108,8 +108,25 @@ if nSessions~=x.dataset.nSessions
 end
 
 if x.S.InputNativeSpace
+	% For native space, it could happen that ROIs/Atlases are not saved for all subjects/sessions, so while setting up, 
+	% we need to load the first existing one
+	indExistSubject = 0;
+	bExistingSession = 0;
+
+	while (indExistSubject < x.dataset.nSubjects) && ~bExistingSession
+		% Go through subjects and sessions until we find an existing atlas/ROI
+		indExistSubject = indExistSubject + 1;
+		indExistSession = 0;
+		while (indExistSession < x.dataset.nSessions) && ~bExistingSession
+			indExistSession = indExistSession + 1;
+			if xASL_exist(fullfile(x.dir.xASLDerivatives,x.SUBJECTS{indExistSubject},listSessions{indExistSession},[x.S.InputAtlasNativeName '.nii']), 'file')
+				bExistingSession = 1;
+			end
+		end
+	end
+
 	% Native space
-	x.S.masks.WBmask = xASL_io_Nifti2Im(fullfile(x.dir.xASLDerivatives,x.SUBJECTS{1},listSessions{1},[x.S.InputAtlasNativeName '.nii']));
+	x.S.masks.WBmask = xASL_io_Nifti2Im(fullfile(x.dir.xASLDerivatives,x.SUBJECTS{indExistSubject},listSessions{indExistSession},[x.S.InputAtlasNativeName '.nii']));
 	x.S.masks.WBmask = sum(x.S.masks.WBmask,4) > 0;
 	x.LeftMask = xASL_io_Nifti2Im(fullfile(x.dir.xASLDerivatives,x.SUBJECTS{1},listSessions{1},'LeftRight_Atlas.nii'));
 	x.LeftMask = (x.S.masks.WBmask .* (x.LeftMask == 1)) > 0;
@@ -202,7 +219,7 @@ namesROIlocal = x.S.NamesROI;
 
 %% 0.b Native space atlas input
 if x.S.InputNativeSpace
-	inputAtlasTmp = xASL_io_Nifti2Im(fullfile(x.dir.xASLDerivatives,x.SUBJECTS{1},listSessions{1},[x.S.InputAtlasNativeName '.nii']));
+	inputAtlasTmp = xASL_io_Nifti2Im(fullfile(x.dir.xASLDerivatives,x.SUBJECTS{indExistSubject},listSessions{indExistSession},[x.S.InputAtlasNativeName '.nii']));
 	if x.S.bSubjectSpecificROI
 		% For subject specific atlases - they are 4D probabilistic PV maps
 		atlasN = size(inputAtlasTmp, 4);
@@ -267,7 +284,7 @@ fprintf('%s\n',['Preparing ROI-based ' x.S.output_ID ' statistics:']);
     end
     
     if x.S.InputNativeSpace
-		VoxelSize = xASL_io_ReadNifti(fullfile(x.dir.xASLDerivatives,x.SUBJECTS{1},listSessions{1},[x.S.InputAtlasNativeName '.nii']));
+		VoxelSize = xASL_io_ReadNifti(fullfile(x.dir.xASLDerivatives,x.SUBJECTS{indExistSubject},listSessions{indExistSession},[x.S.InputAtlasNativeName '.nii']));
 		VoxelSize = [norm(VoxelSize.mat(1:3,1)), norm(VoxelSize.mat(1:3,2)), norm(VoxelSize.mat(1:3,3))];
 	else
 		VoxelSize = [1.5 1.5 1.5];


Functions/BIDS/xASL_bids_Add2ParticipantsTSV.m
```diff
@@ -70,8 +70,12 @@ function xASL_bids_Add2ParticipantsTSV(DataIn, DataName, x, bOverwrite, PathTSV)
         % First get the parameter name
         indicesAre = regexp(DataName, '_');
         indicesAre(end+1) = length(DataName);
-        parameterName = DataName(1:indicesAre(1));
-
+		if strcmpi(DataName(1:3), 'PV_')
```
HenkMutsaerts: Why do you need this exception for PV parameters?

Functions/Statistics/xASL_stat_PrintStats.m
```diff
@@ -108,7 +108,7 @@
 %% 3) Define number of sessions, force to 1 in case of TT or volume metrics
 SingleSessions = {'volume' 'TT' 'PV_pGM' 'PV_pWM' 'pGM' 'pWM' 'mrc1T1' 'mrc2T1'};
 
-HasSingleSessionOnly = sum(cellfun(@(y) ~isempty(regexpi(y,['^' x.S.output_ID])), SingleSessions));
+HasSingleSessionOnly = sum(cellfun(@(y) ~isempty(regexpi(y, ['^' x.S.output_ID], 'once')), SingleSessions));
```
HenkMutsaerts: Is this really correct? We do the sum of something, 'once' will avoid it looking further?

Functions/Statistics/xASL_stat_PrintStats.m
```diff
@@ -183,7 +183,7 @@
                 fprintf(2, ['Could not find session for ' x.S.SubjectSessionID{iSubjSess} '\n']);
 			else
 				% If session ID was identified, we have to double-check that session ID is not part of subject ID and exclude if necessary
-				SubjectID = x.S.SubjectSessionID{iSubjSess}(startSubjectIndex:min(endSubjectIndex, startSessionIndex-2));
+				SubjectID = x.S.SubjectSessionID{iSubjSess}(startSubjectIndex(1):min(endSubjectIndex, startSessionIndex-2));
```
HenkMutsaerts: Why do you have more than 1 startSubjectIndex in the first place, is the subject ID not illegal?

Functions/Statistics/xASL_stat_PrintStats.m
```diff
@@ -392,15 +390,15 @@
     if bPrintSessions
         printMatrix = x.S.SetsID;
         optionsMatrix = x.S.SetsOptions;
-        sampleMatrix = x.S.Sets1_2Sample;
+        %sampleMatrix = x.S.Sets1_2Sample;
```
HenkMutsaerts: Why?

Functions/Statistics/xASL_stat_PrintStats.m
```diff
@@ -392,15 +390,15 @@
...
     else
         SessionColumn = find(strcmpi(x.S.SetsName, 'session'));
         vector2Print = 1:size(x.S.SetsName,2);
         vector2Print = vector2Print(vector2Print~=SessionColumn);
         
         printMatrix = x.S.SetsID(:, vector2Print);
         optionsMatrix = x.S.SetsOptions(:, vector2Print);
-        sampleMatrix = x.S.Sets1_2Sample(:, vector2Print);
+        %sampleMatrix = x.S.Sets1_2Sample(:, vector2Print);
```
HenkMutsaerts: Why?

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
```diff
@@ -108,8 +108,25 @@
 end
 
 if x.S.InputNativeSpace
+	% For native space, it could happen that ROIs/Atlases are not saved for all subjects/sessions, so while setting up,
```
HenkMutsaerts, line 112: Just note that (sorry for saying), you're making it increasingly more difficult to get rid of the space dependency of this function, if you add yet another space-dependent part?

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
```diff
@@ -202,7 +219,7 @@
 
 %% 0.b Native space atlas input
 if x.S.InputNativeSpace
-	inputAtlasTmp = xASL_io_Nifti2Im(fullfile(x.dir.xASLDerivatives,x.SUBJECTS{1},listSessions{1},[x.S.InputAtlasNativeName '.nii']));
+	inputAtlasTmp = xASL_io_Nifti2Im(fullfile(x.dir.xASLDerivatives,x.SUBJECTS{indExistSubject},listSessions{indExistSession},[x.S.InputAtlasNativeName '.nii']));
```
HenkMutsaerts, line 223: Is this indexExistSession something that only occurs for native space processing?

HenkMutsaerts (changes requested):

Functions/Statistics/xASL_stat_PrintStats.m
jan-petr: Yes. that's correct. Before sum, we do ~isempty which requires single occurence.

Functions/BIDS/xASL_bids_Add2ParticipantsTSV.m
jan-petr: We are looking for T1, GM, etc. So the goal is to find rcT1, rGM etc. 

At the same time, we separate the first group of characters separated by underscore. so 'PV_GM' didn't work as it took 'PV' only. We had to avoid separating 'PV' from 'PV_GM'. It couldn't be solved by adding 'PV_GM' to the list

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
jan-petr: Yes. But we need this to work without bugs. And the entire function was badly designed from the start (before adding native space) - sorry for saying.

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
jan-petr: yes

Functions/Statistics/xASL_stat_PrintStats.m
HenkMutsaerts: Are you sure it cannot have multiple occurrences? I'm not fan of more code that makes processing faster (I know as a programming this is beneficial, but that's not the case when easily wanting to read/write what is done processing-wise)

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
HenkMutsaerts: OK then please add it to the list to make space-independent. All these things should become space-independent, as we agreed. This whole function is becoming unreadable

Functions/Statistics/xASL_stat_PrintStats.m
HenkMutsaerts: I checked again. It seems that this would only work if the subject regular expression is not fully correct (the subject regular expression should occur only once in the subjectID)

-> again, I rather have things to crash for good reasons, then fixing things to avoid crashing

Functions/Statistics/xASL_stat_PrintStats.m
HenkMutsaerts: This is a required parameter, that should be defaulted. I don't see why you would want to comment this out.

Functions/Statistics/xASL_stat_PrintStats.m
HenkMutsaerts: This is a required parameter, that should be defaulted. I don't see why you would want to comment this out.

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
HenkMutsaerts: I will approve this, if you add something like

%%%%%%%%%%%%%%% NATIVE SPACE TEMPORARY FIX %%%%%%%%%%%%%

in front and at the end, and add this to the list of things that we want to make space-independent (lesion analysis should also work in standard space)

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
HenkMutsaerts, line 112: So in the future, the ROIs/atlases should always be skipped for all subjects & sessions (can you add this to the list?)

HenkMutsaerts (changes requested):
See comments

HenkMutsaerts (changes requested):
Still two comments from me, can you fix them?

And do you fix the lesion stuff in xASL_stat_GetROiStats in #2092 (then please rename this issue and PR accordingly)? (#2095 also has something about lesion skipping)

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
jan-petr: Added to issue #58.

Modules/Module_Processing/Module_Population/xASL_stat_GetROIstatistics.m
jan-petr: But this is not a temporary fix. It's a necessity! For native space, we simply need to load one of the masks for one of the subjects. For standard space, it works differently.

You wanted to have it all in a single function. So the consequence is that you have this branching code. There's no way around that other than fully redesigning everything.

jan-petr: In any case -> #58 now has a comment that we are dealing with this (as said in the comment above). And I'll explain this in a comment in the code. But the whole part of code is withing "if native space", so makes no sense to write "temporary native space fix" when the entire code section is clearly run only for native space

jan-petr: I've checked the code and added extra comments. And as said before:
We are not doing any specific dirty hacks, but we are just dealing with the decision of having everything in the same function!

Standard space analysis masks are set with a specific image size and hard code voxel size. For native space, we simply need to initialize everything with the study-specific image size and specific voxel size. For that, we simply need extra code and we need extra options and we need to find and load the image. So this is not a temporary fix, but something that will always have to be there.

HenkMutsaerts (approved):

PR #2102, opened 2026-07-16
Fixes #2096 population module

### LINKED ISSUE #2096: Population module unnecessary warnings
### Description

For the participants TSV, it looks for files containing T1, GM, etc. And it looks into file names separating the first group of characters separated by '_'. This doesn't work for 'PV_GM'.

### Tasks

- [x] xASL_bids_Add2ParticipantsTSV: improve text
- [x] xASL_wrp_CreatePopulationTemplates: remove underscores for template parameter names (e.g. PV_GM -> PVGM)
For checking the session column validity, if a session column is missing, we check
if we have structural keys. We expect the format key_unit, e.g. pvGM_mL.
So ensure that, if we load filenames like PV_GM, that we remove the underscore.

- [x] xASL_wrp_CreatePopulationTemplates: fix session-loading

xASL_wrp_CreatePopulationTemplates currently loads images per session, and recreates a template per session, but overwrites this, both as template names and in the QC_RMS.txv

So for 3 subjects with 3 sessions, we get n=3 MaskVascular, CBF, etc instead of n=9 for the Templates/*.nii. E.g. CBF_bs-mean.nii that is created from ASL_3's only.

I suggest to fix this bug to change this default to n=9, to combine all subjects&sessions.
The alternative is that we keep sessions apart, e.g. CBF_ASL_1_bs-mean.nii CBF_ASL_2_bs-mean.nii etc, but the tricky thing here is that for the masks, we want to combine them from all subjects.
E.g. MaskSuscepbility, MaskVascular, those should be averaged across all; and then we will check if a voxel is present in at least 95% of all SubjectSessions (not always the best but a good pragmatic solution).

Agree with combining all subjects&sessions to n=9 templates (in this example)?

JAN: I'm fine combining, but we also need templates per session. Pehaps we can combine masks and generate separate+combined templates.

HENK: OK then I will keep templates per session (only) for now, and ensure that masks use combined templates. Disadvantage is that if we have 10 * ASL_1 and 100 * ASL_2, we will weigh them equally for the masks (so not weighed by the number of subjects that have a session). OK?

JAN: OK. Because we can only make a global mask? That's OK with me. But for DEBBIE, we really want to have templates per session rather than complete. Also for other studies with potential changes between sessions.

### Release notes

Fix minor issues in the population module.

HenkMutsaerts:
### Linked issue

Fixes #2096

### DIFF
--- a/Modules/Module_Processing/Module_Population/xASL_wrp_CreatePopulationTemplates.m
+++ b/Modules/Module_Processing/Module_Population/xASL_wrp_CreatePopulationTemplates.m
@@ -536,6 +536,36 @@ for iScanType=1:length(PreFixList)
         clear IM % Save each session independently
     end % for iSession=1:nSessions
 
+    % Now, we average over multiple sessions/runs, e.g., for creating analyses masks
+    if SessionsExist(iScanType)
+        IM_Masked = [];
+        IM_Unmasked = [];
+        includeSess_Masked = 1;
+        includeSess_Unmasked = 1;
+        for iSession=1:length(listSessions)
+            fileList = xASL_adm_GetFileList(x.D.TemplatesStudyDir, ['^' TemplateNameList{iScanType} '_' listSessions{iSession} x.S.TemplateNumberName '_bs-mean\.nii']);
+            if ~isempty(fileList)
+                IM_Masked(:,:,:,includeSess_Masked) = xASL_io_Nifti2Im(fileList{1});
+                includeSess_Masked = includeSess_Masked+1;
+            end
+            fileList = xASL_adm_GetFileList(x.D.TemplatesStudyDir, ['^' TemplateNameList{iScanType} '_' listSessions{iSession} x.S.TemplateNumberName '_bs-mean_Unmasked\.nii']);
+            if ~isempty(fileList)
+                IM_Unmasked(:,:,:,includeSess_Unmasked) = xASL_io_Nifti2Im(fileList{1});
+                includeSess_Unmasked = includeSess_Unmasked+1;
+            end
+        end
+        if ~isempty(IM_Masked)
+            IM_Masked = mean(IM_Masked, 4);
+            saveFile = fullfile(x.D.TemplatesStudyDir, [TemplateNameList{iScanType} x.S.TemplateNumberName '_bs-mean.nii']);
+            xASL_io_SaveNifti(fileList{1}, saveFile, IM_Masked)
+        end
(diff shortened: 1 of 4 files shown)


Modules/Module_Processing/Module_Population/xASL_wrp_CreatePopulationTemplates.m
```diff
@@ -520,10 +532,42 @@
...
     end % for iSession=1:nSessions
-    fprintf('\n');
+
+    % Now, we average over multiple sessions/runs, e.g., for creating analyses masks
+    if SessionsExist(iScanType)
+        IM_Masked = [];
+        IM_Unmasked = [];
+        includeSess_Masked = 1;
+        includeSess_Unmasked = 1;
+        for iSession=1:length(listSessions)
```
jan-petr, line 545: This is problematic memory-wise to load all images to memory and afterwards calculate mean. We should sum right away - that's faster and less memory consuming. I can fix that....

jan-petr: related to that - There is similar code above that loads into memory the whole set of images for all subjects. This is fully unfeasible for large datasets and will create unattainable memory requirements. That part cannot be rewritten easily because it calls a complicated subfunction. We need to list the basic requirements we expect from this subfunction and we can then still rewrite this relatively easily by calling this subfunction iteratively and iteratively updating the templates. So the functionality and function will be retained, but we will avoid storing all in the memory. Also note that the IM2Column function should be removed as the first step. It reduces memory requirements by constant, while storing all subjects increases memory complexity linearly with the subject number -> the first one doesn't help at all in large studies, the second one is a huge problem.

jan-petr (changes requested):
See comments. Space complexity needs to be solved.

Modules/Module_Processing/Module_Population/xASL_wrp_CreatePopulationTemplates.m
HenkMutsaerts: You're right, but this is not what I changed here. The summing here at the new lines is summing session-averages (e.g. for two sessions, all subjects are averaged for the first session, and for the second session, we already had this; I only added that session 1 and session 2 are averaged together).

So yes, happy to change it eventually, although it will make some functionality harder; but generally I agree with you. Just not for this fix; the code that I now added will still be used after your tweaks :)

jan-petr (approved):
OK

PR #2110, opened 2026-07-28
#2101 im_dilateROI Fix for empty lesions

### LINKED ISSUE #2101: xASL_wrp_Resample2StandardSpace : dilating user-provided lesions/ROIs
### Description

The resampling function in structural module - xASL_wrp_Resample2StandardSpace - dilates lesions until they reach 40mL.

There are two questions:
1. if user provides an empty ROI, which might happen when mutliple automated ROIs are provided, it crashes ExploreASL as it keeps dilating without ending.
2. Providing ROIs is a user responsibility. We are now taking user ROIs and dilating them to an ad-hoc size. This risks undermining the trust in ExploreASL.
 
Proposed solution to address both issues -> do not dilate user-provided ROIs.

HENK: This was implemented as lesions can be single voxels, as annotations (e.g., the location of microbleeds). In this case, a lesion may need dilation because it otherwise gets lost.

40 mm^3 is equal to 3 voxels in all directions in DARTEL space, or around the highest obtainable ASL effective resolution (3x3x4 mm).

We could lower this value, e.g., to 2.5x2.5x2.5 ~ 15? Would this suffice? I wouldn't go lower.

I agree that we should also leave some responsibility to the user, but at the same time, we should avoid that the users report noise; or that single-voxel annotations get lost and you get an error later on. Is there a good compromise? (e.g., the same question can be asked for the min ROI size in standard space of 1 mL).

JAN:
-The main confusion comes from the fact the xASL_im_DilateROI says 40mL. So that's huge. 40mm3 is better. 40mm3 is 5 voxels at 2x2x2mm resolution, so fully fine. We only need to fix the description in the functions.
-The second thing is that if a user provides ROI that is empty for a particular subject, then dilation is skipped. Providing a failsafe isn't necessary as N/A will be reported later.

### Tasks

- [x] xASL_im_DilateROI: fix description: 40 mL-> 0.4 mL
- [x] xASL_im_DilateROI: skip dilation for empty ROI

### Release notes

Fix bug in dilating empty lesions.

jan-petr:
### Linked issue

Close #2101

### DIFF
--- a/Functions/ImageProcessing/xASL_im_dilateROI.m
+++ b/Functions/ImageProcessing/xASL_im_dilateROI.m
@@ -5,11 +5,11 @@ function xASL_im_dilateROI(PathIn, PathOut, minVolume)
 % 
 % INPUT:        PathIn    - Path to input NIfTI image (REQUIRED)
 %               PathOut   - Path to output NIfTI image (OPTIONAL, DEFAULT = PathIn)
-%               minVolume - The ROI is dilated until reaching this volume in ml (OPTIONAL, DEFAULT 40ml)
+%               minVolume - The ROI is dilated until reaching this volume in mm^3 (OPTIONAL, DEFAULT 40mm^3)
 % OUTPUT:       n/a
 % 
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
-% DESCRIPTION:  The function loads a binary image from PathIn and if smaller than the defined volume (40 mL by default) it 
+% DESCRIPTION:  The function loads a binary image from PathIn and if smaller than the defined volume (40 mm^3 by default) it 
 %               dilates it with a 3x3 sphere element until a minimal volume is reached. When it is small enough, it is saved to PathOut.
 %               40 mm^3 is equal to 3 voxels in all directions in DARTEL space, or around the highest obtainable ASL effective resolution (3x3x4 mm).
 %
@@ -40,9 +40,12 @@ pixdim  = prod(HDR.hdr.pixdim(2:4));
 IMvol   = sum(IM(:)).*pixdim;
 
 %% Iteratively dilate until the minimal volume value is reached
-while  IMvol<minVolume
-       IM = xASL_im_DilateErodeFull(IM,'dilate',xASL_im_DilateErodeSphere(2));
-       IMvol   = sum(IM(:)).*pixdim;
+if IMvol > 0
+	% Only do iterative dilation when the Lesion is not empty
+	while  IMvol<minVolume
+		IM = xASL_im_DilateErodeFull(IM,'dilate',xASL_im_DilateErodeSphere(2));
+		IMvol   = sum(IM(:)).*pixdim;
+	end
 end
 
 %% Saves the dilated ROI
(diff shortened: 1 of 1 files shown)


Functions/ImageProcessing/xASL_im_dilateROI.m
```diff
@@ -5,11 +5,11 @@ function xASL_im_dilateROI(PathIn, PathOut, minVolume)
 % 
 % INPUT:        PathIn    - Path to input NIfTI image (REQUIRED)
 %               PathOut   - Path to output NIfTI image (OPTIONAL, DEFAULT = PathIn)
-%               minVolume - The ROI is dilated until reaching this volume in ml (OPTIONAL, DEFAULT 40ml)
+%               minVolume - The ROI is dilated until reaching this volume in mm^3 (OPTIONAL, DEFAULT 40mm^3)
```
HenkMutsaerts: I would replace 40mm^3 by 0.4 mL

Functions/ImageProcessing/xASL_im_dilateROI.m
```diff
@@ -5,11 +5,11 @@ function xASL_im_dilateROI(PathIn, PathOut, minVolume)
...
 % OUTPUT:       n/a
 % 
 % -----------------------------------------------------------------------------------------------------------------------------------------------------
-% DESCRIPTION:  The function loads a binary image from PathIn and if smaller than the defined volume (40 mL by default) it 
+% DESCRIPTION:  The function loads a binary image from PathIn and if smaller than the defined volume (40 mm^3 by default) it
```
HenkMutsaerts: I would replace 40mm^3 by 0.4 mL

Functions/ImageProcessing/xASL_im_dilateROI.m
```diff
@@ -40,9 +40,12 @@ function xASL_im_dilateROI(PathIn, PathOut, minVolume)
 IMvol   = sum(IM(:)).*pixdim;
```
HenkMutsaerts: Should we replace this by xASL_stat_NanSum, or mask with finites, to avoid resampling issues?

Functions/ImageProcessing/xASL_im_dilateROI.m
```diff
@@ -40,9 +40,12 @@ function xASL_im_dilateROI(PathIn, PathOut, minVolume)
...
 
 %% Iteratively dilate until the minimal volume value is reached
-while  IMvol<minVolume
-       IM = xASL_im_DilateErodeFull(IM,'dilate',xASL_im_DilateErodeSphere(2));
-       IMvol   = sum(IM(:)).*pixdim;
+if IMvol > 0
+	% Only do iterative dilation when the Lesion is not empty
+	while  IMvol<minVolume
+		IM = xASL_im_DilateErodeFull(IM,'dilate',xASL_im_DilateErodeSphere(2));
+		IMvol   = sum(IM(:)).*pixdim;
```
HenkMutsaerts: here the same question, if we should mask finites

HenkMutsaerts (approved):
Few minor points, I don't have to review it again

Functions/ImageProcessing/xASL_im_dilateROI.m
jan-petr: Done

Functions/ImageProcessing/xASL_im_dilateROI.m
jan-petr: Done

Functions/ImageProcessing/xASL_im_dilateROI.m
jan-petr: done

Functions/ImageProcessing/xASL_im_dilateROI.m
jan-petr: done

PR #2119, opened 2026-08-04
#2118 wrp_Quantify Disable scaling correction

### LINKED ISSUE #2118: Strong correction with M0 turned off
### Description

When we turn off all quantification steps (for example when we have quantified CBF or ATT on the input), then a special portion of the code is activated that scales the entire mean CBF to 10mL.

This is highly random procedure - unclear why this threshold is used and why it's scaled to 10mL. Especially when M0 scaling is deactivated, it can easily happen to have strong outliers and it doesn't really help anything to scale them.

I propose to either increase the threshold for this scaling - so that we can process ATT images. Or remove this correction completely.

See the proposed code change. But I'd be happy removing this completely - or at least removing this completely when CBF quantification is turned off.

It should be noted that this operates rather randomly a get triggered differently for native and standard space values (as the mean is different due to different image size etc). So I would really welcome removing this random check altogether. I'm of course fine keeping a warning or way higher threshold.

HENK: this is meant that if CBF values are way off, it scales down. I agree that the threshold and resulting mean should be sensible, see my quick calculations below, 10 mL/100/min as mean over the whole image is not weird (assuming that 2/3 of the image is air and 0).

* First question: if quantification is off; we assume that quantification was already done, and we assume to have CBF output.
So your use case, disabling quantification, but having only PWI and ATT, seems weird?

But fine to remove this all together; this is probably historical, something that was useful at the time.

EXPLANATION:

Threshold: average brain should be 1/3.*100+1/3*25+0 ~ 42 mL/100g/min, air = 0, so whole image average CBF should be:
1/3 .* 42 = 14 mL/100g/min.

Current threshold 666/14 = 48-fold scaled too high, which we would correct by 66.* (10/666) to become 10, which is similar to the calculated 14 above.

JAN: Yes, it is roughly reasonable. But the whole message is that if we have CBF values that are 50*times off, we scale it to a random number that can be 2-3 times off. I'm fine issuing a strong warning. But scaling helps nothing at all and it makes it hard to trace any errors.

### Release notes

In quantification without M0, skip normalization when CBF values are off.

jan-petr:
### Linked issue

Close #2118

### DIFF
--- a/Modules/Module_Processing/Module_ASL/xASL_wrp_Quantify.m
+++ b/Modules/Module_Processing/Module_ASL/xASL_wrp_Quantify.m
@@ -364,10 +364,8 @@ end
 
 if x.modules.asl.ApplyQuantification(5)==0
     MeanCBF = xASL_stat_MeanNan(CBF(:));
-    if MeanCBF>2000 % this is the average including air
-        CBF = CBF .* (10./MeanCBF); % protection against extremely high values
+    if MeanCBF>666 % this is the average including air
         warning('M0 division was disabled & CBF image had too high values');
-        fprintf('%s\n',['mean whole image CBF normalized from ' xASL_num2str(MeanCBF) ' to 10 mL/100g/min']);
     end
 end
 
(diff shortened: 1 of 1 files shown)


HenkMutsaerts (approved):
I agree, but then I would make the warning a bit stronger, e.g. by adding before and after the warning:

fprintf('\n\n\n\n');

PR #2124, opened 2026-08-10
#2123 Hematocrit quantification

### LINKED ISSUE #2123: Fix errors in hematocrit correction
### Description

Fix a few small errors found by AI in how Hct is used to calculate BloodT1

### Further discussion

- json sidecars contain BIDS only parameters. JAN: This is already the case.
- Patient-specific parameters. The priority in adding them is
   - Patient BloodT1 (we have to see if this entry is possible through as BloodT1 isn't a BIDS parameter. So for now, we skip this)
   - Study BloodT1
   - Patient Hct
   - Patient age/sex
   - Default BloodT1
- Voxel-wise parameter maps: These are not used so far. The only solution we have in mind is e.g. calculating T2 map based on multi-TE. But that will be an internal parameter. We will solves this in a later issue. Obviously, parameter map has the highest priority.
- Specific ancillary parameters not related to acquisition (T1-GM, T2-GM) are kept in dataPar.json
### Tasks

- [x] y and Y inside the function xASL_quant_Hct2BloodT1
- [x] it incorrectly talks about venous O2 saturation instead of arterial in function description
- [x] it incorreclty says that user provided BloodT1 has priority, while the highest priority is in provided Hematocrit values
- [x] If both exist (bloodT1 and Hct) -> we take bloodT1 and create a warning.
- [x] Create a new issue patient bloodT1 vs study blood T1. Also for T2GM etc. How to solve participants.tsv incompleteness/typos. #2126 

### Release notes

Minor fixes in hematocrit correction.

jan-petr:
### Linked issue

Close #2123

### DIFF
--- a/Functions/Quantification/xASL_quant_DefineQuantificationParameters.m
+++ b/Functions/Quantification/xASL_quant_DefineQuantificationParameters.m
@@ -151,12 +151,12 @@ end
 %% 2.   Arterial blood T1
 % T1 relaxation time of arterial blood
 % There are 3 options for x.Q.T1blood (A has the highest priority:
-% A) users have provided x.Hematocrit (in any of the forms defined in xASL_wrp_Quantify 3.a-c), which is converted to x.Q.T1blood there
-% B) users have provided x.Q.T1blood
+% A) users have provided x.Q.T1blood
+% B) users have provided x.Hematocrit (in any of the forms defined in xASL_wrp_Quantify 3.a-c), which is converted to x.Q.T1blood there
 % C) it doesn't exist and is defaulted here based on MagneticFieldStrength
 
 % We convert x.Q.Hematocrit -> x.Q.T1blood
-if isfield(x.Q, 'Hematocrit')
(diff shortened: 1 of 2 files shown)


HenkMutsaerts (changes requested):
Amazing this AI, we can run it on all code before releasing 2.0.0 :) Matlab does it now as well inside the code editor.

1 thing I would change: blood T1 should have priority over hematocrit; I would define the priority as "closest to the CBF quantification" -> T1, measured Hct, estimated Hct from age/sex.
That's also how I set it up, if this is incorrectly programmed it should change to this priority, instead of changing the explanation.

The venous vs arterial oxygenation: are you sure? Blood T1 measurements, and hematocrit measurements, are often venous. Just to check.

HenkMutsaerts (approved):
Nice. See my cosmetic last commit (while I was reviewing your change in the Matlab code).

I would refactor this function in the new issue at least such that it is processed by priority, now the order inside the function is also still a bit weird.

Can you notify me where the new issue is that I check it, before I forget our phone discussion :)

Thanks

PR #2144, opened 2026-08-21
#2143 GE import fix

### LINKED ISSUE #2143: GE reading with SPM fails
### Description

SPM reading instead of DCMTK adds a space at the end of certain strings.

When we use strcmpi command then this fails with this additional space in the string, so we need to replace it with regexpi that is able to ignore the additional space generate by SPM DICOM reading instead of DCMTK.

### Release notes

Minor fix of GE import. Report with the other GE fix.

jan-petr:
### Linked issue

Close #2143

### DIFF
--- a/Modules/Module_Import/NII2BIDS/xASL_bids_determineImageTypeGE.m
+++ b/Modules/Module_Import/NII2BIDS/xASL_bids_determineImageTypeGE.m
@@ -40,44 +40,45 @@ function imageType = xASL_bids_determineImageTypeGE(jsonPar)
     end
 
     % ["ImageType": ["DERIVED", "PRIMARY", "ASL", "PERFUSION", "ASL"] - deltaM
-    if length(jsonPar.ImageType) == 5 && strcmpi(jsonPar.ImageType{1},'DERIVED') && strcmpi(jsonPar.ImageType{2},'PRIMARY') &&...
-            strcmpi(jsonPar.ImageType{3},'ASL') && strcmpi(jsonPar.ImageType{4},'PERFUSION') && strcmpi(jsonPar.ImageType{5},'ASL')
+    if length(jsonPar.ImageType) == 5 && ~isempty(regexpi(jsonPar.ImageType{1},'DERIVED','once')) && ~isempty(regexpi(jsonPar.ImageType{2},'PRIMARY','once')) &&...
+            ~isempty(regexpi(jsonPar.ImageType{3},'ASL','once')) && ~isempty(regexpi(jsonPar.ImageType{4},'PERFUSION','once')) && ~isempty(regexpi(jsonPar.ImageType{5},'ASL','once'))
         imageType = 'deltam';
 	end
 	
 	% ["ImageType": ["DERIVED", "PRIMARY", "ASL", "PERFUSION", "ASL", "REAL"] - deltaM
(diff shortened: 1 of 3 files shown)


HenkMutsaerts (approved):
So we don't care about the contexts of these fields, just that it contains the text at least once? What if it contains it twice, do you want to catch it (see my other suggested fixes)?

Modules/Module_Import/NII2BIDS/xASL_bids_determineImageTypeGE.m
```diff
@@ -40,44 +40,45 @@
...
+            ~isempty(regexpi(jsonPar.ImageType{3},'ASL','once')) && ~isempty(regexpi(jsonPar.ImageType{4},'PERFUSION','once')) &&...
+			~isempty(regexpi(jsonPar.ImageType{5},'ASL','once')) && ~isempty(regexpi(jsonPar.ImageType{6},'REAL','once'))
         imageType = 'deltam';
     end
 
     % ["DERIVED", "PRIMARY", "ASL", "PERFUSION_ASL"] - deltaM
-    if length(jsonPar.ImageType) == 4 && strcmpi(jsonPar.ImageType{1},'DERIVED') && strcmpi(jsonPar.ImageType{2},'PRIMARY') &&...
-            strcmpi(jsonPar.ImageType{3},'ASL') && strcmpi(jsonPar.ImageType{4},'PERFUSION_ASL')
+    if length(jsonPar.ImageType) == 4 && ~isempty(regexpi(jsonPar.ImageType{1},'DERIVED','once')) && ~isempty(regexpi(jsonPar.ImageType{2},'PRIMARY','once')) &&...
+            ~isempty(regexpi(jsonPar.ImageType{3},'ASL','once')) && ~isempty(regexpi(jsonPar.ImageType{4},'PERFUSION_ASL','once'))
```
jan-petr: ASL at start

Modules/Module_Import/NII2BIDS/xASL_bids_determineImageTypeGE.m
```diff
@@ -40,44 +40,45 @@
     end
 
     % ["ImageType": ["DERIVED", "PRIMARY", "ASL", "PERFUSION", "ASL"] - deltaM
-    if length(jsonPar.ImageType) == 5 && strcmpi(jsonPar.ImageType{1},'DERIVED') && strcmpi(jsonPar.ImageType{2},'PRIMARY') &&...
-            strcmpi(jsonPar.ImageType{3},'ASL') && strcmpi(jsonPar.ImageType{4},'PERFUSION') && strcmpi(jsonPar.ImageType{5},'ASL')
+    if length(jsonPar.ImageType) == 5 && ~isempty(regexpi(jsonPar.ImageType{1},'DERIVED','once')) && ~isempty(regexpi(jsonPar.ImageType{2},'PRIMARY','once')) &&...
+            ~isempty(regexpi(jsonPar.ImageType{3},'ASL','once')) && ~isempty(regexpi(jsonPar.ImageType{4},'PERFUSION','once')) && ~isempty(regexpi(jsonPar.ImageType{5},'ASL','once'))
```
jan-petr: ASL at start

jan-petr:
> So we don't care about the contexts of these fields, just that it contains the text at least once? What if it contains it twice, do you want to catch it (see my other suggested fixes)?

Not sure what context means. But this is fine like this, there's no real posibility to confuse the current implementation with anything else.

The best we can do is to make sure that ASL is not confused with PERFUSION_ASL, but that's all.

PR #2149, opened 2026-08-27
Closes #2148 stat_GetROIstatistics minor fix

### LINKED ISSUE #2148: Incorrect use of number of sessions in the population module
### Description

Inside xASL_stat_GetROIStatistics, we use listSessions that contain a list of all available sessions in the population module. 

For DEBBIE, it might happen that we have sessions ASL_1, ASL_2, ASL_3, ASL_4. But only ASL_2 and ASL_4 are analyzed in the population module. So while cycling through the listSession, we should use it's real length. And not use the total available count of sessions estimated from the ASL module.

See my fix, it's clear.

### Release notes

Minor fix in sessions in Population module.

HenkMutsaerts:
### Linked issue

Closes #2148


jan-petr (approved):
OK

